PR #5, opened 2018-09-04
Add Configuration Validation

jkmassel:
Adds a set of fastlane actions to allow dealing with configuration secrets.

**To Test:**

#### fastlane run configure_setup
1. Add the Release Toolkit to a project's `Pluginfile`, then update your dependencies.
2. Run `bundle exec fastlane run configure_setup` (works from anywhere in the project that calling `fastlane` works). You'll be prompted to either install the sample data, or install from another git repository you have access to. For testing purposes, it's probably easiest to use the sample data.
3. Next, you'll be prompted to add any files you wish to be copied from the secrets repository to your project. Type `y` for yes, then feel free to just us the examples provided. For the source, you have to enter a valid file path, but for the destination, you can use any path, so enter a path that won't overwrite an existing file, just to be safe.
4. You'll keep being asked if you want to add files until you say `n`, so do that now. 
5. The tool will now apply the configuration to the project. If any files will be overwritten, you'll be prompted whether to not to continue.

#### fastlane run configure_validate /  fastlane run configure_update

Run `bundle exec fastlane run configure_validate` anywhere in the project directory that calling `fastlane` works. The configuration should be valid, as the setup tool will not allow you to create an invalid configuration. To test invalid configurations, try the following:

1. The sample data repository contains two additional branches – track either of the non-`master` branches, and switch to it. You should receive an error that the current branch doesn't match the one specified in `.configure`. This prevents us from accidentally building with invalid credentials.

To fix this issue, run `bundle exec fastlane run configure_update`, and when it asks if you want to switch branches, say yes and choose the `master` branch, then try validating again.

2. Change the commit hash in `.configure` to an old commit hash. You should receive an error that the `.configure` file is out of date. This prevents us from missing updates to the secrets repository. 

To fix this issue, run `bundle exec fastlane run configure_update`, and when it asks if you want to switch branches, say no, but when it asks if you want to update the `.configure` file to the latest commit hash, say yes, then try validating again.

3. Make a change in `~/.mobile-secrets` (as a for-instance, you could duplicate a file or folder). You should receive an error saying that the secrets repository has uncommitted changes.

To fix this issue, it's easiest to just delete the file or folder you just made, but if you were to commit your changes, you'd receive the error from step 2, saying that your `.configure` file is out of date.

4. Change a value in one of the files copied from the secrets repo (the change must be done within the project – if you do it within the secrets repo, you'll get the error from step 3). You should receive an error saying that the file doesn't match the file in the secrets repository.  If you were to delete the file entirely, there would be a separate error saying that the file does not exist.

To fix this issue, just undo your change, and the project should validate.

#### fastlane run configure_add_files_to_copy

Running this action will interactively walk you through adding a source / destination pair to the `.configure` file – it's actually invoked as part of the setup action, but can be run separately.

#### fastlane run configure_download

Running this action will update the local secrets repo with changes from the remote.

#### fastlane run configure_apply

Running this action will copy any files that are specified in `.configure` from the repo to the project. By default, this action will ask for confirmation before overwriting any existing files. Running `fastlane run configure_apply force:true` will not ask for confirmation.

Let me know if you have any questions or concerns, and thanks for reading this far!

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_download_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_download_action.rb
@@ -1,0 +1,94 @@
+require 'fastlane/action'
+require 'fastlane_core/ui/ui'
+require 'fileutils'
+
+module Fastlane
+  module Actions
+    class ConfigureDownloadAction < Action
+      def self.run(params = {})
+
+        UI.message "Running Configure Download"
+
+        # If the `~/.mobile-secrets` repository doesn't exist
+        unless File.directory?("#{Dir.home}/.mobile-secrets")
+
+          choice = UI.select("No local secret store exists. Please choose an option: ", [
+              "Install the sample data (default)",
+              "Install from a git repository"
+          ])
+
+          # If they enter anything other than "2", just do the default action
+          case choice
+              when "2" then install_from_git_repository 
+              else install_sample_data
+          end
+        else
+          update_repository # If the repo already exists, just update it
+        end
+      end
+
+      # Install the sample repository
+      def self.install_sample_data
+
+        UI.message "Installing from sample data"
+
+        clone_repository("https://github.com/jkmassel/mobile-secrets-repository")
+
+        UI.success "Sample Data Setup Complete"
+      end
+
+      # Attempt to install a user-provided git repository
+      def self.install_from_git_repository
+
+        UI.message "Installing from a git repository"
+
+        ### Prompt the user for the git repo URL
+        repo_url = UI.input("Git Repository URL:  ")
+
+        clone_repository(repo_url)
+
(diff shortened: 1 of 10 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_download_action.rb
```diff
@@ -0,0 +1,94 @@
...
+          update_repository # If the repo already exists, just update it
+        end
+      end
+
+      # Install the sample repository
+      def self.install_sample_data
+
+        UI.message "Installing from sample data"
+
+        clone_repository("https://github.com/jkmassel/mobile-secrets-repository")
```
jkmassel: We should make this something that falls under the WP repos, rather than my personal. It's just there for testing right now :)

PR #11, opened 2018-12-06
Prompt the user to backup the file before overwriting

jkmassel:
Doesn’t apply when using the `force` flag


loremattei:
@jkmassel I tested it and it works perfectly! 
I see a small issue though: the backup file is not in the .gitignore list and there's the risk that it'll be committed by mistake. 
What do you think about using `.bak` at the end of the file name? This extension has been in `.gitignore` for long time, so it's a safe one to use.

jkmassel:
>What do you think about using .bak at the end of the file name? This extension has been in .gitignore for long time, so it's a safe one to use.

This is a great idea – I've added it!

Also made a change to make the command run a little more nicely – if the files haven't changed at all, there's no prompt to overwrite.

jkmassel:
Added the ability to show a diff before prompting to overwrite the file.

jkmassel:
Thank you for reviewing @loremattei!

PR #12, opened 2018-12-12
Add per-file validation

jkmassel:
This PR adds support for validating individual files, rather than the whole repo. It's backwards-compatible with older `.configure` files, too – so this shouldn't really break anything.

**How to test:**

On an Android project, update the `.configure` file, setting the `pinned_hash` to a commit hash from the secrets repo – pick one from way back, so that there will have been changes that affect the files in the project. If you run `bundle exec fastlane run configure_validate`, you'll get an error listing which files have changed, and prompting you on how to fix the issue. Following the instructions will resolve the problem for you.

Next, pick a commit hash that isn't the most recent, but that's new enough that there are no new changes to files used by this project. The validation will pass.

~iOS `.configure` files will need a manual update to add the `file_dependencies` array, which will be used to keep track of which files are used for those repositories. I'll add those to the projects in order to enable testing them, and will reference the PRs for those changes below. Once these are merged, I'll update this PR with direction on how to test iOS 🙂~

Now that the `.configure` files have been updated, you can test these in the same way you did Android – that'll show you how it responds to changes to already-existing files. To test how files that exist in the repo, but don't yet exist in the project work, I'd suggest testing with WPiOS and pinning a commit prior to the recent screenshot credentials change.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_update_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_update_action.rb
@@ -26,6 +26,9 @@ module Fastlane
         end
 
         UI.success "Configuration Secrets are up to date – don't forget to commit your changes to `.configure`."
+
+        # Apply the changes that are now in the .configure file
+        ConfigureApplyAction::run
       end
 
       def self.prompt_to_switch_branches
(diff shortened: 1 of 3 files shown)


jkmassel:
@loremattei - this is ready for review at your convenience! 😃

jkmassel:
Thanks for the review @loremattei! 😄

PR #28, opened 2019-02-25
Add missing octokit dependency

jtreanor:
`octokit` is used in https://github.com/wordpress-mobile/release-toolkit/blob/develop/lib/fastlane/plugin/wpmreleasetoolkit/helper/ghhelper_helper.rb but not declared as a dependency. This adds it to the gemspec.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,10 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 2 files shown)


loremattei:
Thanks for this @jtreanor!
:shipit:

PR #31, opened 2019-03-01
Re-add SetbranchprotectionAction action and remove BuddyBuild status check

jtreanor:
I removed `SetbranchprotectionAction` in wordpress-mobile/release-toolkit#24 but @loremattei pointed out that we now can't delete release branches 🤦‍♂️ 

This adds it back in but removes the BuudyBuild required check. Since this is used in more than one repo, a hardcoded check here doesn't make sense.

Reverts wordpress-mobile/release-toolkit#24


jtreanor:
@jkmassel Yes, exactly.

PR #35, opened 2019-03-22
Add actions for Android releases

loremattei:
This PR adds some actions that help to handle Android releases. The actions have been extracted from WPAndroid and can be tested in this WooCommerce Android PR: https://github.com/woocommerce/woocommerce-android/pull/924.


loremattei:
I think we can just import everything so that we don't need to add the platform every time we invoke fastlane.

jkmassel:
>I think we can just import everything so that we don't need to add the platform every time we invoke fastlane.

I agree, that'll be nicer to work with!

PR #45, opened 2019-05-24
Fix ios_hotfix_prechecks action + add tests

jkmassel:
For some reason, `bundle exec fastlane new_hotfix_release` wasn't work for me at all today on WPiOS.

AFAIK nothing changed, but suddenly file imports were failing.

This PR fixes that, and also adds a really rudimentary test script. Over time, I'd love to improve this (right now the tests don't even run on CI), but it's a starting point, at least.

**To Test:**
- In this branch, run `cd test && ./test.sh`


loremattei:
I tested the fix and it works perfectly. 

I also love the idea of having tests for the toolkit! It's a core part of the release process of most of our apps, so it has to be really reliable!

Though, I'm not totally sold on the test itself: it verifies that the operation finishes, but not that it actually does the right thing. Also, it pushes to this repository, which is not ideal. 

Do you think that it would be worth to think better about a reasonable test strategy for the whole plugin?

loremattei:
Hey @jkmassel I'm merging this because I'm about to release a new version of the toolkit and we can include this.

PR #47, opened 2019-06-20
Ensure secrets repo is on correct hash before copying files

markpar:
Currently in the WCAndroid/WPAndroid projects, running the command `bundle exec fastlane install configure_apply` ignores the `pinned_hash` and `branch` in `.configure`.  This has the potential to introduce bugs if the `mobile-secrets` repo is at a different hash.  

We can avoid this by checking to see if the secrets repo is on the correct hash (both branch and commit) and, if not, go ahead and check out `pinned_hash` before copying any files.  If the secrets repo is already on the correct branch and commit, then do nothing and go ahead with the file copy.

**To test:**
* Change Fastlane’s Pluginfile in WCAndroid/WPAndroid to point to this fork and branch, like so:
`gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/markpar/release-toolkit', branch: 'add/checkout-correct-hash'`
* `bundle install` or `bundle update` to pull down the changes
* `cd ~/.mobile-secrets`
* `git checkout HEAD~1` (to go back one commit - you could also switch to a new branch here)
* Go back to the app folder and run `bundle exec fastlane run config_apply`

The secrets repo should now be back on the commit specified by `pinned_hash`.  I tried all 4 combinations of right/wrong branch/commit and they all put the secrets repo into the same correct and expected state.

NB:  This change will result in the secrets repo being in a detached head state. Not _too_ big of a deal (you can just `git checkout master` to get back), but I did notice that the Download action will fail in this state, since it does a straight `git pull`, which fails in the detached head state.  If we want, we could address this separately, or not at all depending on how much pain it causes.  🙂 

Big thanks to @jtreanor for assistance!  🎉

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -14,11 +14,13 @@ module Fastlane
         ### Make sure secrets repo is at the proper hash as specified in .configure
         repo_hash = Fastlane::Helper::ConfigureHelper.repo_commit_hash
         file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
+        original_repo_branch = Fastlane::Helper::ConfigureHelper.repo_branch_name
 
         unless repo_hash == file_hash
           sh("cd #{repository_path} && git fetch && git checkout #{file_hash}")
         end
 
+        ### Copy the files
         files_to_copy.each { |x|
 
             source = absolute_secret_store_path(x["file"])
@@ -30,6 +32,9 @@ module Fastlane
                 copy_with_confirmation(source, destination)
             end
         }
+
+        ### Restore secrets repo to original branch
+        sh("cd #{repository_path} && git checkout #{original_repo_branch}")
 
         UI.success "Applied configuration"
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -18,7 +18,7 @@ module Fastlane
         file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
 
         unless repo_branch_name == file_branch_name && repo_hash == file_hash
-          sh("cd #{repository_path} && git checkout #{file_hash}")
+          sh("cd #{repository_path} && git fetch && git checkout #{file_hash}")
         end
 
         files_to_copy.each { |x|
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -34,8 +34,8 @@ module Fastlane
         }
 
         ### Restore secrets repo to original branch.  If it was originally in a 
-        ### detached HEAD state, we need to extract the hash from the branch name.
-        original_repo_branch = original_repo_branch.sub("(HEAD detached at ", "").sub(")", "")
+        ### detached HEAD state, we need to use the hash since there's no branch name.
+        original_repo_branch = repo_commit_hash if (original_repo_branch == nil)
 
         sh("cd #{repository_path} && git checkout #{original_repo_branch}")
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -11,7 +11,7 @@ module Fastlane
     class ConfigureApplyAction < Action
       def self.run(params = {})
 
-        ### Make sure secrets repo is at the proper hash as specified in .configure
+        ### Make sure secrets repo is at the proper hash as specified in .configure.
         repo_hash = Fastlane::Helper::ConfigureHelper.repo_commit_hash
         file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
         original_repo_branch = Fastlane::Helper::ConfigureHelper.repo_branch_name
@@ -33,7 +33,10 @@ module Fastlane
             end
         }
 
-        ### Restore secrets repo to original branch
+        ### Restore secrets repo to original branch.  If it was originally in a 
+        ### detached HEAD state, we need to extract the hash from the branch name.
+        original_repo_branch = original_repo_branch.sub("(HEAD detached at ", "").sub(")", "")
+
         sh("cd #{repository_path} && git checkout #{original_repo_branch}")
 
         UI.success "Applied configuration"
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -12,12 +12,10 @@ module Fastlane
       def self.run(params = {})
 
         ### Make sure secrets repo is at the proper hash as specified in .configure
-        repo_branch_name = Fastlane::Helper::ConfigureHelper.repo_branch_name
-        file_branch_name = Fastlane::Helper::ConfigureHelper.configure_file_branch_name
         repo_hash = Fastlane::Helper::ConfigureHelper.repo_commit_hash
         file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
 
-        unless repo_branch_name == file_branch_name && repo_hash == file_hash
+        unless repo_hash == file_hash
           sh("cd #{repository_path} && git fetch && git checkout #{file_hash}")
         end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -35,7 +35,7 @@ module Fastlane
 
         ### Restore secrets repo to original branch.  If it was originally in a 
         ### detached HEAD state, we need to use the hash since there's no branch name.
-        original_repo_branch = repo_commit_hash if (original_repo_branch == nil)
+        original_repo_branch = repo_hash if (original_repo_branch == nil)
 
         sh("cd #{repository_path} && git checkout #{original_repo_branch}")
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -10,6 +10,17 @@ module Fastlane
   module Actions
     class ConfigureApplyAction < Action
       def self.run(params = {})
+
+        ### Make sure secrets repo is at the proper hash as specified in .configure
+        repo_branch_name = Fastlane::Helper::ConfigureHelper.repo_branch_name
+        file_branch_name = Fastlane::Helper::ConfigureHelper.configure_file_branch_name
+        repo_hash = Fastlane::Helper::ConfigureHelper.repo_commit_hash
+        file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
+
+        unless repo_branch_name == file_branch_name && repo_hash == file_hash
+          sh("cd #{repository_path} && git checkout #{file_hash}")
+        end
+
         files_to_copy.each { |x|
 
             source = absolute_secret_store_path(x["file"])
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_download_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_download_action.rb
@@ -10,16 +10,26 @@ module Fastlane
         UI.message "Running Configure Download"
 
         # If the `~/.mobile-secrets` repository doesn't exist
-        unless File.directory?("#{Dir.home}/.mobile-secrets")
-            UI.user_error!("The local secrets store does not exist. Please clone it to ~/.mobile-secrets before continuing.")
+        unless File.directory?("#{secrets_dir}")
+            UI.user_error!("The local secrets store does not exist. Please clone it to #{secrets_dir} before continuing.")
         else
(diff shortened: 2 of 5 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
```diff
@@ -10,6 +10,17 @@ module Fastlane
...
       def self.run(params = {})
+
+        ### Make sure secrets repo is at the proper hash as specified in .configure
+        repo_branch_name = Fastlane::Helper::ConfigureHelper.repo_branch_name
+        file_branch_name = Fastlane::Helper::ConfigureHelper.configure_file_branch_name
+        repo_hash = Fastlane::Helper::ConfigureHelper.repo_commit_hash
+        file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
+
+        unless repo_branch_name == file_branch_name && repo_hash == file_hash
+          sh("cd #{repository_path} && git checkout #{file_hash}")
```
jtreanor: In addition to checking out the hash, should we also do a `git fetch` here? Its possible that the remote ref won't be available.

jtreanor:
> NB: This change will result in the secrets repo being in a detached head state. Not too big of a deal (you can just git checkout master to get back), but I did notice that the Download action will fail in this state, since it does a straight git pull, which fails in the detached head state. If we want, we could address this separately, or not at all depending on how much pain it causes. 🙂

This is a good point and I think the way you have things here is probably what we want. If none of our projects directly referenced files from the repo and always copied (the iOS projects reference and don't copy), checkout out the branch again could work but as things are now I think it would be best to leave it in the detached state.

@jkmassel What do you think?

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
markpar: I think we should, too!  Good catch - will add that!

markpar:
Hi @jtreanor!  I tested adding a fetch and it’s looking good.  Here’s the console output that shows both the fetch happening, as well as the text about being in a detached head state (i.e. the output from `git`).  

```[07:53:40]: -----------------------------
[07:53:40]: --- Step: configure_apply ---
[07:53:40]: -----------------------------
[07:53:40]: $ cd /Users/mark/.mobile-secrets && git fetch && git checkout b60ba8960378a5ce14a416654c1c80cd60af7403
[07:53:40]: ▸ From https://github.com/markpar/peril-settings
[07:53:40]: ▸ * [new branch]      fooboo2    -> origin/fooboo2
[07:53:40]: ▸ Note: checking out 'b60ba8960378a5ce14a416654c1c80cd60af7403'.
[07:53:40]: ▸ You are in 'detached HEAD' state. You can look around, make experimental
[07:53:40]: ▸ changes and commit them, and you can discard any commits you make in this
[07:53:40]: ▸ state without impacting any branches by performing another checkout.
[07:53:40]: ▸ If you want to create a new branch to retain commits you create, you may
[07:53:40]: ▸ do so (now or later) by using -b with the checkout command again. Example:
[07:53:40]: ▸ git checkout -b <new-branch-name>
[07:53:40]: ▸ HEAD is now at b60ba89 Create rule to post merge instructions
[07:53:40]: Applied configuration
[07:53:40]: Result: true```

File: lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb (hunk 1 above)
Comment (markpar) [reply lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb]: Added it!

File: lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb [hunk 2]
```diff
@@ -10,6 +10,17 @@ module Fastlane
...
     class ConfigureApplyAction < Action
       def self.run(params = {})
+
+        ### Make sure secrets repo is at the proper hash as specified in .configure
+        repo_branch_name = Fastlane::Helper::ConfigureHelper.repo_branch_name
+        file_branch_name = Fastlane::Helper::ConfigureHelper.configure_file_branch_name
+        repo_hash = Fastlane::Helper::ConfigureHelper.repo_commit_hash
+        file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
+
+        unless repo_branch_name == file_branch_name && repo_hash == file_hash
```
jtreanor: Is the branch check here needed? I think if the hashes match, thats all that matters but I may be wrong.

jtreanor:
Works as advertised. Thanks @markpar!

I have left one more small comment and I'll let @jkmassel weigh in on the detached state behaviour before merging.

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
markpar: I think that’s a great optimization and also makes the code more readable.  Removed!

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
```diff
@@ -10,6 +10,15 @@ module Fastlane
   module Actions
     class ConfigureApplyAction < Action
       def self.run(params = {})
+
+        ### Make sure secrets repo is at the proper hash as specified in .configure
+        repo_hash = Fastlane::Helper::ConfigureHelper.repo_commit_hash
+        file_hash = Fastlane::Helper::ConfigureHelper.configure_file_commit_hash
+
+        unless repo_hash == file_hash
+          sh("cd #{repository_path} && git fetch && git checkout #{file_hash}")
```
jtreanor: What happens if this command fails? Does the script exit or carry on with the assumption that the repo is in the correct state?

markpar: If `sh` fails, it logs an error to the console and exits.  The Fastlane docs say that `sh` takes an error callback function, so we could handle that case if we wanted to.

jtreanor: If it exists, I'm happy with that. Thanks for confirming 👍

jkmassel:
Hey y'all – my suggestion would be to always roll the repo back to `master` after each operation. 

Two reason:

1) If it were just one project using the repo, that'd be fine, but we all have multiple projects that need to access it – leaving it in a detached `head` state isn't the worst, but it's unlikely that that'd be correct for the next app accessing it. I'd say about 90% of the time, apps will be ok with having it in the `master` state (given that this bug hasn't really hurt us until recently).
2) If the repo were read-only by the apps, #1 could be debated. But one common-ish operation is to add new folks to the repo, and that should always be done against the latest` master. 

IMHO it just comes down to leaving the repo in a consistent, knowable state after every operation so there's never a question of where it's at.

My 2¢, anyways.

markpar:
This is great info - thanks, @jkmassel!  I can definitely see the value in switching back to `master`.  👍 

Since we know the state of the secrets repo from `repo_branch_name` (and `repo_commit_hash`) before checking out `pinned_hash`, we could roll back to that after the file copy operation.  As you note, the majority of the time, that will be `master`.  (I suggested this because I can envision a case where someone might be doing local development and have their secrets repo on a topic branch, but I’m not sure how common that case is.)

Alternately, we could always switch back to `master` after the file copy, no matter what. 

Generally, I lean towards leaving things how we found them (i.e. switching back to `repo_branch_name`), but you and @jtreanor have more context than I.  WDYT?

jkmassel:
You're right – leaving it exactly how you found it is even more superior to just going back to `master`.

>I suggested this because I can envision a case where someone might be doing local development and have their secrets repo on a topic branch, but I’m not sure how common that case is.

I envisioned the same case when I was originally building this, and the number of times anybody has done that (as far as I know) is zero 😅. 

So while going back to the right branch is great, I don't think it's something we need to spend _too_ much time on! 😄

markpar:
>So while going back to the right branch is great, I don't think it's something we need to spend too much time on! 😄
I think it'll be just an extra line of code or two, so I'll give it a go!

jtreanor:
Thanks for the input @jkmassel!

My concern with going back to the right branch is that after running `configure_apply`, files in `~/.mobile-secrets` may not be what you expect. e.g. for how WPiOS works right now. However, given that we also have the validation step, I think going back to the branch is probably the best approach.

We can consider changing the cases where the secrets are referenced directly so that this inconsistency can't happen.

markpar:
After adding code to switch back to the original branch and testing, it’s working as expected with the exception of one case.  If the secrets repo is already in a detached head state, the command to change back will fail attempting to execute:

`cd /Users/mark/.mobile-secrets && git checkout (HEAD detached at a6693cc)`

This _seems_ like it might be a situation that shouldn’t occur on a build machine, so maybe that’s OK seeing as the 90% case will use the most recent commit in `master`?  I would guesstimate that of the remaining 10%, not a lot of those would have a secrets repo that was already in a detached head state.  🙂 

>We can consider changing the cases where the secrets are referenced directly so that this inconsistency can't happen.

That’s an interesting point, @jtreanor!  I like the idea of preventing the inconsistency from occurring in the first place!  I’d love to chat further about this offline and learn more about how this part of the system works!  👍

markpar:
Here’s what the output looks like when switching back:

```[14:50:52]: -----------------------------
[14:50:52]: --- Step: configure_apply ---
[14:50:52]: -----------------------------
[14:50:52]: $ cd /Users/mark/.mobile-secrets && git fetch && git checkout b60ba8960378a5ce14a416654c1c80cd60af7403
[14:50:52]: ▸ Note: checking out 'b60ba8960378a5ce14a416654c1c80cd60af7403'.
[14:50:52]: ▸ You are in 'detached HEAD' state. You can look around, make experimental
[14:50:52]: ▸ changes and commit them, and you can discard any commits you make in this
[14:50:52]: ▸ state without impacting any branches by performing another checkout.
[14:50:52]: ▸ If you want to create a new branch to retain commits you create, you may
[14:50:52]: ▸ do so (now or later) by using -b with the checkout command again. Example:
[14:50:52]: ▸ git checkout -b <new-branch-name>
[14:50:52]: ▸ HEAD is now at b60ba89 Create rule to post merge instructions
[14:50:52]: $ cd /Users/mark/.mobile-secrets && git checkout fooboo
[14:50:52]: ▸ Previous HEAD position was b60ba89 Create rule to post merge instructions
[14:50:52]: ▸ Switched to branch 'fooboo'
[14:50:52]: Applied configuration
[14:50:52]: Result: true
```

jtreanor:
Thanks for the updates @markpar. I have tested your changes and they work well.

> If the secrets repo is already in a detached head state, the command to change back will fail attempting to execute

This is a good point and I have confirmed this is the case. Can we handle the error? Our options are either to gracefully handle a detached state and return to that ref instead of the branch or give a helpful error so the user knows how to resolve the problem.

I would be happy with either, although I like the idea of avoiding an error where possible.

markpar:
> I like the idea of avoiding an error where possible.

Definitely agree.  🙂   From my quick investigations, I think we will be able to handle this case and return to the prior ref.  👍

markpar:
Here’s the output when handling the detached HEAD case.  

```[10:00:34]: -----------------------------
[10:00:34]: --- Step: configure_apply ---
[10:00:34]: -----------------------------
[10:00:34]: $ cd /Users/mark/.mobile-secrets && git fetch && git checkout 8e1d419707e1b1edd3d0f819bb0542f4e3d97baa
[10:00:34]: ▸ Previous HEAD position was a6693cc try mptest
[10:00:34]: ▸ HEAD is now at 8e1d419 blah.txt
[10:00:34]: $ cd /Users/mark/.mobile-secrets && git checkout a6693cc
[10:00:34]: ▸ Previous HEAD position was 8e1d419 blah.txt
[10:00:34]: ▸ HEAD is now at a6693cc try mptest
[10:00:34]: Applied configuration
[10:00:34]: Result: true
```

WDYT, @jtreanor?

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
```diff
@@ -22,6 +33,12 @@ def self.run(params = {})
             end
         }
 
+        ### Restore secrets repo to original branch.  If it was originally in a 
+        ### detached HEAD state, we need to extract the hash from the branch name.
+        original_repo_branch = original_repo_branch.sub("(HEAD detached at ", "").sub(")", "")
```
jtreanor: This seems a little brittle. Can we use `repo_hash` if the branch isn't present? Perhaps you could add logic to `Fastlane::Helper::ConfigureHelper` so that `repo_branch_name` is set to nil if its not a branch.

markpar: > Perhaps you could add logic to Fastlane::Helper::ConfigureHelper so that repo_branch_name is set to nil if its not a branch.

That’s a great suggestion!  We could, in the helper, check to see if the branch name begins with `(HEAD detached at` and, if so, return `nil`.

jtreanor: > check to see if the branch name begins with `(HEAD detached at` 
Yes, or even better, check that its on a branch. Something like [this](https://stackoverflow.com/questions/52221558/programmatically-check-if-head-is-detached) or [this](https://stackoverflow.com/questions/5167957/is-there-a-better-way-to-find-out-if-a-local-git-branch-exists) perhaps.

markpar: That is super cool.  I haven’t used `git rev-parse` much and its usefulness is blowing my mind.  😄  I’ll look at those posts - thanks!

markpar:
Moved the check into `repo_branch_name` (in configure_helper.rb) and simplified it a bit.  My quick tests succeeded, and I’m verifying the tool commands now.

markpar:
The `download` action needed a nil check, and I consolidated uses of `~/.mobile-secrets` while I was there (let me know if that’s not OK!).   After that, the `download`, `validate`, and `apply` actions all look good to me, in both the valid branch and detached HEAD cases.  

Now that `repo_branch_name` can be nil, `update` action didn’t do too well in the detached HEAD state.  @jtreanor , is this one we would want to support?  If it’s a common use case, I think we could support it.  For now, I detect that state, print an error message with instructions to check out a branch, and then exit.

PR #60, opened 2019-07-17
Make RMagick optional

loremattei:
This PR makes RMagick an optional dependency for the release toolkit. 

The reason for this is that the RMagick pod seems to be quite troublesome to install for a lot of people (and on CI) and, since it's required only by Promo Screenshots so far, it makes sense to require it only when that feature is used. 

There's a companion PR in WPiOS to test the changes here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12133

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
@@ -1,5 +1,12 @@
 require 'tmpdir'
-require 'RMagick'
+begin
+  skip_magick = false
+  require 'RMagick'
+rescue LoadError
+  puts "Please, install RMagick if you aim to generate the PromoScreenshots."
+  puts "\'bundle install --with screenshots\' should do it if your project is configured for PromoScreenshots."
+  skip_magick = true
+end
 require 'json'
 require 'tempfile'
 require 'optparse'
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -1,5 +1,12 @@
 require 'tmpdir'
-require 'RMagick'
+begin
+  skip_magick = false
+  require 'RMagick'
+rescue LoadError
+  puts "Please, install RMagick if you aim to generate the PromoScreenshots."
+  puts "\'bundle install --with screenshots\' should do it if your project is configured for PromoScreenshots."
```
jtreanor: Is this error printed whenever any `fastlane` task is run?

loremattei: Yes, it is. It's printed every time the plugin is loaded, which means every time `fastlane` is run. 
I took a  _very quick_ look at fastlane code and it doesn't seem possible to understand what action has been called at this time. 

Though, now that I re-think at this, we could try to only set the flag here, with no error message, and show the error message  when the action is called.
Wdyt?

jtreanor: Yeah, I think it would be nice if the error was only when the promo screenshots were run. It should be easy to move this to `initialize` using the `skip_magick` variable.

loremattei: Changes here: https://github.com/wordpress-mobile/release-toolkit/pull/63

PR #66, opened 2019-07-25
 Don't throw an error if there are no new translations

jtreanor:
When finalising the WCiOS release today, `fastlane finalise_release` crash with this error:

```
bundler: failed to load command: fastlane (/Users/james/src/woocommerce-ios/vendor/bundle/ruby/2.3.0/bin/fastlane)
FastlaneCore::Interface::FastlaneShellError: [!] Exit status of command 'git commit -m "Updates translation"' was 1 instead of 0.
```

This is because there were no new strings for this release. We should still proceed in this case.


jtreanor:
Ah, I see you already did this 😄

PR #67, opened 2019-07-26
Use configure_apply instead of validate

loremattei:
This PR updates the iOS Build Preflight action to use the `configure_apply` instead of `configure_validate` during the initial check.
This way, the commit referenced in the `.configure` file is checked out in the mobile secrets repository before starting the build.

This PR is related to https://github.com/wordpress-mobile/WordPress-iOS/pull/12207 and can be tested from there.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_build_preflight.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_build_preflight.rb
@@ -4,7 +4,7 @@ module Fastlane
       def self.run(params)
 
         # Validate mobile configuration secrets
-        other_action.configure_validate
+        other_action.configure_apply
 
         Action.sh("cd .. && rm -rf ~/Library/Developer/Xcode/DerivedData")
 


loremattei:
As far as I understand, this is the behaviour we want after the recent changes that have been made after @jtreanor's request. The idea is that the tool should help to make that we're using the validated configuration, i.e. the one in `.configure`.
But double checking the request and the development history of the changes surely helps to make sure we're doing the right changes.

jkmassel:
@jtreanor – could you weigh in on the changes above? :)

jtreanor:
I think this change is a good idea.

> If we plan to still use configure_validate, we should remove the check for whether the mobile secrets hash matches the .configure file – it never will because configure_apply rolls back the changes to always leave the mobile secrets repo on master. That said, just force-applying the changes on every release means we could potentially just get rid of that action altogether!

This is only the case because some of the repos (mainly iOS I think) directly reference files in `~/.mobile-secrets`. I think we should change that so that anything used from the secrets repo is copied and defined by `.configure`. This will be necessary for anything that CI uses anyway but it removes the moving target of the repo from the equation.

So yes, I think we should make this change 😄

loremattei:
Thank you @jkmassel and @jtreanor!

PR #68, opened 2019-08-01
Add model for secrets configuration

jtreanor:
I'm starting to work on some changes to make the configure tool more powerful. As a first step I have extracted the logic for the configuration into a model of its own with tests. There are no functional changes here.

I have also added config to run the tests on CI.

To Test:

- See that CI passes.
- Point a repo with a `.configure` file to this branch (`gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', branch:'configure-object' 
`).
- Run `bundle install`.
- Run `bundle exec fastlane run configure_apply` and `bundle exec fastlane run configure_validate`. It should work exactly as before.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,20 @@
 version: 2.1
 
 orbs:
-  ios: wordpress-mobile/ios@0.0.29
+  bundle-install: toshimaru/bundle-install@0.3.1
 
-jobs: # a collection of steps
-  build: # runs not using Workflows must have a `build` job as entry point
-    executor:
-      name: ios/default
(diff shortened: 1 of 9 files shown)


jtreanor:
Thanks @jkmassel. I have updated the executor to use macOS. It should be good to go now 😄

jtreanor:
Thanks @jkmassel! I've updated to use a macOS executor as you suggested.

I also made some more improvements by introducing a `Configuration::FileReference` model with tests. It will make the upcoming changes safer.

jtreanor:
Thanks @loremattei!

PR #72, opened 2019-08-16
Refactor actions folder

loremattei:
This PR moves some actions from the `actions` folder to the proper subfolder in order to give a better organization to the repository. 

### To Test
1. Update your `Pluginfile` to point to this branch.
2. Run `bundle install`.
3. Run `bundle exec fastlane actions` and verify that the moved actions are still listed. 
4. Try to run some actions and verify that they are executed without errors. For example: `bundle exec fastlane get_pullrequests_list start_tag:13.0`.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit.rb
@@ -4,7 +4,7 @@ module Fastlane
   module Wpmreleasetoolkit
     # Return all .rb files inside the "actions" and "helper" directory
     def self.all_classes
-      Dir[File.expand_path('**/{actions,helper,actions/common,actions/configure,actions/android,helper/android,actions/ios,helper/ios}/*.rb', File.dirname(__FILE__))]
+      Dir[File.expand_path('**/{actions,helper}/**/*.rb', File.dirname(__FILE__))]
     end
   end
 end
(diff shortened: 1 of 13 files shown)


lib/fastlane/plugin/wpmreleasetoolkit.rb
```diff
@@ -4,7 +4,7 @@ module Fastlane
   module Wpmreleasetoolkit
     # Return all .rb files inside the "actions" and "helper" directory
     def self.all_classes
-      Dir[File.expand_path('**/{actions,helper,actions/configure,actions/android,helper/android,actions/ios,helper/ios}/*.rb', File.dirname(__FILE__))]
+      Dir[File.expand_path('**/{actions,helper,actions/common,actions/configure,actions/android,helper/android,actions/ios,helper/ios}/*.rb', File.dirname(__FILE__))]
```
jtreanor: Can this just be?

```
Dir[File.expand_path('**/{actions,helper}/**/*.rb', File.dirname(__FILE__))]
```

I don't think we need to list every subfolder.

loremattei: 👍

loremattei:
Oh, you're right. 
Though, I checked with the current version and it actually listens only the ones in the `actions` folder, which is wrong. 

Screenshot 2019-08-16 at 16 28 24

So, it seems to be a Fastlane issue. I'll bring this up with the people there.

jtreanor:
👍 In the meantime, I think it's fine to continue with this change.

PR #102, opened 2020-01-09
Add configure_apply preflight

jkmassel:
Checks to ensure that the decryption key is present before running `configure_apply` – this prevents inscrutable errors from popping up.


jkmassel:
Thanks for checking on that!

PR #108, opened 2020-03-18
Feature/builds on ci

loremattei:
This PR introduces the `ios_validate_ci_build.rb` action which:
- verifies that the build runs on the tagged commit
- checks if the we're building a final tag or a beta tag. 

The PR also updates `ios_build_prechecks.rb` to make it work on CI (due to `fastlane match`, the repository is dirty as a new helper file is added. Since a clean VM is spinned-up on every build and we do a fresh checkout, skipping the test should be fine).

You can see it working in https://github.com/wordpress-mobile/WordPress-iOS/pull/13658.


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_build_prechecks.rb
```diff
@@ -18,7 +18,7 @@ def self.run(params)
         end
 
         # Check local repo status
-        other_action.ensure_git_status_clean()
+        other_action.ensure_git_status_clean() unless params[:run_on_ci]
```
jkmassel: Rather than declaring an action parameter for this, could we use detect it using https://docs.fastlane.tools/actions/is_ci/?

loremattei: This makes sense!
Updated!

PR #113, opened 2020-04-09
Fix iOS localization

loremattei:
This PR fixes two iOS localization issues: 
- it makes sure that the latest version of the pods is installed before localization.
- it makes sure that `localizable.strings` is committed when there are changes. 

This can be tested in https://github.com/woocommerce/woocommerce-ios/pull/2125


loremattei:
Thank you @oguzkocer!
That's fixed now!

PR #116, opened 2020-04-29
Configure: Skip copying unencrypted files on CI

loremattei:
This PR adds:

-  a check to avoid trying to locate unencrypted files on CI. This allows `configure_apply` to work properly even when the configuration includes files for functions that are not available on CI. 
Files that are not encrypted are not available on CI: they are only available on local environments through the mobile-secrets repository, so missing files should not trigger an error on CI if they are flagged with `encrypt: false`.

- `~` -> home folder expansion for target paths. 

The changes can be tested as a part of https://github.com/wordpress-mobile/WordPress-Android/pull/11773


lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
```diff
@@ -10,7 +10,7 @@ def initialize(params = {})
       end
   
       def source_contents
-        return File.read(secrets_repository_file_path) unless self.encrypt
+        return File.read(secrets_repository_file_path) unless (self.encrypt || ENV.key?('CIRCLECI'))
```
mokagio, line 13: Since this doesn't seem specific to CircleCI, but any CI provider, what do you think about using `CI` instead? It's one of the [CircleCI built-in env vars](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables), and is supported by [GitHub](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables), [Travis CI](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables), and [Buildkite](https://buildkite.com/docs/pipelines/environment-variables) too.

```suggestion
        return File.read(secrets_repository_file_path) unless (self.encrypt || ENV.key?('CI'))
```

spec/spec_helper.rb
```diff
@@ -25,3 +25,21 @@ module SpecHelper
 RSpec.configure do |config|
   config.filter_run_when_matching :focus
 end
+
+def set_circle_env(define_ci)
```
mokagio, line 29: If my suggestion above to use `CI` was implemented, this could become.

```suggestion
def set_ci_env(define_ci)
```

spec/file_reference_spec.rb
```diff
@@ -54,8 +54,19 @@
...
-        allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
-        expect(subject.source_contents).to eq('source contents')
+        set_circle_env(false) do
+          allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
+          expect(subject.source_contents).to eq('source contents')
+        end
+      end
+    end
+
+    describe '#source_contents_on_ci' do
```
mokagio, line 64: I'm sure what's the naming convention for RSpec, but since there's no `source_contents_on_ci` method, it might be misleading to use that name for this `describe` block. What do you think of:

```suggestion
    describe '#source_contents on ci' do
```

lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
```diff
@@ -52,6 +52,7 @@ def encrypted_file_path
       end
   
       def destination_file_path
+        return File.expand_path(self.destination) if self.destination.start_with?('~')
```
mokagio, line 55: I wonder, is it worth adding a test for this behavior?

oguzkocer:
I am skipping this review since @mokagio already reviewed it. Please let me know if you need a second set of eyes.

lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
loremattei: I chose `CIRCLECI` because it's the one used by Fastlane::is:ci ([#](https://github.com/fastlane/fastlane/blob/ce06c2156628637dbd74f43f87e5bf37c1a42111/fastlane_core/lib/fastlane_core/helper.rb#L73)) which we use in other places so that we have a coherent behaviour. 
I actually tried to use that function here, but it doesn't work (it always returns false). 
I dug a bit into the Fastlane code and it looks like the issue is related to some very in depth mechanisms of Fastlane. I'm adding a note to the code to revisit this with future versions of Fastlane.

loremattei: I'll also open a PR for this in Fastlane.

spec/file_reference_spec.rb
loremattei: 👍

loremattei:
Thanks for the review @mokagio! I addressed most of the notes, but I kept the original version for the variable to check for the reasons above. 
I'll try to understand the issue better on the Fastlane side and revisit this after that.

mokagio:
@loremattei, gotcha. That's interesting 🤔 Thanks for the extra context 👍

PR #126, opened 2020-05-30
Update Fastlane to 2.148.1

mokagio:
While working on the Simplenote screenshots, I had made a dedicated lane to upload them to the App Store. It worked fine for at the time of 4.18, but then when I tried yesterday for 4.19 it failed.

```
Error loading plugin 'fastlane-plugin-wpmreleasetoolkit':
  uninitialized constant Faraday::Error::ClientError

Did you mean?  Faraday::ClientError
```

As far as I could trace back with Google searches, without looking at the changelogs of the gems, it looks like there was some incompatibility between `octokit` and `faraday` after certain updates. There are more details [in this issue](https://github.com/github/pages-gem/issues/665), sort of.

I updated `fastlane` to its latest version here, and then used the branch that was giving me trouble to verify the fix.

## To test

1. Download a set of of pre-generated Simplenote iOS screenshots [here](https://github.com/Automattic/simplenote-ios/suites/731298482/artifacts/7342385). It's a `.zip` file, which you'll need to extract.
1. Checkout [this branch on Simplenote iOS](https://github.com/Automattic/simplenote-ios/pull/765)
1. Run `bundle install && bundle exec fastlane create_promo_screenshots source:/path/to/your/extracted/screenshots`

You should get an error:

```
[✔] 🚀
[22:40:18]: Error loading plugin 'fastlane-plugin-wpmreleasetoolkit': uninitialized constant Faraday::Error::ClientError
Did you mean?  Faraday::ClientError
+-----------------------------------+---------+----------------------------------------+
|                                     Used plugins                                     |
+-----------------------------------+---------+----------------------------------------+
| Plugin                            | Version | Action                                 |
+-----------------------------------+---------+----------------------------------------+
| fastlane-plugin-wpmreleasetoolkit | 0.9.2   | No actions found                       |
```

Next, update the `Pluginfile` to point to this branch, `use-faraday-1.x`:

```diff
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.9.2'
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', branch: 'use-faraday-1.x'
```

Try again, `bundle install && bundle exec fastlane create_promo_screenshots source:/path/to/your/extracted/screenshots`, and verify it works.

Note, if you get

```
Git LFS not enabled – Unable to generate promo screenshots. Run `git lfs install && git lfs fetch && git lfs pull` to fix this.
```

that's actually good as it shows that some of the code that wasn't working before actually is 😄 . Run the `git lfs` commands and it should work properly.


.gitignore
```diff
@@ -1,6 +1,13 @@
 .DS_Store
 /tmp
 
+# Ruby / Bundler
+.bundle/
```
mokagio, line 5: Running `bundle install --with development` created this. Ignoring it [seems to be the best practice](https://devcenter.heroku.com/articles/bundler-configuration) and it's what we've been doing [in other repos too](https://github.com/Automattic/simplenote-ios/blob/1bf42d514cab4aa127d8cbb70f25f5fef0a34d77/.gitignore#L38).

.gitignore
```diff
@@ -1,6 +1,13 @@
...
+
+# Fastlane
+fastlane/README.md
+fastlane/report.xml
```
mokagio, line 9: Fastlane generates these at runtime, so ignoring them avoids noise in the diff.

.gitignore
mokagio, line 5: Running `bundle install --with development` created this. Ignoring it [seems to be the best practice](https://devcenter.heroku.com/articles/bundler-configuration) and it's what we've been doing [in other repos too](https://github.com/Automattic/simplenote-ios/blob/1bf42d514cab4aa127d8cbb70f25f5fef0a34d77/.gitignore#L38).

ext/drawText/.gitignore
```diff
@@ -0,0 +1,7 @@
+Build/
+Logs
+Makefile
+ModuleCache.noindex/
+drawText.bundle
+info.plist
+mkmf.log
```
mokagio, line 7: All these were generated by the `compile` Rake task. Because they're runtime generated, I think it's safer to ignore them.

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -35,7 +35,7 @@ Gem::Specification.new do |spec|
   # spec.add_dependency 'your-dependency', '~> 1.0.0'
   spec.add_dependency 'diffy', '~> 3.3'
   spec.add_dependency 'nokogiri', '>= 1.10.4'
-  spec.add_dependency 'octokit', '~> 4.13'
+  spec.add_dependency 'octokit', '~> 4.18'
```
mokagio, line 38: Need to bump this in the `.gemspec` to ensure its resolved to at least 4.18, which is the version that fixes the issue with faraday 1.0.0, which comes from Fastlane.

jkmassel: To avoid this sort of issue in the future, WDYT about relaxing this dependency to just be `~> 4` which would take any compatible version in the 4.xx series that's compatible with the other gems? I don't think we require any specific functionality from `4.18` – moreso just the compatibility?

mokagio:
@jkmassel I tried relaxing the condition, which made sense to me, but the plugin failed to load, like before.

I think the reason has to do with the fact that the issue with Faraday was fixed in 4.18, but if we relax the requirement, we might not get that version, like it happens on that Simplenote branch.

Screen Shot 2020-06-09 at 12 38 20 pm

PR #131, opened 2020-06-11
Add Danger to check version consistency

mokagio:
Here's a mini hack week project. Setting up Danger to help us maintaining the gem (_or maybe just being pedantic_).

I added a check to make sure the version in the `Gemfile.lock` is in line with the one in `version.rb`.

The two version on being aligned is not a problem for the gem distribution, but will result in an extra edit to checkin on the `Gemfile.lock` whenever someone uses the gem for dev, e.g. running `bundle install && bundle exec rspec`.

I chose to use GitHub Action instead of CircleCI because... that's what I've been doing as part of the Peril migration 🤔 . I guess one could [use CircleCI](https://danger.systems/guides/getting_started.html#setting-up-danger-to-run-on-your-ci), but it requires a bit extra setup with the token... 

---

This PR is currently failing to show how the error looks like. I you give me the green light, I'll commit the fix and merge.

Screen Shot 2020-06-12 at 1 02 06 pm


Dangerfile
```diff
@@ -0,0 +1,25 @@
...
+end
+
+def gemfile_lock_version
+  gemfile_lock = File.read('./Gemfile.lock')
+  gemfile_lock.scan(/fastlane-plugin-wpmreleasetoolkit \((\d+.\d+.\d+)\)/).last.first
+end
+
+# Before checking the version, get rid of any change that `bundle install`
+# might have done.
+`git checkout Gemfile.lock &> /dev/null`
```
mokagio, line 15: It's annoying to have to do this, but if `bundle install` run in CI, it will have updated the version in the lockfile, which is exactly what we're trying to figure out.

Dangerfile
```diff
@@ -0,0 +1,25 @@
...
+`git checkout lib/fastlane/plugin/wpmreleasetoolkit/version.rb &> /dev/null`
```
mokagio, line 16: This is not needed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: actions/checkout@v2
+
+      # This action reads the Ruby version from .ruby-version
+      - uses: ruby/setup-ruby@v1
+
+      - uses: actions/cache@v2
+        with:
+          # This is duplicated from the above BUNDLE_PATH, but I don't know how
+          # to read the env var here
+          path: vendor/bundle
```
mokagio, line 22: Need to fix this before going out of draft

Dangerfile
```diff
@@ -0,0 +1,24 @@
...
+  gemfile_lock.scan(/fastlane-plugin-wpmreleasetoolkit \((\d+.\d+.\d+)\)/).last.first
+end
+
+# Before checking the version, get rid of any change that `bundle install`
+# might have done.
+`git checkout Gemfile.lock &> /dev/null`
+
+if version.to_s != gemfile_lock_version.to_s
+  message = %Q{
+The version in the `Gemfile.lock` (`#{gemfile_lock_version}`) doesn't match the one in `version.rb` (`#{version}`).
```
mokagio, line 19: I realize it's a bit wasteful to have those functions run twice (if the numbers don't match), but in the context of this file, I think the better readability trumps the runtime duplication.

mokagio:
Notice that after a0e8f59 there's no more Danger failure and the checks are all ✅

PR #132, opened 2020-06-15
Add Rubocop

mokagio:
_I'm not sure about how to approach this, because there are so many changes that it does. I guess I could split it into multiple PRs with a TODO config file, or... just take the plunge and trust that the only tests are automated and we have enough tests around them._


mokagio:
Closing. @AliSoftware is doing a much more comprehensive job at this in #207.

PR #133, opened 2020-06-16
Translation reviewers builds

loremattei:
This PR adds two actions to merge the pending translations into the main resource file in order to be able to build special app versions for the translation reviewers. 

This can be tested as a part of https://github.com/wordpress-mobile/WordPress-Android/pull/12208 and https://github.com/wordpress-mobile/WordPress-iOS/pull/14282.


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_merge_translators_strings.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      def self.details
+        # Optional:
+        "Merges waiting and fuzzy strings into the main file for translators"
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :strings_folder,
+                                   env_name: "MTS_STRING_FOLDER",
```
mokagio, line 89: I don't know if there's a special convention to decide which env name to use. If that's the case, feel free to discard (and can you point me to where it's defined?)

Anyways, since the action is called `AndroidMergeTranslatorsStrings`, I'd add an `A` to the name here.

```suggestion
                                   env_name: "AMTS_STRING_FOLDER",
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_translators_strings.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      def self.details
+        # Optional:
+        "Merges waiting and fuzzy strings into the main file for translators"
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :strings_folder,
+                                   env_name: "MTS_STRING_FOLDER",
```
mokagio, line 81: Same comment as above regarding the Android action.
```suggestion
                                   env_name: "IMTS_STRING_FOLDER",
```

I doubt we'd ever run in a case where this script runs for both platforms, but just in case, it might be safer to not reuse the env var name, since it's supposed to have different values between Android and iOS?

loremattei:
Good point @mokagio!
I added some tests to cover for the cases I encountered during the development. I'm pretty sure there are many more, but it's hard to figure them out right now without knowing how GlotPress works in depth. 
I don't think it's worth to go more in depth right now, as these functions are only used by translators, so it's easier to push a fix if we'll find issues. 

wdyt?

mokagio:
I think... :shipit:, and thanks for adding some tests.

I agree, it's hard to have good tests because they depend on GlotPress and since this is an internal tool, it'll be fast to ship a fix.

PR #140, opened 2020-06-29
Fix: small typos in docs

AliSoftware:
Just fixing some typos (and nitpicking spacing inconsistencies while at it) in the `screenshot-compositor.md` documentation

### DIFF
--- a/docs/screenshot-compositor.md
+++ b/docs/screenshot-compositor.md
@@ -27,7 +27,7 @@ This argument is a path to the directory containing any metadata, such as locali
 
 > Note: The tool will automatically check these directories, and will be default use any locales found in both directories.
 
-**`metadata_folder`**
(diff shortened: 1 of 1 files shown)


AliSoftware:
Mmmh I guess `Danger` is not able to run on forks (given that the error seems to say that [she](https://danger.systems/guides/faq.html#danger-keeps-getting-referred-to-as-quot-her-she-quot-what-gives) didn't find the commit sha1 while it's the correct one… but it's only on my fork).

I checked the `Dangerfile` and it seems to only check if the plugin's `version.rb` and the version in `Gemfile.lock` are matching. Since I didn't change any of those, that shouldn't be an issue (but depending on how your configured your GH repo that might block the merge of the PR to non-admins though)

PR #142, opened 2020-07-02
Fix: typos in fastlane's close_milestone_action description

AliSoftware:
Found some copy/paste in description and a typo in ENV var.

Not sure if you want to actually fix the typo in ENV var because that would break the scripts of anyone depending on that env var though.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
@@ -19,7 +19,7 @@ module Fastlane
       end
 
       def self.description
-        "Creates a new milestone for the project"
+        "Closes an existing milestone in the project"
       end
 
       def self.authors
@@ -32,7 +32,7 @@ module Fastlane
 
       def self.details
         # Optional:
-        "Creates a new milestone for the project"
+        "Closes an existing milestone in the project"
       end
 
       def self.available_options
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
```diff
@@ -43,7 +43,7 @@ def self.available_options
                                    optional: false,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :milestone,
-                                        env_name: "GHHELPER_MILESTORE",
+                                        env_name: "GHHELPER_MILESTONE",
```
AliSoftware, line 46: Careful accepting this fix as it would need anyone who is depending on this ENV var to update their development environment accordingly

AliSoftware:
Danger failure seems to only be due to the fact that his is a fork and it can't find the commit sha1 on the main repo, only on my fork. (Similar to the issue in https://github.com/wordpress-mobile/release-toolkit/pull/140)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
jkmassel: @loremattei – I think this one is set in every usage anyway during the `finalize_release` lanes, but WDYT?

loremattei: yep! It is. I think we're ok with merging this.

PR #152, opened 2020-07-24
Tracking multiple localizations

jleandroperez:
As per [this Simplenote macOS PR](https://github.com/Automattic/simplenote-macos/pull/605) we'll need to track `Localizable.strings` + `MainMenu.strings`.

In this PR we're extending the `localize_project` and `update_metadata` script(s), so that multiple `.strings` files can be tracked / pushed.

@loremattei Sir, may I bug you with this one?

/cc @mokagio 

Thanks in advance!!


jleandroperez:
Okay, not really sure why I'm getting CI errors. All of them would appear to be unrelated to this particular change.

(Any hings would be super welcomed!!!). Thank you!!

jkmassel:
I disabled tests on the compositor code for now – they're effectively snapshot tests, and they're failing for different reasons in CI and local. I'll investigate, but IMHO it's not super high priority at the moment

jleandroperez:
Thank you @jkmassel !!!

PR #161, opened 2020-08-25
Auto-increment build number

loremattei:
The version scheme on Simplenote MacOS has been historically different from the one on the other apps where `CFBundleVersion` is set with a stripped string without dots.  

In order to make the app compatible with our standard tooling, @mokagio added a `BUILD_NUMBER` field in `Version.public.xcconfig` and switched `CFBundleVersion` to it.

This PR updates the iOSVersionHelper to bump `BUILD_NUMBER` when it's present in the `xcconfig` file.

This can be tested as a part of https://github.com/Automattic/simplenote-macos/pull/657

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source('https://rubygems.org')
 gemspec
 gem 'codecov', :require => false, :group => :test
 gem "danger", "~> 8.0"
-gem "bigdecimal", "~> 1.4"
 
 plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
 eval_gemfile(plugins_path) if File.exist?(plugins_path)
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source('https://rubygems.org')
 gemspec
 gem 'codecov', :require => false, :group => :test
 gem "danger", "~> 8.0"
+gem "bigdecimal", "~> 1.4"
 
 plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -3,6 +3,7 @@ source('https://rubygems.org')
 gemspec
 gem 'codecov', :require => false, :group => :test
 gem "danger", "~> 8.0"
+gem "bigdecimal", "~> 1.4"
```
mokagio: Constraining `bigdecimal` in the `Gemfile` will work for the repo, but downstream dependencies won't inherit this constraint, as can be seen in [the sister PR, which uses `bigdecimal 2.0.0`](https://github.com/Automattic/simplenote-macos/pull/657/files#diff-e79a60dc6b85309ae70a6ea8261eaf95R48).

This can be problematic, as per my [comment](https://github.com/Automattic/simplenote-macos/pull/657#discussion_r477028135) in the sister PR.

I pushed [a commit on another branch that moves this to the `.gemspec`](https://github.com/wordpress-mobile/release-toolkit/commit/f1605543b9d92bfe2a02ef4151a1ca77cb5414f4), feel free to cherry pick it.

If `simplenote-macos` uses that commit, the plugin gets loaded successfully.

mokagio (changes requested):
Thank you for getting onto this.

Gemfile
loremattei: Good point @mokagio!
For some reasons, on my local environment the error went back to be a warning with my configuration, but this your solution seems better anyway. 
I've pushed it and updated the sister PR on Simplenote-MacOS

mokagio (approved):
Ship it.

PR #162, opened 2020-08-28
Fix/android strings import attributes

loremattei:
This PR fixes an issue with Android strings import from libraries where the attributes are not imported to the main `strings.xml`.

This PR updates `android_localize_helper` so that: 
- when a string is imported, all the attributes are brought in. 
- on updates, all the attributes are considered. 
- if the target string has the `tools:ignore` attribute, only the string content is updated and the attributes are not.

This PR can be tested as a part of https://github.com/wordpress-mobile/WordPress-Android/pull/12825.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
@@ -41,22 +41,32 @@ module Fastlane
         return :skipped if skip_string_by_exclusion_list(library, string_name)
       
         # Search for the string in the main file
+        result = :added
         main_strings.xpath('//string').each do | this_string | 
           if (this_string.attr("name") == string_name) then
             # Skip if the string has the content_override tag
             return :skipped if skip_string_by_tag(this_string)
             
-            # Update if needed
-            (if (this_string.content == string_content) then return :found else this_string.content = string_content ; return :updated end)   
+            # If nodes are equivalent, skip
+            return :found if (string_line =~ this_string) 
+            
+            # The string needs an update
+            result = :updated
+            if (this_string.attr("tools:ignore").nil?) 
+              # It can be updated, so remove the current one and move ahead
+              this_string.remove 
+              break 
+            else
+              # It has the tools:ignore flag, so update the content without touching the other attributes
+              this_string.content = string_content
+              return result
+            end 
           end
         end
       
-        # String not found and not in the exclusion list: add to the main file
-        new_element = Nokogiri::XML::Node.new "string", main_strings
-        new_element['name'] = string_name
-        new_element.content = string_content
-        main_strings.xpath('//string').last().add_next_sibling("\n#{" " * 4}#{new_element.to_xml()}")
-        return :added
+        # String not found, or removed because needing update and not in the exclusion list: add to the main file
+        main_strings.xpath('//string').last().add_next_sibling("\n#{" " * 4}#{string_line.to_xml()}")
+        return result
       end
 
       # Verify a string
@@ -167,5 +177,27 @@ module Fastlane
         end
       end
     end
+  end
+end
+
+class Nokogiri::XML::Node
+  # Return true if this node is content-equivalent to other, false otherwise
+  def =~(other)
+    return true if self == other
+    return false unless name == other.name
+    stype = node_type; otype = other.node_type
+    return false unless stype == otype
+    sa = attributes; oa = other.attributes
+    return false unless sa.length == oa.length
+    sa = sa.sort.map{ |n,a| [n,a.value,a.namespace && a.namespace.href] }
+    oa = oa.sort.map{ |n,a| [n,a.value,a.namespace && a.namespace.href] }
+    return false unless sa == oa
+    skids = children; okids = other.children
+    return false unless skids.length == okids.length
+    return false if stype == TEXT_NODE && (content != other.content)
+    sns = namespace; ons = other.namespace
+    return false if !sns ^ !ons
+    return false if sns && (sns.href != ons.href)
+    skids.to_enum.with_index.all?{ |ski,i| ski =~ okids[i] }
   end
 end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
@@ -65,7 +65,7 @@ module Fastlane
         end
       
         # String not found, or removed because needing update and not in the exclusion list: add to the main file
-        main_strings.xpath('//string').last().add_next_sibling("\n#{" " * 4}#{string_line.to_xml()}")
+        main_strings.xpath('//string').last().add_next_sibling("\n#{" " * 4}#{string_line.to_xml().strip}")
         return result
       end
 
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
```diff
@@ -169,3 +179,25 @@ def self.verify_pr_diff(main, library, main_strings, lib_strings, source_diff)
...
+    return false unless sa == oa
+    skids = children; okids = other.children
+    return false unless skids.length == okids.length
+    return false if stype == TEXT_NODE && (content != other.content)
+    sns = namespace; ons = other.namespace
+    return false if !sns ^ !ons
+    return false if sns && (sns.href != ons.href)
+    skids.to_enum.with_index.all?{ |ski,i| ski =~ okids[i] }
+  end
+end
```
mokagio, line 205: I'm guessing this code comes from [this StackOverflow question & answer](https://stackoverflow.com/questions/7825258/determine-if-two-nokogiri-nodes-are-equivalent?rq=1), right?

The author mentions that at the time of writing Nokogiri didn't have support for canonicalization, but [now it does](https://nokogiri.org/rdoc/Nokogiri/XML/Document#canonicalize-instance_method). 

I would be good to add a link to the question so that if we'll ever have time (and understanding of Nogokiri) we can replace this confusing code with something simpler.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
```diff
@@ -41,22 +41,32 @@ def self.merge_string(main_strings, library, string_line)
...
         end
       
-        # String not found and not in the exclusion list: add to the main file
-        new_element = Nokogiri::XML::Node.new "string", main_strings
-        new_element['name'] = string_name
-        new_element.content = string_content
-        main_strings.xpath('//string').last().add_next_sibling("\n#{" " * 4}#{new_element.to_xml()}")
-        return :added
+        # String not found, or removed because needing update and not in the exclusion list: add to the main file
+        main_strings.xpath('//string').last().add_next_sibling("\n#{" " * 4}#{string_line.to_xml()}")
```
mokagio:
```suggestion
        main_strings.xpath('//string').last().add_next_sibling("\n#{" " * 4}#{string_line.to_xml().strip}")
```

As mentioned [in the sister PR](https://github.com/wordpress-mobile/WordPress-Android/pull/12825#pullrequestreview-478274213) I noticed an extra newline when running this new version of the code.

I added a dumb `.strip` to this line and it was gone.

mokagio (approved):
Only a few non blocking suggestions. The code does what it says on the tin. It's tough to be confident about it with no tests, but also adding them would be a bigger job than this PR.

mokagio:
I noticed something else, actually.

When only editing an existing string, this code moves the new value to the bottom of the `strings.xml`

```diff
diff --git a/WordPress/src/main/res/values/strings.xml b/WordPress/src/main/res/valu
index 1a114e8b57..a07eb0a386 100644
--- a/WordPress/src/main/res/values/strings.xml
+++ b/WordPress/src/main/res/values/strings.xml
@@ -2880,7 +2880,7 @@
     <string name="mlp_create_page">Create page</string>
     <string name="mlp_preview_page">Preview</string>

-    <string name="capture_button_alt">Capture</string>
+
     <string name="flip_button_alt">Flip camera</string>
     <string name="flash_button_alt">Flash</string>
     <string name="stickers_button_alt">Stickers</string>
@@ -3193,5 +3193,7 @@ translators: sample content for "Services" page template -->
     <!-- translators: sample content for "About" page template -->
     <string name="gutenberg_native_what_people_say" tools:ignore="UnusedResources">
     <string name="gutenberg_native_wordpress_media_library" tools:ignore="UnusedRes
+    <string name="capture_button_alt">Capture an image</string>
     <!--/Autogenerated:Gutenberg Native-->
 </resources>
```

whereas the previous code edited it in place

```diff
diff --git a/WordPress/src/main/res/values/strings.xml b/WordPress/src/main/res/values/str
index 1a114e8b57..5e0b59ae1d 100644
--- a/WordPress/src/main/res/values/strings.xml
+++ b/WordPress/src/main/res/values/strings.xml
@@ -2880,7 +2880,7 @@
     <string name="mlp_create_page">Create page</string>
     <string name="mlp_preview_page">Preview</string>

-    <string name="capture_button_alt">Capture</string>
+    <string name="capture_button_alt">Capture an image</string>
     <string name="flip_button_alt">Flip camera</string>
     <string name="flash_button_alt">Flash</string>
     <string name="stickers_button_alt">Stickers</string>
```

I think it would be great if we could keep the edit-in-place behavior, because it makes the diff much easier to read, and would likely simplify merge conflict resolutions if there are edits to the file on different branches.

oguzkocer (approved):
I tested this in WPAndorid PR and it worked as expected except for the extra empty line as mentioned by @mokagio. I don't have any other comments, so 👍 from me.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
loremattei: It's probably code that has been around for a while as I found it in GitHub following a link from a blog post :-), but I agree that it's a good idea to reference and mention it.

loremattei:
> I think it would be great if we could keep the edit-in-place behavior, because it makes the diff much easier to read, and would likely simplify merge conflict resolutions if there are edits to the file on different branches.

Yep, I agree, but it turned out to be not that easy when I looked into it. It would make the code more complex and I decided not to do it in order to make this patch simpler, because this is a fix that we need now as it's affecting our releases, but I don't think we should keep this code around for long time. 

Merging the strings like we do now is a complex task and it would require a lot of work on this code to have it done properly (which is: without moving the nodes around, maintaining the order, being aware of comments/areas, etc.). 

With the new upload policy, I think we don't need this anymore (ref: paaHJt-sd-P2). 
We now have a dedicated file to upload the translations which is not used by gradle at all. Currently, we just copy the main `stringsl.xml`, but I think that we should: 
- keep all the source strings in their libraries.
- remove all the imported strings from the main `strings.xml`.
- during the code freeze, re-create the `strings.xml` every time by merging all the `strings.xml` files. 
This way, we don't have to face string updates or manual edits. The file we upload is just a product of a "compilation" which is used by GlotPress, while the developers continue to use the source ones, which our tools don't edit anymore. 

Does it make sense to you?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android_localize_helper.rb
```diff
@@ -180,6 +180,8 @@ def self.verify_pr_diff(main, library, main_strings, lib_strings, source_diff)
   end
 end
 
+# Source: https://stackoverflow.com/questions/7825258/determine-if-two-nokogiri-nodes-are-equivalent?rq=1
+# There may be better solutions now that Ruby supports canonicalization.
```
mokagio, line 184: Thank you for adding this

mokagio (approved):
>  I decided not to do it in order to make this patch simpler, because this is a fix that we need now as it's affecting our releases, but I don't think we should keep this code around for long time.

That's cool 👍 Fixing issues should always have priority

> The file we upload is just a product of a "compilation" which is used by GlotPress, while the developers continue to use the source ones, which our tools don't edit anymore.
>
> Does it make sense to you?

I like the idea of the creating a merged file only for the GlotPress upload purpose.

Something else that came to mind was that we could sort the strings file alphabetically by their `name` attribute. I think that should help with merge conflicts and the like. But, it might be unnecessary if we move towards using a file that's generated at runtime by the code freeze.

PR #169, opened 2020-10-23
Add Android complete_codefreeze_prechecks action

loremattei:
This PR adds the `complete_codefreeze_prechecks`  action for Android which runs the usual prechecks:
- Make sure a release branch is checked out.
- Confirm the version with the user.
- Make sure git repo is clean.

This can be tested in the related WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/13215


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
```diff
@@ -0,0 +1,67 @@
...
+        def self.authors
+          ["loremattei"]
+        end
+  
+        def self.is_supported?(platform)
+          platform == :android
+        end
+      end
+    end
+  end
```
AliSoftware: nit: add CRLF at EOF
```suggestion
  end

```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
```diff
@@ -0,0 +1,67 @@
...
+          
+          require_relative '../../helper/android/android_version_helper.rb'
+          require_relative '../../helper/android/android_git_helper.rb'
+  
+          UI.user_error!("This is not a release branch. Abort.") unless other_action.git_branch.start_with?("release/")
+  
+          version = Fastlane::Helpers::AndroidVersionHelper::get_public_version
+          message = "Completing code freeze for: #{version}\n"
+          if (!params[:skip_confirm])
+            if (!UI.confirm("#{message}Do you want to continue?"))
```
AliSoftware: totally codestyle nitpick but Ruby doesn't require parentheses around `if` condtions, so feels a bit strange to read here 😅  Also, you should prefer `unless x` to `if !x`.

```suggestion
          unless params[:skip_confirm]
            unless UI.confirm("#{message}Do you want to continue?")
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
```diff
@@ -0,0 +1,67 @@
+module Fastlane
+    module Actions
+      class AndroidCompletecodefreezePrechecksAction < Action
+        def self.run(params)
+          UI.message "Skip confirm: #{params[:skip_confirm]}"
+          
+          require_relative '../../helper/android/android_version_helper.rb'
+          require_relative '../../helper/android/android_git_helper.rb'
+  
+          UI.user_error!("This is not a release branch. Abort.") unless other_action.git_branch.start_with?("release/")
```
AliSoftware, line 10: Shouldn't we even go as far as checking that the branch is actually named `"release/#{version}"` (with `version` variable defined as line 12) – to be sure we're not accidentally on the release branch for another version and that the version provided via code and returned by `get_public_version` matches the target version number?

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
```diff
@@ -0,0 +1,67 @@
...
+          ]
+        end
+  
+        def self.output
+          
+        end
+  
+        def self.return_value
+          
+        end
```
AliSoftware:
```suggestion
        def self.return_type
          :string
        end
        
        def self.return_value
          "The version number that has been prechecked."
        end
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
mokagio: Nice idea.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
mokagio: I've been meaning to add Rubocop on this repo. There are a lot of Ruby specific syntax and style optimization that I didn't know about which Rubocop taught me.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
loremattei: We could... but this will kill any easy way to test future changes 😂. This, and the fact that we haven't had issues with the looser check, are the main reasons why we never added stricter checks.

AliSoftware (approved):

loremattei:
Thank for the review and the suggestions @AliSoftware! 🙇‍♂️

PR #174, opened 2020-11-12
Introduce Localization linting action

AliSoftware:
# Why?

This is part of the task about linting our localized strings files.

_Internal references: [paaHJt-D6-p2] + Step 1 of [paNNhX-am-p2]_

# How?

* This action uses [SwiftGen](https://github.com/SwiftGen/SwiftGen) and a custom template to parse the various `.lproj/*.strings` files in an `input_dir` directory provided to the action.
* The custom template we use generates (in a temporary dir) an output for each locale/lproj folder found, which lists the types (`Int`, `String`, …) of the placeholder parameters (`%d`, `%@`, …) for each key.
* Then the action compares the generated output (containing lines like `{key} => [{type1}, {type2}]` e.g. `"You have %lu hidden WordPress sites." => [Int]`) of each locale to a baseline (by default, the baseline is the output of the `en` locale) to detect which keys expect different `String(format:)` parameters according to the different `%` placeholders found in their translations.

Note: Because SwiftGen already handles the complexity of parsing `String(format:)` format string (e.g. `%3$d`, `%2$.3f` and similar fun placeholder formats), it also supports when different locales uses different parameter orders in the translation and takes that into account.

# Output

For each locale (`.lproj`) where inconsistencies are found, the action prints an error with the `diff`, showing which key differs in the parameters types.

You'll see the parameter types extracted from the english translation (e.g. `[Int, Int]` for a translation that contains twho `%d` or `%li` parameters) then the types extracted for the locale where it found an inconsistency (e.g. `[]` if the translation for that key in that locale forgot to use any placeholders), allowing you to revisit the translation of that key for that locale to fix the inconsistency.

<details><summary>Extract of an example output</summary>

```
Inconsistencies found between 'en' and 'zh-Hant':

@@ -60 +60 @@
-"%li words, %li characters" => [Int,Int]
+"%li words, %li characters" => []
@@ -793,0 +794 @@
+"Great! Could you leave us a nice review?\n\nIt really helps." => []
@@ -846 +847 @@
-"If you remove %@, that user will no longer be able to access this site, but any content that was created by %@ will remain on the site." => [String,String]
+"If you remove %@, that user will no longer be able to access this site, but any content that was created by %@ will remain on the site." => []
@@ -848 +849 @@
-"If you want a site but don't want any of the posts and pages you have now, our support team can delete your posts, pages, media, and comments for you.\n\nThis will keep your site and URL active, but give you a fresh start on your content creation. Just contact us to have your current content cleared out." => []
+"If you want a site but don't want any of the posts and pages you have now, our support team can delete your posts, pages, media, and comments for you.\n\nThis will keep your site and URL active, but give you a fresh start on your content creation. Just cont" => []
@@ -1014 +1015 @@
-"Media uploaded (1 file)" => []
+"Media uploaded (1 file)" => [Int]
@@ -2213 +2214 @@
-"You have %lu hidden WordPress sites." => [Int]
+"You have %lu hidden WordPress sites." => []
```
</details>

# Typical Usage

You will typically add a call to `ios_lint_localizations(input_path: …)` to your `Fastfile` right after you pulled the latest translations from glotpress via the `ios_update_metadata` action.

You should provide it the path (relative to the repo root) to the directory containing all your `.lproj` folders. You can also pass it other options, see `bundle exec fastlane action ios_lint_localizations` for more documentation

For example in https://github.com/wordpress-mobile/WordPress-iOS this would likely look like this:

```diff
  lane :finalize_release do | options |
      ios_finalize_prechecks(options)
-     ios_update_metadata(options) unless ios_current_branch_is_hotfix
-     ios_bump_version_beta() unless ios_current_branch_is_hotfix
+     unless ios_current_branch_is_hotfix
+       ios_update_metadata(options)
+       ios_lint_localizations(input_dir: 'WordPress/Resources')
+       ios_bump_version_beta()
+     end
      ios_final_tag(options)
…
```

# Testing the PR

* Point your `Pluginfile` to the latest version of `fastlane-plugin-wpmreleasetoolkit` (e.g. to this branch until this PR is merged)
* Run `bundle install` to install the latest `wpmreleasetoolkit` which includes this new action
* Create a test lane `lint_l10n` and make it call the `ios_lint_localizations` action, providing it the right `input_path`
* Optionally, you can also add something like `UI.message("Continuing after lint")` at the end of that `lint_l10n` lane, especially to confirm that the message won't be printed (and the lane will abort and stop fastlane) if violations are found, and you will only see the message being printed if there are no violations found.
* Run `bundle exec fastlane lint_l10n`. The first run should install SwiftGen in `vendor/swiftgen/6.4.0`, then tell you if it found any violation in `.strings` files
* Run `bundle exec fastlane lint_l10n` again. The second run should not re-install SwiftGen as it has been already installed, but should output the same violations.

```ruby
lane :lint_l10n do | options |
  ios_lint_localizations(input_dir: 'WordPress/Resources')
  UI.message("Debug: the rest of this lane, like this message, should only execute if there were no violations found.")
end
```

Note: I've mainly tested this with WordPress-iOS in my tests, so it would be nice if testers of this PR could test it with other repos too, to ensure it works everywhere in the various configurations we might have.

# Plan about enabling the action in other repos

The next step will be to make PRs in each repo to call the action (by editing each repo's `Fastfile` to call it like explained above), but those will (1) require this PR to be merged first (2) but also for us to fix the existing violations.

Which is also why I'll only start enabling the call to this action in the various repos (i.e. doing the PRs there) next week, **after** the current release is finalized, so that we have time (before the next code freeze after that) to fix the existing violations.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_store_app_sizes.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_store_app_sizes.rb
@@ -100,6 +100,10 @@ module Fastlane
           # Define the shared values you are going to provide
           
         end
+
+        def self.return_type
+          :hash
+        end
   
         def self.return_value
           "Return a Hash containing the details of download and install app size, for various device models, all that for each requested version of the app"
@@ -111,7 +115,7 @@ module Fastlane
         end
   
         def self.is_supported?(platform)
-          platform == :ios
+          [:ios, :mac].include?(platform)
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
@@ -1,0 +1,105 @@
+module Fastlane
+    module Actions
+      class IosLintLocalizationsAction < Action
+        def self.run(params)
+          UI.message "Linting localizations for parameter placeholders consistency..."
+    
+          require_relative '../../helper/ios/ios_l10n_helper.rb'
+          helper = Fastlane::Helpers::IosL10nHelper.new(
+            install_path: resolve_path(params[:install_path]),
+            version: params[:version]
+          )
+          violations = helper.run(
+            input_dir: resolve_path(params[:input_dir]),
+            base_lang: params[:base_lang],
+          )
+        
+          violations.each do |lang, diff|
+            UI.error "Inconsistencies found between '#{params[:base_lang]}' and '#{lang}':\n\n#{diff}\n"
+          end
+
+          violations
+        end
+
+        def self.repo_root
+          `git rev-parse --show-toplevel`.chomp
+        end
+
+        def self.resolve_path(path)
+          if File.relative?(path)
+            File.join(repo_root, path)
+          else
+            path
+          end
+        end
+    
+        #####################################################
+        # @!group Documentation
+        #####################################################
+    
+        def self.description
+          "Lint the different *.lproj/.strings files for different locales and ensure the parameter placeholders are consistent."
+        end
+    
+        def self.details
+          "Compares the translations against a base language to find potential mismatches for the %s/%d/… parameter placeholders between locales."
+        end
+    
+        def self.available_options
+          [
+            FastlaneCore::ConfigItem.new(
+              key: :install_path,
+              env_name: "FL_IOS_LINT_TRANSLATIONS_INSTALL_PATH",
+              description: "The path where to install SwiftGen tooling needed to run the linting process. If a relative path, should be relative to your repo_root",
+              type: String,
+              optional: true,
+              default_value: "vendor/swiftgen/#{Fastlane::Helpers::IosL10nHelper::SWIFTGEN_VERSION}"
+            ),
+            FastlaneCore::ConfigItem.new(
+              key: :version,
+              env_name: "FL_IOS_LINT_TRANSLATIONS_SWIFTGEN_VERSION",
+              description: "The version of SwiftGen to install and use for linting",
+              type: String,
+              optional: true,
+              default_value: Fastlane::Helpers::IosL10nHelper::SWIFTGEN_VERSION
+            ),
+            FastlaneCore::ConfigItem.new(
+              key: :input_dir,
+              env_name: "FL_IOS_LINT_TRANSLATIONS_INPUT_DIR",
+              description: "The path to the directory containing the .lproj folders to lint, relative to your git repo root",
+              type: String,
+              optional: false
+            ),
+            FastlaneCore::ConfigItem.new(
+              key: :base_lang,
+              env_name: "FL_IOS_LINT_TRANSLATIONS_BASE_LANG",
+              description: "The language that should be used as the base language that every other language will be compared against",
+              type: String,
+              optional: true,
+              default_value: Fastlane::Helpers::IosL10nHelper::DEFAULT_BASE_LANG
+            ),
+          ]
+        end
+    
+        def self.output
+          nil
+        end
+    
+        def self.return_type
+          :hash_of_strings
+        end
+
+        def self.return_value
+          "A hash, keyed by language code, whose values are the diff found for said language"
+        end
+    
+        def self.authors
+          ["AliSoftware"]
+        end
+    
+        def self.is_supported?(platform)
+          platform == :ios
+        end
+      end
+    end
+  end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
@@ -23,15 +23,13 @@ module Fastlane
         end
 
         def self.repo_root
-          `git rev-parse --show-toplevel`.chomp
+          @repo_root || `git rev-parse --show-toplevel`.chomp
         end
 
+        # If the path is relative, makes the path absolute by resolving it relative to the repository's root.
+        # If the path is already absolute, it will not affect it and return it as-is.
         def self.resolve_path(path)
-          if File.relative?(path)
-            File.join(repo_root, path)
-          else
-            path
-          end
+          File.absolute_path(path, repo_root)
         end
     
         #####################################################
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
@@ -17,7 +17,9 @@ module Fastlane
           violations.each do |lang, diff|
             UI.error "Inconsistencies found between '#{params[:base_lang]}' and '#{lang}':\n\n#{diff}\n"
           end
-          UI.abort_with_message!('Inconsistencies found during Localization linting. Aborting.') unless violations.empty?
+          if params[:abort_on_violations] && !violations.empty?
+            UI.abort_with_message!('Inconsistencies found during Localization linting. Aborting.')
+          end
           
           violations
         end
@@ -76,6 +78,14 @@ module Fastlane
               type: String,
               optional: true,
               default_value: Fastlane::Helpers::IosL10nHelper::DEFAULT_BASE_LANG
+            ),
+            FastlaneCore::ConfigItem.new(
+              key: :abort_on_violations,
+              env_name: "FL_IOS_LINT_TRANSLATIONS_ABORT",
+              description: "Should we abort the rest of the lane with a global error if any violations are found?",
+              optional: true,
+              default_value: true,
+              is_string: false # https://docs.fastlane.tools/advanced/actions/#boolean-parameters
             ),
           ]
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
@@ -37,7 +37,7 @@ module Fastlane
         #####################################################
     
         def self.description
-          "Lint the different *.lproj/.strings files for different locales and ensure the parameter placeholders are consistent."
+          "Lint the different *.lproj/.strings files for each locale and ensure the parameter placeholders are consistent."
         end
     
         def self.details
@@ -49,7 +49,7 @@ module Fastlane
             FastlaneCore::ConfigItem.new(
               key: :install_path,
               env_name: "FL_IOS_LINT_TRANSLATIONS_INSTALL_PATH",
-              description: "The path where to install SwiftGen tooling needed to run the linting process. If a relative path, should be relative to your repo_root",
+              description: "The path where to install the SwiftGen tooling needed to run the linting process. If a relative path, should be relative to your repo_root",
               type: String,
               optional: true,
               default_value: "vendor/swiftgen/#{Fastlane::Helpers::IosL10nHelper::SWIFTGEN_VERSION}"
@@ -97,7 +97,7 @@ module Fastlane
         end
     
         def self.is_supported?(platform)
-          platform == :ios
+          [:ios, :mac].include?(platform)
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
@@ -17,7 +17,8 @@ module Fastlane
           violations.each do |lang, diff|
             UI.error "Inconsistencies found between '#{params[:base_lang]}' and '#{lang}':\n\n#{diff}\n"
           end
-
+          UI.abort_with_message!('Inconsistencies found during Localization linting. Aborting.') unless violations.empty?
+          
           violations
         end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
@@ -117,7 +117,7 @@ module Fastlane
             'strings' => langs.map do |lang|
               {
                   'inputs' => ["#{lang}.lproj/Localizable.strings"],
-                  'options' => { 'separator' => "____" }, # Choose an unlikely one to avoid creating needlessly complex Stencil Context due to '.' in sentences sued as keys
+                  'options' => { 'separator' => "____" }, # Choose an unlikely one to avoid creating needlessly complex Stencil Context due to '.' in sentences used as keys
                   'outputs' => [{
                       'templatePath' => template_path,
                       'output' => output_filename(lang)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
@@ -1,0 +1,146 @@
+require 'yaml'
+require 'tmpdir'
+
+module Fastlane
+    module Helpers
+      class IosL10nHelper
+        SWIFTGEN_VERSION = '6.4.0'
+        DEFAULT_BASE_LANG = 'en'
+        CONFIG_FILE_NAME = 'swiftgen-stringtypes.yml'
+
+        attr_reader :install_path
+        attr_reader :version
+
+        # @param [String] install_path The path to install SwiftGen to. Usually something like "#{PROJECT_DIR}/vendor/swiftgen"
+        def initialize(install_path:, version: SWIFTGEN_VERSION)
+          @install_path = install_path
+          @version = version || SWIFTGEN_VERSION
+        end
+
+        def check_swiftgen_installed
+          return false unless File.exists?(swiftgen_bin)
+          vers_string = `#{swiftgen_bin} --version`
+          # SwiftGen v6.4.0 (Stencil v0.13.1, StencilSwiftKit v2.7.2, SwiftGenKit v6.4.0)
+          return vers_string.include?("SwiftGen v#{version}")
+        rescue
+          return false
+        end
+
+        def install_swiftgen
+          UI.message "Installing SwiftGen #{version} into #{install_path}"
+          Dir.mktmpdir do |tmpdir|
+            zipfile = File.join(tmpdir, "swiftgen-#{version}.zip")
+            Action.sh('curl', '--fail', '--location', '-o', zipfile, "https://github.com/SwiftGen/SwiftGen/releases/download/#{version}/swiftgen-#{version}.zip")
+            extracted_dir = File.join(tmpdir, "swiftgen-#{version}")
+            Action.sh('unzip', zipfile, '-d', extracted_dir)
+
+            FileUtils.rm_rf(install_path) if File.exists?(install_path)
+            FileUtils.mkdir_p(install_path)
+            FileUtils.cp_r("#{extracted_dir}/.", install_path)
+          end
+        end
+
+        def run(input_dir:, base_lang: DEFAULT_BASE_LANG)
+          check_swiftgen_installed || install_swiftgen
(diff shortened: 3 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_store_app_sizes.rb
```diff
@@ -100,6 +100,10 @@ def self.output
           # Define the shared values you are going to provide
           
         end
+
+        def self.return_type
+          :hash
+        end
```
AliSoftware, line 106: _NB: this is not related to this PR, but since I saw the missing implementation here while I was creating the other new linting action I figured I could do a quick-fix where I was at it._

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -0,0 +1,104 @@
...
+          )
+          violations = helper.run(
+            input_dir: resolve_path(params[:input_dir]),
+            base_lang: params[:base_lang],
+          )
+        
+          violations.each do |lang, diff|
+            UI.error "Inconsistencies found between '#{params[:base_lang]}' and '#{lang}':\n\n#{diff}\n"
+          end
+          UI.abort_with_message!('Inconsistencies found during Localization linting. Aborting.') unless violations.empty?
```
AliSoftware: [Open question]
I considered that localization violations should always be errors and make the lane abort early (as opposed to only be a warning) because those violations will make the app crash in other locales if not fixed; but I'm open to make them a warning instead if you think that would fit our sprint workflow better.

That also might depend on where we want to call this action from in our app's `Fastfiles` (I'm thinking theses would be called during `complete_code_freeze` and should prevent the code freeze from going forward if there are still violations to be fixed, but maybe you have other typical call sites in mind?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_store_app_sizes.rb
```diff
@@ -111,7 +115,7 @@ def self.authors
         end
   
         def self.is_supported?(platform)
-          platform == :ios
+          [:ios, :mac].include?(platform)
```
mokagio, line 118: Totally out of topic question. But it's a bit confusing that this code which supports iOS and macOS lives under `ios/`.

This obviously is not a critique, since there is not `macos` or `apple` folder in the library structure. Just pointing it out for consideration.

I can live with code from a folder named "iOS" running for macOS, too, no dramas 😄 I guess it just stood out because of how small the diff is.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -0,0 +1,104 @@
...
+        def self.resolve_path(path)
+          File.absolute_path(path, repo_root)
+        end
+    
+        #####################################################
+        # @!group Documentation
+        #####################################################
+    
+        def self.description
+          "Lint the different *.lproj/.strings files for different locales and ensure the parameter placeholders are consistent."
```
mokagio: Nitpick

```suggestion
          "Lint the different *.lproj/.strings files for each locale and ensure the parameter placeholders are consistent."
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -0,0 +1,104 @@
...
+        def self.details
+          "Compares the translations against a base language to find potential mismatches for the %s/%d/… parameter placeholders between locales."
+        end
+    
+        def self.available_options
+          [
+            FastlaneCore::ConfigItem.new(
+              key: :install_path,
+              env_name: "FL_IOS_LINT_TRANSLATIONS_INSTALL_PATH",
+              description: "The path where to install SwiftGen tooling needed to run the linting process. If a relative path, should be relative to your repo_root",
```
mokagio: 🤷 ?

```suggestion
              description: "The path where to install the SwiftGen tooling needed to run the linting process. If a relative path, should be relative to your repo_root",
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -0,0 +1,104 @@
...
+        def self.return_value
+          "A hash, keyed by language code, whose values are the diff found for said language"
+        end
+    
+        def self.authors
+          ["AliSoftware"]
+        end
+    
+        def self.is_supported?(platform)
+          platform == :ios
```
mokagio: I suspect [Simplenote macOS](https://github.com/Automattic/simplenote-macos/) could benefit from this kind of linting too, right?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,178 @@
...
+          @install_path = install_path
+          @version = version || SWIFTGEN_VERSION
+        end
+
+        # Check if SwiftGen is installed in the provided `install_path` and if so if the installed version matches the expected `version`
+        #
+        def check_swiftgen_installed
+          return false unless File.exists?(swiftgen_bin)
+          vers_string = `#{swiftgen_bin} --version`
+          # SwiftGen v6.4.0 (Stencil v0.13.1, StencilSwiftKit v2.7.2, SwiftGenKit v6.4.0)
```
mokagio, line 32:
```suggestion
          # The SwiftGen version string has this format:
          #
          # SwiftGen v6.4.0 (Stencil v0.13.1, StencilSwiftKit v2.7.2, SwiftGenKit v6.4.0)
```

Suggesting this because it took me a few extra seconds to understand the comment.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,178 @@
...
+              Dir.glob('*.lproj').map { |dir| File.basename(dir, '.lproj') }
+          end.sort
+      
+          config = {
+            'input_dir' => input_dir,
+            'output_dir' => output_dir,
+            'strings' => langs.map do |lang|
+              {
+                  'inputs' => ["#{lang}.lproj/Localizable.strings"],
+                  'options' => { 'separator' => "____" }, # Choose an unlikely one to avoid creating needlessly complex Stencil Context due to '.' in sentences sued as keys
```
```suggestion
                  'options' => { 'separator' => "____" }, # Choose an unlikely one to avoid creating needlessly complex Stencil Context due to '.' in the sentences uses as keys
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,178 @@
...
+          }
+      
+          # Write SwiftGen config file
+          config_file = File.join(output_dir, CONFIG_FILE_NAME)
+          File.write(config_file, config.to_yaml)
+      
+          return [config_file, langs]
+        end
+
+        # Because we use English copy verbatim as key names, some keys are the same just except for the upper/lowercase.
```
```suggestion
        # Because we use English copy verbatim as key names, some keys are the same except for the upper/lowercase.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,178 @@
...
+            (config_file, langs) = generate_swiftgen_config!(input_dir, tmpdir)
+            Action.sh(swiftgen_bin, 'config', 'run', '--config', config_file)
+            
+            # Run diffs
+            base_file = sort_file_lines!(tmpdir, base_lang)
+            langs.delete(base_lang)
+            return Hash[langs.map do |lang|
+              file = sort_file_lines!(tmpdir, lang)
+              diff = `diff -U0 "#{base_file}" "#{file}"`
+              diff.gsub!(/^(---|\+\+\+).*\n/, '')
```
mokagio, line 179: Would it be worth adding a comment on this substitution of the diff file names? Also regarding the fact that the `@@ from-file-line-numbers to-file-line-numbers @@` is retained so that users can find the issues faster?

mokagio (approved):
Code looks good. All my comments are nits 😆 

I tested the action on the WordPress iOS repo. It behaved as described. Here's the first lines of my output:

```
[15:43:47]: Inconsistencies found between 'en' and 'ar':

@@ -60 +60 @@
-"%li words, %li characters" => [Int,Int]
+"%li words, %li characters" => []
@@ -215 +215 @@
-"Are you sure you want to continue?\n All site data will be removed from your %@." => [String]
+"Are you sure you want to continue?\n All site data will be removed from your %@." => []
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
mokagio: I never had to deal with localization issues, so I don't have a solid experience on which to base my suggestion.

I like the idea of failing early 👍 

The worst thing that I can see happening is the release being delayed because there's a broken string that we release managers don't know how to fix and we need to wait for the translators' input. That's better than crashing on the user 😄 

Plus, maybe we could have a check one or two days before the time to finalize the release to see how the strings that have already been localized pass the lint.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware:
```suggestion
                  'options' => { 'separator' => "____" }, # Choose an unlikely one to avoid creating needlessly complex Stencil Context due to '.' in sentences used as keys
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_store_app_sizes.rb
AliSoftware: Ohh good point. How does that work for simplenote-macos? Do we use `ios_xxx` actions in its `Fastfile` despite it being a `macos` project?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
AliSoftware: Indeed!
```suggestion
          [:ios, :mac].include?(platform)
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
loremattei: I totally agree that localization violations should be errors.

AliSoftware: Pushed a new commit to make it abort by default but allow to opt-out of abort behavior via an optional `abort_on_violations` param.

I think we'll definitively make it error and abort during `finalize_release` (and not during `complete_code_freeze` like I initially said in my PR description, sorry), but I'm still wondering if we should also make it errors for `new_beta_release` (probably yes too, but open for debate)

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: Updated the whole comment in dbf2ab0a0349f6439f74bf8e26b106b1d64fb025

loremattei:
I only tested it briefly so far on WordPress iOS and I was impressed by how many potential crashes we have around 😱. This just confirm how useful this tool is! 😍

Skimming through the report I found some entries that I don't understand. I _think_ they may be false positives, but it may also be that I'm missing something. 

For example: in `bg` I find this entry: 
```
+"Great! Could you leave us a nice review?\n\nIt really helps." => []
```
and this is the actual string:
```
"Great! Could you leave us a nice review?\n\nIt really helps." = "Чудесно! Можете ли да ни оставите позитивен отзив?\n\nНаистина би ни било полезно.";
```

The same entry is in `en-CA`, where the actual string is: 
```
"Great! Could you leave us a nice review?\n\nIt really helps." = "Great! Could you leave us a nice review?\n\nIt really helps.";
```

In `zh-Hans` I see this entry:
```
@@ -235 +235 @@
-"Are you sure you wish to remove the topic '%@'?" => [String]
+"Are you sure you wish to remove the topic '%@'?" => []
``` 
and the actual string is: 
```
"Are you sure you wish to remove the topic '%@'?" = "是否确定要删除主题“％@”？";
``` 

I don't know if there are more, I only checked some entries here and there.

AliSoftware:
@loremattei Thanks for doing a thorough check on those violations to check for false positives! ❤️ 

> I only tested it briefly so far on WordPress iOS and I was impressed by how many potential crashes we have around 😱. This just confirm how useful this tool is! 😍

Indeed! A bit scary!

> For example: in `bg` I find this entry:
> 
> ```
> +"Great! Could you leave us a nice review?\n\nIt really helps." => []
> ```
> 
> and this is the actual string:
> 
> ```
> "Great! Could you leave us a nice review?\n\nIt really helps." = "Чудесно! Можете ли да ни оставите позитивен отзив?\n\nНаистина би ни било полезно.";
> ```
> 
> The same entry is in `en-CA`, where the actual string is:
> 
> ```
> "Great! Could you leave us a nice review?\n\nIt really helps." = "Great! Could you leave us a nice review?\n\nIt really helps.";
> ```

I don't think that's a false positive. The error reported by this `diff` here is only reporting an addition (`+`) without deletion, so it's not a difference between 2 similar lines (2 lines for the similar key but different expected parameters between `en` and the locale) but instead it's a line that is present in those locales but is _not_ present at all in the baseline. The baseline being `en` (and not `en-CA`).

This means that this key is present in many locales, including some regional variants for en like `en-CA` and `en-GB` and `en-AU`, but not in the main `en` (used, amongst others, as a fallback for `en-US`, and as a fallback for any locale we don't support).

It's thus not reporting a placeholder mismatch in keys, but **a key that is actually completely missing in the baseline** (or that is added unnecessarily for all locales other than the baseline? Depends on your POV)

> In `zh-Hans` I see this entry:
> 
> ```
> @@ -235 +235 @@
> -"Are you sure you wish to remove the topic '%@'?" => [String]
> +"Are you sure you wish to remove the topic '%@'?" => []
> ```
> 
> and the actual string is:
> 
> ```
> "Are you sure you wish to remove the topic '%@'?" = "是否确定要删除主题“％@”？";
> ```
> 

Oohhhh this one is super interesting, because it's a trap we would likely _never_ have caught manually without the script!

In fact, what looks like `"%@"` in the translation is not just a `%` followed by a `@`, as shown if I hexdump that content:
```
✓ $ sed -n 735p WordPress/Resources/zh-Hans.lproj/Localizable.strings | xxd     
00000000: 2241 7265 2079 6f75 2073 7572 6520 796f  "Are you sure yo
00000010: 7520 7769 7368 2074 6f20 7265 6d6f 7665  u wish to remove
00000020: 2074 6865 2074 6f70 6963 2027 2540 273f   the topic '%@'?
00000030: 2220 3d20 22e6 98af e590 a6e7 a1ae e5ae  " = "...........
00000040: 9ae8 a681 e588 a0e9 99a4 e4b8 bbe9 a298  ................
00000050: e280 9cef bc85 40e2 809d efbc 9f22 3b0a  ......@......";.
```

On the last line, you can see the `0x40` which is for `@`, but it's is _not_ preceded by `0x25` (`%`), but instead by `0xefbc85` – which is actually the UTF8 encoding of ["FULLWIDTH PERCENT SIGN"](https://www.fileformat.info/info/unicode/char/ff05/index.htm), not a regular `%` character – hence SwiftGen (correctly) not interpreting it as a placeholder and not being fooled by it!

_(FYI, the rest of the string is `0xe2809d` = ["RIGHT DOUBLE QUOTATION MARK"](https://www.fileformat.info/info/unicode/char/201d/index.htm) encoded in UTF8, then `0xefbc9f` =["FULLWIDTH QUESTION MARK"](https://www.fileformat.info/info/unicode/char/ff1f/index.htm) encoded in UTF8, then the expected `0x22` + `0x3b` + `0x0a` = `";\n`)_

So actually, not only it's not a false positive, but it's even a case that we would likely never have caught ourselves if the tooling didn't highlight it for us! 🎉

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
mokagio: +1 for `new_beta_release`, too.

The opt-out parameter is great. Gives us the best of both world. If the lint fails and we have the tool to fix it, we can address it. If it fails but we still need to push a build out, e.g. to test an important bug fix, we can skip.

Excellent 👍

loremattei:
Thanks for the explanation @AliSoftware!

> It's thus not reporting a placeholder mismatch in keys, but a key that is actually completely missing in the baseline

I wonder how this can happen 🤔 This is something which probably requires a bit more investigation, because it should not be possible AFAIK 😅

Having said that, should we consider this a warning rather than an error? 

> On the last line, you can see the 0x40 which is for @, but it's is not preceded by 0x25 (%), but instead by 0xefbc85 – which is actually the UTF8 encoding of "FULLWIDTH PERCENT SIGN", not a regular % character – hence SwiftGen (correctly) not interpreting it as a placeholder and not being fooled by it!

Interesting. Again, something which is probably worth more investigation 🤔

AliSoftware:
> > It's thus not reporting a placeholder mismatch in keys, but a key that is actually completely missing in the baseline
> 
> I wonder how this can happen 🤔 This is something which probably requires a bit more investigation, because it should not be possible AFAIK 😅

One idea I have for this to happen is if the key was added to en.lproj at some point, got translated… and then got removed from code (but still has translations in GlotPress so still get pulled)… but not entirely sure, as if that was the case, wouldn't the key be pulled back from GlotPress by our scripts into `en.lproj` anyway on next run? 🤔  Definitively needs investigation!
 
> Having said that, should we consider this a warning rather than an error?

@loremattei I thought about that. Ideally I think it could make sense to consider this a warning… but well, it depends on what the investigation about the root cause concludes I guess. But in practice it would actually be quite complicated to make it a warning only, because that means I'd have to analyse/parse the `diff` output to detect that there's a line between two `@​@` that only has a `+` but no corresponding `-`… easy to see with human eyes, but the script currently only see the `diff` as an arbitrary textual output from the `diff` command itself, without interpreting it (letting this interpretation to the release manager).

Providing different treatment of warning vs error depending on the interpretation of the `diff` would mean we'd need to go quite deeper into making the script interpret that `diff` more thoroughly (and tbh I'm afraid this automated interpretation if any would raise the risk of false positives e.g. things that should be errors but would be interpreted as warnings instead), so I thought it was probably not worth the trouble and would probably be better to fix the root cause of this in GlotPress once and for all instead.
 
> > On the last line, you can see the 0x40 which is for @, but it's is not preceded by 0x25 (%), but instead by 0xefbc85 – which is actually the UTF8 encoding of "FULLWIDTH PERCENT SIGN", not a regular % character – hence SwiftGen (correctly) not interpreting it as a placeholder and not being fooled by it!
> 
> Interesting. Again, something which is probably worth more investigation 🤔

@loremattei Well wasn't my thorough investigation above about it enough for you 😅  I guess what we could do tho is to ping the translators to make them aware of such a case and to be extra careful to use regular `%` for those cases. Though tbh that's just a one-time case here (only one key with that issue in that language) so not sure it's worth pinging them for that (maybe only if that happens again in the future – which the linter script will definitively catch automatically like here if so! 🎉 ), I think we should just fix the full-width-percent-char to be a regular `%` in GlotPress for this one case, and go on our merry way 🙃

loremattei:
> Providing different treatment of warning vs error depending on the interpretation of the diff would mean we'd need to go quite deeper into making the script interpret that diff more thoroughly (and tbh I'm afraid this automated interpretation if any would raise the risk of false positives e.g. things that should be errors but would be interpreted as warnings instead), so I thought it was probably not worth the trouble and would probably be better to fix the root cause of this in GlotPress once and for all instead.

Yeah! I see! As discussed, it's ok to try to figure out how this happened and then fix it on GlotPress 👍

loremattei (approved):

PR #175, opened 2020-11-19
Fix indentation and remove trailing whitespaces from Git helpers

mokagio:
What I actually wanted to do was the change in the next #176, but before doing that I had to come through this. My text editor pointed these out.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -1,103 +1,103 @@
 module Fastlane
-    module Helpers
-      module AndroidGitHelper
+  module Helpers
+    module AndroidGitHelper
 
-        def self.git_checkout_and_pull(branch)
-          Action.sh("git checkout #{branch}")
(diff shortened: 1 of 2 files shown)


loremattei:
I never know how to review these PRs because the way the diff are presented makes it difficult to figure out if something got deleted by mistake or any other change happened 🤔
What I know is that without a way to check these rules locally, we'll probably have this problem back again in the future. What do you think about adding a `.editor-config` file to this PR?

AliSoftware:
> I never know how to review these PRs because the way the diff are presented makes it difficult to figure out if something got deleted by mistake or any other change happened 🤔

@loremattei did you know about this option in GitHub?

GitHub-Whitespace-Option

AliSoftware (approved):
GitHub shows that it's indeed only whitespace changes, so LGTM 👍 

But I agree with Lore that this PR could be a good occasion to finally add that `.editorconfig` file here indeed!!

loremattei:
> @loremattei did you know about this option in GitHub?

Ah! Right @AliSoftware! I think I actually knew about this, but I keep forgetting that it's there! Thanks for the reminder! It's actually super useful 🙇

PR #176, opened 2020-11-19
Remove period from generated commit titles

mokagio:
Pedantic PR: Trailing punctuation is unnecessary in subject lines.

See:

- https://www.git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines where the example doesn't have trailing punctuation
- https://chris.beams.io/posts/git-commit/#end which makes this convention explicit


loremattei (approved):
LGTM :-) 

I'm curious: who's Chris Beams and why his post should be considered as a guideline for us? (I don't mind having the period or not, so I'm asking only because I'm genuinely curious about it)

PR #177, opened 2020-12-02
Fixing glitches in ios l10n linter action

AliSoftware:
Some edge cases I found when using the linter on WC and WP:
 - `.lproj` folders not containing any .strings file (especially `Base.lproj` in WC)
 - empty diffs due to missing strings file in some translation bundles (for locales where the app strings are not translated at all despite having a lproj, so the runtime will still fallback while we were still trying to compare to empty diff file in the script)

+ Added the ability to apply the linter to only a specific set of language; very useful for debugging and focusing on one locale at a time when I was fixing the existing violations.

Note: we'll need a new release after this merge so I can push my PRs enabling those actions on WPiOS/WCiOS by pointing on the new release

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
@@ -165,7 +165,7 @@ module Fastlane
         def find_diffs(input_dir:, base_lang:, only_langs: nil)
           Dir.mktmpdir('a8c-lint-translations-') do |tmpdir|
             # Run SwiftGen 
-            langs = only_langs.nil? ? nil : only_langs + [base_lang]
+            langs = only_langs.nil? ? nil : (only_langs + [base_lang]).uniq
             (config_file, langs) = generate_swiftgen_config!(input_dir, tmpdir, only_langs: langs)
             Action.sh(swiftgen_bin, 'config', 'run', '--config', config_file)
             
@@ -175,10 +175,7 @@ module Fastlane
             return Hash[langs.map do |lang|
               file = sort_file_lines!(tmpdir, lang)
               # If the lang ends up not having any translation at all (e.g. a `.lproj` without any `.strings` file in it but maybe just a storyboard or assets catalog), ignore it
-              # Note: to check if the file is "almost" empty, i.e. only containing newlines, we only read it if it's small enough (arbitrary max length of 10 bytes) because that
-              # situation typically comes from a template having a couple of newlines at top and bottom between what's usually the content – but is empty in this case – and we don't
-              # want to waste RAM reading a big file of multiple KB for nothing, so considering only files that are "small enough to be candidates for only-newlines files" seems enough.
-              next nil if file.nil? || File.size(file) <= 10 && File.readlines(file).all? { |line| line.chomp.length == 0 }
+              next nil if file.nil? || only_empty_lines?(file)
               # Compute the diff
               diff = `diff -U0 "#{base_file}" "#{file}"`
               # Remove the lines starting with `---`/`+++` which contains the file names (which are temp files we don't want to expose in the final diff to users)
@@ -192,6 +189,15 @@ module Fastlane
           end
         end
 
+        # Returns true if the file only contains empty lines, i.e. lines that only contains whitespace (space, tab, CR, LF)
+        def only_empty_lines?(file)
+          File.open(file) do |f|  
+            while line = f.gets
+              return false if not line.strip.empty?
+            end  
+          end  
+          return true
+        end
       end
     end
 end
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -155,21 +157,26 @@ def sort_file_lines!(dir, lang)
...
+            langs = only_langs.nil? ? nil : only_langs + [base_lang]
+            (config_file, langs) = generate_swiftgen_config!(input_dir, tmpdir, only_langs: langs)
             Action.sh(swiftgen_bin, 'config', 'run', '--config', config_file)
             
             # Run diffs
             base_file = sort_file_lines!(tmpdir, base_lang)
             langs.delete(base_lang)
             return Hash[langs.map do |lang|
               file = sort_file_lines!(tmpdir, lang)
+              # If the lang ends up not having any translation at all (e.g. a `.lproj` without any `.strings` file in it but maybe just an storyboard or assets catalog), ignore it
```
```suggestion
              # If the lang ends up not having any translation at all (e.g. a `.lproj` without any `.strings` file in it but maybe just a storyboard or assets catalog), ignore it
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -155,21 +157,26 @@ def sort_file_lines!(dir, lang)
...
+              next nil if file.nil? || File.size(file) <= 10 && File.readlines(file).all? { |line| line.chomp.length == 0 }
```
mokagio: Where does the 10 value in the `File.size` check come from?

mokagio (approved):
LGTM! 👍 

I have mixed feelings about the `only_langs` parameter name. I was tempted to suggest renaming it to `langs` as to specify the langs to lint, but then it would be confusing to have `lagns: nil` as the default value, having `nil` mean "all languages". So 👌 .

What could we do to unit test this? Also, totally out of scope but just popped to my mind: what can we do to make this a standalone action? I think it would be useful to other people. I don't think there's much WordPress or GlotPress specific logic backed in it, right?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: Right, I should add a comment on this. So basically the idea is to avoid any empty file, including a file that only contains blank lines, but not waste any memory by reading a potentially huge file.
So the idea is that the files are usually either completely empty (size zero, actually rare)… or are only containing a couple of `\n` newlines and that's it – because of the original template having newlines to space things up between header, (in this case empty) content, and footer. So my idea here was to only check if the file was small enough before trying to `readlines` and read its whole content to check if it was only a couple of newlines, vs reading the whole files for all cases including when the file is multiple KBs and consuming RAM for nothing.

Definitively needs an explanation comment though I agree 👍

jkmassel (approved):
This LGTM as well – agree with @mokagio that adding tests would be nice

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -155,21 +157,29 @@ def sort_file_lines!(dir, lang)
...
         # @return [Hash<String, String>] A hash of violations, keyed by language code, whose values are the diff output.
         #
         # @note The returned Hash contains keys only for locales with violations. Locales parsed but without any violations found will not appear in the resulting hash.
         #
-        def find_diffs(input_dir:, base_lang:)
+        def find_diffs(input_dir:, base_lang:, only_langs: nil)
           Dir.mktmpdir('a8c-lint-translations-') do |tmpdir|
             # Run SwiftGen 
-            (config_file, langs) = generate_swiftgen_config!(input_dir, tmpdir)
+            langs = only_langs.nil? ? nil : only_langs + [base_lang]
```
AliSoftware:
```suggestion
            langs = only_langs.nil? ? nil : (only_langs + [base_lang]).uniq
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: I actually ended up thinking of a better and more logical solution to "check if the file is all-empty-lines without reading the whole file if it's too large". Pushed in d50bd33 🙂

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -192,6 +189,15 @@ def find_diffs(input_dir:, base_lang:, only_langs: nil)
...
 
+        # Returns true if the file only contains empty lines, i.e. lines that only contains whitespace (space, tab, CR, LF)
+        def only_empty_lines?(file)
+          File.open(file) do |f|  
+            while line = f.gets
+              return false if not line.strip.empty?
+            end  
+          end  
+          return true
+        end
```
mokagio, line 200: Nice :)

PR #181, opened 2020-12-15
Add some unit tests

AliSoftware:
## What is this about

As promised in https://github.com/wordpress-mobile/release-toolkit/pull/177#pullrequestreview-545829378:

* Adding some tests to our Fastlane actions, especially the new linter action
* Updated some of our gems in `Gemfile.lock` in the process

## How

Added a spec file for the linter spec + `yaml` files to provide definition of fixtures data + expected result. The tests mainly read the YAML file which tells it how to set up a test temporary directory with dummy `.lproj` and `.strings` files with the appropriate contents, then run the linter, then check that the linter result (returned hash) matches the expectation as defined by the YAML.

Note: I chose YAML over JSON mainly because of its ability to write strings directly without the need to quote them – which would otherwise have required a LOT of backslash-escapes in a JSON file to represent the same content.

_Note: Encountered a challenge with Fastlane::Actions subtle internal tweaks that makes it behave differently when running under tests. Especially the `Action.sh` refused to execute when it was run in the context of a unit test, while it worked as expected otherwise. Turns out, Fastlane has a test hardcoded in the implementation of `Actions.sh`, which changes its behavior under tests… but I had to dig quite a bit to find that out 😰 . I included a link to the relevant part in Fastlane's code as a code comment for reference for later._

## To test

* Clone this branch
* Run `rbenv exec bundle install` (I took the occasion to update some gems to latest version)
* Run `rbenv exec bundle exec rspec -fd` (`-fd` option is to format the test results with a more verbose output so you can see the name of each test that is run, and ensure our new tests are part of the set)
* Check that all the tests, and especially the ones about `Fastlane::Actions::IosLintLocalizationsAction`, are all green.

```
Fastlane::Actions::IosLintLocalizationsAction
  SwiftGen Installation Logic
    only installs SwiftGen the first time, when it is not yet installed
  Linter
    succeeds when there are no violations
    detects inconsistent placeholder count
    detects inconsistent placeholder types
    properly handles misleading characters and placeholders in RTL languages
    does not fail if a locale does not have any Localizable.strings
```

OR… [trust CI being all green](https://app.circleci.com/pipelines/github/wordpress-mobile/release-toolkit/466/workflows/98fe951a-ae5e-410f-9518-0dd1b3a9fd24/jobs/451) when it ran rspec 😛


spec/test-data/translations/test-lint-ios-tricky-chars.yaml
```diff
@@ -0,0 +1,39 @@
...
+    "many_placeholders_mix_pos" = "String %2$@, Float %4$f, Int %1#d, Long %5$li, Precise %3$.3f"; // WRONG: %1#d instead of %1$d so not a placeholder, which shifts all other args
+    "repeated_placeholder" = "String %1$@, yes, I repeat, that's %1$@ indeed, not %2$@.";
+  ar: |
+    // Those strings are carefully crafted to be using the right order of characters at byte level, despite bidi making it strange-looking in your IDE
+    "many_placeholders_mix_pos" = "السلسلة %2$@ ، عدد %4$.3f ، عدد صحيح %1$d طويل %5$li دقيق %3$f"; // OK
+    "string_no_pos" = "سلسلة% @ هنا."; // OK
+    "string_pos" = "سلسلة موضعية٪1$@ هنا."; // WRONG: Use of 'ARABIC PERCENT SIGN' character
+    "string_int_pos" = "خيط \u{200F}%1$@\u{200E} عدد صحيح \u{200F}%2$d\u{200E}."; // OK. 200E and 200F are LRM/RLM markers
+    "string_int_pos_inv" = "-عدد صحيح%2$d قبل السلسلة%1$@"; // OK
+    "repeated_placeholder" = "السلسلة   %1$@ ، نعم ، أكرر ، هذه   %1$@ بالفعل ، وليست   %2$@."; // OK
```
AliSoftware, line 23: This is why we needed a linter 😅

spec/test-data/translations/test-lint-ios-tricky-chars.yaml
```diff
@@ -0,0 +1,39 @@
...
+    "many_placeholders_mix_pos" = "String %2$@, Float %4$f, Int %1#d, Long %5$li, Precise %3$.3f"; // WRONG: %1#d instead of %1$d so not a placeholder, which shifts all other args
+    "repeated_placeholder" = "String %1$@, yes, I repeat, that's %1$@ indeed, told you %2$d times.";
+  ar: |
+    // Those strings are carefully crafted to be using the right order of characters at byte level, despite bidi making it strange-looking in your IDE
+    "many_placeholders_mix_pos" = "السلسلة %2$@ ، عدد %4$.3f ، عدد صحيح %1$d طويل %5$li دقيق %3$f"; // OK
+    "string_no_pos" = "سلسلة% @ هنا."; // OK
+    "string_pos" = "سلسلة موضعية٪1$@ هنا."; // WRONG: Use of 'ARABIC PERCENT SIGN' character
+    "string_int_pos" = "خيط \u{200F}%1$@\u{200E} عدد صحيح \u{200F}%2$d\u{200E}."; // OK. 200E and 200F are LRM/RLM markers
+    "string_int_pos_inv" = "-عدد صحيح%2$d قبل السلسلة%1$@"; // OK
+    "repeated_placeholder" = "السلسلة   %1$@ ، نعم ، أكرر ، هذه   %1$@ بالفعل ، وليست   %2$d."; // OK
```
AliSoftware, line 23: This is why we needed a linter 😅

AliSoftware:
Bonus points if you test this PR in parallel of https://github.com/wordpress-mobile/release-toolkit/pull/182 and validate that those unit tests still work even after the other PR is applied.
_Easiest way to do this is probably to checkout this PR's branch then locally cherry-pick the commit from https://github.com/wordpress-mobile/release-toolkit/pull/182 on top of it before running `rspec`_

spec/ios_lint_localizations_spec.rb
```diff
@@ -0,0 +1,126 @@
+require 'spec_helper.rb'
+require 'fileutils'
+require 'tmpdir'
+require 'yaml'
+
+describe Fastlane::Actions::IosLintLocalizationsAction do
+  before do
+    # Ensure `Action.sh` is not skipped during test – so that SwiftGen will be installed by our action as normal
+    # See https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/helper/sh_helper.rb#L45-L85
```
mokagio: How do you feel about adding the commit hash version of this URL, so that it'll always point at the code mentioned as it was at the time of writing.

https://github.com/fastlane/fastlane/blob/e6bd288f17038a39cd1bfc1b70373cab1fa1e173/fastlane/lib/fastlane/helper/sh_helper.rb#L45-L85

I always have mixed feelings about using the commit hash URL version. It's great for the reason above, but it can also lead to confusion if the reader doesn't realized they're looking at code that might not be up to date anymore when triaging an issue.

spec/ios_lint_localizations_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+    allow(FastlaneCore::Helper).to receive(:sh_enabled?).and_return(true)
+  end
+
+  context 'SwiftGen Installation Logic' do
+    it 'only installs SwiftGen the first time, when it is not yet installed' do
+      Dir.mktmpdir('a8c-lint-l10n-tests-swiftgen-install-') do |install_dir|
+        Dir.mktmpdir('a8c-lint-l10n-tests-data-') do |empty_dataset|
+          # Expect install dir to be empty before we start
+          expect(Dir.entries(install_dir)).to eq(['.','..'])
```
mokagio: What do you think of 

```suggestion
          expect(Dir.empty?(install_dir)).to be_true
```

spec/ios_lint_localizations_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+          # First run: expect curl, unzip and cp_r to be called to install SwiftGen before it gets run
```
mokagio: Super nitpick

```suggestion
          # First run: expect curl, unzip and cp_r to be called to install SwiftGen before running the action
```

mokagio (commented):
Love seeing new tests. And what a set of mocking-fu there is 🤩!

I hacked the code to always return `false` when checking if SwiftGen is installed and the test failed 👍 

Something I noticed, though, is that the tests are now considerably slower compared to what they were before.

Here's the result of running RSpec with the `--profile` option on my 2019 16" MacBook Pro.

```
Top 10 slowest examples (13.59 seconds, 99.4% of total time):
  Fastlane::Actions::IosLintLocalizationsAction Linter succeeds when there are no violations
    11.79 seconds ./spec/ios_lint_localizations_spec.rb:65
  Fastlane::Actions::IosLintLocalizationsAction SwiftGen Installation Logic only installs SwiftGen the first time, when it is not yet installed
    1.12 seconds ./spec/ios_lint_localizations_spec.rb:14
  FastlaneCore::Helper::GitHelper can detect a repository with Git-lfs enabled
    0.13488 seconds ./spec/git_helper_spec.rb:25
...

Top 7 slowest example groups:
  Fastlane::Actions::IosLintLocalizationsAction
    2.22 seconds average (13.33 seconds / 6 examples) ./spec/ios_lint_localizations_spec.rb:6
  FastlaneCore::Helper::GitHelper
    0.07145 seconds average (0.28579 seconds / 4 examples) ./spec/git_helper_spec.rb:4
...
```

🤔 what could take so long?

mokagio:
> 🤔 what could take so long?

I checked and I'm pretty sure it's because the first test for the actual lint actually goes off and downloads SwiftGen.

I wonder if the tests should just expect SwiftGen to be available in the machine running them. It would be a great performance boost for the tests. We could have a check in the test itself to verify SwiftGent is installed and fail if it isn't. It seems reasonable to expect that to develop for this plugin, one would have to install all the relevant tooling IMHO (we could always put that test in a group, so that if someone's can't install SwiftGen for some reason, they can still run the other tests).

spec/ios_lint_localizations_spec.rb
AliSoftware: TIL about `Dir.empty?` 👍

AliSoftware:
Yes it takes longer for the first lint test to run because it has to actually download SwiftGen for real in order to actually run the tests indeed.

I thought about making those tests in an optional group that the user can choose to only run if they have SwiftGen installed on their machine, but I'm pretty sure that it will lead to those tests never to be run as we'll always forget to opt-in for them most of the time…

Making them opt out instead of opt in (ie user could disable the group if they can't install SwiftGen but by default the group will try to run) is a neat idea, but I'm always a bit hesitant to make tests fail if a tool is not installed (as opposed to the test detecting an actual logic issue in the behavior)

An alternative would be to still make the tests install SwiftGen when the first linter test is run… but install it in vendor/swiftgen (ie the same place where it's expected to be installed when you actually run the action for real) and not delete the install after the tests are run, as opposed to having the tests install it in a temp dir that gets deleted after the tests have finished 🤔

mokagio:
> An alternative would be to still make the tests install SwiftGen when the first linter test is run… but install it in vendor/swiftgen (ie the same place where it's expected to be installed when you actually run the action for real) and not delete the install after the tests are run, as opposed to having the tests install it in a temp dir that gets deleted after the tests have finished 🤔

I like this idea 👍

AliSoftware:
> I like this idea 👍

Applied in 93cec1afdcdbba3028eb73bc257b2592af92e78d

The first run – as well as every run in CI – will still take longer, because SwiftGen is not installed yet when CI makes a clean clone of the repo so it will always have to install it in CI. But when run locally, or the whole test suite is run multiple times on the same machine, the subsequent test suite executions will be as fast as before.

mokagio (approved):
Excellent work!

~17s on the first run (without `vendor/swiftgen/0.6.4/`), ~1s on the second 👍 

image

PR #186, opened 2021-01-14
Fix/update submodules before build

loremattei:
This PR makes two changes to the `android_update_metadata` action: 
- It inits the submodules in the repository before trying to run a build to check the strings. 
- It removes a call to the `release-checks.sh` script. Over the time, all the checks in that script were deprecated or moved to the `release-toolkit` and the only one which was already in place is running lint on the new strings which is actually duplicated in `update_metadata`.

# To Test
This PR can be tested as a part of the [companion PR in WordPress Android](https://github.com/wordpress-mobile/WordPress-Android/pull/13770).


jkmassel (approved):
Approved with https://github.com/wordpress-mobile/WordPress-Android/pull/13770

Issue #194, opened 2021-01-18
Update the release-toolkit README.md


AliSoftware:
The README of this repo would benefit from being more helpful and providing information for devs about, amongst other things:

* How to run RSpecs
* How to generate YARD doc to get documentation about its actions and the code
* Description of the folder hierarchy / organization in the codebase: actions vs helpers, ios vs android vs common, …
* How to configure VSCode to be able to navigate into the ruby codebase with ease (it's arguably not specific to this repository, but would still be helpful to include a quick paragraph / pointer to tips and plugins that will help contributors a lot in working on this)
 * …

Issue #193, opened 2021-01-18
Add/Improve Unit Tests (RSpec) on actions/helpers


AliSoftware:
There's only a little part of our release toolkit covered with unit tests / specs.

This tasks consists on increasing our unit test coverage for our actions and our helper methods as well as improving the RSpec / Unit test infrastructure / architecture if/where needed.

- [ ] Helpers
  - [ ] `*_version_helper` iOS + Android (🚧  WIP)
  - [ ] `*_git_helper` iOS + Android (🚧  WIP)
  - [ ] Other iOS helpers (🚧  WIP)
  - [ ] Common (non-platform-specific) helpers
- [ ] Actions
  - [ ] Android actions
  - [ ] iOS Actions
  - [ ] Common Actions
  - [ ] Configure Actions

Issue #192, opened 2021-01-18
Add YARD documentation to release-toolkit actions


AliSoftware:
Add YARD documentation to all the methods in `release-toolkit`, especially the methods on actions and helpers.

Part of this task will likely be tackled in parallel to the code rearrangement / refactoring in #191 – as we plan to write YARD documentation for new code as it is being refactored / rewritten. But this issue will cover more than #191 as we need to also document even code that won't really need to change / be refactored.

- [ ] Helpers
  - [ ] `*_version_helper` iOS + Android (🚧  WIP)
  - [ ] `*_git_helper` iOS + Android (🚧  WIP)
  - [ ] Other iOS helpers (🚧  WIP)
  - [ ] Common (non-platform-specific) helpers
- [ ] Actions
  - [ ] Android actions
  - [ ] iOS Actions
  - [ ] Common Actions
  - [ ] Configure Actions

PR #198, opened 2021-02-01
[Toolkit Cleanup] Yard doc: step 1 – config cleanup

AliSoftware:
First step to add YARD documentation https://github.com/wordpress-mobile/release-toolkit/issues/192.
Also starts to also address #194 (README/CONTRIBUTING doc).

 - Adds some basic doc to configure your IDE and dev environment for ruby
 - Configure the rspec and yard gems properly
 - Adds documentation for `*_version_helper.rb` files

## To Test

This is typically quite hard to test as automatically and strictly, as this is mainly about documentation, so it will be a lot of manual proof-reading.

### Proof-reading the documentation

* Run `bundle install` to install the gems, especially `yard`.
* Run `bundle exec yard stats --list-undoc lib/fastlane/plugin/wpmreleasetoolkit/helper/**/*_version_helper.rb` to see if there is any methods or constants missing documentation in those 2 files (ios+android version helpers)
   * The only expected documentation missing should be about modules `Fastlane` and `Fastlane::Helper`, for which I didn't think it would make sense to add specific documentation, but all the rest of those 2 files should be documented
* Run `bundle exec yard doc` to generate the documentation.
   * Ensure that there is no warnings or errors printed in the stdout while generating the documentation.
   * Open the `yard-doc/index.html` file, and proof-read the documentation, especially for `AndroidVersionHelper` and `IosVersionHelper` modules, from within your browser.
   * For each method in those documented modules, ensure that the method description, parameter types and descriptions, and return types make sense and are consistent with the implementation of those methods

> 💡  I highly encourage you to review the added documentation in the `yard-doc/index.html` page of your browser (see above) rather than / in addition to reviewing it directly from the diff in GitHub, both because that would ensure that the rendering is correct once rendered as HTML, but also because I have to admit I had my eyes focused on VSCode and the IDE when writing the code and didn't check every single rendering in the HTML doc myself, so having some fresh eyes looking at it from that side would allow a different perspective.

### Double-checking the methods that have been renamed

This PR also contains some very small renaming of some methods. It's not a big refactoring but more making some methods conform to ruby conventions a bit more.

But because ruby is a duck-typed language without type safety like for compiled languages, and because we don't yet have rspec (unit tests) for those `*_version_helpers` methods yet (that's for #193 to address) and CI won't help here, I would appreciate a thorough code consistency review / confidence check for those method renamings here, i.e. double-check that when I renamed a method, I didn't forget renaming all its call sites too (the best I could do on my end is a project-wide textual Find in VSCode… which is quite error-prone…)

### Incremental PRs and WIP

This is the first PR in a series of incremental PRs to clean up the codebase, especially add documentation and tests.

This PR targets the branch `toolkit-cleanup`, which was just cut from `develop`. I intend to stack all my upcoming incremental cleanup PRs on top of this `toolkit-cleanup` branch, and only merge `toolkit-cleanup` to `develop` once the sum of all incremental PRs feels stable enough, to avoid merging unfinished cleanup and WIP into `develop` directly.

### Confusing names and refactoring

While documenting, I came across some strangely/confusingly named methods especially in the Android helpers (e.g. `calc_next_release_version` which returns a `-rc-` beta version and not a release version name, because it's in fact the name to use after the release cut… but the method name is quite confusing. Some others are also not super clear). Which made the codebase quite hard to decipher and understand at first.

I plan to do some renaming of those to add more clarity, but probably only once I've added a couple of unit tests on those first before refactoring everything (especially given how ruby is not compiled nor typed, making a refactoring quite risky without writing any unit test first).


lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -89,15 +187,45 @@ def self.calc_next_release_base_version(version)
...
+        #   calc_next_release_version({"name": "1.2", "code": 456}) #=> {"name":"1.3-rc-1", "code": 457}
+        #   calc_next_release_version({"name": "1.2.3", "code": 456}) #=> {"name":"1.3-rc-1", "code": 457}
+        #   calc_next_release_version({"name": "1.2", "code": 456}, {"name":"alpha-1.2", "code": 457}) #=> {"name":"1.3-rc-1", "code": 458}
+        #
+        # @param [Hash] version The current version hash, with keys `"name"` and `"code"`
+        # @param [Hash] alpha_version The current alpha version hash, with keys `"name"` and `"code"`, or nil if no alpha version
+        #
+        # @return [Hash] The hash containing the version name and code to use after release cut
+        #
         def self.calc_next_release_version(version, alpha_version = nil)
```
AliSoftware, line 202: This one confused me quite a bit… method named "next release version" but in fact it's returning a `"X.Y-rc-1"` version… so more the name of the "first beta after code freeze" rather than the "release version".

This is why the documentation was also so confusing to write at first. Will probably rename in a separate/future PR.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -127,81 +265,162 @@ def self.bump_version_release
...
               if !found_section
-                if (line.include? section) 
+                if line.include?(section)
                   found_section = true
                 end
               else
-                if (line.include? keyword)
-                  return line unless line.include?("\"#{keyword}\"") or line.include?("P#{keyword}")
+                if line.include?(keyword) && !line.include?("\"#{keyword}\"") && !line.include?("P#{keyword}")
+                  return line.split(' ')[1]
```
AliSoftware, line 368: All call sites were doing the `split(' ')[1]`, so I figured I'd DRY it here – which actually makes more sense so that the method returns the keyword value, and not the "`.gradle` line containing the keyword, with potentially some indentation and spaces in front"

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -216,15 +435,15 @@ def self.update_version(version, section)
                 end
               else
                 if (version_updated < 2)
-                  if (line.include? "versionName") and (!line.include? "\"versionName\"") and (!line.include?"PversionName")
+                  if line.include?("versionName") && !line.include?('"versionName"') && !line.include?("PversionName")
                     version_name = line.split(' ')[1].tr('\"', '')
-                    line.replace line.sub(version_name, version[VERSION_NAME].to_s)
+                    line.sub!(version_name, version[VERSION_NAME].to_s)
```
AliSoftware, line 442: While `String#sub` returns a new string in ruby, `String#sub!` modifies the string in-place. So more idiomatic and performant than calling `replace` on the same string with the returned value like before.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -89,15 +187,45 @@ def self.calc_next_release_base_version(version)
...
+        # Compute the name and code of the next hotfix version.
+        #
+        # @param [String] hotfix_version_name The next version name we want for the hotfix
+        # @param [String] hotfix_version_code The next version code we want for the hotfix
+        #
+        # @return [Hash] The predicted next hotfix version, as a Hash containing the keys `"name"` and `"code"`
+        #
         def self.calc_next_hotfix_version(hotfix_version_name, hotfix_version_code)
-          { VERSION_NAME => hotfix_version_name, VERSION_CODE => hotfix_version_code}
+          { VERSION_NAME => hotfix_version_name, VERSION_CODE => hotfix_version_code }
```
AliSoftware, line 215: This one was also quite confusing, name suggesting we compute the next hotfix so that we will increment the versionName and code… but in practice we're not. Which matches how it's used at call site, but is damn confusingly and misleadingly named.

Will probably either rename the method to a more apt name in a subsequent PR, or more likely try to match `ios` here and update the call site to let this method do the hotfix increment as its name suggests, instead of making the increment at call site. That will require some refactoring of the Action tho, which is why I kept that for a future, separate PR.

.yardopts
```diff
@@ -0,0 +1,5 @@
+--no-save
+--private --protected
+--name-tag env:"Environment Variables used"
+--markup markdown
+--output-dir yard-doc
```
jkmassel: Nit – do we care about the lack of tailing newline?

Gemfile
```diff
@@ -1,8 +1,13 @@
 source('https://rubygems.org')
 
 gemspec
-gem 'codecov', :require => false, :group => :test
 gem "danger", "~> 8.0"
 
+group :test do
```
jkmassel, line 6: Nice cleanup

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -1,56 +1,122 @@
 module Fastlane
     module Helpers
+      # A module containing helper methods to manipulate/extract/bump Android version strings in gradle files
+      #
       module AndroidVersionHelper
+        # The key used in internal version Hash objects to hold the versionName value
```
jkmassel, line 6: These were rather unclear to me when reading the docs on their own (it's much more clear in context of reading the whole file).

For the purposes of documentation, WDYT about adding an example for each (or is there a way to make them internal to this module and thus not require public documentation?) Understanding how these fit into the context might help?

That said, I could see some of this being replaced a level lower by a parser, and thus only operating on a higher-level abstraction, so take it with a grain of salt?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -110,12 +238,22 @@ def self.calc_prev_release_version(version)
...
+        #
+        # @return [Bool] True if the version number has a non-zero 3rd component, meaning that it is a hotfix version.
+        #
+        def self.is_hotfix?(version)
+          return false if is_alpha_version?(version)
           vp = get_version_parts(version[VERSION_NAME])
           return (vp.length > 2) && (vp[HOTFIX_NUMBER] != 0)
         end
   
+        # Prints the current and next release version names to stdout, then return the next release version
```
jkmassel:
```suggestion
        # Prints the current and next release version names to stdout, then returns the next release version
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -1,56 +1,122 @@
...
-          name = get_version_name_from_file(gradle_path, section)
-          code = get_version_build_from_file(gradle_path, section)
+          name = get_version_name_from_gradle_file(gradle_path, section)
+          code = get_version_build_from_gradle_file(gradle_path, section)
           return { VERSION_NAME => name, VERSION_CODE => code }
         end
 
+        # Extract the version name and code from the `defaultConfig` of the `$PROJECT_NAME/build.gradle` file
+        #
+        # @return [Hash] A hash with 2 keys `"name"`` and `"code"`` containing the extracted version name and code, respectively,
```
jkmassel:
```suggestion
        # @return [Hash] A hash with 2 keys `"name"` and `"code"` containing the extracted version name and code, respectively,
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -1,56 +1,122 @@
...
         end
   
-        def self.is_alpha_version(version)
+        # Determines if a version name corresponds to an alpha version (starts with `"alpha-"`` prefix)
+        #
+        # @param [String] version The version name to check
+        #
+        # @return [Bool] true if the version name starts with the `ALPHA_PREFIX`, false otherwise.
+        #
+        #private
```
jkmassel: This looks odd in the rendered HTML file – is this supposed to be perhaps:
```suggestion
        # @private
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -1,56 +1,122 @@
...
   
-        def self.is_beta_version(version)
+
+        # Check if this versionName corresponds to a beta, i.e. contains some `-rc` suffix
+        #
+        # @param [String] version The versionName string to check for
+        #
+        # @return [Bool] True if the version string contains `-rc`, indicating it is a beta version.
+        #
+        #public
```
jkmassel: Same thing here, though not sure if this is correct or not?

```suggestion
        # @public
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
jkmassel: Also from a consistency point of view – all of the method docs end in a `.` (like a sentence) where these don't– not sure if intentional though 🙂

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -62,21 +128,53 @@ def self.calc_next_alpha_version(version, alpha_version)
...
         end
   
+        # Compute the version name and code to use for the next beta (`X.Y.Z-rc-N`).
+        #
+        # - The next version name corresponds to the `version`'s name with the value after the `-rc-` suffix incremented by one,
+        #     or with `-rc-1` added if there was no previous rc suffix (if `version` was not a beta but a release)
+        # - The next version code corresponds to the `alpha_version`'s (or `version`'s if `alpha_version` is nil) code, incremented by one.
+        #
+        # @example
+        #   calc_next_beta_version({"name": "1.2.3", "code": 456}) #=> {"name": "1.2.3-rc-1", "code": 457}
```
jkmassel, line 138: The syntax highlighting on this example is kind of strange, and I'm not sure why:

Screen Shot 2021-02-01 at 12 45 35 PM

It doesn't look like anything is wrong with the example code though

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -89,15 +187,45 @@ def self.calc_next_release_base_version(version)
           { VERSION_NAME => "#{vp[MAJOR_NUMBER]}.#{vp[MINOR_NUMBER]}", VERSION_CODE => version[VERSION_CODE] }
         end
 
+        # Compute the name of the next version to use after code freeze, by incrementing the current version name and making it a `-rc-1`
+        #
+        # @example
+        #   calc_next_release_version({"name": "1.2", "code": 456}) #=> {"name":"1.3-rc-1", "code": 457}
```
jkmassel, line 193: Same highlighting issue here, which is what makes me think the problem is Yard, not the code.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -127,81 +265,162 @@ def self.bump_version_release
           return verified_version
         end
 
+        # Update the build.gradle file with new versionName and versionCode values, both or the defaultConfig and vanilla flavors
```
jkmassel:
```suggestion
        # Update the build.gradle file with new versionName and versionCode values, both or the `defaultConfig` and `vanilla` flavors
```

In other places you've used code-quotes for `defaultConfig` –not sure if you want to stay consistent here? Same note for `vanilla`.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -127,81 +265,162 @@ def self.bump_version_release
...
           self.update_version(new_version_beta, ENV["HAS_ALPHA_VERSION"].nil? ? "defaultConfig" : "vanilla {")
           self.update_version(new_version_alpha, "defaultConfig") unless new_version_alpha.nil?
         end
 
+        # Compute the name of the previous hotfix version.
+        #
+        # @param [String] version_name The current version name we want to decrement
+        #
+        # @return [String] The predicted previous hotfix version, in the form of "X.Y.Z", or "X.Y" if Z is 0.
+        #         Corresponds to decrementing the 3rd component Z of the version, striping it if it ends up being zero.
```
jkmassel:
```suggestion
        #         Corresponds to decrementing the 3rd component Z of the version, stripping it if it ends up being zero.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -127,81 +265,162 @@ def self.bump_version_release
...
+        # The path to the build.gradle file for the project.
+        #
+        # @env PROJECT_ROOT_FOLDER The path to the root of the project (the folder containing the `.git` directory).
+        # @env PROJECT_NAME The name of the project, i.e. the name of the subdirectory containing the project's `build.gradle` file.
+        #
+        # @return [String] The path of the `build.gradle` file inside the project subfolder in the project's repo
+        #
         def self.gradle_path 
-          "#{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]}/build.gradle"
+          File.join(ENV["PROJECT_ROOT_FOLDER"] || '.', ENV["PROJECT_NAME"], 'build.gradle')
```
jkmassel: If I'm not mistaken, wouldn't this would presume that the root `build.gradle` is in the `fastlane` directory if `ENV["PROJECT_ROOT_FOLDER"]` isn't set?

I'm not sure that matches the setup of our repos.

IMHO it would make sense to make this a failing error – it could explode with a helpful error message if `ENV["PROJECT_ROOT_FOLDER"] ` isn't available?

WDYT?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -127,81 +265,162 @@ def self.bump_version_release
...
 
-        # FIXME: This implementation is very fragile. This should be done parsing the file in a proper way. 
-        # Leveraging gradle itself is probably the easiest way.
+        # Update both the versionName and versionCode of the build.gradle file to the specified version.
+        #
+        # @param [Hash] version The version hash, containing values for keys "name" and "code"
+        # @param [String] section The name of the section to update in the build.gradle file, e.g. "defaultConfig" or "vanilla"
+        #
+        # @todo This implementation is very fragile. This should be done parsing the file in a proper way. 
+        #       Leveraging gradle itself is probably the easiest way.
```
jkmassel, line 424: > Leveraging gradle itself is probably the easiest way.

👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
```diff
@@ -1,18 +1,40 @@
 module Fastlane
     module Helpers
+      # A module containing helper methods to manipulate/extract/bump iOS version strings in xcconfig files
+      #
       module IosVersionHelper
+        # @!group The indices for various parts of the version parts array.
```
jkmassel: This is a different heading than on Android – it kind of helps clarify things, though same note applies for your consideration.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
```diff
@@ -41,55 +76,113 @@ def self.calc_prev_release_version(version)
...
           vp[HOTFIX_NUMBER] -= 1 unless vp[HOTFIX_NUMBER] == 0
           return "#{vp[MAJOR_NUMBER]}.#{vp[MINOR_NUMBER]}.#{vp[HOTFIX_NUMBER]}" unless vp[HOTFIX_NUMBER] == 0
           "#{vp[MAJOR_NUMBER]}.#{vp[MINOR_NUMBER]}"
         end
 
+        # Create an internal version number, for which the build number is based on today's date.
+        #
+        # @param [String] version The current version to create an internal version name for.
+        #
+        # @return [String] The internal version, in the form of "X.Y.Z.YYYYMMDD".
```
jkmassel, line 134: Not necessarily important to _this_ PR, but this assumes that we never have more than one beta in a day. Regrettably, this assumption is incorrect 😅

Might be worth addressing down the road?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
```diff
@@ -101,6 +194,11 @@ def self.bump_version_release
           return verified_version
         end
 
+        # Updates the app_version entry in the `Deliverfile`
+        #
+        # @param [String] new_version The new value to set the `app_version` entry to.
+        # @raise [UserError] If the Deliverfile was not found.
+        #
         def self.update_fastlane_deliver(new_version)
           fd_file = "./fastlane/Deliverfile"
```
jkmassel, line 201: This path seems mildly fragile, but not necessarily the job of this PR to fix it.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
```diff
@@ -131,8 +246,18 @@ def self.update_xc_config(file_path, new_version, new_version_short)
...
+        #----------------------------------------
+        private
+        #----------------------------------------
+
   
+        # Split a version string into its 4 parts, ensuring its parts count is valid
+        #
+        # @param [String] version The version string to split into parts
+        # @return [Array<String>] An array of exactly 4 elements, containing each part of the version string.
+        # @note If the original version string contains less than 4 parts, the returned array is filled with zeros at the end to always contain 4 items.
```
jkmassel, line 256: This answered my question about the line above as soon as I thought of it. Nice!

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -282,8 +282,8 @@ def draw_text_to_canvas(canvas, text, width, height, x_position, y_position, fon
       #
       # @example
       #
-      #.  image = open_image("image-path")
-      #.  mask  = open_image("mask-path")
+      #   image = open_image("image-path")
```
jkmassel, line 285: Thanks for cleaning up my mess here 😂

jkmassel (approved):
This generally looks good – I left some nitpicks for your consideration (as well as a few suggested fixes). 

One global note – the use of periods to demarcate sentences vs not – it seems somewhat inconsistent, but also it's IMHO a silly thing to get too caught up in and spend tons of time on.

I took a look at the method renaming and nothing seemed out of place there – I also checked in both WPAndroid and WooAndroid and neither had references to this version (though WPAndroid has its own copy of the Android version helpers file...)

I don't foresee this requiring a re-review, so feel free to merge when you feel the feedback is sufficiently considered and/or addressed.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: I already have on my list to get rid of those in a future PR and replacing them either with using symbols as keys as it's customary in Ruby (`:name` and `:code`, and don't bother having constants for them), or alternatively create a `Version = Struct.new(:name,:code)` type and use it everywhere in place of a Hash. But preferred to keep all that for a future PR than cram everything in here.

Also a `Android::Version` object with name and code properties + instance methods would probably make more sense than using a Helper in the first place, something like:

```ruby
class Version
  # [String] Doc here
  attr_reader :name
  # [String] Doc here
  attr_reader :code

  # @return [Version] Next release version
  def next_release
    …
  end
  …
end
```
(Or maybe make the internal properties be `:major, :minor, :hotfix` + `alpha?` and `suffix`, and make `def name` be computed from all that to assemble the parts; which would make all the internal math of incrementing values even easier)

That kind of change will make it easier to document without having to repeat the documentation for "name" and "code" and the Hash everywhere, and would feel much more idiomatic. That one is on my list, not sure when I'll get to it but definitively part of the plan.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: Tbh this are more remnants of personal/debug annotations to mark for myself which methods I explicitly and manually checked to see if they were called from outside of this file or not, and could later be put inside a `private` section when I will reorder the methods in the file (because in Ruby the `private` keyword applies to _all_ the methods that follow it until the end of the file/module/scope so we have to groupe them all at the end of the scope; something I plan to do in a future, separate PR)

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: Yeah I noticed that as well and no idea why. Probably an issue with the parser/renderer in yard?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: … and for `build.gradle` probably 😜

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: Good point and good idea

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: FWIW I'm not even sure I'll go down that rabbit hole during my cleanup project though. This FIXME has been here for a while and I've looked quickly at how to do it on the net and all solutions involve modifying the .gradle files themselves to add a task that would print the value. Which would make our lanes depend on the presence of such task, and would not solve the other similar method and technique we use when writing/updating the version name and code… so probably gonna keep it there for a while still.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
mokagio: Definitely an issue with YARD

image

Looking at the HTML source, you can see that it somehow puts all of the text from the first `"` into a `string` `span`.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
```diff
@@ -1,18 +1,40 @@
...
           "#{vp[MAJOR_NUMBER]}.#{vp[MINOR_NUMBER]}.#{vp[HOTFIX_NUMBER]}"
         end
   
+        # Compute the name of the next release version.
+        #
+        # @param [String] version The current version that we want to increment
+        #
+        # @return [String] The predicted next version, in the form of "X.Y".
+        #         Corresponds to incrementing the minor part, except if it reached 10
+        #         (in that case we go to the next major version, as decided in our versioning conventions)
```
mokagio, line 34: I remember searching for a link to these conventions but without luck. The best I could come up with was [the WordPress versions page](https://codex.wordpress.org/WordPress_Versions).

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_build_prechecks.rb
```diff
@@ -14,7 +14,7 @@ def self.run(params)
         beta_version = Fastlane::Helpers::AndroidVersionHelper.get_release_version() unless !params[:beta] and !params[:final]
         alpha_version = Fastlane::Helpers::AndroidVersionHelper.get_alpha_version() unless !params[:alpha]
         
-        if (params[:final] and Fastlane::Helpers::AndroidVersionHelper.is_beta_version(beta_version))
+        if (params[:final] and Fastlane::Helpers::AndroidVersionHelper.is_beta_version?(beta_version))
```
mokagio, line 17: Yey for using Ruby's `?` and `!`s method suffixes.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
```diff
@@ -101,6 +194,11 @@ def self.bump_version_release
           return verified_version
         end
 
+        # Updates the app_version entry in the `Deliverfile`
```
mokagio:
```suggestion
        # Updates the `app_version` entry in the `Deliverfile`
```

.gitignore
```diff
@@ -10,16 +10,19 @@ fastlane/README.md
 fastlane/report.xml
 
 # Testing
-/test_results
-/coverage
+test_results/
+coverage/
+
+# YARD Documentation
```
mokagio, line 16: +1 for writing "YARD" with all upper case letters, because it's an acronym (Yay! Another Ruby Documentation tool) 🤓

mokagio (approved):
Note: given @jkmassel already review this, I only gave it a quick pass.

Look good. Thank you for doing this. I bet you have a much better understanding than I do of how the toolkit works now. And I'm grateful I'll be able to benefit from it by reading the documentation if I'll land on this code in the future 🙇‍♂️

---

As an aside, looking at some of the undocumented code I noticed some `=> Object` when the return type is actually a `String`, e.g.:

image

That's obviously due to missing documentation and is not an error per-se, but it made me miss the types I got so used to with Swift, so I opened [this issue for discussion](https://github.com/wordpress-mobile/release-toolkit/issues/201).

.gitignore
AliSoftware: Yeah, I like [going the extra (0.00056818) mile](https://www.google.com/search?q=1+yard+in+miles) by using correct case 🥁 😆

.yardopts
AliSoftware: My OCD does indeed 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
AliSoftware: Tracked by https://github.com/wordpress-mobile/release-toolkit/issues/202

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: Tracked in https://github.com/wordpress-mobile/release-toolkit/issues/203

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: Addressed in 0813fe61d73fbbf2263a961a7d85a027a4ac4dac

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
AliSoftware: I actually ended up removing it, as (1) it doesn't add much value in the generated doc, and (2) it will likely disappear anyway once I get to https://github.com/wordpress-mobile/release-toolkit/issues/203

PR #199, opened 2021-02-01
[Toolkit Cleanup] Yard doc: step 2 – rename module Helpers -> Helper

AliSoftware:
This PR sits on top of https://github.com/wordpress-mobile/release-toolkit/pull/198 and simply renames the `module Helpers` into `module Helper` for consistency both in code and in the generated documentation.

> ⚠️  This PR is ready for review but PLEASE DO NOT MERGE it yourself. Once it's approved I will take care of merging it as it should only be merged after #198 is itself merged.

### Why?

* In fastlane's own source code, the helpers live inside `Fastlane::Helper` (singular).
* In our codebase, some helpers are named under `Fastlane::Helper` but others are named `Fastlane::Helpers`.

This discrepancy makes the code inconsistent but also makes the YARD documentation incoherent, by splitting some modules and helpers in one module and other helpers in another.

### How?

This is mainly a "Find & Replace All" on the codebase for `module Helpers` and `Fastlane::Helpers` to its singular form instead. I made it a dedicated PR for easier review because it touches many files and already makes quite a big diff all by itself.

### To Test

* Review the code changes
* Run `bundle install && bundle exec rspec` to run the existing tests and ensure they still compile and pass
* Optionally you could try running a couple of actions (`bundle exec fastlane run <action_name>` to ensure that they still work, though be careful about potential side effects triggered by those actions (we wouldn't want to cut a release or push a new beta unexpectedly)
* Run `bundle exec yard doc` to generate the documentation. Open `yard-doc/index.html` in your browser and ensure that the modules and helpers are consistently classified under a common `Fastlane` > `Helper` hierarchy

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
@@ -9,17 +9,17 @@ module Fastlane
         require_relative '../../helper/android/android_git_helper.rb'
 
         # Checkout develop and update
-        Fastlane::Helpers::AndroidGitHelper::git_checkout_and_pull("develop")
+        Fastlane::Helper::AndroidGitHelper::git_checkout_and_pull("develop")
 
         # Check versions
-        release_version = Fastlane::Helpers::AndroidVersionHelper::get_release_version
-        message = "The following current version has been detected: #{release_version[Fastlane::Helpers::AndroidVersionHelper::VERSION_NAME]}\n"
-        alpha_release_version = Fastlane::Helpers::AndroidVersionHelper::get_alpha_version
-        message << "The following Alpha version has been detected: #{alpha_release_version[Fastlane::Helpers::AndroidVersionHelper::VERSION_NAME]}\n" unless alpha_release_version.nil?
+        release_version = Fastlane::Helper::AndroidVersionHelper::get_release_version
+        message = "The following current version has been detected: #{release_version[Fastlane::Helper::AndroidVersionHelper::VERSION_NAME]}\n"
+        alpha_release_version = Fastlane::Helper::AndroidVersionHelper::get_alpha_version
+        message << "The following Alpha version has been detected: #{alpha_release_version[Fastlane::Helper::AndroidVersionHelper::VERSION_NAME]}\n" unless alpha_release_version.nil?
         
         # Check branch
-        app_version = Fastlane::Helpers::AndroidVersionHelper::get_public_version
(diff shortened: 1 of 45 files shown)


AliSoftware:
⚠️ This PR is ready for review but PLEASE DO NOT MERGE it yourself. Once it's approved I will take care of merging it as it should only be merged _after_ #198 is itself merged.

jkmassel (approved):
This is a very nice alignment change. Review notes:

✅  Review the code changes [ they were very boring and exactly what I expected]
✅  [...] Run the existing tests and ensure they still compile and pass
✅  Run bundle exec yard doc to generate the documentation. Open yard-doc/index.html in your browser and ensure that the modules and helpers are consistently classified under a common Fastlane > Helper hierarchy. [they were, and this does make it significantly nicer]

I didn't run any lanes, as I have confidence that this change works well – I instead used `rg helpers && rg Helpers` to try to find any unresolved instances of this. All that came back were comments and files in the .xcproject for `drawText`.

One thing to note – we'll likely need to update WPAndroid for this, as it contains calls to these helpers (though AFAICT only its own local copies – the `Fastfile` doesn't seem to reference them, so it might just be possible to delete the `helpers` directory from that project altogether – just noting it here because we should take special care when bringing this update to that project).

loremattei (approved):
LGTM!
I double checked that there are no remaining instances of `Helpers` in the project. 

@jkmassel raised a very good point about WPAndroid. I'd say it's a good time to remove all the local instances and pull everything in the toolkit. IIRC, for the most part we can just get rid of the local instances and call the toolkit equivalent without any other chance.

AliSoftware:
That's a good point. I'll definitively try to get rid of the versions in WPAndroid.

Unfortunately quickly `diff`-ing between `WordPress-Android/fastlane/helper/android_*_helper.rb` and `release-toolkit/…/helper/android/*_helper.rb` files I can see the diff is quite not empty and some changes in code are substantial, so I'll have to take some time to compare them and to ensure the code is supposed to behave the same (hopefully with the one in release-toolkit being more modern/recent) before blindly deleting the one in WPAndroid.
Also deleting the local helpers will mean either also immediately deleting the local actions in WPAndroid repo too (so, will need to `diff` and compare them point by point with the release-toolkit ones too first, as with helpers), or to first modify the local actions to call the release-toolkit helpers as a first step, before deleting the local actions as a second, iterative step.

So might not be as trivial / immediate (I mean I could also just blindly delete the local actions and helpers, sacrifice a lamb and pray that using the release-toolkit ones will behave the same… but I prefer to be safe than sorry 😅, even if that means taking more time to double-check). But definitively on my roadmap for sure.

AliSoftware:
WPAndroid local helpers and actions removal tracked in https://github.com/wordpress-mobile/release-toolkit/issues/204

PR #200, opened 2021-02-01
[Toolkit Cleanup] Yard doc: step 3 - reorganize modules for helpers

AliSoftware:
This PR is part of the Toolkit cleanup and builds on top of #199 and part of issue #192.

Its goal is to reorganize `module` declarations, to make them consistent and so that they create a better organized module tree structure.

> ⚠️  This PR is ready for review but PLEASE DO NOT MERGE it yourself. Once it's approved I will take care of merging it as it should only be merged after #199 is itself merged.

### Why?

When generating the documentation, it's more readable to have all the code related to iOS all grouped in dedicated `Fastlane::Helper::Ios::*` modules (and likewise for Android), so that when looking at the documentation, the are all under the same tree / module, as opposed to be in a higher-level `Fastlane::Helper` module and using the prefix in the module name as differentiator.

image

### How

* Refactored every `module AndroidXYZ` declaration to split it into `module Android` + `module XYZ`, and replaced all calls to `AndroidXYZ::Foo` with `Android::XYZ::Foo`
* Likewise for `module IosXYZ` modules
* Moved `android_localize_helper` from the `helper/` root folder to its `helper/android` subfolder
* Note: I didn't yet move `helper/an_metadata_update_helper.rb` in this PR because this one will likely require separate treatment (it could be renaming the `an_` prefix to `android` and namespacing the whole module, but it might also disappear completely, as it doesn't seem specific to Android after all and the other file `metadata_update_helper.rb` seems to be a duplicate; but I'll need to triple-check that there's no subtle differences before unifying them and tidy all that up.)
* I also only reorganized those modules for helpers only now, because those helpers are private to the actions and not exposed so those should be non-breaking changes, as compared to moving and renaming fastlane actions. I will consider reorganizing modules for actions too in a separate PR, but doing this for actions has bigger consequences because it might be a breaking for Fastfiles using them.

### To Review

I suggest you to enable to "Ignore Whitespaces" mode when reviewing this PR's diff, given that the addition of a level of modules means an additional level of indentation, which makes every like shown as changed if you don't ignore whitespaces

### To Test

* Run `bundle exec rspec` and ensure the tests are still passing.
* Run `bundle exec yard doc`
* Open the generated documentation `open yard-doc/index.html` in your browser
* Ensure the documentation is still consistent. For now only the `VersionHelper` submodules in `Ios` and `Android` modules for `Helper` are fully documented, but we want to ensure than even the `Helper` submodules that are not fully documented yet are at least classified in the proper hierarchy. Note that this applies only to helpers for now (not to actions)


AliSoftware:
⚠️ This PR is ready for review but PLEASE DO NOT MERGE it yourself. Once it's approved I will take care of merging it as it should only be merged _after_ #199 is itself merged.

jkmassel (approved):
This LGTM in every way except one:

`module Ios` will always hurt my heart a little 😅

GitHub seems to be confused by the change to the `AndroidLocalizeHelper`, so I ran in through `diff` locally, and everything seems ok there. 

I also double-checked the codebase for references to the old helper names and they consistently came back empty.

In the WP and WC projects, these searches also came back empty (except for WPAndroid, which has its own copy of all the helpers, as noted in previous PRs)

AliSoftware:
Yeah I have no idea how we should case iOS as a module name to both respect the original name casing and the module naming conventions… we could call it `Apple` but that would make quite a big change especially if for consistency we then apply the same convention not only to helpers but also to actions… and I'm not yet ready to go that far at that stage of the refactoring 😅

Issue #203, opened 2021-02-02
Refactor *VersionHelpers to use an object


AliSoftware:
Today `ios_version_helper` and `android_version_helper` consist of freeform, module-static methods which take various and heterogeneous types of parameters depending on the method (sometimes a `String` for versionName, sometimes a `Hash`, etc)

It would make way more sense, and make the code way easier to understand, if instead of freeform methods, we created a `class Version` which would expose the appropriate instance properties (e.g. for Android, `name` and `code`) and instance methods to manipulate those versions, inspired by `Gem::Version` and its methods.

This would make the implementation way easier to understand and document (not having to repeat the parameter type documentation methods taking similar inputs), but also clearer at call-site and more idiomatic.

---

Some ideas / inspiration for possible API, e.g. for the Android one in the `Fastlane::Helper::Android` module:

```ruby
class Version
  # [String] Doc here
  attr_reader :name
  # [Int] Doc here
  attr_reader :code

  # @return [Version] Next release version
  def next_release
    …
  end
  …
end
```

Or, alternatively, something like:

```ruby
class Version
  attr_reader :major, :minor, :hotfix
  attr_reader :is_alpha?
  attr_reader :suffix
  attr_reader :code

  # @return [String] The versionName. `"alpha-major.minor.patch-suffix"` if `is_alpha?`, `"major.minor.patch-rc-suffix"` otherwise.
  def name
    (sprefix, *ssuffixes) = is_alpha? ? ["alpha", suffix] : [nil, "rc", suffix]
    [sprefix, "#{major}.#{minor}.#{patch}", *ssuffixes].compact.join('-')
  end
  …
  # And a lot of instance methods to bump versions up and down, etc
end
```

AliSoftware:
Another benefit of refactoring this to be objects, and to have instance methods on them, is that it would be easier to refactor that to not rely on the various env vars like `ENV['XX_CONFIG_FILE']`, and to instead take the paths for those config files as parameter to the constructor of those `Version` objects. Which would then allow us to easily call both getters and setters (get current version, set new version for upcoming beta or release, etc) on the instance without relying to any global state

AliSoftware:
Some work have already been started in [the `203-version` branch here](https://github.com/wordpress-mobile/release-toolkit/compare/develop...cleanup/203-version) to work on this refactoring.

[...]

PR #206, opened 2021-02-04
[Tooling/Cleanup] Cleanup git helpers

AliSoftware:
This PR cleans up all the `ios_git_helper` + `android_git_helper` + `git_helper` methods: refactoring + YARD doc.

## Why?

 - Many were duplicated in the platform-specific helper files while there was not reason not to DRY them in one common place
 - Many methods in there were actually not really related to git. There were doing other actions then pushing the changes to git all in one place, which is probably why they ended up in those git_helpers, but most of the actions those helper were doing were something other than git in the first place
 - Many methods didn't have very explicit name, and none of them were documented

## How?

 - Refactoring all the common methods into the common `git_helper`
 - Cleaning up the calls to avoid repetition (many of those ended up just calling core methods like `commit(message:files:push:)`
 - Update the call sites, and for some of the methods which were not that much related to git, move the atomic operations directly into the actions
 - Added YARD documentation

## To Review and Test

That's the trickiest part. This is very hard to test and we don't have `rspec` for most of those methods (see "Why not TDD" below) either.

The best we can do is:
 - `bundle exec rspec` to ensure the existing tests still pass (even if there are not many tests covering the changes made there)
 - `be yard stats --list-undoc lib/fastlane/plugin/wpmreleasetoolkit/**/*git_helper.rb` to check that everything about git helpers is documented (only items that are undocumented should be the `Fastlane`, `Fastlane::Helper`, and `Fastlane::Helper::[Ios|Android]` modules)
 - `bundle exec yard doc` to generate the doc, then `open yard-doc/index.html` and inspect the documentation for the helpers
 - Review the code manually to ensure I didn't introduce typos / syntax errors in the code and didn't miss any call sites when refactoring methods… things that we can only really check by proof-reading, given that Ruby is not type-checked… 😒 
   - I split my progress in nice isolated commits, so you might prefer doing a per-commit review instead of the whole shebang.
 - Optionally, point your `Pluginfile` to this branch and run a couple of actions (ideally the ones that would not risk having any side-effects) from an app repo to ensure those still work. Unfortunately, most of the actions and helpers impacted here do have side-effects, so that might be tricky…

### Please do thorough review 🙏

This PR should be ready to be reviewed, but I pondered about opening it as Draft or not because I didn't proof-read the whole thing after I finished everything.

* I did proof-read my individual commits and double-checked their changes, but didn't proof-read the whole PR (as I use to do) now that all the dust have settled (also because it's 11pm and my brain is nothing but a blur now…).
* I didn't actually test any of the changes either, so many of those changes were done while flying blind… which doesn't inspire me much confidence… _(That's the issue with a non-type-checked language that doesn't have unit tests 😓)_

This means that, for example, a review directly by reading diff in GitHub might not be enough to ensure that I didn't forget to update all call sites… Which is why I'd appreciate a more thorough review of those changes, especially about ensuring that the _behavior_ of helper methods are still unchanged (e.g. a parameter previously expecting just the version number for creating the release branch, but now expecting the full branch name or `release: version` instead…), there are no call sites I forgot to update, and there's no syntax error (e.g. missing comma) that slipped through…

_(At some point I'd love to add rubocop to at least catch simple things like stupid syntax errors or a missing commas… but that's whole PR on its own…)_

### Work in Progress

Things that I will probably fix in a separate incremental PR (except if I get to address those before that Draft PR gets its final review…):
* There are still a couple of methods in the platform-specific `git_helpers`, especially the helper methods around metadata+localization, and the one helper method to commit files related to version bumps (which are different set of files depending on the platforms, but are still worth DRYing in a helper rather than moving in each action because there are multiple actions using them)
* In particular, the `update_metadata` methods (iOS+Android) and the `localize_project` method (iOS) are still calling non-git-related code (invoking the script).

### Why not TDD?

I initially considered doing TDD and adding tests _first_, before refactoring. That's what tests are for after all.

Unfortunately, that ended up being very unproductive, given that the previous state of the code was really not easily testable and the amount of refactoring and thus the amount of methods (that I was gonna add tests for) that I was gonna delete, move and refactor entirely. That would have made any test written prior refactoring be obsolete by the end of it.

I know that it would have been ideal to write the tests prior to refactoring – that's the whole benefit of having tests after all – but that was too cumbersome to do, so I decided to keep the writing of the tests for a future PR addressing #193 .

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_tag_build.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_tag_build.rb
@@ -7,7 +7,8 @@ module Fastlane
   
         itc_ver = Fastlane::Helper::Ios::VersionHelper.get_build_version()
         int_ver = Fastlane::Helper::Ios::VersionHelper.get_internal_version() unless ENV["INTERNAL_CONFIG_FILE"].nil?
-        Fastlane::Helper::Ios::GitHelper.tag_build(itc_ver, int_ver)
+        Fastlane::Helper::GitHelper.create_tag(itc_ver)
+        Fastlane::Helper::GitHelper.create_tag(int_ver) unless int_ver.nil?
       end
   
       #####################################################
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -2,16 +2,6 @@ module Fastlane
   module Helper
     module Android
       module GitHelper
-        def self.update_metadata(validate_translation_command)
-          Action.sh("./tools/update-translations.sh")
-          Action.sh("git submodule update --init --recursive")
-          Action.sh("./gradlew #{validate_translation_command}")
-          Action.sh("git add #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]}/src/main/res")
-          Action.sh("git diff-index --quiet HEAD || git commit -m \"Updates translations\"")
-
-          Action.sh("git push origin HEAD")
-        end
-
         # Commit and push the files that are modified when we bump version numbers on an iOS project
         #
         # This typically commits and pushes the `build.gradle` file inside the project subfolder.
@@ -25,6 +15,16 @@ module Fastlane
             files: File.join(ENV["PROJECT_ROOT_FOLDER"], ENV["PROJECT_NAME"], "build.gradle"),
             push: true
           )
+        end
+
+        def self.update_metadata(validate_translation_command)
+          Action.sh("./tools/update-translations.sh")
+          Action.sh("git submodule update --init --recursive")
+          Action.sh("./gradlew #{validate_translation_command}")
+          Action.sh("git add #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]}/src/main/res")
+          Action.sh("git diff-index --quiet HEAD || git commit -m \"Updates translations\"")
+
+          Action.sh("git push origin HEAD")
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -53,11 +53,6 @@ module Fastlane
           Action.sh("git push origin HEAD")
         end
 
-        def self.tag_build(release_version, alpha_version)
-          Fastlane::Helper::GitHelper.create_tag(release_version)
-          Fastlane::Helper::GitHelper.create_tag(alpha_version) unless alpha_version.nil?
-        end
-
         def self.check_on_branch(branch_name)
           current_branch_name=Action.sh("git symbolic-ref -q HEAD")
           UI.user_error!("This command works only on #{branch_name} branch") unless current_branch_name.include?(branch_name)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -2,23 +2,6 @@ module Fastlane
   module Helper
     module Android
       module GitHelper
-        def self.git_checkout_and_pull(branch)
-          Action.sh("git checkout #{branch}")
-          Action.sh("git pull")
-        end
-
-        def self.git_checkout_and_pull_release_branch_for(version)
-          branch_name = "release/#{version}"
-          Action.sh("git pull")
-          begin
-            Action.sh("git checkout #{branch_name}")
-            Action.sh("git pull origin #{branch_name}")
-            return true
-          rescue
-            return false
-          end
-        end
-
         def self.do_release_branch(branch_name)
           if (check_branch_exists(branch_name) == true) then
             UI.message("Branch #{branch_name} already exists. Skipping creation.")
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -35,11 +35,6 @@ module Fastlane
           Action.sh("git commit -m \"Bump version number\"")
           Action.sh("git push origin HEAD")
         end
-
-        def self.check_on_branch(branch_name)
-          current_branch_name=Action.sh("git symbolic-ref -q HEAD")
-          UI.user_error!("This command works only on #{branch_name} branch") unless current_branch_name.include?(branch_name)
-        end
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -1,6 +1,8 @@
 module Fastlane
   module Helper
     module Android
+      # Helper methods to execute git-related operations that are specific to Android projects
+      #
       module GitHelper
         # Commit and push the files that are modified when we bump version numbers on an iOS project
         #
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -12,28 +12,12 @@ module Fastlane
           Action.sh("git push origin HEAD")
         end
 
-        def self.bump_version_release()
-          Action.sh("cd #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]} && git add ./build.gradle")
-          Action.sh("git commit -m \"Bump version number\"")
-          Action.sh("git push origin HEAD")
-        end
-
-        def self.bump_version_beta()
-          Action.sh("cd #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]} && git add ./build.gradle")
-          Action.sh("git commit -m \"Bump version number\"")
-          Action.sh("git push origin HEAD")
-        end
-
-        def self.bump_version_hotfix(version)
-          Action.sh("cd #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]} && git add ./build.gradle")
-          Action.sh("git commit -m \"Bump version number\"")
-          Action.sh("git push origin HEAD")
-        end
-
-        def self.bump_version_final()
-          Action.sh("cd #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]} && git add ./build.gradle")
-          Action.sh("git commit -m \"Bump version number\"")
-          Action.sh("git push origin HEAD")
+        def self.commit_version_bump()
+          Fastlane::Helper::GitHelper.commit(
+            message: "Bump version number",
+            files: File.join(ENV["PROJECT_ROOT_FOLDER"], ENV["PROJECT_NAME"], "build.gradle"),
+            push: true
+          )
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -12,6 +12,13 @@ module Fastlane
           Action.sh("git push origin HEAD")
         end
 
+        # Commit and push the files that are modified when we bump version numbers on an iOS project
+        #
+        # This typically commits and pushes the `build.gradle` file inside the project subfolder.
+        #
+        # @env PROJECT_ROOT_FOLDER The path to the git root of the project
+        # @env PROJECT_NAME The name of the directory containing the project code (especially containing the `build.gradle` file)
+        #
         def self.commit_version_bump()
           Fastlane::Helper::GitHelper.commit(
             message: "Bump version number",
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
@@ -2,23 +2,6 @@ module Fastlane
   module Helper
     module Android
       module GitHelper
-        def self.do_release_branch(branch_name)
-          if (check_branch_exists(branch_name) == true) then
-            UI.message("Branch #{branch_name} already exists. Skipping creation.")
-            Action.sh("git checkout #{branch_name}")
-            Action.sh("git pull origin #{branch_name}")
-          else
-            Action.sh("git checkout -b #{branch_name}")
-
-            # Push to origin
-            Action.sh("git push -u origin #{branch_name}")
-          end
-        end
-
-        def self.check_branch_exists(branch_name)
-          !Action.sh("git branch --list #{branch_name}").empty?
-        end
-
         def self.update_metadata(validate_translation_command)
           Action.sh("./tools/update-translations.sh")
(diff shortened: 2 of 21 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
```diff
@@ -17,14 +17,20 @@ def self.commit_version_bump()
...
-          Action.sh("./gradlew #{validate_translation_command}")
-          Action.sh("git add #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]}/src/main/res")
-          Action.sh("git diff-index --quiet HEAD || git commit -m \"Updates translations\"")
-
-          Action.sh("git push origin HEAD")
+          Action.sh("git", "submodule", "update", "--init", "--recursive")
+          Action.sh("./gradlew", validate_translation_command)
+          
+          res_dir = File.join(ENV["PROJECT_ROOT_FOLDER"], ENV["PROJECT_NAME"], "src", "main", "res")
+          Fastlane::Helper::GitHelper.commit(message: "Update translations", files: res_dir, push: true)
```
jkmassel, line 38: 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -44,16 +44,22 @@ def self.localize_project()
...
-          Action.sh("cd fastlane && ./download_metadata.swift")
-          Action.sh("git add ./fastlane/metadata/")
-          Action.sh("git diff-index --quiet HEAD || git commit -m \"Updates metadata translation\"")
+          strings_files = Dir.chdir(File.join(ENV["PROJECT_ROOT_FOLDER"], ENV["PROJECT_NAME"])) do
+            Dir.glob("*.lproj/*.strings")
+          end
+          Fastlane::Helper::GitHelper.commit(message: "Update translations", files: strings_files, push: false)
 
-          Action.sh("git push origin HEAD")
+          Action.sh("cd fastlane && ./download_metadata.swift")
```
jkmassel, line 68: This is a IMHO a high-value improvement opportunity – should we mark it with #todo or something? 😄

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -68,8 +70,13 @@ def self.commit(message:, files: nil, push: false)
         elsif !files.nil? && !files.empty?
           Action.sh("git", "add", *files)
         end
-        Action.sh("git", "commit", *args, "-m", message)
-        Action.sh("git", "push", "origin", "HEAD") if push
+        begin
+          Action.sh("git", "commit", *args, "-m", message)
```
jkmassel, line 84: I see that the args are variable, but is it possible to use the `GitHelper::commit` action here as we have in other places?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
```diff
@@ -12,28 +12,12 @@ def self.update_metadata(validate_translation_command)
...
-          Action.sh("cd #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]} && git add ./build.gradle")
-          Action.sh("git commit -m \"Bump version number\"")
-          Action.sh("git push origin HEAD")
-        end
-
-        def self.bump_version_final()
-          Action.sh("cd #{ENV["PROJECT_ROOT_FOLDER"]}#{ENV["PROJECT_NAME"]} && git add ./build.gradle")
-          Action.sh("git commit -m \"Bump version number\"")
-          Action.sh("git push origin HEAD")
+        def self.commit_version_bump()
```
jkmassel, line 14: Nice cleanup!

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -30,31 +30,27 @@ def self.checkout_and_pull(branch)
...
+      # Create a new branch named `branch_name`, cutting it from branch/commit/tag `from`, and push it
+      #
+      # If the branch with that name already exists, it will instead switch to it and pull new commits.
+      #
+      # @param [String] branch_name The full name of the new branch to create, e.g "release/1.2"
+      # @param [String?] from The branch or tag from which to cut the branch from.
+      #        If `nil`, will cut the new branch from the current commit. Otherwise, will checkout that commit/branch/tag before cutting the branch.
+      # @param [Bool] push If true, will also push the branch to `origin`, tracking the upstream branch with the local one.
+      #
+      def self.create_branch(branch_name, from: nil, push: true)
```
jkmassel, line 52: I like that this is now consolidated by its purpose, not it's use case – will definitely help reduce code and complexity!

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -49,7 +49,7 @@ def self.cut_release_branch(branch_name)
       # @param [String] tag_version The name of the tag to cut the hotfix from
       # @param [String] new_version The name of the new version, e.g. "1.2.3"
       #
-      def self.create_branch_for_hotfix(tag_version, new_version)
+      def self.cut_hotfix_branch(tag_version, new_version)
```
jkmassel: Nitpick – given the previous change to name things by purpose rather than use case, I wonder if this would make more sense as `branch_from_tag` or something?

Also – it seems like `create_branch` might already be able to do this from a tag, so do we need it at all?

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -30,6 +30,17 @@ def self.checkout_and_pull(branch)
         return false
       end
 
+      def self.cut_release_branch(branch_name)
+        if branch_exists?(branch_name)
+          UI.message("Branch #{branch_name} already exists. Skipping creation.")
```
jkmassel, line 54: It seems like if I run `git checkout -b cleanup/git-helpers` in this branch, it already fails with `fatal: A branch named 'cleanup/git-helpers' already exists.`

Might be overthinking this, but could we use the `create_branch` in place of this?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_tag_build.rb
```diff
@@ -7,7 +7,8 @@ def self.run(params)
   
         itc_ver = Fastlane::Helper::Ios::VersionHelper.get_build_version()
         int_ver = Fastlane::Helper::Ios::VersionHelper.get_internal_version() unless ENV["INTERNAL_CONFIG_FILE"].nil?
-        Fastlane::Helper::Ios::GitHelper.tag_build(itc_ver, int_ver)
+        Fastlane::Helper::GitHelper.create_tag(itc_ver)
+        Fastlane::Helper::GitHelper.create_tag(int_ver) unless int_ver.nil?
```
jkmassel, line 11: Nitpick, but I wonder if it'd make sense to group the `itc_ver` and `int_ver` stuff together and only have one conditional? Might make it easier for future readers?

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -64,6 +64,16 @@ def self.commit(message:, files: nil, push: false)
         Action.sh("git", "commit", *args, "-m", message)
         Action.sh("git", "push", "origin", "HEAD") if push
       end
+
+      # Creates a tag for the given version, and optionally push it to the remote.
+      #
+      # @param [String] version The name of the tag to push, e.g. "1.2"
+      # @param [Bool] push If true 9the default), the tag will also be pushed to `origin`
```
jkmassel: `s/9the default)/(the default)` (GitHub isn't showing me the suggestions UI for some reason, sorry)

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -13,6 +13,23 @@ def self.has_git_lfs
...
 
+      # Switch to the given branch and pull its latest commits.
+      #
+      # @param [String,Hash] branch Name of the branch to pull.
+      #        If you provide a Hash with a single key=>value pair, it will build the branch name as `"#{key}/#{value}"`,
+      #        i.e. `checkout_and_pull(release: version)` is equivalent to `checkout_and_pull("release/#{version}")`.    
+      #
+      # @return [Bool] True if it succeeded switching and pulling, false if there was an error during the switch or pull.
+      #
+      def self.checkout_and_pull(branch)
```
jkmassel, line 34: I wonder if this would make more sense to have as `checkout_and_pull!` and crash if it doesn't work instead? 

I can't think of a lot of times where we'd want to carry on if this failed, but WDYT?

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -30,6 +30,20 @@ def self.checkout_and_pull(branch)
...
 
+      # Create a new branch in preparation to do a hotfix.
+      #
+      # - Cuts the new branch from the tag `tag_version`
+      # - The name of the new branch will be `release/#{new_verison}`
+      #
+      # @param [String] tag_version The name of the tag to cut the hotfix from
+      # @param [String] new_version The name of the new version, e.g. "1.2.3"
+      #
+      def self.create_branch_for_hotfix(tag_version, new_version)
```
jkmassel: Mentioned elsewhere already, but this seems to have overlap with `create_branch` – I wonder if there's some reuse possible here?

jkmassel (changes requested):
A few nitpicks for your consideration and one very minor actual fix.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
AliSoftware: Note that as said in the PR description:
 - I plan to at least extract that call to the script away from `ios_git_helper` in a subsequent PR (I actually hoped I'd be able to do that today but didn't have the time after all) – in which I will even try to completely get rid of those `[platform]_git_helper.rb` and the remaining things that I left there so far
 - I explicitly don't plan to make any changes in the way we do localization (i.e. via calling .swift or .sh scripts that are in project repos) during this Great Refactoring™ , as this will be part of a separate Project Thread that I plan to tackle after this one, but that will be a big project all on its own. That's why I plan to still keep the call to `./download_metadata.swift` so far and refactor this only in the next Project dedicated about Localization Tooling

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: This IS the `GitHelper::commit` method, this code extract you reviewed is part of that method's implementation details 😄

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: This seems to be a leftover of a previous intermediate step during my refactor. Did you review that change commit-per-commit? Normally this is supposed to be gone and replaced by `create_branch` in a later commit in this PR… 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_tag_build.rb
AliSoftware: Probably something I will pick up when I tackle https://github.com/wordpress-mobile/release-toolkit/issues/203

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware:
```suggestion
      # @param [Bool] push If true (the default), the tag will also be pushed to `origin`
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: Yeah, I mainly did that to mimick `commit(message:files:push:)` I think.

But in act we actually use that feature at some call sites – e.g. in `get_user_build_version` from `android_beta_build_prechecks.rb:53` – so it's still useful to have this return a `Bool` instead of crashing.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: Outdated diff, seems like you reviewed that code during a commit-by-commit review, but that is not supposed to be there anymore in the final PR diff 🤔

AliSoftware:
No idea why danger is suddenly failing in GitHub Actions?! 🧐 
[EDIT] Just amended and force-pushed the last commit (with no change but its commit message) and it's green again 🤷

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: Ditto https://github.com/wordpress-mobile/release-toolkit/pull/206#discussion_r571265425 – seems this comment was made on an outdated diff

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
jkmassel: 🤦

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
jkmassel: Oh for sure – I was thinking in the context of the project scope potentially shrinking that we may not get back to it any time soon, so it might be good to flag for future travellers, sorry!

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
jkmassel: Yep can confirm that's the case, sorry!

jkmassel (approved):
LGTM! :shipit:

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
AliSoftware: I might add a `@todo` indeed

mokagio:
I opened this but then realise I don't have enough time for a thorough review right now.

I just wanted to drop a not regarding "why not TDD?"

> I know that it would have been ideal to write the tests prior to refactoring – that's the whole benefit of having tests after all – but that was too cumbersome to do, so I decided to keep the writing of the tests for a future PR addressing #193 .

No worries. All the work you're doing is improving this foundational codebase and making it easy to add tests later 👍 

In app development, I'd say "Maybe write a few UI tests? Even if they hit the production APIs they'll still give you a bit of trust that nothing major broke during the refactor"

There could be a chance to do something similar here. Setup a test repo, install Fastlane, fetch this branch for the plugin, then write a few bash scripts that run the actions and check the how the repository changed.

Now that I ~said it out loud~ wrote it down, it sounds like a lot of work... Still, an idea worth putting out there. Maybe there's no need for it anymore or maybe there still is and that could help us in doing and testing the work. I don't know 🤷‍♂️

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
AliSoftware: Fixed in e3264c721dfc9ec5ca78adbd3e08ec04c9c9f919 :)

AliSoftware:
@mokagio I'm merging this to progress on my dependant PR chain but don't hesitate to git it a review even post-merge if want and have time, I can always address any feedback in another PR down the chain ;)

PR #207, opened 2021-02-08
[Tooling/Cleanup] Enable Rubocop and fixes first steps

AliSoftware:
Part of #208 

## What

This PR:
 - Updates the `.rubocop.yml` file with what I feel like sensible values for some rules.
   - Some of them might be a bit too lenient (especially method lengths and similar) but I figured it's better to start large and fix all the other ones (which require minimal, localized changes for most of them, as opposed to a class too long requiring bigger refactor)
 - Adds the `.rubocop_todo.yml` file, with the list of violations that we still need to fix

## How

This is only the first step of fixing rubocop violations. Basically my process is the following, which I will continue to use to iterate for further steps in the series of rubocop PRs:

 - Pick a category of violation in the `.rubocop_todo.yml` file that I feel is reasonable to fix
 - Remove that rule exception from the `.rubocop_todo.yml`
 - Run `bundle exec rubocop --auto-correct`
 - Commit the changes made by autocorrect in a commit named `"[rubocop] fix <rule name>"`
 - Wash, Rince, Repeat.

## PR dependency

This PR sits on top of https://github.com/wordpress-mobile/release-toolkit/pull/206 and thus targets its head branch `cleanup/git-helpers`.
There will be more PRs to come on top of this one, to add further rubocop violations fixes iteratively.

## How to Review

Given this, this PR might be better to be reviewed commit per commit, and potentially enable "Ignore Whitespace" during review too.

Alternatively, you can just focus on reviewing the list of enabled rules and trust that `rubocop --auto-correct` did its automated job correctly in fixing the violations, as it might be unnecessary to double-guess every change made by autocorrect itself…


.rubocop.yml
```diff
@@ -1,6 +1,48 @@
+inherit_from: .rubocop_todo.yml
+
 require: 
 - rubocop-rspec
 
-Metrics/BlockLength:
+########## Code Style
+
+# We are ok using single and double quotes interchangeably
```
jkmassel: _Are_ we though? 😁 

I'd vote to move this into the todo pile but if I'm alone in this, feel free to ignore me!

.gitignore
```diff
@@ -1,6 +1,9 @@
 .DS_Store
 /tmp
 
+# RubyMine IDE
```
jkmassel: nit:  this could be generated by any Jetbrains (or derived) product

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -5,8 +5,8 @@ module Fastlane
   UI = FastlaneCore::UI unless Fastlane.const_defined?("UI")
 
   module Helper
-    class GhhelperHelper
-      def self.GHClient()
+    class GithubHelper
```
jkmassel, line 8: 👍

jkmassel (approved):
This LGTM – I had a nit and a question, but neither are blocking – what's nice about this is we could always change our minds later on something any significant negative affects.

I'd say :shipit: and we can iterate if needed – reading through this, the most dangerous part IMHO is the `ghhelper` rename stuff.

I'm excited to have a linter to yell at me for this project!

jkmassel:
Also GitHub Actions is broken, but we can merge around that

.rubocop.yml
AliSoftware: Yeah maybe we should indeed.

mokagio: That's quite a hot topic in Ruby land IIRC 🔥 🍿

.rubocop.yml
```diff
@@ -1,6 +1,48 @@
...
+
+Metrics/AbcSize:
+  Max: 140
+
+Metrics/BlockLength:
+  Max: 80
+
+Metrics/ClassLength:
+  Max: 300
+
```
mokagio, line 25:
```suggestion
```

Extra new line here?

mokagio (approved):
Let's get this merged and start benefitting from a linter 🗣📯

Have you ever looked into using an [ignore revs file](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt). As far as I understand, it's a way to add this kind of high noise but no behavior change commits to a list that will ignore them when running `git blame`. I heard about it [in a podcast](https://www.bikeshed.fm/258), but you can read [this post's intro](https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame) to get an idea.

AliSoftware:
Ah yeah I heard about ignore revs file a while ago too but totally forgot about it… that might be a good idea indeed 👍  Will look into it, probably add it as part of the last steps once all violations have been addressed and I have a full list of commits to ignore for blame.

.rubocop.yml
```diff
@@ -1,6 +1,48 @@
...
+Metrics/ParameterLists:
+  Max: 10
+
+########## Complexity
+
+Metrics/CyclomaticComplexity:
+  Max: 20
+
+Metrics/PerceivedComplexity:
+  Max: 20
```
loremattei: Do we need a new line here?

loremattei (approved):
LGTM! 
Left just a nit commit. 
CI failed, but the error doesn't seem to be related to these changes. I restarted it. If it's green, I'm ok with this!

AliSoftware:
Addressed feedback so all should be good now, but please let me merge this on green, as I need to first address feedback then merge the dependant PR https://github.com/wordpress-mobile/release-toolkit/pull/206 first – which might actually fix the odd Danger CI failure, let's see – before I rebase this on on top of it for merge.

.rubocop.yml
AliSoftware: yup; fixed.

.gitignore
AliSoftware: fixed comment in b6ecae4cfab106309864f5ad8ccd5801220f913d

.rubocop.yml
AliSoftware: Moved it back into the todo rules list in e4edfda2caea356df27dfd9a8221b13c7ff2e30c

I'll probably actually enable it and apply autocorrect on it at some point, better be consistent even if the rule might sometimes feel nitpicky

AliSoftware:
Fixed the Danger / CI failure via d893cdf

Issue #208, opened 2021-02-10
Enable rubocop on the project and fix violations


AliSoftware:

AliSoftware:
Addressed via https://github.com/wordpress-mobile/release-toolkit/pull/246 a while ago.

PR #211, opened 2021-02-11
[Cleanup/rubocop] Step3: Spacing and Empty Lines

AliSoftware:
This continues work about #208 and builds on top of step 2.

This intermediate/incremental PR focuses on layout violations about Spacing and Empty Lines

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -10,7 +10,6 @@ module Fastlane
   module Actions
     class ConfigureApplyAction < Action
       def self.run(params = {})
-
         # Preflight
         UI.user_error!('Decryption key could not be found') if Fastlane::Helper::ConfigureHelper.encryption_key.nil?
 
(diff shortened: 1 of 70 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
```diff
@@ -77,10 +77,10 @@ def self.apply_file(file_reference, force)
           date_string = Time.now.strftime('%m-%d-%Y--%H-%M-%S')
 
           backup_path = base
-                        .concat('-')            # Handy-dandy separator
-                        .concat(date_string)    # date string to allow multiple backups
-                        .concat(extension)      # and the original file extension
-                        .concat('.bak')        # add the .bak file extension - easier to .gitignore
+                        .concat('-') # Handy-dandy separator
```
jkmassel, line 80: I'm not a huge fan of how it doesn't allow aligned comments like this, but it's worth the trade-off everywhere else 😅

jkmassel (approved):
LGTM – nice and easy!

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
AliSoftware: Yeah I saw that too. Thought about adding a `# rubocop:disable` inline exception for those lines but figured it was not worth it after all.

PR #212, opened 2021-02-15
Add action to trigger CI jobs via API

loremattei:
This PR adds a short action to trigger CI jobs via API calls in order to make it possible to move away from starting builds using tags. 

This can be tested as a part of https://github.com/wordpress-mobile/WordPress-iOS/pull/15890


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/circleci_trigger_job_action.rb
```diff
@@ -0,0 +1,59 @@
...
+            #####################################################
+  
+            def self.description
+                "Triggers a job on CircleCI"
+            end
+  
+            def self.available_options
+            [
+                FastlaneCore::ConfigItem.new(key: :circle_ci_token,
+                                            env_name: "FL_CCI_TOKEN",
```
AliSoftware: I suggest we use more explicit/descriptive env var names. It's already sometimes easy to get lost in release-toolkit when you don't have all the context, so seeing just `CCI` in a user's env vars config (e.g. `.env` files) or wherever this will be used to inject it (if users end up relying on the env var rather than passing the value directly) will end up being a bit obscure imho.

```suggestion
                                            env_name: "FL_CIRCLECI_TOKEN",
```

Of course same recommendation for all other `ConfigItems` below.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/circleci_trigger_job_action.rb
```diff
@@ -0,0 +1,59 @@
...
+                                            description: "The GitHub repository you want to work on",
+                                            is_string: true),  
+                FastlaneCore::ConfigItem.new(key: :branch,
+                                            env_name: "FL_CCI_BRANCH",
+                                            description: "The branch on which you want to work on",
+                                            is_string: true),
+                FastlaneCore::ConfigItem.new(key: :job_params,
+                                            env_name: "FL_CCI_JOB_PARAMS",
+                                            description: "Parameters to send to the CircleCI pipeline",
+                                            is_string: false,
```
AliSoftware: `is_string` is deprecated. You should use `type:` parameter instead. For example replace `is_string: true` with `type: String`; Replace `is_string: false` with the appropriate `type: <expected ruby type here>`.

The only use case left for `is_string: false` instead of `type: X` is when the expected type is a Boolean (because Ruby doesn't have formal `Boolean` _type name_ – only `TrueClass` and `FalseClass` and boolean literals)

For more information, see https://docs.fastlane.tools/advanced/actions/#using-fastlanecoreconfiguration

lib/fastlane/plugin/wpmreleasetoolkit/helper/ci_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+                @login = login
+                @organization = organization
+                @repository = repository
+            end 
+
+            def command_uri
+                URI.parse("https://circleci.com/api/v2/project/github/#{@organization}/#{@repository}/pipeline")
+            end
+
+            def trig_job(branch, parameters=nil)
```
AliSoftware: I suggest to use more explicit names, and avoid obscure or unclear method names.
```suggestion
            def trigger_job(branch, parameters=nil)
```

Also, it could be nice to have named ruby parameters, that makes the code more readable. You can do that by using `:` after the parameter name when declaring the method:
```suggestion
            def trigger_job(branch:, parameters: nil)
```
then update the call sites to `trigger_job(branch: 'develop', parameters: { … }`

lib/fastlane/plugin/wpmreleasetoolkit/helper/ci_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+                    'Accept' => 'application/json',
+                    'Circle-Token' => @login
+                }
+
+                Net::HTTP.start(command_uri.host, command_uri.port, use_ssl: true) do |http|
+                    request = Net::HTTP::Post.new(command_uri.request_uri, headers)
+                    body = { "branch": branch, "parameters": parameters }
+                    request.body = body.to_json
+                    response = http.request(request)
+                    return response
```
AliSoftware: Ruby has implicit returns, so this would be enough:
```suggestion
                    http.request(request)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ci_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+end
+
+module Fastlane
+    module Helper
+        class CircleCIHelper 
+            include CIHelper
+
+            attr_accessor :login, :organization, :repository
+
+            def initialize(login, repository, organization = "wordpress-mobile")
```
AliSoftware: Could be nice to make use of named parameters here too
```suggestion
            def initialize(login:, repository:, organization: "wordpress-mobile")
```

spec/ci_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        describe Fastlane::Helper::CircleCIHelper.new("my_circleci_token", "test_repo", "my_org") do 
+            it { is_expected.to have_attributes(:organization => "my_org", "repository" => "test_repo", "command_uri" => URI.parse("https://circleci.com/api/v2/project/github/my_org/test_repo/pipeline"))}
+        end
+    end
+
+    context "Main Commands" do
+        describe Fastlane::Helper::CircleCIHelper.new("my_circleci_token", "test_repo") do 
+            it "Trig Job" do  
+                stub = stub_request(:post, "https://circleci.com/api/v2/project/github/wordpress-mobile/test_repo/pipeline").with(:body => {"branch":"develop","parameters":nil}, :headers => {'Content-Type' => 'application/json', 'Accept' => 'application/json', 'Circle-Token' => 'my_circleci_token'}).to_return(:body => "efg")
```
AliSoftware, line 30: Nice! Thanks `webmock` 👍

AliSoftware (commented):
I know this is still a draft, but here's an initial review anyway.

In addition to inline comments:
 - You should make sure to add YARD documentation to all your methods at some point. I'm doing a pass to add it to existing methods in my `cleanup-toolkit` intermediate branch already, so better also add YARD doc to anything we add from now on
 - As a heads up, I haven't yet merged all my changes to the rubocop lint rules into `develop` but amongst the main rules you should try to apply on your own for new actions are: (1) using 2-spaces for indentation (2) use single quotes `'` instead of double-quotes when you don't use string interpolation

Will do a more in-depth pass after testing it alongside https://github.com/wordpress-mobile/WordPress-iOS/pull/15890

mokagio:
Love that this is happening! (_Nothing to add to @AliSoftware's comments_ 😅)

lib/fastlane/plugin/wpmreleasetoolkit/helper/ci_helper.rb
loremattei: yeah! I don't really like it. Explicitly using return seems to be better to communicate what to expect.

loremattei:
Thanks for the thoughtful review @AliSoftware!
This should be ready for another look.

spec/ci_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+describe Fastlane::Helper::CircleCIHelper do
+    it_behaves_like "a CI provider" do 
+        let(:model) { Fastlane::Helper::CircleCIHelper.new(login: "my_circleci_token", repository: "test_repo") }
+    end
+
+    context "initialization" do 
+        describe Fastlane::Helper::CircleCIHelper.new(login: "my_circleci_token", repository: "test_repo") do 
+            it { is_expected.to have_attributes(:organization => "wordpress-mobile", "repository" => "test_repo", "command_uri" => URI.parse("https://circleci.com/api/v2/project/github/wordpress-mobile/test_repo/pipeline"))}
+        end
```
AliSoftware, line 20: TIL we can provide an _instance_ to `describe` and use it inside `it { … }` block. I always only used module/class names like `describe Fastlane::Helper::CircleCIHelper do` only. 🤯

AliSoftware (approved):

spec/ci_helper_spec.rb
mokagio: That's pretty cool (and as meta as only Ruby can get ❤️ )

PR #213, opened 2021-02-18
Cleanup - Step 4

AliSoftware:
Continuing work on #207

This time applying some fixes manually as some of the rules I enabled not being `--auto-correct`-able.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,13 +3,20 @@ inherit_from: .rubocop_todo.yml
 require: 
 - rubocop-rspec
 
+########## Lint / CodeStyle
+
 # This gem is included as part of the `Pluginfile`, which is `required` separately,
 # so makes it hard to keep it ordered with the rest of the gems
 Gemspec/OrderedDependencies:
   Exclude:
     - 'fastlane-plugin-wpmreleasetoolkit.gemspec'
 
-########## Max Lengths Rules
+# Especially for actions and methods like `self.run(params)` it's common to have the argument unused; but that
+# comes with the parent class API / contract, so we feel it more readable to avoid the _ unused prefix for those.
+Lint/UnusedMethodArgument:
+  Enabled: false
+
+########## Metrics / Max Lengths Rules
 
 Metrics/LineLength:
   Max: 300
@@ -32,7 +39,7 @@ Metrics/ModuleLength:
 Metrics/ParameterLists:
   Max: 10
 
-########## Complexity
+########## Metrics / Complexity
 
 Metrics/CyclomaticComplexity:
   Max: 20
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb
@@ -37,7 +37,7 @@ module Fastlane
         target = self.create_target_file_path(orig)
 
         # Clear if older exists
-        File.delete(target) if File.exists? target
+        File.delete(target) if File.exist? target
 
         # Create the new one
         begin
@@ -47,7 +47,7 @@ module Fastlane
             end
           end
         rescue
-          File.delete(target) if File.exists? target
+          File.delete(target) if File.exist? target
           raise
         end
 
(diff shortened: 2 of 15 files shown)


.rubocop.yml
```diff
@@ -3,13 +3,20 @@ inherit_from: .rubocop_todo.yml
...
 # so makes it hard to keep it ordered with the rest of the gems
 Gemspec/OrderedDependencies:
   Exclude:
     - 'fastlane-plugin-wpmreleasetoolkit.gemspec'
 
-########## Max Lengths Rules
+# Especially for actions and methods like `self.run(params)` it's common to have the argument unused; but that
+# comes with the parent class API / contract, so we feel it more readable to avoid the _ unused prefix for those.
+Lint/UnusedMethodArgument:
+  Enabled: false
```
AliSoftware, line 17: iow if we did enable this, we would have to write `def self.run(_params)` and similar everywhere, naming each unused parameters with a leading underscore. Which felt odd to me especially on contract-abiding methods like `Action` subclasses

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -196,7 +196,7 @@ def find_diffs(input_dir:, base_lang:, only_langs: nil)
         # Returns true if the file only contains empty lines, i.e. lines that only contains whitespace (space, tab, CR, LF)
         def only_empty_lines?(file)
           File.open(file) do |f|
-            while line = f.gets
+            while (line = f.gets)
```
jkmassel, line 199: My Ruby might not be up-to-par, but I'm not sure I understand the difference here 😅 

But if the linter is happy with it, I am too if it's functionally identical

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb
```diff
@@ -71,17 +71,17 @@ def self.create_block_parsers(release_version, block_files)
         @blocks = Array.new
 
         # Inits default handler
-        @blocks.push (Fastlane::Helper::UnknownMetadataBlock.new)
+        @blocks.push Fastlane::Helper::UnknownMetadataBlock.new
```
jkmassel, line 74: I'm not sure I understand this one – `push` is a function, so AFAICT shouldn't the brackets be allowed? 

I don't object to it – I just don't understand 😅

jkmassel (approved):
LGTM – had two questions though

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: That's a fair question. The parentheses are needed here mainly to silence a warning and be explicit that it's not a mistake that we do both affectation `=` … and comparison (interpreting the whole expression as a Boolean / truthiness test).

So the linter wants to be sure we didn't mean `==` and accidentally type `=`. Adding the parentheses achieves this, by making the affectation a sub-expression in parentheses that is then interpreted as a (truthiness) value of its own when interpreted as the condition of the `while`.

You can see it as a more compact way of writing `while (line = f.gets) != nil` – except I _think_ (?) that the linter would not recommend this alternative and tell us that the `!= nil` is implicit and can be removed… so here goes.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb
AliSoftware: Yeah that one surprised me as well when the linter warned about it.

The thing is, we could have kept the parentheses, but if we did, we would have needed to remove the spaces between `push` and the open bracket, otherwise it's seen as a sub-expression

```ruby
# Here it's interpreted as a function call on which we just omit the parentheses around parameters, OK
@blocks.push Fastlane::Helper::UnknownMetadataBlock.new

# Here it's interpreted as a function call on which we explicitly use the parentheses around parameters, still OK
@blocks.push(Fastlane::Helper::UnknownMetadataBlock.new)

# But here it's interpreted as a function call on which we omit the parentheses around parameters…
# and then pass a single parameter to it that happens to be wrapped in parentheses, i.e. interpreted as a sub-expression
@blocks.push (Fastlane::Helper::UnknownMetadataBlock.new)

# In a way, the syntax above is seen as equivalent to this, where the parameter would itself be
# in parentheses making it be needlessly interpreted as a sub-expression:
@blocks.push((Fastlane::Helper::UnknownMetadataBlock.new))
```

PR #214, opened 2021-02-18
[rubocop] Step 5

AliSoftware:
Continuing work on #207 and rubocop.

⚠️ Some of the rules that have been incrementally enabled in this PR were not auto-corrected by rubocop but were corrected manually – so will need more attentive review.

This also includes a small refactor (consisting of the extraction of long portion of code into a separate method, to reduce the code complexity and method length, see https://github.com/wordpress-mobile/release-toolkit/pull/214/commits/f84285d778a7d770c7adcf9fc836784777ef705e) that might also need careful review – especially to ensure that all the variables used in the original code were all passed to the method and that the extracted code didn't modify any variable that would then later be used after it in the original code, etc.

The corresponding commits are labeled accordingly (`[rubocop] fix … (manual)` and the like) – in case you want to review commit-per-commit.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,6 +19,9 @@ Lint/UnusedMethodArgument:
 Style/EmptyMethod:
   EnforcedStyle: expanded
 
+Style/TrailingCommaInArrayLiteral:
+  EnforcedStyleForMultiline: consistent_comma
+
 # Enabling the 'frozen_string_literal: true' comment is nice but should be done with extra care
 # because enabling it on code that accidentally mutates string literals will crash at runtime
 # (e.g. `name = 'Hello'` then later `name << ' world'` will crash if the comment is enabled)
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,6 +19,13 @@ Lint/UnusedMethodArgument:
 Style/EmptyMethod:
   EnforcedStyle: expanded
 
+# Enabling the 'frozen_string_literal: true' comment is nice but should be done with extra care
+# because enabling it on code that accidentally mutates string literals will crash at runtime
+# (e.g. `name = 'Hello'` then later `name << ' world'` will crash if the comment is enabled)
+# So we might enable it at some point, be we will need very careful review that this doesn't break everything.
+Style/FrozenStringLiteralComment:
+  Enabled: false
+
 ########## Metrics / Max Lengths Rules
 
 Metrics/LineLength:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,9 @@ Gemspec/OrderedDependencies:
 Lint/UnusedMethodArgument:
   Enabled: false
 
+Style/EmptyMethod:
+  EnforcedStyle: expanded
+
 ########## Metrics / Max Lengths Rules
 
 Metrics/LineLength:
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -435,20 +435,6 @@ Style/SpecialGlobalVars:
   Exclude:
     - 'lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb'
 
-# Offense count: 12
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle.
-# SupportedStyles: single_quotes, double_quotes
-Style/StringLiteralsInInterpolation:
-  Exclude:
-    - 'lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb'
-    - 'lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_update_metadata_source.rb'
-    - 'lib/fastlane/plugin/wpmreleasetoolkit/actions/common/promo_screenshots_action.rb'
-    - 'lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb'
-    - 'lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb'
-    - 'lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb'
-    - 'lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_update_helper.rb'
-
 # Offense count: 4
 # Cop supports --auto-correct.
 # Configuration parameters: .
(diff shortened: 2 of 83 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -410,14 +410,14 @@ def resolve_path(path)
       end
 
       def resolve_text_into_path(text, locale)
-        localizedFile = sprintf(text, locale)
+        localizedFile = format(text, locale)
```
jkmassel, line 387: I had to look up `format`, but it looks like this is a mechanical transformation – functionally identical to `sprintf` in this context. 👍

.rubocop.yml
```diff
@@ -19,6 +19,13 @@ Lint/UnusedMethodArgument:
 Style/EmptyMethod:
   EnforcedStyle: expanded
 
+# Enabling the 'frozen_string_literal: true' comment is nice but should be done with extra care
+# because enabling it on code that accidentally mutates string literals will crash at runtime
+# (e.g. `name = 'Hello'` then later `name << ' world'` will crash if the comment is enabled)
+# So we might enable it at some point, be we will need very careful review that this doesn't break everything.
```
jkmassel, line 28: 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
```diff
@@ -137,12 +137,16 @@ def initialize(block_key, content_file_path, release_version)
...
         values = key.split('_')
-        key.start_with?(@rel_note_key) && values.length == 4 && (Integer(values[3].sub(/^[0]*/, '')) != nil rescue false)
+        key.start_with?(@rel_note_key) && values.length == 4 && is_int?(values[3].sub(/^[0]*/, ''))
       end
 
       def generate_block(fw)
         super(fw) unless File.zero?(@content_file_path)
       end
+
+      def is_int?(value)
```
jkmassel, line 147: This looks ok to me – functionally identical to what came before.

The conditionals that reference this though IMHO could do with some cleanup and documentation though in the future 😅

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/promo_screenshots_action.rb
```diff
@@ -238,6 +173,75 @@ def self.available_options
       def self.is_supported?(platform)
         true
       end
+
+      def self.build_entries(config, languages, output_directory, params)
```
jkmassel: I like this, but AFAICT `config` isn't used directly and we could just pass the entries. 

Might not be worth fixing just now

spec/ios_lint_localizations_spec.rb
```diff
@@ -37,7 +37,7 @@
             fi
           SCRIPT
           FileUtils.mkdir_p File.join(install_dir, 'bin')
-          File.write(File.join(install_dir, 'bin/swiftgen'), script, perm: 0766)
+          File.write(File.join(install_dir, 'bin/swiftgen'), script, perm: 0o766)
```
jkmassel, line 41: Is there a format that can do like `0x766` or something? 

Not a huge deal, but I feel like if I look at this in six months I'm gonna be like "what _is_ that string?"

But then I'd probably just look at the `File.write` call and figure it out in like 2 seconds, so not a huge deal – more just curious as it seems unusual.

.rubocop_todo.yml
```diff
@@ -464,13 +464,6 @@ Style/TernaryParentheses:
     - 'lib/fastlane/plugin/wpmreleasetoolkit/actions/common/extract_release_notes_for_version_action.rb'
     - 'lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb'
 
-# Offense count: 20
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyleForMultiline.
-# SupportedStylesForMultiline: comma, consistent_comma, no_comma
-Style/TrailingCommaInArrayLiteral:
```
jkmassel, line 772: I'm surprised by this one because typically we'd leave it in case we needed to add another entry later.

I presume this was just the less-destructive approach?

jkmassel (approved):
This LGTM – I left a few questions, but nothing blocking

spec/ios_lint_localizations_spec.rb
AliSoftware: > Is there a format that can do like 0x766 or something?

`0x` is the prefix for hexadecimal, while `0o` is the prefix for octal, which is what is used with `chmod`-like permissions. So `0o` is that format that can do it :P

Note: I looked at https://ruby-doc.org/core-2.5.0/File/Constants.html to see if there were constants we could bitwise-or to provide the same value in a more readable way, but those seems to be focused on providing constants for things like the open modes (read/read-write/append, text/binary) not on the Unix permissions.

Agree that since `0o` is not as widely used/known as `0x` it might need a quick comment though.

```suggestion
          # note: `0o` is octal notation, used to specify chmod-like flags
          File.write(File.join(install_dir, 'bin/swiftgen'), script, perm: 0o766)
```

.rubocop_todo.yml
AliSoftware: It was. But good point, we could switch the rule the other way and require them all the time, I'm fine with it too, will see how much changes it would make to turn it on.

AliSoftware: rule flipped in f6499f759a22dca36b72b136bca963c66b9f25cf, agree it's nicer that way

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
AliSoftware: Yep, I used `--safe-auto-correct` (as opposed to just `--auto-correct`) for that one and indeed checked too, they behave the same there (which the rule's autocorrect was smart enough to ensure apparently) 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
AliSoftware: Yeah, lots of ternary operator shenanigans I agree. I plan to tackle that probably during #203.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/promo_screenshots_action.rb
AliSoftware: Nice catch! Fixed in fd8d9bde9600caa7a283eb1e46600811c3a03d50

(That refactoring might go even further I guess, but didn't want to spend too much time on it this time around as I preffered to focus on move things forward in order to put rubocop stuff behind me ASAP. Might revisit this in the future)

AliSoftware:
Thanks for the review @jkmassel 👍  I should have addressed all your questions and suggestions 🙂

PR #217, opened 2021-02-24
[Cleanup/rubocop] 6

AliSoftware:
Quick additional round of fixes.

Note that I think I'll stop doing rubocop after that to focus on #203 and other improvements


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_merge_translators_strings.rb
```diff
@@ -9,7 +9,7 @@ class AndroidMergeTranslatorsStringsAction < Action
       def self.run(params)
         folder_path = File.expand_path(params[:strings_folder])
 
-        subfolders = Dir.entries("#{folder_path}")
+        subfolders = Dir.entries(folder_path.to_s)
```
mokagio: I'm assuming Rubocop added `.to_s`. It's not explicit (yet, maybe in Ruby 3.x at some point?) but [`File.expand_path` returns a string](https://www.rubydoc.info/stdlib/core/File.expand_path) at least as far as one can see from the examples, so I think we could simplify to 
```suggestion
        subfolders = Dir.entries(folder_path)
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_merge_translators_strings.rb
```diff
@@ -24,7 +24,7 @@ def self.merge_folder(strings_folder)
         tmp_main_file = main_file + '.tmp'
         FileUtils.cp(main_file, tmp_main_file)
 
-        join_files = Dir.glob(File.join("#{strings_folder}", 'strings-*.xml'))
+        join_files = Dir.glob(File.join(strings_folder.to_s, 'strings-*.xml'))
```
mokagio: Unlike above, it's not clear whether it's safe to assume `strings_folder` will always be a string. I'm going to leave the suggestion anyways, might not be safe, but it's reasonable.

```suggestion
        join_files = Dir.glob(File.join(strings_folder, 'strings-*.xml'))
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_download_action.rb
```diff
@@ -9,7 +9,7 @@ def self.run(params = {})
         UI.message 'Running Configure Download'
 
         # If the `~/.mobile-secrets` repository doesn't exist
-        unless File.directory?("#{secrets_dir}")
+        unless File.directory?(secrets_dir.to_s)
```
mokagio:
```suggestion
        unless File.directory?(secrets_dir)
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_translators_strings.rb
```diff
@@ -8,7 +8,7 @@ class IosMergeTranslatorsStringsAction < Action
       def self.run(params)
         folder_path = File.expand_path(params[:strings_folder])
 
-        subfolders = Dir.entries("#{folder_path}")
+        subfolders = Dir.entries(folder_path.to_s)
```
mokagio:
```suggestion
        subfolders = Dir.entries(folder_path)
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_translators_strings.rb
```diff
@@ -21,7 +21,7 @@ def self.merge_folder(strings_folder)
 
         UI.message("Merging in: #{strings_folder}")
 
-        join_files = Dir.glob(File.join("#{strings_folder}", 'Localizable_*.strings')) - [tmp_main_file]
+        join_files = Dir.glob(File.join(strings_folder.to_s, 'Localizable_*.strings')) - [tmp_main_file]
```
mokagio:
```suggestion
        join_files = Dir.glob(File.join(strings_folder, 'Localizable_*.strings')) - [tmp_main_file]
```

mokagio (approved):
I left some nitpicks regarding possibly unnecessary `.to_s` usages. Curious to hear whether there's a good case to keep them and learn more about Ruby in the process.

Nothing worth blocking the PR, though.

.rubocop.yml
```diff
@@ -29,6 +29,9 @@ Style/TrailingCommaInArrayLiteral:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
+Style/AsciiComments:
+  Enabled: false
```
mokagio, line 39: I'm happy with this but I'm curious about the rationale?

mokagio: Generally, I'd like to have a rationale for every disabled rule or any other variation from [the default](https://github.com/rubocop/rubocop/blob/3482522e2be022eee2321722f4e049ab4bcb3bf1/config/default.yml#L2721-L2728).

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_merge_translators_strings.rb
AliSoftware: Indeed this was added by auto correct but you're right that they should not be needed, will remove.

.rubocop.yml
AliSoftware: My rationale is:
- We already use some Unicode characters in existing comments, like em dash — which I'm a big fan of, I have to admit — and ellipsis… which I use here and there too.
- It feels reasonable to consider that at some point we might also want to use emojis, as well as some Unicode characters necessary to draw nicer Monodraw graphs — and you know how I love those too 😛.
- And in the era of Unicode I don't see any reason why we shouldn't allow them. (iow, I don't really see the rationale for the default value of this rule.)

I'll add a comment to explain that rationale in the config file anyway indeed.

TL;DR:
"Jealous Girlfriend" meme variation with em dash

AliSoftware: Rationale added as comment in 70fb5d5a168126b3438351fc7b2328c5e20df4aa

mokagio: 🤣 👍

.rubocop.yml
```diff
@@ -16,9 +16,13 @@ Gemspec/OrderedDependencies:
 Lint/UnusedMethodArgument:
   Enabled: false
 
+# We have a couple of empty methods especially when we implement an API contract on our Action subclasses
+# and those look nicer to us when still written as expended (in separate lines) to suggest us to fill them at some point
+# rather than the compact representation of having them one-liner `def foo; end` that is rubocop's default.
 Style/EmptyMethod:
   EnforcedStyle: expanded
 
+# Trailing commas in array literals helps having smaller diffs when we want to add additional parameters
```
mokagio, line 25: 👍 

Totally out of topic and not really a Ruby friendly thing, but I do love how Elm does trailing commas

```elm
[ "this"
, "makes"
, "for"
, "clear"
, "diffs"
]
```

PR #218, opened 2021-02-24
Merge toolkit-cleanup into develop

AliSoftware:
⚠️  Please review and merge #217 first.

Once #217 is merged into `toolkit-cleanup`, I think it's time to merge all the intermediate work back to `develop` (and start using it next time we do a release).

_Note: I'll still continue to do other PRs for that project but probably directly on develop by now, and better not keep the long-lived branch too long._


.rspec
```diff
@@ -0,0 +1,2 @@
+--color
```
mokagio, line 1: Yes for 🎨 !

I think I came across a study (or maybe just an anecdote 🤔 ) on the value of having red for failed and green for passing tests from a psychological point of view. Something like green is  calmer so you'll want to get to green fast. Maybe.

mokagio (approved):
I only skimmed the diff, since all the work that made into `toolkit-cleanup` already went through code review.

Approving to avoid congestions, even though this is pending #217, which I just approved with a question.

.rspec
AliSoftware: The only issue with this, especially with a distributed company like us, is that colors might have different meanings depending on cultures; especially, in China, red is positive and green is negative (*), so the opposite meaning of what western culture is used to. I wonder if rspec and similar tools take the locale into account and switch those accordingly, but I doubt it. AFAIK we don't have anyone working on this codebase living in China, but still a good anecdote to know about.

(*) that's why, for example, [in Apple's Stocks app, if you have your device's locale and region set to China, the colors used for positive vs negative fluctuations are switched compared to a device set in any other region](https://discussions.apple.com/thread/3282538).

loremattei (approved):
Everything here seems to come from reviewed PRs, so I'm ok with :shipit:

Thanks for all this amazing clean-up @AliSoftware 🙇

PR #219, opened 2021-03-01
Initial editorconfig setup

loremattei:
This PR builds up on [the initial proposal](https://github.com/wordpress-mobile/WordPress-Android/pull/12414) by @oguzkocer which was put on hold because of time issues before his sabbatical.

I'm starting from the ruleset defined in that PR because we mostly agreed on it. 

I'm moving this to `release-toolkit` because I think it would be nice that we share the rules across all the repositories, so that if we make a change, it's easily deployed without having the open PRs on every project. 
So far, the best solution I've found is to rely on `.editorconfig`'s standard behaviour of looking up for a configuration file in every parent folder until it finds one with `root=true`. 
So, my proposal would be to add `.editorconfig` to this repository and update the configuration tasks we already have on every app (`rake dependencies` and `gradle configureApply`) to download it in the parent folder. 

This way, we'll have a shared base configuration and every project will be able to tweak it's own rules if they want to by adding a specific `.editorconfig`.

If you have a better solution to share the file across different repos, please let me know :-)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,11 +12,8 @@ trim_trailing_whitespace = true
 max_line_length=120
 
 # Ruby specific rules
-[*.rb]
+[{*.rb,Fastfile,Gemfile}]
 indent_style = space
 indent_size = 2
 
-# Fastlane specific rules
-[**Fastfile]
-indent_style = space
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+# top-most EditorConfig file
+root = true
+
+# Apply to all files
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Kotlin specific rules
+[*.{kt,kts}]
+max_line_length=120
+
+# Ruby specific rules
+[*.rb]
+indent_style = space
+indent_size = 2
+
+# Fastlane specific rules
+[**Fastfile]
+indent_style = space
+


.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+indent_style = space
+indent_size = 2
+
+# Fastlane specific rules
+[**Fastfile]
+indent_style = space
+
+# YML specific fules
+[*.yml]
+trim_trailing_whitespace=true
```
AliSoftware: Isn't that already provided by the rules in `[*]`?

loremattei: Oh, right! Fixed :-)

AliSoftware:
I like the idea of having one shared config from release-toolkit and have each repo update it 👍 

I'm not sure to understand what you describe to provide a override per repo though: the idea would be that `rake dependencies` (or whichever) will already copy the file from release-toolkit to the root of WPAndroid (and other) repos, right? But then devs would not be able to have _another_ different repo-specific `.editorconfig`  with the same name at the same location… 🤔 

Or were you thinking that the one from release-toolkit would be copied/installed… in `$HOME` or some place like that? (I am not a fan of that option because that overrides the user/machine/system-wide environment and assumes that people don't work on other projects than the our projects using release-toolkit.)

.editorconfig
```diff
@@ -0,0 +1,22 @@
...
+max_line_length=120
+
+# Ruby specific rules
+[*.rb]
+indent_style = space
+indent_size = 2
+
+# Fastlane specific rules
+[**Fastfile]
+indent_style = space
```
AliSoftware: I'm wondering if we could not group the `Fastfile` in the same group as `*.rb` (after all, Fastfiles are ruby files) to ensure the same rules in both without having to repeat everything.

Also, I think we should applying the same rules to `Gemfile` while we're at it.

.editorconfig
```diff
@@ -0,0 +1,22 @@
+# top-most EditorConfig file
+root = true
+
+# Apply to all files
+[*]
+end_of_line = lf
```
AliSoftware, line 6: I'm not sure about that. I thought the standard those days (esp in macOS) was CRLF, not just LF? 🤔

loremattei:
> Or were you thinking that the one from release-toolkit would be copied/installed… in $HOME or some place like that? (I am not a fan of that option because that overrides the user/machine/system-wide environment and assumes that people don't work on other projects than the our projects using release-toolkit.)

Sort of. I'm thinking about copying to the parent folder of the root folder of the project. This way, it will be used also for projects that don't have a specific rake or configuration script (like: many libraries). 

I hope people don't use a 100% flat strategy with all the projects in the $HOME folder. In a setup where the GitHub organization is mirrored, the file will go into the `wordpress-mobile` folder for example.

.editorconfig
AliSoftware: Actually, nvm, I checked and macOS (as well as Unix) uses `0x0A`, aka `LF` nowadays, so this is the right one after all.

jkmassel (approved):
This LGTM pending @AliSoftware's suggestions about Gemfile/Fastfile in the `rb` section

.editorconfig
oguzkocer: I remember trying this at the time, but iirc it wasn't working in my vim setup. I might just need to tweak something in my setup, but thought I'd mention it in case @loremattei hits a similar issue.

oguzkocer (approved):

.editorconfig
loremattei: Good point @AliSoftware!
I've updated the configuration in [d088c4b](https://github.com/wordpress-mobile/release-toolkit/pull/219/commits/d088c4be98928e05e9978b94e28e15dcd060235f).

@oguzkocer this seems to work for me on VSCode and MacVim. Please let me know if it doesn't work on your configuration. It breaks the purpose of this file if it doesn't work on every environment 😅, so I'll revert to the previous configuration in that case.

AliSoftware:
> I hope people don't use a 100% flat strategy with all the projects in the `$HOME` folder.

Actually, I realize that… I personally do this. Well not in my `$HOME` folder, but I still have all my GitHub repos in the same `~/Documents/Dev` subfolder personally (mixing the repos of our mobile apps, a8c tooling repos like `peril-settings`/`circleci-orbs`/`release-toolkit`… but also non-mobile/non-apps-division repos, and also non-a8c repos like `SwiftGen` and `fastlane`… all in the same parent folder).

I might not be the only one having such a setup and I bet not everybody creates subfolders for each GH org… especially since some of a8c repos belong to different GH orgs to begin with (`woocommerce` vs `wordpress-mobile` vs `automattic` orgs) but it doesn't always make sense to keep those separate/isolated in subdirs on your local filesystem… 🤷‍♂️ 

Anyway, not a real blocker for this particular PR – we can still have a reference `.editorconfig` here anyway – only food for thoughts about the _next_ step.

AliSoftware (approved):

loremattei:
>  I still have all my GitHub repos in the same ~/Documents/Dev subfolder personally (mixing the repos of our mobile apps, a8c tooling repos like peril-settings/circleci-orbs/release-toolkit… but also non-mobile/non-apps-division repos, and also non-a8c repos like SwiftGen and fastlane… all in the same parent folder).

@AliSoftware no, I can't believe that _you_ do this 😂  
Just kidding obviously, I see the point. I'll merge this PR here and open dedicated PRs to add/update `.editorconfig` in the main apps repositories. 
As a final step, I'll publish a post to explain how to setup your folder structure if you want to use this `.editorconfig` as a baseline for the other repositories.

AliSoftware:
Random idea: maybe another option to allow override could be to have `rake dependencies` (or whatever task we use for that) check if there's a `.editorconfig-local` file and if there is, then instead of doing `cp release-toolkit/.editorconfig project/.editorconfig` we would do a `cat release-toolkit/.editorconfig project/.editorconfig-local >/project/.editorconfig`?

(Though not sure it's worth the trouble. It's just that I really dislike having tools do things outside of their directory's scope and creating files in places outside their "sandbox" 🤷 feels a bit "bad citizen" attitude to me 😅 )

loremattei:
This would make the repo dirty every time is run, right? 🤔

Since the rules we've added affect mostly our work, I'd say that we can start with adding the file to the main repos and then see how we can iterate on it after we have some real world usage data (for example, if we don't end up updating the rules often, it's not worth it to build complex rules or scripts for it).

AliSoftware:
> This would make the repo dirty every time is run, right? 🤔
I don't see why it would? It should only generate changes in app repos only when either we update the `.editorconfig` file in release-toolkit or users update their local `.editorconfig-local` file…

> for example, if we don't end up updating the rules often, it's not worth it to build complex rules or scripts for it

I agree, probably not worth it. Let's update them by hand for now.

---

btw another idea could be doing the scripting the other way around, like a simple shell script in `mobile-toolbox` that would create all the PRs on all the repos to propagate any change. Something like this pseudo-code, to give you the idea (definitively not saying we should do this now, I agree it's prob not worth it; but just dumping the idea here in case decide to automate it one day in the future):

```sh
SRCFILE=<absolute-path-to-release-toolkit>/.editorconfig
for repo in wordpress-ios wordpress-android woocommmerce-ios … ; do
  cd $repo
  git checkout develop && git pull
  git checkout -b update-editor-config
  cp "$SRCFILE" .editorconfig
  git add .editorconfig && git commit -m "Update editorconfig file"
  git push -u origin update-editor-config
  open "https://github.com/$org/$repo/compare/develop...update-editor-config?title=Update%20editorconfig"
done
```

PR #221, opened 2021-03-04
Release/0.15.0

loremattei:
Merge Release/0.15.0 into Trunk


AliSoftware (approved):
Reviewed the commit list and focused only on your commits only (skipping the ones from me about cleanup), and LGTM :shipit: 

_Note for potential other reviewers: I was a bit confused by GitHub's commit list at first (2 first commits are part of #212 but was was wondering where the rest of #212 went)… but realized that GitHub lists the commits ordered by **date**, instead of by **ancestor** like I'm used to… so that explains it and all is good after all._

PR #222, opened 2021-03-05
Add action to download Android files from a GH version

loremattei:
This PR adds the `android_download_file_by_version.rb` action which downloads a file from a certain GitHub release. The release title is inferred from a key in `build.gradle`.

The action is supposed to be used to support translations for internal libraries which are imported as binary dependencies. 
Each library has its own `strings.xml` which needs to be merged to the client one in order to be uploaded to GlotPress for translation. 

The action can be tested as a part of [the related WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14200).

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_localize_libs_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_localize_libs_action.rb
@@ -10,7 +10,7 @@ module Fastlane
 
         any_changes = false
         libraries_strings_path.each do |lib|
-          (any_changes = Fastlane::Helper::AndroidLocalizeHelper.merge_lib(main_strings_path, lib)) || any_changes
+          (any_changes = Fastlane::Helper::Android::LocalizeHelper.merge_lib(main_strings_path, lib)) || any_changes
         end
 
         UI.message("Changes have been applied to #{main_strings_path}. Please, verify it!") if any_changes
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -11,7 +11,7 @@ module Fastlane
 
         Fastlane::Helper::GithubHelper.download_file_from_release(
           repository: params[:repository],
-          release: "v#{version}",
+          release: "#{params[:github_release_prefix]}#{version}",
           file_path: params[:file_path],
           download_folder: params[:download_folder]
         )
@@ -53,6 +53,10 @@ module Fastlane
                         type: String,
                         optional: true,
                         default_value: Dir.tmpdir()),
+          FastlaneCore::ConfigItem.new(key: :github_release_prefix,
+                        description: "The prefix which is used in the GitHub release title",
+                        type: String,
+                        optional: true),
         ]
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -9,9 +9,9 @@ module Fastlane
         UI.user_error!("Can't find any reference for key #{params[:import_key]}") if version.nil?
         UI.message "Downloading #{params[:file_path]} from #{params[:repository]} at version #{version} to #{params[:download_folder]}"
 
-        Fastlane::Helper::GithubHelper.download_file_from_release(
+        Fastlane::Helper::GithubHelper.download_file_from_tag(
           repository: params[:repository],
-          release: "#{params[:github_release_prefix]}#{version}",
+          tag: "#{params[:github_release_prefix]}#{version}",
           file_path: params[:file_path],
           download_folder: params[:download_folder]
         )
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -5,7 +5,7 @@ module Fastlane
         require_relative '../../helper/android/android_localize_helper.rb'
         require_relative '../../helper/github_helper'
 
-        version = Fastlane::Helper::Android::LocalizeHelper.get_library_version_from_gradle_config(params[:import_key])
+        version = Fastlane::Helper::Android::LocalizeHelper.get_library_version_from_gradle_config(import_key: params[:import_key])
         UI.user_error!("Can't find any reference for key #{params[:import_key]}") if version.nil?
         UI.message "Downloading #{params[:file_path]} from #{params[:repository]} at version #{version} to #{params[:download_folder]}"
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -5,7 +5,7 @@ module Fastlane
         require_relative '../../helper/android/android_localize_helper.rb'
         require_relative '../../helper/github_helper'
 
-        version = Fastlane::Helper::Android::LocalizeHelper.get_library_version_from_gradle_config(import_key: params[:import_key])
+        version = Fastlane::Helper::Android::VersionHelper.get_library_version_from_gradle_config(import_key: params[:import_key])
         UI.user_error!("Can't find any reference for key #{params[:import_key]}") if version.nil?
         UI.message "Downloading #{params[:file_path]} from #{params[:repository]} at version #{version} to #{params[:download_folder]}"
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -64,7 +64,7 @@ module Fastlane
       end
 
       def self.return_value
-        'The location of the downloaded file'
+        'The path where the file was downloaded to (typically <download_folder>/<basename(file_path)>), or nil if there was an issue downloading the file'
       end
 
       def self.authors
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -1,0 +1,75 @@
+module Fastlane
+  module Actions
+    class AndroidDownloadFileByVersionAction < Action
+      def self.run(params)
+        require_relative '../../helper/android/android_localize_helper.rb'
+        require_relative '../../helper/github_helper'
+
+        version = Fastlane::Helper::Android::LocalizeHelper.get_library_version_from_gradle_config(params[:import_key])
+        UI.user_error!("Can't find any reference for key #{params[:import_key]}") if version.nil?
+        UI.message "Downloading #{params[:file_path]} from #{params[:repository]} at version #{version} to #{params[:download_folder]}"
+
+        Fastlane::Helper::GithubHelper.download_file_from_release(
+          repository: params[:repository],
+          release: "v#{version}",
+          file_path: params[:file_path],
+          download_folder: params[:download_folder]
+        )
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        'Downloads a file from a GitHub release based on the version used by the client app'
+      end
+
+      def self.details
+        'This action extracts the version of the library which is imported by the client app' \
+        'and downloads the request file from the relevant GitHub release'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :library_name,
+                        description: "The display name of the library",
+                        type: String,
+                        optional: false),
+          FastlaneCore::ConfigItem.new(key: :import_key,
+                        description: "The key which is used in gradle.build to reference the version of the library to import",
+                        type: String,
+                        optional: false),
+          FastlaneCore::ConfigItem.new(key: :repository,
+                        description: "The GitHub repository which hosts the library project",
+                        type: String,
+                        optional: false),
+          FastlaneCore::ConfigItem.new(key: :file_path,
+                        description: "The path of the file to download",
+                        type: String,
+                        optional: false),
+          FastlaneCore::ConfigItem.new(key: :download_folder,
+                        description: "The download folder",
+                        type: String,
+                        optional: true,
+                        default_value: Dir.tmpdir()),
+        ]
+      end
+
+      def self.output
+      end
+
+      def self.return_value
+        'The location of the downloaded file'
+      end
+
+      def self.authors
+        ['loremattei']
+      end
+
+      def self.is_supported?(platform)
+        platform == :android
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -41,7 +41,7 @@ module Fastlane
                         type: String,
                         optional: false),
           FastlaneCore::ConfigItem.new(key: :repository,
-                        description: "The GitHub repository which hosts the library project",
+                        description: "The GitHub repository slug ('<orgname>/<reponame>') which hosts the library project",
                         type: String,
                         optional: false),
           FastlaneCore::ConfigItem.new(key: :file_path,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
@@ -37,7 +37,7 @@ module Fastlane
                         type: String,
                         optional: false),
           FastlaneCore::ConfigItem.new(key: :import_key,
-                        description: "The key which is used in gradle.build to reference the version of the library to import",
+                        description: "The key which is used in build.gradle to reference the version of the library to import",
                         type: String,
                         optional: false),
           FastlaneCore::ConfigItem.new(key: :repository,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -171,22 +171,6 @@ module Fastlane
             end
           end
         end
-
-        # Extract the value of a import key from build.gradle
-        #
-        # @param [String] import_key The key to look for
-        # @return [String] The value of the key, or nil if not found
-        #
-        def self.get_library_version_from_gradle_config(import_key:)
-          gradle_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'build.gradle')
-
-          return nil unless File.exists?(gradle_file_path)
-
-          File.open(gradle_file_path, 'r') do | f |
-            text = f.read
-            text.match(/^(?:\w*\.)?#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.first
-          end
-        end
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -172,7 +172,11 @@ module Fastlane
           end
         end
 
-
+        # Extract the value of a import key from build.gradle
+        #
+        # @param [String] import_key The key to look for
+        # @return [String] The value of the key, or nil if not found
+        #
         def self.get_library_version_from_gradle_config(import_key:)
           gradle_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'build.gradle')
 
@@ -183,7 +187,6 @@ module Fastlane
             text.match(/#{import_key}\s?=\s?'(.*?)'/m)&.captures&.first
           end
         end
-
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -184,7 +184,7 @@ module Fastlane
 
           File.open(gradle_file_path, 'r') do | f |
             text = f.read
-            text.match(/#{import_key}\s?=\s?'(.*?)'/m)&.captures&.first
+            text.match(/#{Regexp.escape(import_key)}\s?=\s?'(.*?)'/m)&.captures&.first
           end
         end
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -184,7 +184,7 @@ module Fastlane
 
           File.open(gradle_file_path, 'r') do | f |
             text = f.read
-            text.match(/#{Regexp.escape(import_key)}\s?=\s?'(.*?)'/m)&.captures&.first
+            text.match(/#{Regexp.escape(import_key)}\s?=\s?['"](.*?)["']/m)&.captures&.first
           end
         end
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -184,7 +184,7 @@ module Fastlane
 
           File.open(gradle_file_path, 'r') do | f |
             text = f.read
-            text.match(/#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.first
+            text.match(/^(?:\w*\.)?#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.first
           end
         end
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -171,6 +171,17 @@ module Fastlane
             end
           end
         end
+
+
+        def self.get_library_version_from_gradle_config(import_key)
+          gradle_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'build.gradle')
+
+          File.open(gradle_file_path, 'r') do | f |
+            text = f.read
+            text.match(/#{import_key}\s?=\s?'(.*?)'/m)&.captures&.first
+          end
+        end
+
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -173,8 +173,10 @@ module Fastlane
         end
 
 
-        def self.get_library_version_from_gradle_config(import_key)
+        def self.get_library_version_from_gradle_config(import_key:)
           gradle_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'build.gradle')
+
+          return nil unless File.exists?(gradle_file_path)
 
           File.open(gradle_file_path, 'r') do | f |
             text = f.read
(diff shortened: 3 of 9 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_localize_libs_action.rb
```diff
@@ -10,7 +10,7 @@ def self.run(params)
 
         any_changes = false
         libraries_strings_path.each do |lib|
-          (any_changes = Fastlane::Helper::AndroidLocalizeHelper.merge_lib(main_strings_path, lib)) || any_changes
+          (any_changes = Fastlane::Helper::Android::LocalizeHelper.merge_lib(main_strings_path, lib)) || any_changes
```
AliSoftware, line 13: :oops: thanks for fixing this ❤️  – seems it slipped thru the cracks during last refactoring.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -171,6 +171,22 @@ def self.verify_pr_diff(main, library, main_strings, lib_strings, source_diff)
...
+        #
+        def self.get_library_version_from_gradle_config(import_key:)
+          gradle_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'build.gradle')
+
+          return nil unless File.exists?(gradle_file_path)
+
+          File.open(gradle_file_path, 'r') do | f |
+            text = f.read
+            text.match(/#{import_key}\s?=\s?'(.*?)'/m)&.captures&.first
+          end
```
AliSoftware: This works, but I'm pretty sure there are more efficient ways to do this. Especially this reads the whole file in memory, no matter how big the file is.

You should be able to use `File.readlines.each do |line| … end` to iterate line by line, and `return` inside the loop as soon as you find a match instead. Way more memory-efficient.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -171,6 +171,22 @@ def self.verify_pr_diff(main, library, main_strings, lib_strings, source_diff)
...
+        # @return [String] The value of the key, or nil if not found
+        #
+        def self.get_library_version_from_gradle_config(import_key:)
+          gradle_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'build.gradle')
+
+          return nil unless File.exists?(gradle_file_path)
+
+          File.open(gradle_file_path, 'r') do | f |
+            text = f.read
+            text.match(/#{import_key}\s?=\s?'(.*?)'/m)&.captures&.first
```
AliSoftware: I like that line very much, great ruby use of `&.` 🙂 

Wonder if we should not anchor the import_key though, to avoid accidentally matching a larger key `abc` when searching for `bc`? (probably quite unlikely to happen in practice given the kind of import keys we're gonna use, but still). Something like: `/^\s*#{import_key}\s*=\s*'(.*?)'/m`. We also might want to `RegExp.escape` the import_key in case it contains special characters like `.` that we don't want to be interpreted as regex.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -71,6 +72,34 @@ def self.create_release(repository, version, release_notes, assets, prerelease)
...
+        file_path = file_path.delete_prefix('/').chomp('/').concat('/')
+        release = release.concat('/')
+        file_name = File.basename(file_path)
+        download_path = File.join(download_folder, file_name)
+
+        begin
+          open(URI.join('https://raw.githubusercontent.com/', repository, release, file_path).to_s.chomp('/')) do |remote_file|
+            open(download_path, 'wb') do |file|
+              file.write(remote_file.read)
+            end
```
AliSoftware: I _think_ you can simplify that using `File.write(dest, content)` ie:

```suggestion
          File.write(download_path, remote_file.read)
```

spec/android_localize_helper_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    it 'returns nil when the key is not present' do
+      allow(File).to receive(:exists?).and_return(true)
+      allow(File).to receive(:open).with('./build.gradle', 'r').and_yield(StringIO.new('my-test-key = \'my_test_value\''))
+      expect(subject.get_library_version_from_gradle_config(import_key: 'my-test-key2')).to be_nil
+    end
+
+    it 'returns the key content when the key is present' do
+      allow(File).to receive(:exists?).and_return(true)
+      allow(File).to receive(:open).with('./build.gradle', 'r').and_yield(StringIO.new('my-test-key = \'my_test_value\''))
```
AliSoftware: I think it should be worth testing with more dummy content than just a one-liner fake gradle file. Probably instead mock with content like this:

```
test-key-1 = 'foo'
oh-my-test-key = 'bad'
my-test-key   =  'good'
```

Also, what about when `my-test-key` uses double-quotes instead of simple-quotes? I'm guessing that so far our various `build.gradle` files in all our app repos might be using `fooVersion = 'x.y.z'` syntax… but I can easily imagine someone using double-quotes one day (especially when replacing the hash with a version or vice-versa and/if their IDE does some auto-lint/formatting and changes those)

spec/github_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper.rb'
+require 'webmock/rspec'
+
+describe Fastlane::Helper::GithubHelper do
+  describe 'download_file_from_release' do
+    it 'fails if it does not find the right release on GitHub' do
+      stub = stub_request(:get, 'https://raw.githubusercontent.com/repo-test/project-test/1.0/test-file.xml').to_return(:status => [404, "Not Found"])
```
AliSoftware, line 7: nit / brain dump: this makes me wonder: what if there's a 302 or something like that? Is that automatically handled by `open-uri`? (though the only case I can think of for a 302 in that case is if we were to rename the repository, in which case GitHub auto-redirects URLs pointing to the old repo name to point to the new one. A use case that we're unlikely to encounter anytime soon… but just curious)

spec/github_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      expect(stub).to have_been_made.once
+    end
+
+    it 'writes the raw content to a file' do
+      stub = stub_request(:get, 'https://raw.githubusercontent.com/repo-test/project-test/1.0/test-file.xml').to_return(:status => 200, body: 'my-test-content')
+      allow(File).to receive(:open).with('./test-file.xml', 'wb') do | file |
+        expect(Fastlane::Helper::GithubHelper.download_file_from_release(repository: 'repo-test/project-test', release: '1.0', file_path: 'test-file.xml', download_folder: './')).to eq('./test-file.xml')
+        expect(file).to receive(:write).with('my-test-content')
+        expect(stub).to have_been_made.once
+      end
```
AliSoftware: I feel like this kind of testing is a bit too tied with implementation details (relying on the fact that you call `File.open` and `file.write`).
What about downloading to a temp file instead?

```suggestion
      Dir.mktmpdir('a8c-download-repo-file-') do |tmpdir|
        dst_file = File.join(tmpdir, 'test-file.xml')
        expect(Fastlane::Helper::GithubHelper.download_file_from_release(repository: 'repo-test/project-test', release: '1.0', file_path: 'test-file.xml', download_folder: tmpdir)).to eq(dst_file)
        expect(stub).to have_been_made.once
        expect(File.read(dst_file)).to eq('my-test-content')
      end
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
loremattei: Yeah! I was torn about this, but I ended up thinking that reading the whole file is better in this very specific case. It's not supposed to work on random files, in fact it only works on the `build.gradle` file in the project root folder. The size of this file for our most complex project (WPAndroid) is less than 7KB, so size doesn't seem to be an issue. 
Because of this, I preferred the simpler code here (and I'd bet that modern OSes will actually read files of this size in one step anyway at a lower level to optimize for I/O operations as the size is close to a single page on modern devices).

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
loremattei: Oh! TIL about `RegExp.escape` and it sounds like a perfect solution here. Actually, the `import_key` that I expect to be passed here could be part of a larger key (there's usually a `ext.` prefix). Escaping the `.` would make it possible to include the prefix in case we had issues to find the right value using only the short key.

AliSoftware: Alternatively maybe you'd be interested in taking some inspiration on how I do a very similar thing in my WIP branch here: https://github.com/wordpress-mobile/release-toolkit/blob/cleanup/203-version/lib/fastlane/plugin/wpmreleasetoolkit/models/android/version_set.rb#L29-L52 (iow when iterating on each line, call `.strip` on the line being checked to clear leading and trailing whitespaces, then anchor the text to search on the stripped/trimmed line)

AliSoftware: (though, that being said, I'm also gonna borrow a couple of your ideas in that WIP branch at some point, like replace my use of `RegExp.last_match` with your `&.captures&.first` instead 👍 – love that one)

spec/android_localize_helper_spec.rb
loremattei: Good point! Let's update the regex to handle double quotes :-)

spec/android_localize_helper_spec.rb
```diff
@@ -8,14 +8,31 @@
...
 
     it 'returns nil when the key is not present' do
+      test_file_content = 'my-test-key-foo = \'foo\''
+      test_file_content += 'my-test-key-bad = \'bad\''
+      test_file_content += 'my-test-key = \'my_test_value\''
+
       allow(File).to receive(:exists?).and_return(true)
-      allow(File).to receive(:open).with('./build.gradle', 'r').and_yield(StringIO.new('my-test-key = \'my_test_value\''))
+      allow(File).to receive(:open).with('./build.gradle', 'r').and_yield(StringIO.new(test_file_content))
       expect(subject.get_library_version_from_gradle_config(import_key: 'my-test-key2')).to be_nil
```
AliSoftware: What about testing with `my-test-key-b` so that the unit test also ensure that you don't accidentally pick `my-test-key-bad` by accident?

spec/android_localize_helper_spec.rb
```diff
@@ -8,14 +8,31 @@
...
     end
 
     it 'returns the key content when the key is present' do
+      test_file_content = 'my-test-key-foo = \'foo\''
+      test_file_content += 'my-test-key-bad = \'bad\''
+      test_file_content += 'my-test-key = \'my_test_value\''
```
AliSoftware: tip: you could use either `%{}` (to avoid having to escape each single quote) or even heredoc (`<<~EOF` … `EOF`) to make this look nicer and more readable

```suggestion
      test_file_content = <<~CONTENT
        my-test-key-foo = 'foo'
        my-test-key-bad = 'bad'
        my-test-key = 'my_test_value'
      CONTENT
```

spec/android_localize_helper_spec.rb
```diff
@@ -8,14 +8,31 @@
...
+
+      allow(File).to receive(:exists?).and_return(true)
+      allow(File).to receive(:open).with('./build.gradle', 'r').and_yield(StringIO.new(test_file_content))
+      expect(subject.get_library_version_from_gradle_config(import_key: 'my-test-key')).to eq('my_test_value')
+
+      # Make sure it handles double quotes
+      test_file_content = 'my-test-key-foo = \'foo\''
+      test_file_content += 'some-other-content = "dummy"'
+      test_file_content += 'my-test-key = "my_test_value"'
```
AliSoftware: ditto about heredoc tip

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -184,7 +184,7 @@ def self.get_library_version_from_gradle_config(import_key:)
 
           File.open(gradle_file_path, 'r') do | f |
             text = f.read
-            text.match(/#{Regexp.escape(import_key)}\s?=\s?'(.*?)'/m)&.captures&.first
+            text.match(/#{Regexp.escape(import_key)}\s?=\s?['"](.*?)["']/m)&.captures&.first
```
AliSoftware: 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
```diff
@@ -0,0 +1,75 @@
...
+        require_relative '../../helper/android/android_localize_helper.rb'
+        require_relative '../../helper/github_helper'
+
+        version = Fastlane::Helper::Android::LocalizeHelper.get_library_version_from_gradle_config(import_key: params[:import_key])
+        UI.user_error!("Can't find any reference for key #{params[:import_key]}") if version.nil?
+        UI.message "Downloading #{params[:file_path]} from #{params[:repository]} at version #{version} to #{params[:download_folder]}"
+
+        Fastlane::Helper::GithubHelper.download_file_from_release(
+          repository: params[:repository],
+          release: "v#{version}",
```
AliSoftware: I don't think any of our repos use the `v` prefix for release names… or do they? 🤔 

```suggestion
          release: version,
```

spec/android_localize_helper_spec.rb
loremattei: Ah! Nice 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
loremattei: Oh, right! `mobile-gutenberg`, which is the repo which we need this right now for, actually uses the `v` prefix. 
I've added better handling for both cases.

loremattei:
@AliSoftware thanks for the review! I think this is ready for another pass.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :library_name,
+                        description: "The display name of the library",
+                        type: String,
+                        optional: false),
+          FastlaneCore::ConfigItem.new(key: :import_key,
+                        description: "The key which is used in gradle.build to reference the version of the library to import",
```
AliSoftware:
```suggestion
                        description: "The key which is used in build.gradle to reference the version of the library to import",
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
```diff
@@ -0,0 +1,79 @@
...
+                        type: String,
+                        optional: false),
+          FastlaneCore::ConfigItem.new(key: :repository,
+                        description: "The GitHub repository which hosts the library project",
```
AliSoftware: The doc should tell what format is expected here. The full github URL to the repo? Just the repo slug like `org/reponame`?

```suggestion
                        description: "The GitHub repository slug ('<orgname>/<reponame>') which hosts the library project",
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
```diff
@@ -0,0 +1,79 @@
...
+                        type: String,
+                        optional: true),
+        ]
+      end
+
+      def self.output
+      end
+
+      def self.return_value
+        'The location of the downloaded file'
```
AliSoftware:
```suggestion
        'The path where the file was downloaded to (typically <download_folder>/<basename(file_path)>), or nil if there was an issue downloading the file'
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
AliSoftware: - what if the `build.gradle` file contains more than one whitespace between the key and the `=` sign, or between `=` and the value? These should be `\s*` imho.
- what if there's a line containing the import key but not starting with it? I think the regex should start with `^\s*` to allow leading whitespaces but no other prefix or chars before.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -71,6 +72,32 @@ def self.create_release(repository, version, release_notes, assets, prerelease)
           github_client().upload_asset(release[:url], file_path, content_type: 'application/octet-stream')
         end
       end
+
+      # Downloads a file from the given GitHub release
+      #
+      # @param [String] repository The repository name (including the organization)
+      # @param [String] relase The title of the GitHub release
```
```suggestion
      # @param [String] release The title of the GitHub release
```

Technically, the `release` parameter is used in a place of the URL where the commit-ish is expected, not the title of the GitHub release. You can have a GitHub release titled `"The best update ever"` pointing to a tag named `"v1.2"`, and what you pass when you construct the githubusercontent URL is the tag name `v1.2` not the title. So this doc might be misleading.

In fact, that part of the URL expects any "commitish" (as git calls them), so it could be a tag name like you designed it to here, but it would also work with a commit sha1, or a branch name. (And maybe one day we might want to download a file from a specific branch rather than tag, so that should be valid)

```suggestion
      # @param [String] commit The git tag name, branch name, or commit sha1 to download
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -71,6 +72,32 @@ def self.create_release(repository, version, release_notes, assets, prerelease)
...
+      # @param [String] file_path The path, inside the project folder, of the file to download
+      # @param [Striog] download_folder The folder which the file should be downloaded into
+      # @return [String] The path of the downloaded file, or nil if something went wrong
+      #
+      def self.download_file_from_release(repository:, release:, file_path:, download_folder:)
```
```suggestion
      def self.download_file_from_release(repository:, commit:, file_path:, download_folder:)
```
(see suggestion above; obv also update the impl and call sites)

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -71,6 +72,32 @@ def self.create_release(repository, version, release_notes, assets, prerelease)
...
+        repository = repository.delete_prefix('/').chomp('/').concat('/')
+        file_path = file_path.delete_prefix('/').chomp('/').concat('/')
+        release = release.concat('/')
```
AliSoftware: Isn't `URI.join` supposed to add the `/` at the right places when needed for you? You're right to `delete_prefix`, since if any argument you pass to `URI.join` has a leading `/` it will take over as being considered the URI's path's root, but for the suffix there is no need to neither `chomp` any extra `/` nor `concat` one yourself.

```suggestion
        repository.delete_prefix!('/')
        file_path.delete_prefix!('/')
```

spec/android_localize_helper_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it 'returns nil when the key is not present' do
+      test_file_content = <<~CONTENT
+        my-test-key-foo = 'foo'
+        my-test-key-bad = 'bad'
+        my-test-key = 'my_test_value'
+      CONTENT
+
+      allow(File).to receive(:exists?).and_return(true)
+      allow(File).to receive(:open).with('./build.gradle', 'r').and_yield(StringIO.new(test_file_content))
+      expect(subject.get_library_version_from_gradle_config(import_key: 'my-test-key-b')).to be_nil
```
AliSoftware:
```suggestion
      expect(subject.get_library_version_from_gradle_config(import_key: 'my-test-key-b')).to be_nil
      expect(subject.get_library_version_from_gradle_config(import_key: 'test-key')).to be_nil
```
I think the 2nd expectation I added here will fail (mis-catching `my-test-key`'s value) with your current code (while I expect it to pass and return `nil` instead), so worth doing some red-green-refactoring to make it work here :)

spec/android_localize_helper_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    it 'returns the key content when the key is present' do
+      test_file_content = <<~CONTENT
+        my-test-key-foo = 'foo'
+        my-test-key-bad = 'bad'
+        my-test-key = 'my_test_value'
```
AliSoftware:
```suggestion
        my-test-key     =    'my_test_value'
```
Just to test more-than-one-space cases

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
loremattei: >what if the build.gradle file contains more than one whitespace between the key and the = sign, or between = and the value? These should be \s* imho.

It shouldn't have more than one whitespace tbh, but there's no harm with changing the regex. 

> what if there's a line containing the import key but not starting with it? I think the regex should start with ^\s* to allow leading whitespaces but no other prefix or chars before.

Hmm... not sure about this now. There may be cases where the import key we want to pass has not the prefix part. Let's keep it as it is now and iterate if required when the whole setup is completed on the Gutenberg repository. We'll have a better picture at that point.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
loremattei: Right. But does `release` make more sense here?. We are going to use this for downloading from releases, so doesn't passing a release title to a parameter which is called `commit` sound confusing?

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
loremattei: Hmm... I thought so. But I added the `/` because `URI.join` didn't work as expected without them. 
I tried to follow your suggestion again in `07fb871`, but it failed. It looks like `URI.join` doesn't behave as expected in this case.

spec/android_localize_helper_spec.rb
loremattei: I was torn about this tbh. I ended up adding your tests in `30846bb`, but I'm not sure the additional complexity in the regex is really worth it, because these are cases which have a really low probability in the actual code.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
AliSoftware: Ah I see, didn't realize this was intentional that you might want to pick a key… that might or might not have a prefix. Makes sense then 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: We're not passing a release **title** though. We're passing a tag name 😉  So keeping this named as `release` is _more_ confusing to me than calling it `commit`, because if you were to pass the release title to that parameter this would not work for GH releases that happens to not have the same name as the tag they're attached to ([example](https://github.com/realm/SwiftLint/releases/tag/0.43.1)).

I agree that the name `commit` might not be the nicest one, but I feel that we should still use a more appropriate name – e.g. maybe name this parameter `tag:` instead?

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: This is because you used `delete_prefix` – and not `delete_prefix!` – in 07fb871, so not surprising it failed 😉 

`delete_prefix!` with a exclamation mark at the end – like in my suggestion above – means (and that's a convention in Ruby) that it will mutate the object this is called on, while the version without the exclamation mark – the one you used – returns a new modified value instead of mutating the original (which your code then ignored, not making anything of the return value).

In other words, `x.delete_prefix!('/')` is a synonym for `x = x.delete_prefix('/')`. (You can use that other variant if you think it's more readable, both are valid.)

loremattei: Isn't `x = x.delete_prefix('/')` the same of `repository = repository.delete_prefix('/')` and `file_path = file_path.delete_prefix('/')` which I used in 07fb871?

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
loremattei: Yeah, I understand your point, but it's just something we don't do. Anyway, exactly because we don't do it, I think it's not worth it to spend more time discussing this 😅 I've switched it to tag in `6223829`.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: 🤦‍♂️🤦‍♂️ Sorry… I misread the diff of that commit the first time around 🤦 nvm then

loremattei: Yeah, no problem. I'm still surprised that `URI.join` doesn't work as expected tbh. 🤷‍♂️

spec/android_localize_helper_spec.rb
loremattei: I've added a test to cover the prefix case in `6d8d19a`.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
```diff
@@ -0,0 +1,79 @@
+module Fastlane
+  module Actions
+    class AndroidDownloadFileByVersionAction < Action
+      def self.run(params)
+        require_relative '../../helper/android/android_localize_helper.rb'
+        require_relative '../../helper/github_helper'
+
+        version = Fastlane::Helper::Android::LocalizeHelper.get_library_version_from_gradle_config(import_key: params[:import_key])
```
AliSoftware: Just realized that this namespace/module doesn't seem like the ideal place to host such a method? (I know this actions is mostly planned to be used to localize libraries, but not sure how the specific step of "extracting the lib version from gradle file" needs to be specific to Localization helpers.)

What do you think about moving it to `android_version_helper` (since it's dealing with parsing versions from gradle file, which this helper already provide similar methods about)?

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
```diff
@@ -0,0 +1,79 @@
...
+        UI.user_error!("Can't find any reference for key #{params[:import_key]}") if version.nil?
+        UI.message "Downloading #{params[:file_path]} from #{params[:repository]} at version #{version} to #{params[:download_folder]}"
+
+        Fastlane::Helper::GithubHelper.download_file_from_release(
+          repository: params[:repository],
+          release: "#{params[:github_release_prefix]}#{version}",
+          file_path: params[:file_path],
+          download_folder: params[:download_folder]
+        )
```
AliSoftware:
```suggestion
        Fastlane::Helper::GithubHelper.download_file_from_tag(
          repository: params[:repository],
          tag: "#{params[:github_release_prefix]}#{version}",
          file_path: params[:file_path],
          download_folder: params[:download_folder]
        )
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -71,6 +72,32 @@ def self.create_release(repository, version, release_notes, assets, prerelease)
           github_client().upload_asset(release[:url], file_path, content_type: 'application/octet-stream')
         end
       end
+
+      # Downloads a file from the given GitHub tag
+      #
+      # @param [String] repository The repository name (including the organization)
+      # @param [String] tag The name of the tag we're downloading from
+      # @param [String] file_path The path, inside the project folder, of the file to download
+      # @param [Striog] download_folder The folder which the file should be downloaded into
```
```suggestion
      # @param [String] download_folder The folder which the file should be downloaded into
```

AliSoftware (commented):
Quick new review pass:

- Found a place where you forgot to rename the call site for `download_file_from_release`
- Suggest to move `get_lib_version…` helper method to version helper module
- [Doc improvement](https://github.com/wordpress-mobile/release-toolkit/pull/222/files#r594232213) still pending

But apart for those, the rest LGTM so far, so once those get addressed we should be good to go! 🎉

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_file_by_version.rb
loremattei: Good catch! TIL that in VSCode, global search uses the actual content of the files opened in the editor, even if they are not saved :-) 
I found this and actually changed it, but I forgot to save the file and when I double checked using the search function, this didn't show up 🤦

AliSoftware: Yeah, that's one of the reason why I missed a couple of things in the first place when I did my refactoring of release-toolkit – as you already saw in this PR with [this](https://github.com/wordpress-mobile/release-toolkit/pull/222#discussion_r589602025) 😅 

_Switching to RubyMine was a big deal and help in that regard (and I'll probably never be using VSCode ever again to edit/refactor Ruby code from now on, as RubyMine makes these kind of things less error-prone 😉)_

loremattei:
@AliSoftware everything should be now addressed. Thanks!

AliSoftware (approved):
You forgot a small doc nit (ece5ca3), so I took the liberty of commiting it, and now LGTM! :shipit:

PR #226, opened 2021-03-23
Fix tag deletion logic

jkmassel:
This PR reworks the tag deletion logic to rework the approach:

1. Nothing is deleted without confirmation
2. Previously, the logic was to delete all local tags and re-download them – this approach isn't compatible with confirmation, so it needs to go – we can handle it later on in the process anyway.
3. There's some logic to better handle string/array input

It should be fairly clear how this works, but if you want to test it in the context of a project, you can run `bundle exec fastlane run ios_clear_intermediate_tags` in a project with your release toolkit dependency pointing at this branch (or create some tags on your local copy of the release toolkit to test with).

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
@@ -6,9 +6,6 @@ module Fastlane
 
         require_relative '../../helper/git_helper.rb'
 
-        # Delete the current version's tag (if it exists)
-        Fastlane::Helper::GitHelper.delete_tags(params[:version])
-
         # Delete 4-parts version names starting with our version number
         parts = params[:version].split('.')
         pattern = parts.fill('*', parts.length...4).join('.') # "1.2.*.*" or "1.2.3.*"
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
@@ -5,18 +5,20 @@ module Fastlane
         UI.message("Deleting tags for version: #{params[:version]}")
 
         require_relative '../../helper/git_helper.rb'
-        Fastlane::Helper::Ios::GitHelper.delete_tags(params[:version])
 
-        # Cleanup local tags and refetch them to make sure we're up-to-date with remote
-        local_tags = Fastlane::Helper::GitHelper.list_local_tags()
-        Fastlane::Helper::GitHelper.delete_tags(local_tags, delete_on_remote: true)
-        Fastlane::Helper::GitHelper.fetch_all_tags()
+        # Delete the current version's tag (if it exists)
+        Fastlane::Helper::GitHelper.delete_tags(params[:version])
 
-        # Now delete intermediate tags (4-parts version names starting with our version number)
+        # Delete 4-parts version names starting with our version number
         parts = params[:version].split('.')
         pattern = parts.fill('*', parts.length...4).join('.') # "1.2.*.*" or "1.2.3.*"
+
         intermediate_tags = Fastlane::Helper::GitHelper.list_local_tags(matching: pattern)
-        Fastlane::Helper::GitHelper.delete_tags(intermediate_tags, delete_on_remote: true)
+        tag_count = intermediate_tags.count
+
+        if tag_count > 0 && UI.confirm("Are you sure you want to delete #{tag_count} tags?")
+          Fastlane::Helper::GitHelper.delete_tags(intermediate_tags, delete_on_remote: true)
+        end
       end
 
       #####################################################
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
@@ -6,9 +6,6 @@ module Fastlane
 
         require_relative '../../helper/git_helper.rb'
 
-        # Download all the remote tags prior to starting – that way we don't miss any on the server
-        Fastlane::Helper::GitHelper::fetch_all_tags
-
         # Delete 4-parts version names starting with our version number
         parts = params[:version].split('.')
         pattern = parts.fill('*', parts.length...4).join('.') # "1.2.*.*" or "1.2.3.*"
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
@@ -6,6 +6,9 @@ module Fastlane
 
         require_relative '../../helper/git_helper.rb'
 
+        # Download all the remote tags prior to starting – that way we don't miss any on the server
+        Fastlane::Helper::GitHelper::fetch_all_tags
+
         # Delete 4-parts version names starting with our version number
         parts = params[:version].split('.')
         pattern = parts.fill('*', parts.length...4).join('.') # "1.2.*.*" or "1.2.3.*"
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -123,22 +123,14 @@ module Fastlane
       # @param [Array<String>] tag_names The list of tags to delete
       # @param [Bool] delete_on_remote If true, will also delete the tag from the remote. Otherwise, it will only be deleted locally.
       #
-      def self.delete_tags(tag_names, delete_on_remote: false)
-
-        tag_names = [tag_names] unless tag_names.is_a? Array
+      def self.delete_tags(tags_to_delete, delete_on_remote: false)
 
         g = Git.open(Dir.pwd)
-        local_tag_names = g.tags.map do |tag|
-          tag.name
-        end
+        local_tag_names = g.tags.map(&:name)
 
-        tag_names.each do |tag|
-          Action.sh('git', 'tag', '-d', *tag) if local_tag_names.include? tag
-        end
-
-        if delete_on_remote
-          remote_refs = tag_names.map { |tag| ":refs/tags/#{tag}" }
-          Action.sh('git', 'push', 'origin', *remote_refs)
+        Array(tags_to_delete).select { |tag| local_tag_names.include? tag }.each do |tag|
+          g.delete_tag(tag)
+          g.push('origin', ":refs/tags/#{tag}") if delete_on_remote 
         end
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -131,8 +131,8 @@ module Fastlane
         g = Git.open(Dir.pwd)
         local_tag_names = g.tags.map(&:name)
 
-        Array(tags_to_delete).select { |tag| local_tag_names.include? tag }.each do |tag|
-          g.delete_tag(tag)
+        Array(tags_to_delete).each do |tag|
+          g.delete_tag(tag) if local_tag_names.include?(tag)
           g.push('origin', ":refs/tags/#{tag}") if delete_on_remote 
         end
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -124,7 +124,6 @@ module Fastlane
       # @param [Bool] delete_on_remote If true, will also delete the tag from the remote. Otherwise, it will only be deleted locally.
       #
       def self.delete_tags(tags_to_delete, delete_on_remote: false)
-
         g = Git.open(Dir.pwd)
         local_tag_names = g.tags.map(&:name)
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -124,7 +124,18 @@ module Fastlane
       # @param [Bool] delete_on_remote If true, will also delete the tag from the remote. Otherwise, it will only be deleted locally.
       #
       def self.delete_tags(tag_names, delete_on_remote: false)
-        Action.sh('git', 'tag', '-d', *tag_names)
+
+        tag_names = [tag_names] unless tag_names.is_a? Array
+
+        g = Git.open(Dir.pwd)
+        local_tag_names = g.tags.map do |tag|
+          tag.name
+        end
+
+        tag_names.each do |tag|
+          Action.sh('git', 'tag', '-d', *tag) if local_tag_names.include? tag
+        end
+
         if delete_on_remote
           remote_refs = tag_names.map { |tag| ":refs/tags/#{tag}" }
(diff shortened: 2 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
```diff
@@ -5,18 +5,20 @@ def self.run(params)
         UI.message("Deleting tags for version: #{params[:version]}")
 
         require_relative '../../helper/git_helper.rb'
-        Fastlane::Helper::Ios::GitHelper.delete_tags(params[:version])
 
-        # Cleanup local tags and refetch them to make sure we're up-to-date with remote
-        local_tags = Fastlane::Helper::GitHelper.list_local_tags()
```
jkmassel, line 11: The root problem here is that we're calling with `delete_on_remote: true`, but we can simplify this quite a bit.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
```diff
@@ -5,18 +5,20 @@ def self.run(params)
...
-        # Now delete intermediate tags (4-parts version names starting with our version number)
+        # Delete 4-parts version names starting with our version number
         parts = params[:version].split('.')
         pattern = parts.fill('*', parts.length...4).join('.') # "1.2.*.*" or "1.2.3.*"
+
         intermediate_tags = Fastlane::Helper::GitHelper.list_local_tags(matching: pattern)
-        Fastlane::Helper::GitHelper.delete_tags(intermediate_tags, delete_on_remote: true)
+        tag_count = intermediate_tags.count
+
+        if tag_count > 0 && UI.confirm("Are you sure you want to delete #{tag_count} tags?")
```
jkmassel, line 19: Because this is destructive, we need to double-check.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -124,7 +124,18 @@ def self.list_local_tags(matching: '*')
       # @param [Bool] delete_on_remote If true, will also delete the tag from the remote. Otherwise, it will only be deleted locally.
       #
       def self.delete_tags(tag_names, delete_on_remote: false)
-        Action.sh('git', 'tag', '-d', *tag_names)
+
+        tag_names = [tag_names] unless tag_names.is_a? Array
```
jkmassel: Because we accept both arrays and strings, we need to always have an array to continue

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
```diff
@@ -5,18 +5,20 @@ def self.run(params)
...
+          Fastlane::Helper::GitHelper.delete_tags(intermediate_tags, delete_on_remote: true)
+        end
```
AliSoftware, line 21: What about listing the actual list of tags that we are gonna delete instead?
```suggestion
        unless intermediate_tags.empty?
          UI.message("About to delete the following intermediate tags:" + intermediate_tags.map { |t| " - #{t}" }.join("\n")
          if UI.confirm("Are you sure you want to delete those tags?")
            Fastlane::Helper::GitHelper.delete_tags(intermediate_tags, delete_on_remote: true)
          end
        end
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware:
```suggestion
        tag_names = Array(tag_names)
```
Should work both if tag_names is already an Array or not – [See this SO post](https://stackoverflow.com/a/18358925)

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -124,7 +124,18 @@ def self.list_local_tags(matching: '*')
...
+
+        g = Git.open(Dir.pwd)
+        local_tag_names = g.tags.map do |tag|
+          tag.name
+        end
```
```suggestion
        local_tag_names = g.tags.map(&:name)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -124,7 +124,18 @@ def self.list_local_tags(matching: '*')
...
+
+        tag_names.each do |tag|
+          Action.sh('git', 'tag', '-d', *tag) if local_tag_names.include? tag
```
AliSoftware: (1) Since you already started to use the `Git` gem above (👍 ), can't we use `g` to delete the tags here, instead of `Action.sh`?

(2) Any benefit of iterating over the array only to pass one tag on each iteration, instead of passing all the tags directly in one call?
```ruby
local_tags_to_delete = tag_names & local_tag_names # '&' is the 'set/array intersection' operator here
Action.sh('git', 'tag', '-d', *local_tags_to_delete) # Or even better if you find in the gem doc a way to do that using the gem, didn't look it up
```
(btw, maybe we could find better name for those local variables to avoid confusion, e.g. `tag_names` -> `tags_to_delete` and `local_tag_names` -> `all_local_tag_names`?)

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -124,7 +124,18 @@ def self.list_local_tags(matching: '*')
...
+        end
+
         if delete_on_remote
           remote_refs = tag_names.map { |tag| ":refs/tags/#{tag}" }
           Action.sh('git', 'push', 'origin', *remote_refs)
```
AliSoftware: Ditto, since you already use the `Git` gem, that would be a good occasion to migrate this call to the `Git` gem's method equivalent that would make it nicer.

AliSoftware (commented):
Didn't test it on actual repo yet, but here's my take at the code.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
jkmassel: Ah nice - addressed in d3d0d673deebb9a288050efa0da9e92bdb279403

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
jkmassel: Right – this is much nicer. Addressed in d3d0d673deebb9a288050efa0da9e92bdb279403

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
jkmassel: 1. Good call – done in d3d0d673deebb9a288050efa0da9e92bdb279403.
2. Unfortunately the `git` gem only supports one-at-a-time, but it shouldn't be a huge deal as there are typically few tags to delete.

Also agree with the variable naming – it's simplified quite a bit now.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
jkmassel: This appears in the `fastlane` output for me like this, which is nice:

```
[12:51:29]: -----------------------------------------
[12:51:29]: --- Step: ios_clear_intermediate_tags ---
[12:51:29]: -----------------------------------------
[12:51:29]: To not be asked about this value, you can specify it using 'version'
[12:51:29]: The version of the tags to clear: 16.9
[12:51:31]: Deleting tags for version: 16.9
16.9
[12:51:32]: $ git tag --list 16.9.\*.\*
[12:51:32]: ▸ 16.9.0.0
[12:51:32]: ▸ 16.9.0.1
```

AliSoftware: Ah, nice.
🤔  That logging will likely disappear if we also switch `list_local_tags(matching:)` from using `Action.sh` to using `Git` gem in the future too though, right?

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -123,22 +123,14 @@ def self.list_local_tags(matching: '*')
...
-        end
+        local_tag_names = g.tags.map(&:name)
 
-        if delete_on_remote
-          remote_refs = tag_names.map { |tag| ":refs/tags/#{tag}" }
-          Action.sh('git', 'push', 'origin', *remote_refs)
+        Array(tags_to_delete).select { |tag| local_tag_names.include? tag }.each do |tag|
+          g.delete_tag(tag)
+          g.push('origin', ":refs/tags/#{tag}") if delete_on_remote 
         end
```
AliSoftware: ⚠️  That will only delete remote tags if they are also present locally though, because you call the `g.push` on the filtered / `.select`'d list.
So if I am not the one who created the intermediate tags nor did I do a `git pull --tags --force` for a while to fetch those tags on my local machine, then the remote tags won't get deleted when I run this from my Mac…

```suggestion
        Array(tags_to_delete).each do |tag|
          g.delete_tag(tag) if local_tag_names.include?(tag)
          g.push('origin', ":refs/tags/#{tag}") if delete_on_remote 
        end
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
```diff
@@ -5,18 +5,20 @@ def self.run(params)
...
 
         require_relative '../../helper/git_helper.rb'
-        Fastlane::Helper::Ios::GitHelper.delete_tags(params[:version])
 
-        # Cleanup local tags and refetch them to make sure we're up-to-date with remote
-        local_tags = Fastlane::Helper::GitHelper.list_local_tags()
-        Fastlane::Helper::GitHelper.delete_tags(local_tags, delete_on_remote: true)
-        Fastlane::Helper::GitHelper.fetch_all_tags()
+        # Delete the current version's tag (if it exists)
+        Fastlane::Helper::GitHelper.delete_tags(params[:version])
```
loremattei: Hmmm... this was introduced during the refactoring, but I'm not sure why we need it 🤔 
I don't expect a tag to be present at this time and, if it is, I don't think we want to delete it silently.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
loremattei: Yeah, I agree. I haven't seen weird tags pushed to the repo for a while now. If they restart to appear, we can re-add this check.

loremattei: But, we may still want to download the remote tags, otherwise we may not delete tags which have been pushed to GitHub, but are not on the local machine.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
loremattei: I'm not sure, but it looks like[ currently `tags_to_delete`](https://github.com/wordpress-mobile/release-toolkit/blob/d3d0d673deebb9a288050efa0da9e92bdb279403/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb#L20) are already [read locally](https://github.com/wordpress-mobile/release-toolkit/blob/d3d0d673deebb9a288050efa0da9e92bdb279403/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb#L16), right? 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
jkmassel: Yeah, good question – so my thinking on this is that if we're re-running this, it's likely due to a failed finalization, in which case we want it to be idempotent. Though what you say about not deleting it silently seems good from a "confirm destructive operations" standpoint (even if we'll re-create the tag right away).

WDYT about keeping it, but asking for confirmation?

loremattei: With the new flow, the final tag is created remotely when the release is published. But also with the old flow (which we still have on Woo and Simplenote) creating the tag is the last step, so we should never need it. 

Tbh, I think this was just a mistake during the refactoring, but if you really want to keep it for any reasons, I agree that we'd need at least a confirmation because it would be a really weird situation.
Also, in this case, I think it makes more sense to delete the tag remotely as well... otherwise we are just going to defer the failure.

jkmassel: I still think it might be useful, but for the purposes of this PR (which is to fix a bug), it doesn't need to be here. We can always make the finalization idempotent another time. 

Removed in 629cc8c19e78db695cce8cf3297ad0fb5c9dfa10

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
jkmassel: Yeah that's correct – right now it's sort of a side-effect that we're relying on. For now it's IMHO ok, but at some point we should probably run around clean it up (though I don't think this is necessarily the time to do so)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_clear_intermediate_tags.rb
jkmassel: Good point – added in 395a1fabd5357f9485466c4cd3e5e33b9e3d3afd

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -123,11 +123,14 @@ def self.list_local_tags(matching: '*')
       # @param [Array<String>] tag_names The list of tags to delete
       # @param [Bool] delete_on_remote If true, will also delete the tag from the remote. Otherwise, it will only be deleted locally.
       #
-      def self.delete_tags(tag_names, delete_on_remote: false)
-        Action.sh('git', 'tag', '-d', *tag_names)
-        if delete_on_remote
-          remote_refs = tag_names.map { |tag| ":refs/tags/#{tag}" }
-          Action.sh('git', 'push', 'origin', *remote_refs)
+      def self.delete_tags(tags_to_delete, delete_on_remote: false)
+
```
```suggestion
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: True (that was just not yet the case in the state the PR was in when I reviewed and wrote that, but Jeremy fixed it since).

Tho tbh I find the code suggested above easier to read than the one with `.select`… but maybe that's just me 🤷

AliSoftware (approved):
LGTM!

My [edit suggestion in `delete_tags`](https://github.com/wordpress-mobile/release-toolkit/pull/226#discussion_r601765733) still applies, even tho the mentioned issue doesn't hold anymore (so change is not functionally necessary anymore, but more for readability); I still find it odd for the _helper_ to rely on the _action_ – on the fact that action has fetched tags locally before calling helper – while usually it's the other way around 🤷 , but that's not a blocker, so take it or leave it.

jkmassel:
Yeah good thought @AliSoftware – addressed in 593a6cd51f2aa4ef88046471ba0c3ee4a027b0c0

AliSoftware (approved):

PR #227, opened 2021-03-23
Lookup PRs using the GitHub search API

jkmassel:
Now that our PRs are consistently tagged with a milestone, we can fetch the PR list based on that, rather than by parsing commits.

This PR updates the release toolkit to use this method. The changes are backward-incompatible, but won't silently fail.

**To Test:**
- Download this PR and run `bundle exec fastlane run get_prs_list repository:wordpress-mobile/wordpress-ios report_path:~/test.txt tag:16.9`. Note that all of the PR data is downloaded in the same format. [note: you'll need `GHHELPER_ACCESS` set as an environment variable with your GitHub token]
- Compare the results of the command above with a pre-existing file on your local machine. The ordering might be different, but the format should be identical (you can see this in the code as well – the line to generate the file is unchanged).

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
@@ -6,34 +6,19 @@ module Fastlane
     class GetPrsListAction < Action
       def self.run(params)
         repository = params[:repository]
-        start_tag = params[:start_tag]
-        end_tag = params[:end_tag]
-        report_path = params[:report_path]
+        report_path = File.expand_path(params[:report_path])
+        tag = params[:tag]
 
         # Get commit list
-        commit_list = sh("git log --pretty=oneline #{start_tag}..#{end_tag}")
+        pr_list = Fastlane::Helper::GithubHelper.get_prs_for_milestone(repository, tag)
 
-        # Extract PRs
-        pr_list = []
-        commit_list.split("\n").each do |commit|
-          if commit.include?('Merge pull request #')
-            # PR found, so extract PR number
-            pr_list.push(commit.partition('#').last.split(' ')[0])
-          end
-        end
-
-        # Get infos from GitHub and put into the target file
-        client = Fastlane::Helper::GithubHelper.github_client()
         File.open(report_path, 'w') do |file|
-          pr_list.each do |pr_number|
-            begin
-              data = client.pull_request(repository, pr_number.to_i)
-              file.puts("##{data[:number]}: #{data[:title]} @#{data[:user][:login]} #{data[:html_url]}")
-            rescue
-              UI.message("Could not find a PR with number #{pr_number.to_i}. Usually this is due to a bad reference in a commit message, but you probably want to check.")
-            end
+          pr_list.each do |data|
+            file.puts("##{data[:number]}: #{data[:title]} @#{data[:user][:login]} #{data[:html_url]}")
           end
         end
+
+        UI.success("Found #{pr_list.count} PRs in #{tag} – saved to #{report_path}")
       end
 
       def self.description
@@ -60,18 +45,14 @@ module Fastlane
                                        description: 'The remote path of the GH repository on which we work',
                                        optional: false,
                                        type: String),
-          FastlaneCore::ConfigItem.new(key: :start_tag,
-                                       description: 'The tag from which the report starts',
-                                       optional: false,
-                                       is_string: true),
-          FastlaneCore::ConfigItem.new(key: :end_tag,
-                                       description: 'The tag to which the report ends',
-                                       optional: true,
-                                       default_value: '.',
-                                       is_string: true),
           FastlaneCore::ConfigItem.new(key: :report_path,
                                        env_name: 'GHHELPER_REPORTPATH',
                                        description: 'The path of the report file',
+                                       optional: false,
+                                       is_string: true),
+          FastlaneCore::ConfigItem.new(key: :tag,
+                                       env_name: 'GHHELPER_CURRENT_VERSION',
+                                       description: 'The version number to fetch PRs for',
                                        optional: false,
                                        is_string: true),
         ]
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
@@ -7,10 +7,10 @@ module Fastlane
       def self.run(params)
         repository = params[:repository]
         report_path = File.expand_path(params[:report_path])
-        tag = params[:tag]
+        milestone = params[:milestone]
 
         # Get commit list
-        pr_list = Fastlane::Helper::GithubHelper.get_prs_for_milestone(repository, tag)
+        pr_list = Fastlane::Helper::GithubHelper.get_prs_for_milestone(repository, milestone)
 
         File.open(report_path, 'w') do |file|
           pr_list.each do |data|
@@ -18,11 +18,11 @@ module Fastlane
           end
         end
 
-        UI.success("Found #{pr_list.count} PRs in #{tag} – saved to #{report_path}")
+        UI.success("Found #{pr_list.count} PRs in #{milestone} – saved to #{report_path}")
       end
 
       def self.description
-        'Generate the list of the PRs from `start_tag` to `end_tag`'
+        'Generate the list of the PRs in the given `repository` for the given `milestone` at the given `report_path'
       end
 
       def self.authors
@@ -35,23 +35,22 @@ module Fastlane
 
       def self.details
         # Optional:
-        'Generate the list of the PRs from `start_tag` to `end_tag`'
+        'Generate the list of the PRs in the given `repository` for the given `milestone` at the given `report_path'
       end
 
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :repository,
                                        env_name: 'GHHELPER_REPOSITORY',
-                                       description: 'The remote path of the GH repository on which we work',
+                                       description: 'The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)',
                                        optional: false,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :report_path,
-                                       env_name: 'GHHELPER_REPORTPATH',
-                                       description: 'The path of the report file',
+                                       description: 'The path that the list of PRs should be written to',
                                        optional: false,
                                        is_string: true),
           FastlaneCore::ConfigItem.new(key: :milestone,
-                                       description: 'The milestone to fetch PRs for',
+                                       description: 'The name of the milestone we want to fetch the list of PRs for (e.g.: `16.9`)',
                                        optional: false,
                                        is_string: true),
         ]
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
@@ -22,7 +22,7 @@ module Fastlane
       end
 
       def self.description
-        'Generate the list of the PRs in the given `repository` for the given `milestone` at the given `report_path'
+        'Generate the list of the PRs in the given `repository` for the given `milestone` at the given `report_path`'
       end
 
       def self.authors
@@ -35,7 +35,7 @@ module Fastlane
 
       def self.details
         # Optional:
-        'Generate the list of the PRs in the given `repository` for the given `milestone` at the given `report_path'
+        description
       end
 
       def self.available_options
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
@@ -46,7 +46,7 @@ module Fastlane
                                        optional: false,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :report_path,
-                                       description: 'The path that the list of PRs should be written to',
+                                       description: 'The path where the list of PRs should be written to',
                                        optional: false,
                                        is_string: true),
           FastlaneCore::ConfigItem.new(key: :milestone,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
@@ -50,9 +50,8 @@ module Fastlane
                                        description: 'The path of the report file',
                                        optional: false,
                                        is_string: true),
-          FastlaneCore::ConfigItem.new(key: :tag,
-                                       env_name: 'GHHELPER_CURRENT_VERSION',
-                                       description: 'The version number to fetch PRs for',
+          FastlaneCore::ConfigItem.new(key: :milestone,
+                                       description: 'The milestone to fetch PRs for',
                                        optional: false,
                                        is_string: true),
         ]
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -14,6 +14,9 @@ module Fastlane
         user = client.user
         UI.message("Logged in as: #{user.name}")
 
+        # Auto-paginate to ensure we're not missing data
+        client.auto_paginate = true
+
         client
       end
 
@@ -26,6 +29,16 @@ module Fastlane
         end
 
         return mile
+      end
+
+      # Downloads a file from the given GitHub tag
+      #
+      # @param String repository The repository name (including the organization) [ex: wordpress-mobile/wordpress-ios]
+      # @param String tag The name of the tag we're downloading from [ex: 16.9]
+      # @return [<Sawyer::Resource>] A list of the PRs for the given milestone, sorted by number
+      #
+      def self.get_prs_for_milestone(repository, release)
+        github_client().search_issues("type:pr milestone:16.9 repo:wordpress-mobile/wordpress-ios")[:items].sort_by(&:number)
       end
 
       def self.get_last_milestone(repository)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -33,8 +33,8 @@ module Fastlane
 
       # Fetch all the PRs for a given milestone
       #
-      # @param String repository The repository name (including the organization) [ex: wordpress-mobile/wordpress-ios]
-      # @param String milestone The name of the tag we're downloading from [ex: 16.9]
+      # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
+      # @param [String] milestone The name of the milestone we want to fetch the list of PRs for (e.g.: `16.9`)
       # @return [<Sawyer::Resource>] A list of the PRs for the given milestone, sorted by number
       #
       def self.get_prs_for_milestone(repository, milestone)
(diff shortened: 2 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -14,6 +14,9 @@ def self.github_client
         user = client.user
         UI.message("Logged in as: #{user.name}")
 
+        # Auto-paginate to ensure we're not missing data
+        client.auto_paginate = true
```
jkmassel, line 18: We don't want to silently miss anything if there are more than 100 PRs, so I've turned this on

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -28,6 +31,16 @@ def self.get_milestone(repository, release)
...
       end
 
+      # Downloads a file from the given GitHub tag
+      #
+      # @param String repository The repository name (including the organization) [ex: wordpress-mobile/wordpress-ios]
+      # @param String tag The name of the tag we're downloading from [ex: 16.9]
+      # @return [<Sawyer::Resource>] A list of the PRs for the given milestone, sorted by number
+      #
+      def self.get_prs_for_milestone(repository, release)
+        github_client().search_issues("type:pr milestone:16.9 repo:wordpress-mobile/wordpress-ios")[:items].sort_by(&:number)
```
AliSoftware: You left hardcoded values here both for the repo slug and the tag name to fetch 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -28,6 +31,16 @@ def self.get_milestone(repository, release)
         return mile
       end
 
+      # Downloads a file from the given GitHub tag
+      #
+      # @param String repository The repository name (including the organization) [ex: wordpress-mobile/wordpress-ios]
+      # @param String tag The name of the tag we're downloading from [ex: 16.9]
+      # @return [<Sawyer::Resource>] A list of the PRs for the given milestone, sorted by number
+      #
+      def self.get_prs_for_milestone(repository, release)
```
AliSoftware: Since what we search here is based on the milestone, better names that param `milestone` rather than `release`

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -28,6 +31,16 @@ def self.get_milestone(repository, release)
         return mile
       end
 
+      # Downloads a file from the given GitHub tag
+      #
+      # @param String repository The repository name (including the organization) [ex: wordpress-mobile/wordpress-ios]
+      # @param String tag The name of the tag we're downloading from [ex: 16.9]
```
AliSoftware: Doc param name mismatch

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -60,20 +45,16 @@ def self.available_options
...
-                                       description: 'The tag to which the report ends',
-                                       optional: true,
-                                       default_value: '.',
-                                       is_string: true),
           FastlaneCore::ConfigItem.new(key: :report_path,
                                        env_name: 'GHHELPER_REPORTPATH',
                                        description: 'The path of the report file',
                                        optional: false,
                                        is_string: true),
+          FastlaneCore::ConfigItem.new(key: :tag,
```
AliSoftware: Given that what we indeed use for the search is the milestone not the (potentially not-yet-existing) tag, I think it would be better to name this param either `milestone` or `version` rather than `tag`.

And maybe also mention in the ConfigItem's description that this will use the milestone — because sure for our case in WC/WP/SN we name our milestones by the name of the version, but that might not be the case for other users of our toolkit (eg non-a8c, but also even A12s like if Tumblr/Orangina starts to use our toolkit but doesn't follow the exact same conventions at first, etc)

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -28,6 +31,16 @@ def self.get_milestone(repository, release)
         return mile
       end
 
+      # Downloads a file from the given GitHub tag
```
AliSoftware: Actually even the doc description is incorrect (copy/pasted too quickly, right?)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
AliSoftware: Looking at the codebase more closely, I see we have a `get_milestone(release)` helper method, so alternatively we could make this param be named `version` and then use that helper method to find the milestone corresponding to that version before calling your other helper method to list PRs.

…though I'd argue that the implementation for that `get_milestone(version)` method is similarly tied to our conventions (and also its implementation could be improved via a `.reverse.find` instead of the `.each` to make it more functional… but I digress…)

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: 🤦 

Fixed in f4c890bceaa0a8c909dc58303c6ea22604ac24b6

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: I agree - fixed

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: Uhhhh....you caught me 😅

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
jkmassel: Yeah, I was initially going to re-work `get_milestone`, but it needed more refactoring than I wanted to spend on it to fix _this_ issue. I feel like it could definitely be replaced by something a bit simpler for sure :)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -60,20 +45,15 @@ def self.available_options
                                        description: 'The remote path of the GH repository on which we work',
```
AliSoftware, line 60: More like the repo slug (`<org name>/<reponame>`) rather than the "remote path" (What was that even supposed to mean? URL?)

Especially since this is used in the milestone filter/query and the "path"/URL would not work for that

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -28,6 +31,16 @@ def self.get_milestone(repository, release)
         return mile
       end
 
+      # Fetch all the PRs for a given milestone
+      #
+      # @param String repository The repository name (including the organization) [ex: wordpress-mobile/wordpress-ios]
+      # @param String milestone The name of the tag we're downloading from [ex: 16.9]
```
AliSoftware: Still wrong description for this `milestone` param 😛

Also, not sure that YARD properly parses the doc when types are not specified between square brackets


```suggestion
      # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
      # @param [String] milestone The name of the milestone we want to fetch the list of PRs for (e.g.: `16.9`)
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
AliSoftware: Yeah let's keep the expected param for this be the milestone directly (and not rely on `get_milestone` for now), I agree.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -6,34 +6,19 @@ module Actions
     class GetPrsListAction < Action
       def self.run(params)
         repository = params[:repository]
-        start_tag = params[:start_tag]
-        end_tag = params[:end_tag]
-        report_path = params[:report_path]
+        report_path = File.expand_path(params[:report_path])
+        tag = params[:tag]
```
mokagio: Since this PR already introduce backward incompatible changes, may I suggest to rename this option to "milestone" to better describe what using it does?

```suggestion
        milestone = params[:milestone]
```

mokagio: Oh, I see that there's actually already [a new `:milestone` option in the `ConfigItem` array below](https://github.com/wordpress-mobile/release-toolkit/pull/227/files#diff-c3c92f31a3aed155aab1aba9315c428344a5079f853257ac495f27c81c12b317R53-R56).

I'm guessing is just something that got lost during the implementation, then.

~~I also guess it went unnoticed because of Fastlane's dynamic nature and the fact that if you pass it an unexpected option (`tag` in this instance, as the PR description suggests) it will still use it.~~

In fact, if I run the lane using `tag` as suggested in the description, it fails:

```
[10:53:44]: You passed invalid parameters to 'get_prs_list'.
[10:53:44]: Check out the error below and available options by running `fastlane action get_prs_list`

[!] Could not find option 'tag' in the list of available options: repository, report_path, milestone
```

Running the same command with `milestone` instead of `tag` fails, too, but that's because there's no match for `params[:tag]`

```
[!] Could not find option 'tag' in the list of available options: repository, report_path, milestone
```

With the suggestions I left, basically `%s/tag/milestone/`, the code works.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -6,34 +6,19 @@ module Actions
...
-            end
+          pr_list.each do |data|
+            file.puts("##{data[:number]}: #{data[:title]} @#{data[:user][:login]} #{data[:html_url]}")
           end
         end
+
+        UI.success("Found #{pr_list.count} PRs in #{tag} – saved to #{report_path}")
       end
 
       def self.description
```
mokagio, line 24: I can't leave a suggestion here because the next line is outside the diff, but it would be appropriate to update the description currently it mentions `start_tag` and `end_tag` which are removed in this PR.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -6,34 +6,19 @@ module Actions
...
         repository = params[:repository]
-        start_tag = params[:start_tag]
-        end_tag = params[:end_tag]
-        report_path = params[:report_path]
+        report_path = File.expand_path(params[:report_path])
+        tag = params[:tag]
 
         # Get commit list
-        commit_list = sh("git log --pretty=oneline #{start_tag}..#{end_tag}")
+        pr_list = Fastlane::Helper::GithubHelper.get_prs_for_milestone(repository, tag)
```
mokagio:
```suggestion
        pr_list = Fastlane::Helper::GithubHelper.get_prs_for_milestone(repository, milestone)
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -6,34 +6,19 @@ module Actions
...
-              file.puts("##{data[:number]}: #{data[:title]} @#{data[:user][:login]} #{data[:html_url]}")
-            rescue
-              UI.message("Could not find a PR with number #{pr_number.to_i}. Usually this is due to a bad reference in a commit message, but you probably want to check.")
-            end
+          pr_list.each do |data|
+            file.puts("##{data[:number]}: #{data[:title]} @#{data[:user][:login]} #{data[:html_url]}")
           end
         end
+
+        UI.success("Found #{pr_list.count} PRs in #{tag} – saved to #{report_path}")
```
mokagio:
```suggestion
        UI.success("Found #{pr_list.count} PRs in #{milestone} – saved to #{report_path}")
```

mokagio (changes requested):
Love the idea 👍 

Once the `tag` vs `milestone` parameter access is fixed, this should be GTG.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: > not sure that YARD properly parses the doc when types are not specified between square brackets
TIL – addressed in 8c3e4affab2acde6067ba4af4de8d044591a350d

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
jkmassel: Yeah, good point. This was what it was originally – I just copied the description from the helper YARDdoc in 8c3e4affab2acde6067ba4af4de8d044591a350d, which should be quite clear I think

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
jkmassel: Fixed in 8c3e4affab2acde6067ba4af4de8d044591a350d, thanks!!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
jkmassel: Yeah, I agree. Done in 8c3e4affab2acde6067ba4af4de8d044591a350d

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
jkmassel: Addressed in 8c3e4affab2acde6067ba4af4de8d044591a350d

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
jkmassel: Addressed in 8c3e4affab2acde6067ba4af4de8d044591a350d

jkmassel:
Welp, this is what I get for being in too much of a rush – thoroughly schooled on a simple PR 😅 

Thanks gents – should all be addressed

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -7,22 +7,22 @@ class GetPrsListAction < Action
...
           end
         end
 
-        UI.success("Found #{pr_list.count} PRs in #{tag} – saved to #{report_path}")
+        UI.success("Found #{pr_list.count} PRs in #{milestone} – saved to #{report_path}")
       end
 
       def self.description
-        'Generate the list of the PRs from `start_tag` to `end_tag`'
+        'Generate the list of the PRs in the given `repository` for the given `milestone` at the given `report_path'
```
mokagio: Excellent!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -35,23 +35,22 @@ def self.return_value
 
       def self.details
         # Optional:
-        'Generate the list of the PRs from `start_tag` to `end_tag`'
+        'Generate the list of the PRs in the given `repository` for the given `milestone` at the given `report_path'
```
mokagio: I wonder if we can call `description` here? To DRY, given there's not difference.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -35,23 +35,22 @@ def self.return_value
...
           FastlaneCore::ConfigItem.new(key: :repository,
                                        env_name: 'GHHELPER_REPOSITORY',
-                                       description: 'The remote path of the GH repository on which we work',
+                                       description: 'The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)',
                                        optional: false,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :report_path,
-                                       env_name: 'GHHELPER_REPORTPATH',
-                                       description: 'The path of the report file',
+                                       description: 'The path that the list of PRs should be written to',
```
mokagio: Nitpick: 
```suggestion
                                       description: 'The path where the list of PRs should be written to',
```

mokagio (approved):
Thank you for considering my suggestions.

I tested this with the new changes and verified it works!

🚀

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
AliSoftware: I might be wrong but I feel like fastlane might even be smart enough to fallback to `self.description` when it prints the action's documentation if `self.details` is empty 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
```diff
@@ -6,38 +6,23 @@ module Actions
...
         File.open(report_path, 'w') do |file|
-          pr_list.each do |pr_number|
-            begin
-              data = client.pull_request(repository, pr_number.to_i)
-              file.puts("##{data[:number]}: #{data[:title]} @#{data[:user][:login]} #{data[:html_url]}")
-            rescue
-              UI.message("Could not find a PR with number #{pr_number.to_i}. Usually this is due to a bad reference in a commit message, but you probably want to check.")
-            end
+          pr_list.each do |data|
+            file.puts("##{data[:number]}: #{data[:title]} @#{data[:user][:login]} #{data[:html_url]}")
```
AliSoftware, line 17: Just in case you want to get fancy, you could replace this whole `File.open(…) do … end` block from lines 15-18 with this:
```ruby
lines = pr_list.map { |pr| "##{pr[:number]}: #{pr[:title]} @#{pr[:user][:login]} #{pr[:html_url]}" }
File.write(report_path, lines.join("\n"))
```

_(feel free to ignore, that's just me being pedantic 😇  and both are ok)_

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -28,6 +31,16 @@ def self.get_milestone(repository, release)
...
       end
 
+      # Fetch all the PRs for a given milestone
+      #
+      # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
+      # @param [String] milestone The name of the milestone we want to fetch the list of PRs for (e.g.: `16.9`)
+      # @return [<Sawyer::Resource>] A list of the PRs for the given milestone, sorted by number
+      #
+      def self.get_prs_for_milestone(repository, milestone)
+        github_client().search_issues("type:pr milestone:#{milestone} repo:#{repository}")[:items].sort_by(&:number)
```
AliSoftware: Quite unlikely to happen in our use cases – except if we want to get the PRs for a frozen milestone with ❄️ in its name 🤔 , but that shouldn't happen in our current process anyway – so not a big deal for our case, but just realized that we should wrap the `milestone` name in quotes in case it contains spaces or special chars.

```suggestion
        github_client.search_issues(%(type:pr milestone:"#{milestone}" repo:#{repository}))[:items].sort_by(&:number)
```

AliSoftware (approved):
Not much to add to Gio's comments, apart from the small nitpicks, LGTM 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_list_action.rb
jkmassel: For now, it's explicit, but we can always change it in the future 🤷

mokagio: 👍

PR #228, opened 2021-03-31
Update regex used to find library version in gradle path

JavonDavis:
After digging into it an issue with the gutenberg-mobile it seems since the line with the gutenberg mobile version begins with spaces, the current regex won't match that to find the version, this is a proposed change to fix that.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
@@ -305,7 +305,7 @@ module Fastlane
 
           File.open(gradle_file_path, 'r') do | f |
             text = f.read
-            text.match(/^(?:\w*\.)?#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.first
+            text.match(/^(\s*)(?:\w*\.)?#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.last
           end
         end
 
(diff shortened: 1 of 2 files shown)


AliSoftware (changes requested):
Thanks for the fix!

Could you also update `spec/android_version_helper_spec.rb` to add a case for this? So that we don't risk having similar regression in the future. Thanks 🙂

loremattei:
Thank you @JavonDavis!

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -305,7 +305,7 @@ def self.get_library_version_from_gradle_config(import_key:)
 
           File.open(gradle_file_path, 'r') do | f |
             text = f.read
-            text.match(/^(?:\w*\.)?#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.first
+            text.match(/^(\s*)(?:\w*\.)?#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.last
```
AliSoftware: 🤔 nit: I just realized: we don't need to have a capture group for this whitespace after all, so I think this _should_ work too (while simplifying the regex slightly)?
```suggestion
            text.match(/^\s*(?:\w*\.)?#{Regexp.escape(import_key)}\s*=\s*['"](.*?)["']/m)&.captures&.first
```
(ndb, take it or leave it; since both solutions work, as long as we fix the issue, either are good… just thought I'd mention this as alternative for simplification)

JavonDavis: Yep you're right I overlooked that! I'll merge it in, thanks @AliSoftware

JavonDavis:
This is ready for another look @AliSoftware

AliSoftware (approved):
👌  :shipit:

PR #232, opened 2021-04-08
Moves integer check to superclass

JavonDavis:
Title says it all


lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
```diff
@@ -125,6 +125,10 @@ def generate_block(fw)
       def extract_key(line)
         line.split(' ')[1].tr('\"', '')
       end
+
+      def is_int?(value)
+        true if Integer(string) rescue false
+      end
```
AliSoftware, line 131: This is not the superclass, but a sibling class.

`StandardMetadataBlock` is the superclass you want to move this method to.
Both `ReleaseNoteMetadataBlock` and `ReleaseNoteShortMetadataBlock` are subclasses of that so moving the `is_int?` from the latter to the former will make it work for the long release notes entry… but will break it for the short release notes; we need for this method to be declared on `StandardMetadataBlock` so that it's available on both subclasses instead.

AliSoftware (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
JavonDavis: Hmm @AliSoftware can you clarify a bit, since we have 

https://github.com/wordpress-mobile/release-toolkit/blob/develop/lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb#L130

Doesn't that make `ReleaseNoteMetadataBlock` the superclass of `ReleaseNoteShortMetadataBlock` and since the is_int function was only being used in those two classes I didn't see the need to push that out to `StandardMetadataBlock` or am I misunderstanding something in Ruby here?

AliSoftware: Ohhh right, I can't read 🤦  (should stop working so late ^^) Yep that should work then 👍

AliSoftware (approved):

PR #236, opened 2021-04-19
Update logic to compute the iOS localization paths to return full path

mokagio:
While running the code freeze for Simplenote 4.35, the lane failed on me with the following error:

```
	 7: from /Users/gio/Developer/a8c/snios/vendor/bundle/ruby/2.6.0/gems/fastlane-2.180.1/fastlane/lib/fastlane/runner.rb:263:in `block (2 levels) in execute_action'
	 6: from /Users/gio/Developer/a8c/snios/vendor/bundle/ruby/2.6.0/bundler/gems/release-toolkit-413bdfd2a22e/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_localize_project.rb:9:in `run'
	 5: from /Users/gio/Developer/a8c/snios/vendor/bundle/ruby/2.6.0/bundler/gems/release-toolkit-413bdfd2a22e/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:47:in `localize_project'
	 4: from /Users/gio/Developer/a8c/snios/vendor/bundle/ruby/2.6.0/bundler/gems/release-toolkit-413bdfd2a22e/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb:81:in `commit'
	 3: from /Users/gio/.rbenv/versions/2.6.4/lib/ruby/2.6.0/forwardable.rb:230:in `sh'
	 2: from /Users/gio/Developer/a8c/snios/vendor/bundle/ruby/2.6.0/gems/fastlane-2.180.1/fastlane/lib/fastlane/helper/sh_helper.rb:80:in `sh_control_output'
	 1: from /Users/gio/Developer/a8c/snios/vendor/bundle/ruby/2.6.0/gems/fastlane-2.180.1/fastlane_core/lib/fastlane_core/ui/ui.rb:17:in `method_missing'
/Users/gio/Developer/a8c/snios/vendor/bundle/ruby/2.6.0/gems/fastlane-2.180.1/fastlane_core/lib/fastlane_core/ui/interface.rb:153:in `shell_error!': [!] Exit status of command 'git add de.lproj/Localizable.strings he.lproj/Localizable.strings zh-Hant-TW.lproj/Localizable.strings ar.lproj/Localizable.strings el.lproj/Localizable.strings ja.lproj/Localizable.strings fa.lproj/Localizable.strings en.lproj/Localizable.strings en.lproj/InfoPlist.strings es.lproj/Localizable.strings it.lproj/Localizable.strings sv.lproj/Localizable.strings ko.lproj/Localizable.strings tr.lproj/Localizable.strings pt-BR.lproj/Localizable.strings ru.lproj/Localizable.strings cy.lproj/Localizable.strings fr.lproj/Localizable.strings id.lproj/Localizable.strings nl.lproj/Localizable.strings zh-Hans-CN.lproj/Localizable.strings' was 128 instead of 0. (FastlaneCore::Interface::FastlaneShellError)
fatal: pathspec 'de.lproj/Localizable.strings' did not match any files
```

Because we ran the glob in the localization folder, the strings path was relative to that folder, but we need it relative to the project root.

I used this branch to finish the code freeze and it worked. See [this commit](https://github.com/Automattic/simplenote-ios/pull/1251/commits/0c38c496671f05ab54fe782ff0c49e9bf0fac8da).

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
@@ -70,9 +70,9 @@ module Fastlane
         end
 
         def self.get_from_env!(key:)
-          value = ENV[key]
-          UI.user_error! "Could not find value for \"#{key}\" in environment." if value.nil?
-          return value
+          ENV.fetch(key) do
+            UI.user_error! "Could not find value for \"#{key}\" in environment."
+          end
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
@@ -61,14 +61,14 @@ module Fastlane
 
           Fastlane::Helper::GitHelper.commit(message: 'Update metadata translations', files: './fastlane/metadata/', push: true)
         end
-      end
-    end
 
-    private
-
-    def self.strings_files
-      Dir.chdir(File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'])) do
-        Dir.glob('*.lproj/*.strings')
+        def self.strings_files
+          Dir.glob(
+            File.join(
+              ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], '*.lproj', '*.strings'
+            )
+          )
+        end
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
@@ -39,7 +39,7 @@ module Fastlane
         #       in the release-toolkit instead, and move this code away from `ios_git_helper`.
         #
         def self.localize_project
-          Action.sh("cd #{ENV['PROJECT_ROOT_FOLDER']} && ./Scripts/localize.py")
+          Action.sh("cd #{get_from_env!(key: 'PROJECT_ROOT_FOLDER')} && ./Scripts/localize.py")
 
           Fastlane::Helper::GitHelper.commit(message: 'Update strings for localization', files: strings_files, push: true) || UI.message('No new strings, skipping commit')
         end
@@ -53,7 +53,7 @@ module Fastlane
         #       in the release-toolkit instead, and move this code away from `ios_git_helper`.
         #
         def self.update_metadata
-          Action.sh("cd #{ENV['PROJECT_ROOT_FOLDER']} && ./Scripts/update-translations.rb")
+          Action.sh("cd #{get_from_env!(key: 'PROJECT_ROOT_FOLDER')} && ./Scripts/update-translations.rb")
 
           Fastlane::Helper::GitHelper.commit(message: 'Update translations', files: strings_files, push: false)
 
@@ -63,11 +63,16 @@ module Fastlane
         end
 
         def self.strings_files
-          Dir.glob(
-            File.join(
-              ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], '**', '*.strings'
-            )
-          )
+          project_root = get_from_env!(key: 'PROJECT_ROOT_FOLDER')
+          project_name = get_from_env!(key: 'PROJECT_NAME')
+
+          Dir.glob(File.join(project_root, project_name, '**', '*.strings'))
+        end
+
+        def self.get_from_env!(key:)
+          value = ENV[key]
+          UI.user_error! "Could not find value for \"#{key}\" in environment." if value.nil?
+          return value
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
@@ -41,9 +41,6 @@ module Fastlane
         def self.localize_project
           Action.sh("cd #{ENV['PROJECT_ROOT_FOLDER']} && ./Scripts/localize.py")
 
-          strings_files = Dir.chdir(File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'])) do
-            Dir.glob('*.lproj/*.strings')
-          end
           Fastlane::Helper::GitHelper.commit(message: 'Update strings for localization', files: strings_files, push: true) || UI.message('No new strings, skipping commit')
         end
 
@@ -58,15 +55,20 @@ module Fastlane
         def self.update_metadata
           Action.sh("cd #{ENV['PROJECT_ROOT_FOLDER']} && ./Scripts/update-translations.rb")
 
-          strings_files = Dir.chdir(File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'])) do
-            Dir.glob('*.lproj/*.strings')
-          end
           Fastlane::Helper::GitHelper.commit(message: 'Update translations', files: strings_files, push: false)
 
           Action.sh('cd fastlane && ./download_metadata.swift')
 
           Fastlane::Helper::GitHelper.commit(message: 'Update metadata translations', files: './fastlane/metadata/', push: true)
         end
+      end
+    end
+
+    private
+
+    def self.strings_files
+      Dir.chdir(File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'])) do
+        Dir.glob('*.lproj/*.strings')
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
@@ -65,7 +65,7 @@ module Fastlane
         def self.strings_files
           Dir.glob(
             File.join(
-              ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], '*.lproj', '*.strings'
+              ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], '**', '*.strings'
             )
           )
         end
--- a/spec/ios_git_helper_spec.rb
+++ b/spec/ios_git_helper_spec.rb
@@ -1,0 +1,21 @@
+require_relative './spec_helper'
+
+describe Fastlane::Helper::Ios::GitHelper do
+  describe '#get_from_env!' do
+    let(:key) { 'a_key' }
+
+    it 'shows an error when the value is not in the environment' do
+      ENV[key] = nil
+
+      expect(FastlaneCore::UI).to receive(:user_error!)
+      described_class.get_from_env!(key: key)
+    end
+
+    it 'returns the value when in the environment' do
+      ENV[key] = 'abc123'
+
+      expect(FastlaneCore::UI).not_to receive(:user_error!)
+      expect(described_class.get_from_env!(key: key)).to eq('abc123')
+    end
+  end
+end


lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -41,9 +41,9 @@ def self.commit_version_bump(include_deliverfile: true, include_metadata: true)
         def self.localize_project
           Action.sh("cd #{ENV['PROJECT_ROOT_FOLDER']} && ./Scripts/localize.py")
 
-          strings_files = Dir.chdir(File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'])) do
-            Dir.glob('*.lproj/*.strings')
-          end
+          localizations_path = File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'])
+          strings_files = Dir.glob(File.join(localizations_path, '*.lproj/*.strings'))
```
AliSoftware: You started using `File.join` for first part, but didn't use it for the `*.lproj/*.strings' part then ^^ 
```suggestion
          strings_files = Dir.glob(File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], '*.lproj', '*.strings'))
```

Note: slight risk with this solution is we open up to wildcard injection: if either of the `PROJECT_ROOT_FOLDER` or `PROJECT_NAME` env var contains a wildcard glob character (`*` or `?` typically), this part of the path too will end up being interpreted by `Dir.glob`. That's very likely why this was done via a `Dir.chrdir` before.

We can consider a low risk and very unlikely situation (honestly I'm good with it) and keep that. Or we can re-introduce `Dir.chdir` to avoid that risk 🤷 , something like this:
```suggestion
          loc_path = File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'])
          strings_files = Dir.chdir(loc_path) { Dir.glob('*.lproj/*.strings') }.map { |p| File.join(loc_path, p) }
```

Honestly I think the first solution is fine and simpler to read, and many ruby projects use `glob` that way. Just wanted to point out about the possible injection so that we're aware of it at least.

AliSoftware (changes requested):
I'm a bit surprised that the `strings_files`  variable ever worked, in any project.

1. As you said, in the original code this returned relative paths, so that wouldn't work anyway
2. But even in the modified code, this will only find lproj files that are directly under the `PROJECT_NAME` folder. Which is not the case for all our apps

For example:
 - For `Simplenote`, the files seems indeed to be under `$PROJECT_ROOT_FOLDER/Simplenote/*.lproj`.
 - But for WooCommerce-iOS, they are under `$PROJECT_ROOT_FOLDER/WooCommerce/Resources/*.lproj`
 - For `WordPress`, the main strings files are also under `./WordPress/Resources/*.lproj`, but there are also `*.lproj` folders for widgets and extensions under `./WordPress/WordPressIntents/`, `./WordPress/WordPress*Extension/{Resources/}`, `./WordPress/*Widget/{Supporting Files/}`… do we want to commit them too? (depends what the `localize.py` script does, really)

So overall, I think this change would not fix apps other that Simplenote. Not sure why we never encountered the issue earlier, as the path was already wrong for all apps, but with this fix it will be correct for SN but still not for WP nor WC iinm.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
AliSoftware: That being said, as pointed out in the PR review, in WC and WP those are not the path to the `.lproj` files either, so that won't probably fix it for them.

mokagio: > You started using File.join for first part, but didn't use it for the `*.lproj/*.strings` part

🙃

mokagio:
@AliSoftware 

> So overall, I think this change would not fix apps other that Simplenote. Not sure why we never encountered the issue earlier, as the path was already wrong for all apps

🤔 As a matter of fact, I did not test the change outside of Simplenote... And I don't particularly like the use of env var either, but I just ~hacked~ updated the code to solve my immediate issue.

This looks like it needs more investigation... I'm going to revert it to draft. I don't have time to look into this just now, though.

Thanks for looking over this.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
jkmassel: > Note: slight risk with this solution is we open up to wildcard injection: if either of the PROJECT_ROOT_FOLDER or PROJECT_NAME env var contains a wildcard glob character (* or ? typically), this part of the path too will end up being interpreted by Dir.glob. That's very likely why this was done via a Dir.chrdir before.

IMHO this is pretty unlikely, so I'd say just keep it clean for now 🤷‍♂️

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -41,9 +41,9 @@ def self.commit_version_bump(include_deliverfile: true, include_metadata: true)
         def self.localize_project
           Action.sh("cd #{ENV['PROJECT_ROOT_FOLDER']} && ./Scripts/localize.py")
```
jkmassel: While we're in here, WDYT about adding a check that these environment variables are present before starting this method? IMHO it would make anything that's missed far more obvious without having to dig into the source. Just thinking basically a runtime assertion (until we can refactor this to use proper arguments and stop assuming certain `ENV` keys exist, anyway...)

jkmassel (approved):
LGTM, but I left one potential improvement

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
mokagio: Oh yeah! I think we've been too relaxed with `ENV` vars and are paying the price now that we're changing things.

AliSoftware:
Encountered an issue that I think is related when I handled the WCiOS 6.5 submission today:

```
[17:49:46]: ---------------------------------
[17:49:46]: --- Step: ios_update_metadata ---
[17:49:46]: ---------------------------------
[17:49:46]: $ cd ./ && ./Scripts/update-translations.rb
[17:49:47]: ▸ Updating ar
[17:49:48]: ▸ WooCommerce/Resources/ar.lproj/Localizable.strings: OK
…

[17:50:11]: $ git commit -m Update\ translations
[17:50:11]: ▸ On branch release/6.5
[17:50:11]: ▸ Your branch is up to date with 'origin/release/6.5'.
[17:50:11]: ▸ Changes not staged for commit:
[17:50:11]: ▸ modified:   WooCommerce/Resources/ar.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/de.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/es.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/fr.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/he.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/id.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/it.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/ja.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/ko.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/nl.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/pt-BR.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/ru.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/sv.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/tr.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/zh-Hans.lproj/Localizable.strings
[17:50:11]: ▸ modified:   WooCommerce/Resources/zh-Hant.lproj/Localizable.strings
[17:50:11]: ▸ no changes added to commit
```
Which means that the translations didn't make it to the binary (and I only realized this once I submitted the build to Apple… so will have to redo another build with that now…)

This definitively needs to be fixed ASAP.

Maybe we can use some `Dir.glob('**/*.strings')` for the list of files to commit so that it words regardless of where the project has its strings file?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -58,15 +55,20 @@ def self.localize_project
...
           Action.sh('cd fastlane && ./download_metadata.swift')
 
           Fastlane::Helper::GitHelper.commit(message: 'Update metadata translations', files: './fastlane/metadata/', push: true)
         end
+
+        def self.strings_files
+          Dir.glob(
+            File.join(
+              ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], '*.lproj', '*.strings'
+            )
```
AliSoftware: I don't think this will work because I think `*` might be interpreted literally/verbatim here (not entirely sure if the shell substitution that would interpret that `*` as a wildcard comes into play here when we call the `commit` helper). Also, as mentioned previously, won't work for all the repos.

Maybe this instead would cover all cases?
```suggestion
            Dir.glob(File.join(
              ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], '**', '*.lproj', '*.strings'
            ))
```

jkmassel: Interestingly this has a conflict on WPiOS, but it appears to work the way we want 🙂

I ran it locally and it fixed the issue.

AliSoftware:
Unsurprisingly, ran into it again during today's `code_freeze`. Just wanted to comment in order to highlight that it's not just happening during new betas and finalize release, but also during release branch cut… so practically during every step of our release process 😓  and indeed starts to become urgent to fix as this led to almost not shipping localizations (or submitting them into GlotPress in time for being translated) more than once, which wouldn't be pretty.

```
[15:19:53]: $ git commit -m Update\ strings\ for\ localization
[15:19:53]: ▸ On branch release/6.6
[15:19:53]: ▸ Your branch is up to date with 'origin/release/6.6'.
[15:19:53]: ▸ Changes not staged for commit:
[15:19:53]: ▸ modified:   WooCommerce/Resources/en.lproj/Localizable.strings
[15:19:53]: ▸ no changes added to commit
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
mokagio: @jkmassel what conflict are you seeing? 

I tried it just now (with @AliSoftware's `**` suggestion) and it seemed like it worked fine, as far as I can see. It definitely found and added the files: https://github.com/wordpress-mobile/WordPress-iOS/pull/16432.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
mokagio: Partially addressed in 0239b8d. I didn't go down the full refactoring with DI route to prioritizie shipping the fix.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -56,17 +53,27 @@ def self.localize_project
...
+          project_name = get_from_env!(key: 'PROJECT_NAME')
+
+          Dir.glob(File.join(project_root, project_name, '**', '*.strings'))
+        end
+
+        def self.get_from_env!(key:)
+          ENV.fetch(key) do
+            UI.user_error! "Could not find value for \"#{key}\" in environment."
+          end
+        end
```
mokagio, line 76: This is not the nicest method, and it's probably duplicated somewhere in the library.

As a followup, I think we may want to add this to `ENV` directly and find a nice name (`fetch!`?) for it and make it flexible.

spec/ios_git_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_relative './spec_helper'
+
+describe Fastlane::Helper::Ios::GitHelper do
```
mokagio, line 3: Rubocop is telling me:

```
warning| RSpec/FilePath: Spec path should end with `fastlane/helper/ios/git_helper*_spec.rb`.
```

Something I'm not sure is why it's saying is for this spec but not for, say, `git_helper_spec.rb`. 🤔 

Anyways, if we were to change this, it would be appropriate to change all the others, too. Or, we can just add it to the excluded rules list. I'm more leaning towards the first option. Keen to hear what you think. Regardless, that should be a dedicated PR.

mokagio (commented):
@jkmassel @AliSoftware here's what I consider a decent hotfix.

I manually tested it on [WordPress](https://github.com/wordpress-mobile/WordPress-iOS/pull/16432), [WooCommerce](https://github.com/woocommerce/woocommerce-ios/pull/4107), and [Simplenote](https://github.com/Automattic/simplenote-ios/pull/1269) (the latter, including the missing `ENV` values behavior). Feel free to have a go at it as well.

If this works as I think it does, I'd rather ship this as it is and followup to address architecture improvements or with things like extracting the `get_from_env!` method.

spec/ios_git_helper_spec.rb
AliSoftware: This is because it expects the module tree declared in this file to match the path subdirectories structure; so if the file contains a test for `describe Foo::Bar::Baz` then it expects the file to be in `Foo/bar/baz_spec.rb`

Indeed I can confirm that I've run into that one in the past and that this convention doesn't hold for many of our spec files in the current state of the codebase. I think most of them have been ignored in `rubocop-todo.yml` for now — because that's part of the many things we'd need to refactor quite a bit to fix all the violations and I haven't had time to fix them all. For this one I agree we'd need to align all the files to that convention at once, but [what do we say to rubocop spec violations…](https://66.media.tumblr.com/54876958f77d3a5e40d21f374ce31b68/tumblr_pqpq3mSIJ51rp74xfo2_400.gif)…

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
AliSoftware: How does the output look if you just use `ENV.fetch` without a block? According to the ruby documentation, `fetch` will raise a `KeyError` if you don't provide a block nor default value; wondering if Fastlane doesn't already catch that raised `KeyError` and prints a nice message for it already instead of you having to handle the `UI.user_error!` explicitly here?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -56,17 +53,27 @@ def self.localize_project
         #       in the release-toolkit instead, and move this code away from `ios_git_helper`.
         #
         def self.update_metadata
-          Action.sh("cd #{ENV['PROJECT_ROOT_FOLDER']} && ./Scripts/update-translations.rb")
+          Action.sh("cd #{get_from_env!(key: 'PROJECT_ROOT_FOLDER')} && ./Scripts/update-translations.rb")
```
AliSoftware, line 56: Can't wait to get rid of the python script, of the odd requirement of having to `cd` to the right dir first for the script to work as expected, and for that script to be hosted in the app repo… 

soon meme

AliSoftware (approved):
Code changes LGTM. Didn't actually test this from any of the host app repos yet (writing from my phone here) so maybe we should pick an app repo, point it to that branch and check it works as expected, but I'm pretty confident the logic is right (and we need this to move forward)

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
AliSoftware: (That being said I agree to ship this as is, and bike shed and circle back on this detail only later in separate PR if needs be)

spec/ios_git_helper_spec.rb
mokagio: > I think most of them have been ignored in `rubocop-todo.yml` for now

Oh, right! I didn't think of looking in there 👍 

> but what do we say to rubocop spec violations…

🤣 (_still waiting for the last book!!!_)

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
jkmassel: Just to update this - it was a conflict with some deprecated localizations – removed in https://github.com/wordpress-mobile/WordPress-iOS/pull/16415 and everything worked great after that.

PR #229, opened 2021-03-31
Update Android Translations Download Action

AliSoftware:
This PR adds a new action to replace the `tools/update-translations.sh` script that we currently usually have in the Android apps' repositories, while fixing the issue with translation linter (ref: paqN3M-hG-p2).

## What it does

 - Replicates in ruby the behavior of the `update-translations.sh` shell script (which is currently duplicated in each repo and was  run by the `android_update_metadata` action).
 - Replicate the other things that the `android_update_metadata` action does, like calling the gradle lint task, updating the submodules, and commiting.
 - Replicates the ability to merge multiple GlotPress exports – which is used in WordPress to create translator builds and was handled by `android_merge_translator_strings` action.
 - Fix the issue with translation linter by replicating each attributes from the `values/strings.xml` file to the `strings.xml` files downloaded from GlotPress.
 - Deprecate the `android_update_metadata` and the `android_merge_translator_strings` actions that this new action aims to completely replace. Hopefully we can remove them (and clean up `android_git_helper.rb`) once all our apps have migrated to the new action

## Test Instructions

ℹ️ _An earlier version of this action has been tested via [this commit of WCAndroid](https://github.com/woocommerce/woocommerce-android/commit/fdeea4fe0f27b58fc052bfe6ef1fd9e13b2a4ab6) during the WCAndroid 6.3 `finalize_release`, but the implementation in this PR/branch has evolved since._
ℹ️  _There is already [a Draft PR](https://github.com/woocommerce/woocommerce-android/pull/3821) to test those changes on WooCommerce-Android – which you can use as inspiration to replicate the same test steps in other repos_

⚠️  I'm especially interested in you testing in `WordPress-Android`, because this one has additional use cases it needs to cover for translator builds.

### Groundwork

- Create a temporary branch locally (no need to push it)
- Point your release-toolkit on an app repo on this branch
- Add a `formatted="false"` attribute to at least one of the `<string>` tags in your `values/strings.xml` original file
- Update the translations using the legacy action – `bundle exec fastlane run android_update_metadata` – so that your current `values-*/strings.xml` files are up-to-date with the current GlotPress content (will make for an easier diff down below). Commit the changes.

### Edit your Fastfile to test the action

- Adjust the `SUPPORTED_LOCALES` constant in your `Fastfile` to add `android:` key and value for each entry, where the value is supposed to be the locale code used by the Android runtime, i.e. the code used in `values-{code}` folder names.

[...]

PR #237, opened 2021-04-20
Fix error in update appstore strings lanes

JavonDavis:
This PR just fixes the incorrect variable name that was causing the updates to the appstore strings to fail but I think we'll be deprecating this entire path in https://github.com/wordpress-mobile/release-toolkit/pull/229 but I think it still makes sense to have this fix in

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb
@@ -15,7 +15,8 @@ module Fastlane
         # Do
         check_source_files(params[:source_files])
         temp_po_name = create_temp_po(params)
-        swap_po(params[:po_file_path], temp_po_name)
+        UI.message "Temp #{temp_po_name} updated!"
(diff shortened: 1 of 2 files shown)


jkmassel (approved):

mokagio:
>  I think it still makes sense to have this fix in

Totally 👍

AliSoftware:
Thanks for the Fix!

> I think we'll be deprecating this entire path in #229 but I think it still makes sense to have this fix in

FYI, #229 is for something else, so this fix was still required 😉 
 - #229 introduces an new version of the actions that download strings from GlotPress and generate `strings.xml` files from those (the strings that are used by the app itself)
 - While this `an_metadata_update_helper` is used to generate the `AppStoreStrings.po` file (which contains metadata for the stores, like release notes, as opposed to strings used by the app itself)

So those are two different actions, and the merge of 229 (to generate `strings.xml`) won't make this other action (to generate the `.po` file) obsolete. So that fix was definitively needed anyway!

JavonDavis:
Ahhh you're right @AliSoftware ! Thanks for clarifying that 😄

PR #238, opened 2021-04-27
Merge release 0.17.2 into develop

JavonDavis:
Merges release 0.17.2 into develop

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -299,6 +299,7 @@ module Fastlane
         #
         def self.apply_substitutions(tag)
           tag.content = tag.content.gsub('...', '…')
+          tag.content = tag.content.gsub('-', '&#8211;')
         end
         private_class_method :apply_substitutions
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/version.rb
@@ -1,5 +1,5 @@
 module Fastlane
   module Wpmreleasetoolkit
-    VERSION = '0.17.1'
+    VERSION = '0.17.2'
   end
 end
--- a/spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
+++ b/spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
@@ -13,7 +13,6 @@ Language: fr
     <string name="shipping_label_shipping_method_flat_rate">Forfait</string>
     <string name="shipping_label_shipping_carrier_flat_fee_banner_message">Le client a payé %1$s (%2$s) pour la livraison</string>
     <string name="shipping_label_woo_discount_bottomsheet_message" formatted="false">Lors de l’achat d’étiquettes d’expédition avec WooCommerce, vous bénéficiez de 5 à 40 %% de réduction par rapport aux tarifs de la Poste.</string>
-    <string name="shipping_label_woo_discount_bottomsheet_title">Qu’est&amp;#8211;ce que la remise WooCommerce  Services ?</string>
     <string name="shipping_label_payments_saving_dialog_message">Veuillez patienter…</string>
     <string name="shipping_label_payments_saving_dialog_title">Enregistrement des paramètres</string>
     <string name="shipping_label_shipping_carrier_rates_generic_error">Une erreur est survenue lors du chargement des options de livraison</string>
--- a/spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
+++ b/spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
@@ -13,7 +13,7 @@ Language: fr
     <string name="shipping_label_shipping_method_flat_rate">Forfait</string>
     <string name="shipping_label_shipping_carrier_flat_fee_banner_message">Le client a payé %1$s (%2$s) pour la livraison</string>
     <string name="shipping_label_woo_discount_bottomsheet_message" formatted="false">Lors de l’achat d’étiquettes d’expédition avec WooCommerce, vous bénéficiez de 5 à 40 %% de réduction par rapport aux tarifs de la Poste.</string>
-    <string name="shipping_label_woo_discount_bottomsheet_title">Qu’est-ce que la remise WooCommerce  Services ?</string>
+    <string name="shipping_label_woo_discount_bottomsheet_title">Qu’est&amp;#8211;ce que la remise WooCommerce  Services ?</string>
     <string name="shipping_label_payments_saving_dialog_message">Veuillez patienter…</string>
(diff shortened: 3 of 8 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -299,6 +299,7 @@ def self.merge_xml_documents(all_xmls)
         #
         def self.apply_substitutions(tag)
           tag.content = tag.content.gsub('...', '…')
+          tag.content = tag.content.gsub('-', '&#8211;')
```
mokagio, line 302: TIL.

How come this substitution then becomes `&amp;#8211;` in the generated strings?

mokagio (commented):
Hey @JavonDavis ! Thank you for releasing a new version of the toolkit 🎉 

Could you explain why the en dash substitution is necessary, and how does it relate to removing the problematic strings that resulted in test failures?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
AliSoftware: @JavonDavis Wait, who added this change? I remember explicitly explaining in another PR why this was a bad idea 😓  because it will solve some linter errors but introduce others (en dash should only be used instead of standard dash in certain circumstances – especially, for ranges – but not all, so replacing all instances is semantically and typographically incorrect, and will introduce the bug for the other use case)

We discussed in 229 that this kind of misuse of the dash (use of regular dash when en dash is expected for a range, use of en dash when regular dash is expected for hyphenation and other cases, …) had to be solved in GlotPress directly, either by the translators or by ourselves (it's an easy enough change that we don't need to know the language to edit the GlotPress entry), but definitively not automatically and unconditionally for all cases.

AliSoftware, line 302: @JavonDavis Not sure why you added this change… but I explained in #229 [why this is probably a bad idea](https://github.com/wordpress-mobile/release-toolkit/pull/229#issuecomment-820180415) 😓

This kind of misuse of the proper dash in translations (use of regular dash when en dash is expected for a range, use of en dash when regular dash is expected for hyphenation and other cases, …) have to be solved on case by case basis in GlotPress directly, either by the translators or by ourselves (it's an easy enough change that we don't need to know the language to edit the GlotPress entry), but definitively not automatically and unconditionally for all cases, because the right dash to use depends on the use case and typographical context.

_Besides, even if we wanted to do that, the substitution is incorrect here because Nokogiri (the XML gem) will treat that as verbatim text (you're telling it to replace the node `content` after all) so the `&` character followed by `#` followed by `8211` followed by `;`. Which will be encoded to `&amp;#8211;` once serialized in XML, as should be…_

lib/fastlane/plugin/wpmreleasetoolkit/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fastlane
   module Wpmreleasetoolkit
-    VERSION = '0.17.1'
+    VERSION = '0.17.2'
```
AliSoftware, line 3: I would suggest that we bump to `0.18.0` instead, because this new version not only contains bug fixes (patch bump) but also new features (the new action), which according to semver should lead to minor bump.

spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
```diff
@@ -24,7 +23,7 @@ Language: fr
     <string name="shipping_label_create_price_total">Total de la commande</string>
     <string name="shipping_label_create_price_woo_discount_content_description">En savoir plus sur la remise WooCommerce Services</string>
     <string name="shipping_label_create_price_woo_discount">Remise WooCommerce Services</string>
-    <string name="shipping_label_create_price_subtotal">Sous-total</string>
+    <string name="shipping_label_create_price_subtotal">Sous&amp;#8211;total</string>
```
AliSoftware, line 26: This result is incorrect for 2 reasons:

1. This will lead to the literal text once decoded being `Sous&#8211total;` because the `&amp;` will be replaced by `&` but not escape the `#8211` hex code which will stay verbatim.
2. This is typically an example of context where the regular dash (aka hyphen) is the correct use, not the en dash.

The short version is that `en dash` is for ranges or rhetorical pause, while hyphen is for compound words (like here).

spec/test-data/translations/glotpress-download/expected/values-pt-rBR/strings.xml
```diff
@@ -10,13 +10,12 @@ Language: pt_BR
     <string name="shipping_label_payments_saving_dialog_message">Aguarde…</string>
     <string name="shipping_label_payments_saving_dialog_title">Salvando suas configurações</string>
     <string name="shipping_label_payments_expiration_date">Expira em %1$s</string>
-    <string name="shipping_label_payments_email_receipts_checkbox">Enviar os recibos de compra da etiqueta por e-mail para %1$s (%2$s) em %3$s</string>
+    <string name="shipping_label_payments_email_receipts_checkbox">Enviar os recibos de compra da etiqueta por e&amp;#8211;mail para %1$s (%2$s) em %3$s</string>
```
AliSoftware, line 13: this is also a compound word, not a range, so the use of the hyphen was correct and en dash isn't

AliSoftware (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
AliSoftware: Because we're replacing `content`, which is an XML text node. It's not up to us to do the serialization/encoding, it's what the XML lib (Nokogiri) handles for us. So by setting `tag.content` to a value with `&#8211;` we are telling Nokogiri that this is the literal text that we want. Then Nokogiri does its job that ensures that the text content is properly escaped (encode `<` as `&lt;`, `>` as `&gt;`, `&` as `&amp;`, …) when it serializes the nodes as XML document.

JavonDavis: Hey @AliSoftware ! You're right!, thanks for catching this 🤦 !

lib/fastlane/plugin/wpmreleasetoolkit/version.rb
JavonDavis: I agree here and will bump this to 0.18 instead

spec/test-data/translations/glotpress-download/expected/values-pt-rBR/strings.xml
JavonDavis: I misunderstood the linter error here a bit and was too quick to just overwrite here, I'll undo this 👍

JavonDavis:
Closing in favor of #240

PR #242, opened 2021-04-28
Add Git helper method to check if a path results ignored

mokagio:
Adds a bit of security to the secrets management by preventing a user from decrypting a secret to a location that is not `.gitgnored`.

It covers the `configure_apply`, `_update`, and `_add_files_to_copy` actions.

I decided not to edit the `.gitignore` via the automation to leave the user the flexibility to change path or use a pattern in the `.gitignore`.

To test this implementation, you can checkout [this demo PR](https://github.com/Automattic/simplenote-ios/pull/1263) or do something along the lines of:

1. Get a repo on this version of the release toolkit
2. Update the `.configure` with a file to decrypt to a location that's not ignored
3. Run the `configure_apply` and `configure_update` actions to verify they show an error
4. Run the `configure_add_files_to_copy` and pass a destination path not ignored to verify it shows an error

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
@@ -54,39 +54,43 @@ module Fastlane
       ### Check with the user whether we should overwrite the file, if it exists
       ###
       def self.apply_file(file_reference, force)
-        # If the file doesn't exist or force is true, we don't need to confirm
-        if !File.file?(file_reference.destination) || force
-          file_reference.apply
-          return # Don't continue if we were able to copy the file without conflict
-        end
+        begin
+          # If the file doesn't exist or force is true, we don't need to confirm
+          if !File.file?(file_reference.destination) || force
+            file_reference.apply
+            return # Don't continue if we were able to copy the file without conflict
+          end
 
-        unless file_reference.needs_apply?
-          return # Nothing to do if the files are identical
-        end
+          unless file_reference.needs_apply?
+            return # Nothing to do if the files are identical
+          end
 
-        puts Diffy::Diff.new(file_reference.destination_contents, file_reference.source_contents) if UI.confirm("#{file_reference.destination} has changes that need to be merged. Would you like to see a diff?")
+          puts Diffy::Diff.new(file_reference.destination_contents, file_reference.source_contents) if UI.confirm("#{file_reference.destination} has changes that need to be merged. Would you like to see a diff?")
 
-        if UI.confirm("Would you like to make a backup of #{file_reference.destination}?")
-          extension = File.extname(file_reference.destination)
-          base = File.basename(Pathname.new(file_reference.destination), extension)
+          if UI.confirm("Would you like to make a backup of #{file_reference.destination}?")
+            extension = File.extname(file_reference.destination)
+            base = File.basename(Pathname.new(file_reference.destination), extension)
 
-          date_string = Time.now.strftime('%m-%d-%Y--%H-%M-%S')
+            date_string = Time.now.strftime('%m-%d-%Y--%H-%M-%S')
 
-          backup_path = base
-                        .concat('-') # Handy-dandy separator
-                        .concat(date_string) # date string to allow multiple backups
-                        .concat(extension) # and the original file extension
-                        .concat('.bak') # add the .bak file extension - easier to .gitignore
+            backup_path = base
+              .concat('-') # Handy-dandy separator
+              .concat(date_string) # date string to allow multiple backups
+              .concat(extension) # and the original file extension
+              .concat('.bak') # add the .bak file extension - easier to .gitignore
 
-          # Create the destination directory if it doesn't exist
-          FileUtils.mkdir_p(Pathname.new(file_reference.destination).dirname)
-          FileUtils.cp(file_reference.destination, backup_path)
-        end
+            # Create the destination directory if it doesn't exist
+            FileUtils.mkdir_p(Pathname.new(file_reference.destination).dirname)
+            FileUtils.cp(file_reference.destination, backup_path)
+          end
 
-        if UI.confirm("Would you like to overwrite #{file_reference.destination}?")
-          file_reference.apply
-        else
-          UI.message "Skipping #{file_reference.destination}"
+          if UI.confirm("Would you like to overwrite #{file_reference.destination}?")
+            file_reference.apply
+          else
+            UI.message "Skipping #{file_reference.destination}"
+          end
+        rescue => e
+          UI.user_error!(e)
         end
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
@@ -225,6 +225,10 @@ module Fastlane
 
         UI.user_error!('You must pass a `destination` to `add_file`') unless params[:destination]
 
+        unless Fastlane::Helper::GitHelper.is_ignored?(path: params[:destination])
+          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]})"
+        end
+
         new_config = self.configuration
         new_config.add_file_to_copy(params[:source], params[:destination], params[:encrypt])
         update_configuration(new_config)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
@@ -226,7 +226,7 @@ module Fastlane
         UI.user_error!('You must pass a `destination` to `add_file`') unless params[:destination]
 
         unless Fastlane::Helper::GitHelper.is_ignored?(path: params[:destination])
-          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]})"
+          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]}). Please edit your `.gitignore` manually to fix this."
         end
 
         new_config = self.configuration
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
@@ -226,7 +226,7 @@ module Fastlane
         UI.user_error!('You must pass a `destination` to `add_file`') unless params[:destination]
 
         unless Fastlane::Helper::GitHelper.is_ignored?(path: params[:destination])
-          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]}). Please edit your `.gitignore` manually to fix this."
+          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]}). Please either edit your `.configure` file to use an already-ignored destination, or add that destination to the `.gitignore` manually to fix this."
         end
 
         new_config = self.configuration
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -165,6 +165,23 @@ module Fastlane
         current_branch_name = Action.sh('git', 'symbolic-ref', '-q', 'HEAD')
         UI.user_error!("This command works only on #{branch_name} branch") unless current_branch_name.include?(branch_name)
       end
+
+      # Checks whether a given path is ignored by Git, relying on Git's
+      # `check-ignore` under the hood.
+      #
+      # @param [String] path The path to check against the the `.gitignore`
+      #
+      # @return [Bool] True if the given path is ignored, false otherwise.
+      def self.is_ignored(path:)
+        begin
+          Action.sh('git', 'check-ignore', path)
+        rescue
+          # if there was an error, either the given path doesn't result as
+          # ignored or the command was called outside of a Git repo (as per the
+          # manpage). In both cases, we want to fail
+          false
+        end
+      end
     end
   end
 end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -175,6 +175,7 @@ module Fastlane
       def self.is_ignored?(path:)
         begin
           Action.sh('git', 'check-ignore', path)
+          return true
         rescue
           # if there was an error, either the given path doesn't result as
           # ignored or the command was called outside of a Git repo (as per the
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -173,15 +173,8 @@ module Fastlane
       #
       # @return [Bool] True if the given path is ignored, false otherwise.
       def self.is_ignored?(path:)
-        begin
-          Action.sh('git', 'check-ignore', path)
-          return true
-        rescue
-          # if there was an error, either the given path doesn't result as
-          # ignored or the command was called outside of a Git repo (as per the
-          # manpage). In both cases, we want to fail
-          false
-        end
+        system("git check-ignore #{path}")
+        $CHILD_STATUS.exitstatus == 0
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -173,8 +173,7 @@ module Fastlane
       #
       # @return [Bool] True if the given path is ignored, false otherwise.
       def self.is_ignored?(path:)
-        system("git check-ignore #{path}")
-        $CHILD_STATUS.exitstatus == 0
+        Actions.sh('git', 'check-ignore', path) { |status, _, _| status.success? }
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -172,7 +172,7 @@ module Fastlane
       # @param [String] path The path to check against the the `.gitignore`
       #
       # @return [Bool] True if the given path is ignored, false otherwise.
-      def self.is_ignored(path:)
+      def self.is_ignored?(path:)
         begin
           Action.sh('git', 'check-ignore', path)
         rescue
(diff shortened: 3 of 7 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
```diff
@@ -54,39 +54,43 @@ def self.prepare_repository
       ### Check with the user whether we should overwrite the file, if it exists
       ###
       def self.apply_file(file_reference, force)
```
mokagio, line 56: Unfortunately, the diff is noisy. The only change here was wrapping the logic in a `begin – rescue` block, to gracefully handle the expected error raised if a file is not ignored without dumping the stacktrace on the user

spec/file_reference_spec.rb
```diff
@@ -29,11 +29,30 @@
...
+      it 'copies the source to the destination' do
+        allow(Fastlane::Helper::GitHelper).to receive(:is_ignored?)
+          .with(path: subject.destination_file_path)
+          .and_return(true)
+
+        allow(FileUtils).to receive(:mkdir_p)
+        allow(subject).to receive(:source_contents).and_return('source contents')
+        expect(File).to receive(:write).with(subject.destination_file_path, 'source contents')
+        subject.apply
+      end
```
mokagio: I don't like how heavy on runtime mocking these tests are, they're more about the implementation (call `File.write`) than the actual behavior.

On the other hand, this was simpler to achieve than putting in place a temp folders etc.

lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
```diff
@@ -225,6 +225,10 @@ def self.add_file(params)
 
         UI.user_error!('You must pass a `destination` to `add_file`') unless params[:destination]
```
mokagio, line 226: I was tempted to refactor this method to use keyword arguments instead of an hash with parameters. I decided to save it for a later refactor.

cc @AliSoftware in case that's something that's already on your TODO list.

spec/git_helper_spec.rb
```diff
@@ -90,4 +91,58 @@
...
+
+  describe '#is_ignored?' do
+    let(:path) { 'dummy.txt' }
+
+    it 'returns false when the path is not ignored' do
+      setup_git_repo(
+        dummy_file_path: path,
+        in_gitignore: false
+      )
+      expect(Fastlane::Helper::GitHelper.is_ignored?(path: path)).to be false
```
mokagio, line 106: I could have used `described_class` here like in the spec above, but I decided not to because the rest of the file doesn't.

(Also, I remembered about `described_class` after writing these tests... 😅 )

AliSoftware: We still haven't configured CircleCI to run rubocop on each PR (something I really need to take care of before we start reintroducing too many new violations via all our recent updates after my previous cleanup!!) but I know that this would have been flagged by rubocop…

mokagio: 🤔 I actually did run Rubocop before submitting my PR (and found some issues, see 79516ac)

What were you expecting it to flag? Right now, it's all green:

(_I tried to upload a screenshot but I'm getting "Something went really wrong" from the editor._ _I also tried to screenshot and upload that..._ 🤦 😅 ) 




I may be able to do it tomorrow / Friday. Were you thinking something simple like "run on every file that changed" or more advanced like comment in line with the violations?

Oh 🤔 Now that I think of it, maybe we should setup Hound?

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
AliSoftware: "Ignore whitespace" to the rescue! 
image

spec/git_helper_spec.rb
AliSoftware: > What were you expecting it to flag? Right now, it's all green:

I was expecting [this rule](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribedClass) to be triggered 🤔 

> Oh 🤔 Now that I think of it, maybe we should setup Hound?

Woof!! (That means "definitively!!" in case you don't speak dog)

lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
```diff
@@ -225,6 +225,10 @@ def self.add_file(params)
...
 
+        unless Fastlane::Helper::GitHelper.is_ignored?(path: params[:destination])
+          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]})"
```
AliSoftware:
```suggestion
          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]}). Please edit your `.gitignore` manually to fix this."
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -159,6 +159,17 @@ def self.ensure_on_branch!(branch_name)
...
+
+      # Checks whether a given path is ignored by Git, relying on Git's
+      # `check-ignore` under the hood.
+      #
+      # @param [String] path The path to check against the the `.gitignore`
+      #
+      # @return [Bool] True if the given path is ignored, false otherwise.
+      def self.is_ignored?(path:)
+        system("git check-ignore #{path}")
+        $CHILD_STATUS.exitstatus == 0
```
```suggestion
        sh('git', 'check-ignore', path) { |status, _, _| status.success? }
```

(Not tested, so worth checking my syntax, but according to https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/helper/sh_helper.rb#L11-L28 I think that's the right one)

lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
```diff
@@ -42,7 +42,18 @@ def update
...
+      # "Applies" the instruction described in the instance to the file system.
+      # That is, copies the content of the source `file` to the `destination`
+      # path.
+      #
+      # @raise [StandardError] For security reasons, it will raise if
+      # `destination` is not ignored under Git.
       def apply
+        # Only decrypt the file if the destination is ignored in Git
+        unless Fastlane::Helper::GitHelper.is_ignored?(path: destination_file_path)
+          raise StandardError.new "Attempted to decrypt #{file} to #{destination_file_path} which is not ignored under Git"
```
AliSoftware: Surprised that rubocop doesn't suggest you ere to directly use the String literal without `StandardError.new`, as this is the more idiomatic way iinm. That file might be excluded from rubocop though (because I never had time to finish fixing all the existing violations back then)
```suggestion
          raise "Attempted to decrypt #{file} to #{destination_file_path} which is not ignored under Git"
```

spec/git_helper_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'tmpdir'
 require_relative './spec_helper'
 
+# rubocop:disable Metrics/BlockLength
```
AliSoftware: I'm surprised this is expected, as I am pretty sure I explicitly excluded `*_spec.rb` files from this rule in `.rubocop.yml` already 🤔

spec/git_helper_spec.rb
```diff
@@ -90,4 +91,58 @@
...
+          in_gitignore: true,
+          gitignore_committed: true
+        )
+        expect(Fastlane::Helper::GitHelper.is_ignored?(path: path)).to be true
+      end
+      # rubocop:enable RSpec/ExampleLength
+    end
+  end
+end
+# rubocop:enable Metrics/BlockLength
```
AliSoftware: (don't forget to remove this one if you do remove the disable after my previous comment)

spec/git_helper_spec.rb
```diff
@@ -90,4 +91,58 @@
...
+
+def setup_git_repo(dummy_file_path:, in_gitignore:, gitignore_committed: false)
```
AliSoftware: param naming suggestion:
```suggestion
def setup_git_repo(dummy_file_path:, add_file_to_gitignore:, commit_gitignore: false)
```
(+ adapt implementation accordingly)

AliSoftware (commented):
Mostly LGTM, just left some nits and improvement suggestions though.

spec/git_helper_spec.rb
mokagio: You are right. Turns out my `develop` wasn't up to date when created this branch. I rebased and it didn't warn me. 

Fixed up in 830d39c.

lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
mokagio: Rubocop _did_ suggest it, once I enabled it in my Vim setup... 😅 

It also pointed out a few more violations which I'll address later in this PR, as I'm running out of time today. I'll re-request your review once done.

Thank you for pointing this out 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
mokagio: Thanks. That looks neat. Is there any benefit to using that syntax compared to this one?

AliSoftware: `sh` is a method provided by Fastlane, so it's nicer to use it because it wraps the `system` call around a couple of things, including reporting results in logs via `UI.xxx` messages, capturing errors for reporting too, disabling real execution by default when running it in unit tests…

mokagio: I implemented this in b9d543a 👍 

I'm not convinced by your argument, though 😄 

For the purpose of this method, I think using vanilla Ruby was better. It doesn't depend on anything external so it's less likely to break because there's one less moving part. Also, I had to [update the tests](https://github.com/wordpress-mobile/release-toolkit/pull/242/commits/b9d543a0a5a3d60380c36c0cab4750de7517c920#diff-7aed954270eabe63c0117e184d02a920e7c50e090136cd8a33fe121bd7e87ca9R95-R97) to account for the usage of `Action.sh` by calling `allow(FastlaneCore::Helper).to receive(:sh_enabled?).and_return(true)`, which adds an avoidable dependency on the code implementation to tests that had no dependency to it before that.

Neither of my points is very strong, though. It's possible that using Fastlane will actually prevent future bugs coming from Ruby itself, it also delegates the non-trivial job of properly shelling out to a codebase used by millions of people, which should make it quite stable at that.

As for the tests, well the dependency is minor and it's nice that is not about `sh` itself but rather something internal. That is, if we were to use a different helper method or drop `Action.sh` altogether, the tests will likely still pass.

**I'm happy to go with this change, because, after all, this code is deeply embedded in the  Fastlane ecosystem, so it makes sense to leverage some of its utilities.** 😄 🚗 💨

---

@AliSoftware, you mention "reporting results in logs via UI.xxx messages, capturing errors for reporting too".

How does that work? I tried to run the action [in my test PR with these changes](https://github.com/Automattic/simplenote-ios/pull/1263) with `--verbose` but I didn't get anything extra in the logs from this `sh` call 🤔

mokagio:
@AliSoftware I think I addressed all your great comments. I'd appreciate a second look 🙏 👀

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: Yeah I see your point it's debatable whether to use `sh` or not.

I always try to avoid calling `Kernel#system` directly tho as it might be open to injection iirc (due to it running a subshell and going thru shell expansion), so that's why `popen2e` — which is what `sh` uses — and its family of similar `popen*` functions (which directly spawns the subprocess without an intermediate subshell) is recommended instead, and has the advantages of:

1. not be vulnerable to shell expansion like `system` is, and instead takes an array or arguments and spawn the subprocess and pass it its argv list directly (without going thru a shell re-splitting the args and having to handle quoting and escapes and all);
2. be able to capture both `stdout` and `stderr` (and exit `status`) all at once or separately, unlike `system`;
3. Not rely on a global variable — like `$CHILD_STATUS` — to read the result.

If you're curious about how sh is implemented and what call it wraps to end up invoking the external scripts or command lines, you can take a look [here](https://github.com/fastlane/fastlane/blob/e874a47c6e2e0e61590a03d3b71e75e5a505d1ce/fastlane/lib/fastlane/helper/sh_helper.rb).

In the end for this use case it might not end up being a big deal, but better be consistent and avoid spawning an intermediate shell (that evaluates the command and args and quotes and escapes to split that into args, like with `system`) for nothing and invoke the command directly instead (like with `popen2e` or `sh`) 😇

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -165,6 +165,16 @@ def self.ensure_on_branch!(branch_name)
         current_branch_name = Action.sh('git', 'symbolic-ref', '-q', 'HEAD')
         UI.user_error!("This command works only on #{branch_name} branch") unless current_branch_name.include?(branch_name)
       end
+
+      # Checks whether a given path is ignored by Git, relying on Git's
+      # `check-ignore` under the hood.
+      #
+      # @param [String] path The path to check against the the `.gitignore`
```
```suggestion
      # @param [String] path The path to check against `.gitignore`
```

spec/git_helper_spec.rb
```diff
@@ -90,4 +90,61 @@
...
+  `touch .gitignore`
+  `git add .gitignore && git commit -m 'Add .gitignore'`
+
+  `echo abc > #{dummy_file_path}`
+
+  # no point in commiting the gitignore if the file shouldn't be in it
+  return unless add_file_to_gitignore
+
+  `echo #{dummy_file_path} > .gitignore`
+  `git add .gitignore && git commit -m 'Update .gitignore'` if commit_gitignore
```
AliSoftware, line 149: First time I read this I thought "it would be nice and safer if all this was done in a temporary directory". Then I went back to read the whole spec file and saw that the tmpdir switch is done via `before` and `after` on the whole spec, so should be ok after all, just thought I'd leave that comment around so that other reviewers aren't tricked by same train of thought 😛

AliSoftware (commented):
Code LGTM 👍 
Haven't taken the time to run the test steps yet though, so will circle back later today to test it before approving

lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
```diff
@@ -42,7 +42,18 @@ def update
...
+      # "Applies" the instruction described in the instance to the file system.
+      # That is, copies the content of the source `file` to the `destination`
+      # path.
+      #
+      # @raise [StandardError] For security reasons, it will raise if
+      # `destination` is not ignored under Git.
       def apply
+        # Only decrypt the file if the destination is ignored in Git
+        unless Fastlane::Helper::GitHelper.is_ignored?(path: destination_file_path)
+          raise "Attempted to decrypt #{file} to #{destination_file_path} which is not ignored under Git"
```
AliSoftware:
```suggestion
          raise "Attempted to decrypt #{file} to #{destination_file_path} which is not ignored under Git. Please either edit your `.configure` file to use an already-ignored destination, or add that destination `.gitignore` manually to fix this."
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/configure_helper.rb
```diff
@@ -225,6 +225,10 @@ def self.add_file(params)
 
         UI.user_error!('You must pass a `destination` to `add_file`') unless params[:destination]
 
+        unless Fastlane::Helper::GitHelper.is_ignored?(path: params[:destination])
+          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]}). Please edit your `.gitignore` manually to fix this."
```
AliSoftware:
```suggestion
          UI.user_error! "Attempted to add a file to a location which is not ignored under Git (#{params[:destination]}). Please either edit your `.configure` file to use an already-ignored destination, or add that destination `.gitignore` manually to fix this."
```
I know it's a bit lengthy, but I prefer more hints/documentation than having people (including future us) losing time wondering what's the expected fix 🙂 

PS: Feel free to reword, or introduce a `\n` somewhere in the message, and/or use another ruby string syntax (e.g. heredoc) to format this in a way that makes rubocop happy if needed.

AliSoftware (approved):
Just added 2 wording nit/suggestions for better error messages; otherwise LGTM and works as expected 👍 

I even tried to trick the script by using `.configure-files/../foo.json` as the destination to see if it would fail see the `../` trick, but it saw right through me and didn't let me have it 👌 

<details><summary>Couldn't outsmart it! 😸 </summary>

```
$ fl run configure_apply
[15:47:29]: -----------------------------
[15:47:29]: --- Step: configure_apply ---
[15:47:29]: -----------------------------
[15:47:29]: $ git check-ignore /Users/olivier/Documents/Dev/apps/woocommerce-ios/.configure-files/../app_store_connect_fastlane_api_key.json
[15:47:29]: Exit status of command 'git check-ignore /Users/olivier/Documents/Dev/apps/woocommerce-ios/.configure-files/../app_store_connect_fastlane_api_key.json' was 1 instead of 0.

[!] Attempted to decrypt iOS/app_store_connect_fastlane_api_key.json to /Users/olivier/Documents/Dev/apps/woocommerce-ios/.configure-files/../app_store_connect_fastlane_api_key.json which is not ignored under Git
```
```
$ fl run configure_add_files_to_copy
[15:48:09]: -----------------------------------------
[15:48:09]: --- Step: configure_add_files_to_copy ---
[15:48:09]: -----------------------------------------
[15:48:09]: Do you want to specify additional files that should be copied from the secrets repository into your project? (y/n)
y
[15:48:16]: -----------------------------------------------------------------------------------------
[15:48:16]: --- Please provide the location of the source file relative to the secrets repository ---
[15:48:16]: -----------------------------------------------------------------------------------------
[15:48:16]: Example: google-services.json
[15:48:16]: Source File Path:
iOS/tracks-demo/Secrets.swift
[15:48:59]: -------------------------------------------------------------------------------
[15:48:59]: --- Please provide the destination of the file relative to the project root ---
[15:48:59]: -------------------------------------------------------------------------------
[15:48:59]: Example: WordPress/google-services.json
[15:48:59]: Destination File Path:
.configure-files/../whoops.swift
[15:49:15]: Encrypt file?: (y/n)
y
[15:49:16]: $ git check-ignore .configure-files/../whoops.swift
[15:49:17]: Exit status of command 'git check-ignore .configure-files/../whoops.swift' was 1 instead of 0.


[!] Attempted to add a file to a location which is not ignored under Git (.configure-files/../whoops.swift). Please edit your `.gitignore` manually to fix this.
```
</details>

spec/git_helper_spec.rb
mokagio: When I wrote the first version of this test, it took me at least 5 minutes or beard scratching to figure out why it wasn't working, which was because of the temp dir being set in the `before`.

There's some lessons here for me, nothing new, but stuff that bears repeating:

- Always read (or at least skim) a test file before starting to work on it
- Beware of global setup code, it's sneaky and easy to overlook

AliSoftware: > There's some lessons here for me, nothing new, but stuff that bears repeating:

you forgot:

3. Keep your beard, it's always handy to have it around to be able to scratch it.

PR #243, opened 2021-04-29
Add Hound CI config file

mokagio:
Configures Hound, as discussed [here](https://github.com/wordpress-mobile/release-toolkit/pull/242#discussion_r622018937).

Opened #244 to test the integration. Current status: **not working**.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,9 @@
+rubocop:
+  # This should stay in line with the version resolved in the Gemfile.lock,
+  # which you can verify with `bundle info rubocop`.
+  #
+  # Notice that Hound supports only a subset of the RuboCop versions. You can
+  # find the list of supported versions here:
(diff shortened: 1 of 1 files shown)


jkmassel:
WDYT about just dumping hound for this and using our existing CI system to run `rubocop`? That way we get guarantees about which version we run (by locking in the `Gemfile`), and don't have to mess around with a third system?

/cc @AliSoftware

AliSoftware:
I'm ok with running `rubocop` as a step on CI.

The nice thing with Hound is that it reports back on the PR via inline comments iinm, which is nicer than having to dig into the CI log to try to understand what's wrong. But also… if the CI lint step fails, it might be easier to just launch RubyMine and run its rubocop integration and see and fix the violations from there directly, so probably not a big deal after all. So I say let's keep things simple and not waste too much time on debugging Hound indeed.

mokagio:
Closed in favor of #246.

PR #244, opened 2021-04-29
Test PR for #243

mokagio:
There should be comments from Hound here. 🐶

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,9 @@
+rubocop:
+  # This should stay in line with the version resolved in the Gemfile.lock,
+  # which you can verify with `bundle info rubocop`.
+  #
+  # Notice that Hound supports only a subset of the RuboCop versions. You can
+  # find the list of supported versions here:
+  # http://help.houndci.com/en/articles/2461415-supported-linters
+  version: 0.75.0
+  config_file: .rubocop.yml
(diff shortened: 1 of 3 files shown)


mokagio:
🤔 none of the violations which I see locally running `bundle exec rubocop` are reported by Hound in this PR...

AliSoftware:
Are we sure that Hound:
- uses the config file from the branch of the PR, as opposed to always use the one in default branch (develop) — which didn't exist yet until 243 is merged?
- doesn't need any extra setup in the repo settings to enable the Hound app for the repo?

(Never actually configured Hound and have yet to read their doc about setup instructions)

mokagio:
> Never actually configured Hound 

Me neither. Their docs are quite sparse, too.

I had a look in the `wordpress-mobile` org settings and the Hound app is already enabled. The repo also appear in the lists of repos in https://houndci.com/repos:

image

> uses the config file from the branch of the PR, as opposed to always use the one in default branch (develop) — which didn't exist yet until 243 is merged?

Good questions 🤔 🤔 🤔 I might need to setup a repo on my user to verify this

mokagio:
Ah! 

99% sure this is the actual cause of the missing violations:

image

I'm tempted to disable that switch and ~~watch the world burn~~ see what happens. But this is the end of my day so I might save it for another time. 

Alternatively, we could easily™ setup the Ruby version of Danger and run RuboCop with it in CircleCI.

Or... pause this for now and do "real" work 😆

PR #246, opened 2021-04-30
Run Rubocop in CI

mokagio:
This plugs Rubocop into our existing Danger + GitHub Actions configuration, with inline comments.

You can see it in action in #247.

image

There is one confusing bit, which is that we get two reports from Danger in the checks:

image

I tried to play with the plugin [configs](https://github.com/ashfurrow/danger-rubocop#methods), using `report_danger` but that broke the inline comments 🤔 I'd say we can live with it for now.


Gemfile.lock
```diff
@@ -65,7 +65,7 @@ GEM
       colored2 (~> 3.1)
     crack (0.4.5)
       rexml
-    danger (8.2.1)
+    danger (8.2.3)
```
mokagio, line 68: Updated Danger while I was at it.

Dangerfile
```diff
@@ -22,3 +22,11 @@ Please run `bundle install` to make sure they match.
 }
   fail(message)
 end
+
+# Lint with Rubocop and report violations inline in GitHub
+github.dismiss_out_of_range_messages # This way, fixed violations should go
+rubocop.lint(
+  files: git.modified_files + git.added_files,
+  inline_comment: true,
+  fail_on_inline_comment: true # Make the inline comments failures
```
mokagio, line 32: I decided to configure it strict, so that we won't miss lint violations because they'll make the build fail.

AliSoftware: This list of files miss some cases:

* It forgets to remove deleted files. One would think that just not listing the deleted files in the list will just not include it, but I think I've seen in the past some cases where the deleted files were also present in the modified files list, at least I have example of Dangerfiles I used in past projects where I explicitly removed them from the list to be sure, so better safe than sorry.
```
  files: (git.modified_files - git.deleted_files) + git.added_files
```

* It also doesn't handle renamed files. Those are a more complicated beast, because `renamed_files` is not just a flat array, but a list of old name => new name pairs, so we need to `map` to get the proper list. (I sadly didn't manage to dig example of past Dangerfiles on how I did that in past project, so probably needs some digging to figure out the right way)


([EDIT] Sorry, meant to post this as a review not a single comment. eh 🤷  )

mokagio: Thanks @AliSoftware. I'll test these out in #247...

AliSoftware: Reopened #247 to push a file renaming and test this out, and indeed it appears that it doesn't lint the file if it has been renamed as I feared. See https://github.com/wordpress-mobile/release-toolkit/pull/247#issuecomment-831972829 for more details.

mokagio: Thanks a lot for looking into this @AliSoftware 🙇 .

I started working on this because I hoped it would have been a quick fix, but it looks like it isn't 😞 .

I don't have bandwidth to look into it right now, but I definitely want to see this happen here. Feel free to take over the PR if you have time / ideas on how to fix the setup. Otherwise, I'll get to this likely past mid-May.

AliSoftware: Submitted a patch to `danger-rubocop` FYI: https://github.com/ashfurrow/danger-rubocop/pull/50

AliSoftware:
@mokagio I pushed fad07fa7c181e60623f91521e0d8b3fba59b9f5e to fix the renaming case. See it in action in https://github.com/wordpress-mobile/release-toolkit/pull/247

Even with the limitation on inline comments highlighted in #247, I think it is acceptable to merge this as is (which is why I approved the PR) – even if there appears to be a limit in the number of inline comments at once (the important thing is that the PR won't go green until we fixed all the violations – as otherwise the next inline comments will start to appear as we fix the first ones).

We can always iterate, but I think it's more important to have this enabled asap at least so that we don't drift away too much from a clean and linted codebase otherwise without that check at all. I'll let you the honor of merging the PR if you agree.

AliSoftware (approved):

mokagio:
_Speed of iteration trumps quality of iteration_. Let's get this in 👍 

Thank you for helping out @AliSoftware !

Dangerfile
AliSoftware: (…which has already been merged and released into a new version of danger-rubocop! 🎉 )

PR #247, opened 2021-04-30
PR to test #246

mokagio:

PR #251, opened 2021-05-05
PR to test #246 - take 2

AliSoftware:
Same as https://github.com/wordpress-mobile/release-toolkit/pull/247 but creating a brand new PR to see if that changes anything wrt to GitHub actions creating inline comments on a brand new PR (as opposed to trying to update existing PR on which it ran with different configuration in priori runs)


AliSoftware:
Still only 2 inline comments despite 11 violations… even on a fresh PR… 😢

PR #252, opened 2021-05-06
Allow secret decryption to location outside Git repository

mokagio:
In _previous PR_, we prevented `configure` from decrypting files to a location that is not ignored by Git. This relies under the hood on `git check-ignore`.

Soon after merging, I realized that this solution was incomplete because `git check-ignore` will fail when given a path that is not inside the repository.

That's no good for us because we want to be able to decrypt files to `~/.configure/$PROJECT_NAME`.

You can see this in action in https://github.com/Automattic/simplenote-ios/pull/1276, which includes "testing" instructions.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -192,7 +192,7 @@ module Fastlane
         end
       end
 
-      OUTSIDE_OF_REPO_ERROR = 'is outside repository at'.freeze
+      OUTSIDE_OF_REPO_ERROR = 'is outside repository'.freeze
       NOT_A_REPO_ERROR = 'fatal: not a git repository (or any of the parent directories): .git'.freeze
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -23,11 +23,14 @@ module Fastlane
         return current_dir.root? == false
       end
 
-      # Travels back the hierarchy of the given path until it finds an existing ancestor, or it reaches the root of the file system.
+      # Travels back the hierarchy of the given path until it finds an existing
+      # ancestor, or it reaches the root of the file system.
       #
       # @param [String] path The path to inspect
       #
-      # @return [Pathname] The first existing ancestor, or `path` itself if it exists
+      # @return [Pathname] The first existing ancestor, or `path` itself if it
+      #         exists
+      #
       def self.first_existing_ancestor_of(path:)
         p = Pathname(path).expand_path
         p = p.parent until p.exist? || p.root?
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -204,16 +204,12 @@ module Fastlane
       #
       # @return [Bool] True if the given path is ignored, false otherwise.
       def self.is_ignored?(path:)
-        Actions.sh('git', 'check-ignore', path) do |status, output, _|
-          return true if output.strip == NOT_A_REPO_ERROR
-          return true if output.include? OUTSIDE_OF_REPO_ERROR
+        return true unless is_git_repo?(path: path)
 
+        Actions.sh('git', 'check-ignore', path) do |status, _, _|
           status.success?
         end
       end
-
-      OUTSIDE_OF_REPO_ERROR = 'is outside repository'.freeze
-      NOT_A_REPO_ERROR = 'fatal: not a git repository (or any of the parent directories): .git'.freeze
     end
   end
 end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -15,12 +15,19 @@ module Fastlane
       #         (i.e. a local working copy) or a subdirectory of one.
       #
       def self.is_git_repo?(path: Dir.pwd)
-        working_path = first_existing_ancestor_of(path: path)
-        current_dir = working_path.directory? ? working_path : working_path.dirname
+        # If the path doesn't exist, find its first ancestor.
+        path = first_existing_ancestor_of(path: path)
+        # Get the path's directory, so we can look in it for the Git folder
+        dir = path.directory? ? path : path.dirname
 
-        current_dir = current_dir.parent until Dir.entries(current_dir).include?('.git') || current_dir.root?
+        # Recursively look for the Git folder until it's found or we read the
+        # the file system root
+        dir = dir.parent until Dir.entries(dir).include?('.git') || dir.root?
 
-        return current_dir.root? == false
+        # If we reached the root, we haven't found a repo. (Technically, there
+        # could be a repo in the root of the system, but that's a usecase that
+        # we don't need to support at this time)
+        return dir.root? == false
       end
 
       # Travels back the hierarchy of the given path until it finds an existing
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -19,7 +19,7 @@ module Fastlane
         args += ['-C', File.file?(path) ? File.dirname(path) : path] unless path.nil?
         args += ['rev-parse']
 
-        Action.sh(args, print_command: true, print_command_output: true, error_callback: ->(e) { puts e }) do |status, _, _|
+        Action.sh(args, print_command_output: false) do |status, _, _|
           status.success?
         end
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -10,7 +10,9 @@ module Fastlane
       # @return [Bool] True if the current directory is the root of a git repo (i.e. a local working copy) or a subdirectory of one.
       #
       def self.is_git_repo?
-        system 'git rev-parse --git-dir 1> /dev/null 2>/dev/null'
+        Action.sh('git', 'rev-parse', '--git-dir', print_command: false, print_command_output: false) do |status, _, _|
+          status.success?
+        end
       end
 
       # Check if the current directory has git-lfs enabled
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -16,12 +16,32 @@ module Fastlane
       #
       def self.is_git_repo?(path: nil)
         args = ['git']
-        args += ['-C', File.file?(path) ? File.dirname(path) : path] unless path.nil?
+
+        unless path.nil?
+          # `path` might represent a file that doesn't yet exist but is in a
+          # valid Git repo. In such case, the `git rev-parse` we're currently
+          # building would fail. To avoid that, let's perform the check on the
+          # first existing ancestor of the given path – which will be the path
+          # itself if it exists.
+          path = first_existing_ancestor_of(path: path)
+          args += ['-C', File.file?(path) ? File.dirname(path) : path]
+        end
+
         args += ['rev-parse']
 
         Action.sh(args, print_command_output: false) do |status, _, _|
           status.success?
         end
+      end
+
+      # Travels back the hierarchy of the given path until it finds an existing ancestor, or it reaches the root of the file system.
+      #
+      # @param [String] path The path to inspect
+      #
+      # @return [String] The first existing ancestor, or `path` itself if it exists
+      def self.first_existing_ancestor_of(path:)
+        path = Pathname(path).expand_path.parent until File.exist?(path) || path == Pathname('/')
+        return path.to_s
       end
 
       # Check if the current directory has git-lfs enabled
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -202,7 +202,7 @@ module Fastlane
       #
       # @param [String] path The path to check against `.gitignore`
       #
-      # @return [Bool] True if the given path is ignored or outside the Git repository, false otherwise.
+      # @return [Bool] True if the given path is ignored or outside a Git repository, false otherwise.
       def self.is_ignored?(path:)
         return true unless is_git_repo?(path: path)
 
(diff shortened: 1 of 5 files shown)


mokagio:
🤔 the [tests fail in CI](image), but are green on my machine:

https://circleci.com/gh/wordpress-mobile/release-toolkit/733?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

spec/git_helper_spec.rb
```diff
@@ -132,6 +132,23 @@
       end
       # rubocop:enable RSpec/ExampleLength
     end
+
+    it 'returns true when the path is outside the repository folder' do
+      # This path is in the parent directory, which is not a Git repo
+      path = File.join(@path, '..', 'dummy.txt')
```
mokagio, line 171: 🤔 maybe the reason the tests fails has to do with how `@path` is generated in macOS vs the CI environment where they run?

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -173,8 +173,16 @@ def self.ensure_on_branch!(branch_name)
...
+        Actions.sh('git', 'check-ignore', path) do |status, output, _|
+          return true if output.strip == NOT_A_REPO_ERROR
+          return true if output.include? OUTSIDE_OF_REPO_ERROR
+
+          status.success?
+        end
       end
+
+      OUTSIDE_OF_REPO_ERROR = 'is outside repository at'.freeze
+      NOT_A_REPO_ERROR = 'fatal: not a git repository (or any of the parent directories): .git'.freeze
```
mokagio: @jkmassel @AliSoftware what do you think of this approach to return `true` if a path is outside of the Git repository? 

I'm not particularly happy about checking error message strings, but the exit status of `git check-ignore` is not refined enough (see `git check-ignore --help`):

image

I thought about doing some checks on the path here manually, but it felt like reinventing the wheel.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -173,8 +173,16 @@ def self.ensure_on_branch!(branch_name)
       #
       # @return [Bool] True if the given path is ignored, false otherwise.
```
mokagio:
```suggestion
      # @return [Bool] True if the given path is ignored or outside the Git repository, false otherwise.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: I'm not a fan of matching error message either.

What about testing if this is a git repo separately, before invoking `check-ignore`?

I'm especially thinking about something like `git -C #{File.dirname(path)} rev-parse --show-toplevel` , or `git -C #{File.dirname(path)} rev-parse --git-dir`… or even `git -C #{File.dirname(path)} rev-parse` without any other flag 🤔

_PS: If you don't know about it, `-C` allows to execute the `git` command from within a specific directory. Avoids having to manually  `cd` into that directory before (and `cd -` back after) the git command._

```
$ git -C $(dirname ~/app_store_connect_fastlane_api_key.json) rev-parse                                         
fatal: not a git repository (or any of the parent directories): .git

$ echo $?
128
                                                                                                                                                                                                                                
$ git -C $(dirname ~/Documents/Dev/apps/woocommerce-ios/.configure-files/app_store_connect_fastlane_api_key.json) rev-parse
                                                                                                                                                                                                                                
$ echo $?                                                                                                                  
0
```

mokagio: Nice suggestion. I had a note to look into `rev-parse`, but started with the apparently simpler solution from this PR.

What attracts me about your idea is that it will allow us to prevent decrypting into _any_ repository, which is a nice benefit to have. Just in case someone misconfigures the tool and decrypts a file into a local repo on their machine, then end up pushing and leaking from there.

🤔 This also makes me think we might want to raise if a user is trying to decrypt to a non generic path, namely `/Users/$USERNAME`.  

Thank you

mokagio: Hey @AliSoftware, I'm afraid the `rev-parse` option is a dead end for what we're trying to achieve here.

I implemented it `a7a3961`, but discovered an issue: `-C $PATH` expects the path to exist already. This is a problem for the use case in which we want to decrypt new secrets to a path that doesn't exist yet.

I added a dedicated example, `af5fe9e`, and extra logs in the `rev-parse` implementation, `a265e09` to showcase this.

[Here's the CI build failing](https://app.circleci.com/pipelines/github/wordpress-mobile/release-toolkit/765/workflows/a7bd7906-9c57-410b-a4f1-2bca7ec750ac/jobs/739/steps). When used with `-C` and a path that doesn't exist yet, Git errors with:

```
fatal: cannot change to '$PATH': No such file or directory
```

That's disappointing 😞 but I can't blame Git for working that way...

I reverted to the original error output diff implementation mainly to have CI show that it works, but I bumped into a different issue in CI which I'll need to look into next. 🙃

AliSoftware: I see. Unfortunate but makes sense.

Then other option I see is to `File.dirname` recursively until `File.exist?` is true then finally use that for rev-parse 🧐

spec/configure_helper_spec.rb
```diff
@@ -1,18 +1,41 @@
...
+    puts "tmp path is #{tmpdir_path}"
+    FileUtils.rm_rf(tmpdir_path)
+  end
+
+  attr_reader :tmpdir_path
+  attr_reader :pwd_before_spec_run
+end
+
 describe Fastlane::Helper::ConfigureHelper do
+  include_context 'with temp dir'
```
mokagio: What do you think of this? Worth following up with another PR that uses it in the other files that have a temp dir setup?

I'd extract it to it's own file, in that case

spec/configure_helper_spec.rb
```diff
@@ -1,18 +1,41 @@
...
     let(:destination) { 'path/to/destination' }
 
+    # rubocop:disable RSpec/ExampleLength
     it 'shows the user an error when the destination is not ignored in Git' do
       allow(Fastlane::Helper::GitHelper).to receive(:is_ignored?)
         .with(path: destination)
         .and_return(false)
 
+      allow(Fastlane::Helper::FilesystemHelper).to receive(:project_path)
+        .and_return(Pathname.new(tmpdir_path))
```
mokagio: Under the hood, if you follow the chain of calls, you'll eventually get to this method which traverse the caller's parent folders graph (list, tree? 🤔 ) backwards till it find one which contains a repo and identifies it as the `project_path`.

In the context of a temp folder that breaks apart, of course and the tests end up stuck.

mokagio: 🤔 I should probably make this a comment in the test, actually...

.rubocop_todo.yml
```diff
@@ -145,6 +145,7 @@ RSpec/FilePath:
     - 'spec/ios_merge_translators_strings_spec.rb'
     - 'spec/release_notes_helper_spec.rb'
     - 'spec/android_localize_helper_spec.rb'
+    - 'spec/configure_helper_spec.rb'
```
mokagio: This will be superseded by #254, but I didn't know about that PR at the time of writing this.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
mokagio: I'll give it a shot and report 🧐

spec/configure_helper_spec.rb
AliSoftware: tbh I'm not a fan of having the tmpdir created in `before` and removed in `after`, and the before.ater changing the pwd.

Maybe that's just me but personally I prefer a block-based approach, by simply using:
```ruby
if 'does something' do
  Dir.mktmpdir('a8c-release-toolkit-configure-') do |tmpdir|
    … the test code here, using tmpdir to construct paths with `File.join(tmpdir, …)` and use those absolute paths as necessary
  end
end
```

And if the test to be run doesn't take a path as parameter (so you can't DI a `File.join(tmpdir, …)` as parameter to tell the command where to operate), and instead operates in the pwd so you do need to `chdir` beforehand, then I'd do something like this:

```ruby
def in_tmp_dir
  Dir.mktmpdir('a8c-release-toolkit-tests-') do |tmpdir|
    Dir.chdir(tmpdir) do
      yield
    end
  end
end
```
And then you'd wrap all the examples which require to be run inside a tmpdir inside `in_tmp_dir do … end`

The block version of `mktmpdir` takes care of deleting the temp dir once the block exits, and it does this even if there has been a crash or an exception raised in the middle of the block. Likewise, the block version of `chdir` takes care of `cd`-ing back into the previous pwd when it exits the block, also doing so even if an exception was raised.

imho, using block-based syntax helps readability by being explicit that the context is scoped inside a tmpdir. And avoid you to have to deal with all the edge cases like ensuring cleanup even if exceptions have been raised, not having to store previous pwd, and avoiding global state. Sure it adds a level of indentation, but it adds scoping clarity 🤷

spec/configure_helper_spec.rb
AliSoftware: Why not run `git init` inside the tmpdir to make it a git repo? A step which you could add as part of the `in_tmp_dir` scoper/helper I suggested in my previous comment for example.

.rubocop_todo.yml
AliSoftware: 254 is now merged, in case you want to rebase 🙂

spec/configure_helper_spec.rb
mokagio: I like this approach a lot, actually.

Totally agree on the value of being explicit in the tests. While working in this area of the code I've been bitten a few times by setups that were done back at the start of the test and I didn't notice.

Will get to work on it...

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -5,12 +5,42 @@ module Helper
...
+          # one is part of a Git repository.
+          path = first_existing_ancestor_of(path: path)
+          args += ['-C', File.file?(path) ? File.dirname(path) : path]
+        end
+
+        args += ['rev-parse']
+
+        Action.sh(args, print_command_output: false) do |status, _, _|
+          status.success?
+        end
```
mokagio: Unfortunately, because of [how `sh` works internally](https://github.com/fastlane/fastlane/blob/7ba2d6a31e47c7919b3847fdbbde72feec68d213/fastlane/lib/fastlane/helper/sh_helper.rb#L67-L81) this setup will result in an error message in the Fastlane logs in case of a valid path that doesn't exist yet:

image

I'm wondering whether I should add a `UI.message` of sort to let the users know that's an expected error? Or maybe go through the longer process of adding an option for suppressing that output in Fastlane?

.rubocop_todo.yml
mokagio: I did rebase, but the diff is still here 🤔 

Regardless, other manual RuboCop fixes I did in this branch are now gone 💪

mokagio:
@jkmassel @AliSoftware this is finally ready to review. Thank you for your patience.

> You can see this in action in https://github.com/Automattic/simplenote-ios/pull/1276, which includes "testing" instructions.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
jkmassel: We use [this approach](https://github.com/wordpress-mobile/release-toolkit/blob/c476c50f507093ed30070740803083a4d4a42b49/lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb#L10-L30) to determining the root of a git repo elsewhere, I wonder if we could use something along those lines instead of relying on `git` to return specific error codes?

WDYT?

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -171,9 +202,13 @@ def self.ensure_on_branch!(branch_name)
       #
       # @param [String] path The path to check against `.gitignore`
       #
-      # @return [Bool] True if the given path is ignored, false otherwise.
+      # @return [Bool] True if the given path is ignored or outside the Git repository, false otherwise.
```
jkmassel:
```suggestion
      # @return [Bool] True if the given path is ignored or outside a Git repository, false otherwise.
```

spec/git_helper_spec.rb
```diff
@@ -130,14 +141,43 @@
         expect(Fastlane::Helper::GitHelper.is_ignored?(path: path)).to be true
       end
     end
+
+    # This test ensures we support the usecase of the `configure` tool, which
+    # can create new files by decrypting secrets. We need the ability to tell
+    # if a path result as ignored, regardless of whether it exists yet.
```
jkmassel:
```suggestion
    # if a path result is ignored, regardless of whether it exists yet.
```

spec/git_helper_spec.rb
```diff
@@ -130,14 +141,43 @@
...
+
+      setup_git_repo(dummy_file_path: path, add_file_to_gitignore: false)
+      expect(Fastlane::Helper::GitHelper.is_ignored?(path: path)).to be true
+    end
+
+    # This is sort of redundant given the previous example already ensures the
+    # same logic. But, we'll be using paths starting with `~` as part of our
+    # configurations, so it felt appopriate to explicitly ensure this important
+    # use case is respected.
+    it 'returns true when the path is in the home folder ' do
```
jkmassel, line 181: 👍

jkmassel (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
mokagio: Thanks @jkmassel 🙇 

One thing to point out first:

> instead of relying on `git` to return specific error codes?

This implementation relies on `git rev-parse` succeeding or failing, it does not inspect the error code. I think this is an important difference because it makes the logic simpler, it either succeeded or not, no need for extra info. Looking at the error code was one of the first things I tried (it was the error message, actually) but I wasn't happy with it either. Thanks to @AliSoftware's suggestions, I was able to get to this more flexible "did it succeed?" state.

Having said that, you make a legit point anyways. I'm of two minds on whether it's best to delegate to Git or implement the logic ourselves...

It's fair to say that we can trust Git to know how to do this job, and it would save us from reinventing the wheel. At the same time, shelling out is never nice and, like in this instance, we can run into awkward behavior regarding the output of the shelled-out command.

🤔 

I consider this an implementation detail, so I'm going to give it a go in a different branch. In the best case scenario, we like what I come up with and we merge it in here. If instead it may require extra work, we could still merge this working implementation and iterate on that approach async.

mokagio:
@AliSoftware @jkmassel with 1.0 approaching, it would be great to get this one merged 😄 I think I addressed all your comments. Thanks 🙇‍♂️ .

jkmassel (approved):

PR #253, opened 2021-05-10
Make GH Releases point to right commit

AliSoftware:
## Why?

When creating a GitHub release via the API with a tag value of `x.y`, the created release will either (1) point to the existing tag `x.y` if that tag already exists at the time the GH release is created, or (2) create the tag if it did not exist, **making that tag point to the default branch of the repo** (e.g. `develop`) unless otherwise specified.

Before we migrated to our new process of creating using the API to trigger CI builds, our process was to create the tag first, so we were in case 1. With the new process, we don't create the tag beforehand, so that makes us land in case 2.

Unfortunately we didn't specify a target commit for the tag to be created. This resulted in **the publication of our GitHub releases to create tags on `develop`** (instead of the commit we were on when CI created the release build). **This means that our tags were pointing to the incorrect commit**.

## How

Make the API call specify an explicit commit sha as the target of the GitHub release. Defaults to the SHA of the current commit (HEAD).

## To test

 - Go to an app repository, and create some dummy files for some fake release notes and assets
```sh
echo "This is a test release that should not be published and is only used to test a feature of the release toolkit." >rn.txt
echo "Fake asset 1" >asset1.txt
echo "Fake asset 2" >asset2.txt
```
 - Ensure you have a GitHub token set in your `GHHELPER_ACCESS` env var. (If not, `export GHHELPER_ACCESS=…` your token, you probably have one in one of your `~/.*-env.default` files)
 - Run the following command, replacing `<org>/<repo>` with the slug of your app repo:
```
bundle exec fastlane run create_release repository:<org>/<repo> version:test-do-not-publish release_notes_file_path:rn.txt release_assets:asset1.txt,asset2.txt prerelease:true
```
 - Check in the Releases tab of the repo on GitHub that the draft release was created. Edit the created release, and observe that the `@ Target:` dropdown would suggest that the tag will be pointing to `develop` when the release will be published. Discard the GitHub draft release.
 - Update your `fastlane/Pluginfile` to point the toolkit to this PR's branch & `bundle install`.
 - Re-run the fastlane command above
 - Check the Releases tab on GitHub again, and check that the new release draft now targets the commit SHA that your local repo was on when you ran the command (compare with `git rev-parse HEAD`).
 - Discard the GitHub release draft and the temporary changes you made to your app repo.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
@@ -61,7 +61,7 @@ module Fastlane
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :target,
                                        env_name: 'GHHELPER_TARGET_COMMITISH',
-                                       description: 'The branch name or commit sha1 the new tag should point to - if that tag does not exist yet when publishing the release. If omitted, will default to the current HEAD commit at the time of this call',
+                                       description: 'The branch name or commit SHA the new tag should point to - if that tag does not exist yet when publishing the release. If omitted, will default to the current HEAD commit at the time of this call',
                                        optional: true,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :release_notes_file_path,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
@@ -19,7 +19,14 @@ module Fastlane
           UI.user_error!("Can't find file #{file_path}!") unless File.exist?(file_path)
         end
 
-        Fastlane::Helper::GithubHelper.create_release(repository, version, release_notes, assets, prerelease)
+        Fastlane::Helper::GithubHelper.create_release(
+          repository: repository,
+          version: version,
+          target: params[:target],
+          body: release_notes,
+          assets: assets,
+          prerelease: prerelease
+        )
         UI.message('Done')
       end
 
@@ -44,19 +51,24 @@ module Fastlane
         [
           FastlaneCore::ConfigItem.new(key: :repository,
                                        env_name: 'GHHELPER_REPOSITORY',
-                                       description: 'The remote path of the GH repository on which we work',
+                                       description: 'The slug (`<org>/<repo>`) of the GitHub repository we want to create the release on',
                                        optional: false,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :version,
                                        env_name: 'GHHELPER_CREATE_RELEASE_VERSION',
                                        description: 'The version of the release',
                                        optional: false,
-                                       is_string: true),
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :target,
+                                       env_name: 'GHHELPER_TARGET_COMMITISH',
+                                       description: 'The branch name or commit sha1 the new tag should point to - if that tag does not exist yet when publishing the release. If omitted, will default to the current HEAD commit at the time of this call',
+                                       optional: true,
+                                       type: String),
           FastlaneCore::ConfigItem.new(key: :release_notes_file_path,
                                        env_name: 'GHHELPER_CREATE_RELEASE_NOTES',
                                        description: 'The path to the file that contains the release notes',
                                        optional: true,
-                                       is_string: true),
+                                       type: String),
           FastlaneCore::ConfigItem.new(key: :release_assets,
                                        env_name: 'GHHELPER_CREATE_RELEASE_ASSETS',
                                        description: 'Assets to upload',
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
@@ -23,7 +23,7 @@ module Fastlane
           repository: repository,
           version: version,
           target: params[:target],
-          body: release_notes,
+          description: release_notes,
           assets: assets,
           prerelease: prerelease
         )
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -95,6 +95,15 @@ module Fastlane
         end
       end
 
+      # Get the sha1 of a given git ref. Typically useful to get the sha1 of the current HEAD commit.
+      #
+      # @param [String] ref The git ref (commit, branch name, 'HEAD', …) to resolve as a sha1
+      # @return [String] The commit sha1 of the ref
+      #
+      def self.get_commit_sha(ref: 'HEAD')
+        Action.sh('git', 'rev-parse', ref)
+      end
+
       # Creates a tag for the given version, and optionally push it to the remote.
       #
       # @param [String] version The name of the tag to push, e.g. "1.2"
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -95,10 +95,10 @@ module Fastlane
         end
       end
 
-      # Get the sha1 of a given git ref. Typically useful to get the sha1 of the current HEAD commit.
+      # Get the SHA of a given git ref. Typically useful to get the SHA of the current HEAD commit.
       #
-      # @param [String] ref The git ref (commit, branch name, 'HEAD', …) to resolve as a sha1
-      # @return [String] The commit sha1 of the ref
+      # @param [String] ref The git ref (commit, branch name, 'HEAD', …) to resolve as a SHA
+      # @return [String] The commit SHA of the ref
       #
       def self.get_commit_sha(ref: 'HEAD')
         Action.sh('git', 'rev-parse', ref).chomp
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -101,7 +101,7 @@ module Fastlane
       # @return [String] The commit sha1 of the ref
       #
       def self.get_commit_sha(ref: 'HEAD')
-        Action.sh('git', 'rev-parse', ref)
+        Action.sh('git', 'rev-parse', ref).chomp
       end
 
       # Creates a tag for the given version, and optionally push it to the remote.
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -101,7 +101,7 @@ module Fastlane
       # @return [String] The commit SHA of the ref
       #
       def self.get_commit_sha(ref: 'HEAD')
-        Action.sh('git', 'rev-parse', ref).chomp
+        Git.open(Dir.pwd).revparse(ref)
       end
 
       # Creates a tag for the given version, and optionally push it to the remote.
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -79,8 +79,27 @@ module Fastlane
         github_client().create_milestone(repository, newmilestone_number, options)
       end
 
-      def self.create_release(repository, version, release_notes, assets, prerelease)
-        release = github_client().create_release(repository, version, name: version, draft: true, prerelease: prerelease, body: release_notes)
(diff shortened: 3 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -95,14 +114,14 @@ def self.create_release(repository, version, release_notes, assets, prerelease)
...
-        tag = tag.concat('/')
+        repository = repository.delete_prefix('/').chomp('/')
+        file_path = file_path.delete_prefix('/').chomp('/')
         file_name = File.basename(file_path)
         download_path = File.join(download_folder, file_name)
 
         begin
-          open(URI.join('https://raw.githubusercontent.com/', repository, tag, file_path).to_s.chomp('/')) do |remote_file|
+          uri = URI.parse("https://raw.githubusercontent.com/#{repository}/#{tag}/#{file_path}")
+          uri.open do |remote_file|
```
AliSoftware: These changes are unrelated to this PR but were required to fix a rubocop hard violation on ` Security/Open: The use of Kernel#open is a serious security risk.` – see [more info here](https://docs.rubocop.org/rubocop/cops_security.html#securityopen) — that would otherwise prevent the CI to go green.

This is also separately addressed via https://github.com/wordpress-mobile/release-toolkit/pull/254 but I needed to at least address this `Kernel#open` violation in this PR to avoid danger to prevent us merging this PR. If https://github.com/wordpress-mobile/release-toolkit/pull/254 gets merged first, making that fix land in `develop`, this part of the diff should disappear from this PR.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
```diff
@@ -44,19 +51,24 @@ def self.available_options
...
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :version,
                                        env_name: 'GHHELPER_CREATE_RELEASE_VERSION',
                                        description: 'The version of the release',
                                        optional: false,
-                                       is_string: true),
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :target,
+                                       env_name: 'GHHELPER_TARGET_COMMITISH',
+                                       description: 'The branch name or commit sha1 the new tag should point to - if that tag does not exist yet when publishing the release. If omitted, will default to the current HEAD commit at the time of this call',
```
mokagio: Nitpick. "sha1" is an acronym for [Secure Hash Algorithm 1](https://en.wikipedia.org/wiki/SHA-1) (TIL the actual of those letters), as such, it should be all capital letters<sup>*</sup>.

```suggestion
                                       description: 'The branch name or commit SHA-1 the new tag should point to - if that tag does not exist yet when publishing the release. If omitted, will default to the current HEAD commit at the time of this call',
```

<sup>*</sup> – I'm not sure whether there is an actual rule or convention for this, or if it's a personal preference I build reading code and text over time. 😳 The only thing I can say to back this persnickety suggestion of mine is that both the [Wikipedia entry](https://en.wikipedia.org/wiki/SHA-1) and [the Git documentation](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) use "SHA-1". The only exception is a Ruby variable assignment (`sha1 = ...`) which makes sense in the context. That example actually refers to [the `Digest::SHA1` class](https://ruby-doc.org/stdlib-2.7.0/libdoc/digest/rdoc/Digest/SHA1.html) which uses the uppercase style, too.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -79,8 +79,27 @@ def self.create_milestone(repository, newmilestone_number, newmilestone_duedate,
...
+      #
+      def self.create_release(repository:, version:, target: nil, body:, assets:, prerelease:)
+        release = github_client().create_release(
+          repository,
+          version, # tag name
+          name: version, # release name
+          target_commitish: target || Fastlane::Helper::GitHelper.get_commit_sha,
+          draft: true,
+          prerelease: prerelease,
+          body: body
```
mokagio: I don't like the name "body" and was going to suggest something like "description" or "message". Then I looked up [the GitHub APIs](https://docs.github.com/en/rest/reference/repos#create-a-release) and they use "body" as well, so... 🤐 🤷‍♂️ 😅

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
```diff
@@ -44,19 +51,24 @@ def self.available_options
...
+                                       optional: true,
+                                       type: String),
           FastlaneCore::ConfigItem.new(key: :release_notes_file_path,
                                        env_name: 'GHHELPER_CREATE_RELEASE_NOTES',
                                        description: 'The path to the file that contains the release notes',
                                        optional: true,
-                                       is_string: true),
+                                       type: String),
```
mokagio, line 71: Nice cleanup along the way

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -79,8 +79,27 @@ def self.create_milestone(repository, newmilestone_number, newmilestone_duedate,
...
+      # @param [String] body The text to use as the release's body / description (typically the release notes)
+      # @param [Array<String>] assets List of file paths to attach as assets to the release
+      # @param [TrueClass|FalseClass] prerelease Indicates if this should be created as a pre-release (i.e. for alpha/beta)
+      #
+      def self.create_release(repository:, version:, target: nil, body:, assets:, prerelease:)
+        release = github_client().create_release(
+          repository,
+          version, # tag name
+          name: version, # release name
+          target_commitish: target || Fastlane::Helper::GitHelper.get_commit_sha,
```
mokagio: TIL "commitish":

image

mokagio: Is there a use case for us to support the commitish format? That is, do we need support for creating a release on the tip of a branch? As far as I can see, I don't think we do.

Would removing the support for the branch simplify things? If not the coding itself, it may make the naming clearer, from `target` to `sha1` for example.

mokagio (approved):
Left some no-blocking questions and suggestions.

Before:

image

After:

image

The target matches the Git `HEAD` when running the command: 

Screen Shot 2021-05-13 at 12 11 05 pm

👌 and thank you for the handy snippet, it made testing faster 🎩 🙇

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: Yeah that's the reason I used body, to match GH API naming. But description works too so I might use that instead

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: Yeah I used that name because both options were supported out of the box anyway. I mean that parameter is used verbatim without any transformation and directly passed to the call to the GitHub API, and that API support both. So I didn't see a reason why we'd forcefully limit that parameter to only be a SHA1. Besides, how would you do that sort of limitation anyway and why? And how would you name that param? `sha1`? But you said that was supposed to be uppercased… so `SHA1`? Feels then strange for a param name to be uppercased…

One thing we could consider doing is to not have the action param / ConfigItem at all, and always use the current SHA without any way to override it. After all that's currently the only use case we have so far. But I thought it was nice and cheap to expose to option to override for the future (I know I know YAGNI 😅)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
```diff
@@ -44,19 +51,24 @@ def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :repository,
                                        env_name: 'GHHELPER_REPOSITORY',
-                                       description: 'The remote path of the GH repository on which we work',
+                                       description: 'The slug (`<org>/<repo>`) of the GitHub repository we want to create the release on',
```
jkmassel, line 54: 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -95,6 +95,15 @@ def self.commit(message:, files: nil, push: false)
         end
       end
 
+      # Get the sha1 of a given git ref. Typically useful to get the sha1 of the current HEAD commit.
+      #
+      # @param [String] ref The git ref (commit, branch name, 'HEAD', …) to resolve as a sha1
+      # @return [String] The commit sha1 of the ref
+      #
+      def self.get_commit_sha(ref: 'HEAD')
```
jkmassel, line 103: This may be overly nitpicky, but [the release toolkit already depends on the `git` gem](https://github.com/wordpress-mobile/release-toolkit/blob/c476c50f507093ed30070740803083a4d4a42b49/fastlane-plugin-wpmreleasetoolkit.gemspec#L37) – could we use that instead of adding our own helper?

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -79,8 +79,27 @@ def self.create_milestone(repository, newmilestone_number, newmilestone_duedate,
...
+      #
+      # @param [String] repository The repository to create the GitHub release on. Typically a repo slug (<org>/<repo>).
+      # @param [String] version The version for which to create this release. Will be used both as the name of the tag and the name of the release.
+      # @param [String?] target The commit sha1 or branch name that this release will point to when it's published and creates the tag.
+      #        If nil (the default), will use the repo's current HEAD commit at the time this method is called.
+      #        Unused if the tag already exists.
+      # @param [String] body The text to use as the release's body / description (typically the release notes)
+      # @param [Array<String>] assets List of file paths to attach as assets to the release
+      # @param [TrueClass|FalseClass] prerelease Indicates if this should be created as a pre-release (i.e. for alpha/beta)
+      #
```
jkmassel, line 92: 😍 these docs tho

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
```diff
@@ -19,7 +19,14 @@ def self.run(params)
           UI.user_error!("Can't find file #{file_path}!") unless File.exist?(file_path)
         end
 
-        Fastlane::Helper::GithubHelper.create_release(repository, version, release_notes, assets, prerelease)
+        Fastlane::Helper::GithubHelper.create_release(
+          repository: repository,
+          version: version,
+          target: params[:target],
```
jkmassel, line 25: This makes this PR a breaking change, right? Might be a good PR to introduce the `RELEASE_NOTES.txt` file 🙂

jkmassel (approved):
Left a few minor notes, but generally this LGTM.

My big concern is that I think this is breaking, so we should make a release note for it

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
AliSoftware: How so? This is a new param that ends up being optional. So the API is still backwards compatible with any call site not providing `target:`. In fact, we won't have to update the call sites in the app repos at all to benefit from this bugfix, since the default value for this param (`nil` to mean "current commit") is what we'll be using so far.

If anything, this is a change in (1) behavior (before it made those point to `develop`, now to current commit) but that's a bugfix, so in semver this warrants a patch bump (2) API-additive change, not an API-breaking change, since old call sites are still compatible, the new param being optional with default value. So in semver this warrants a minor, non-breaking bump, not a major one anyway 🙂

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: Yeah good point will look into it.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
AliSoftware: Fixed in caf7fbd7749c607f56a61ff6ad0757617f41d203 – also just used the name `SHA` because who knows maybe one day they'll migrate from SHA1 to SHA256…

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: Updated via 539e1626babf33ab4101245c40dc2e5b260378e9

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: Fixed in 2809f2f5d9492c1bb5332d05b533aaba04cbba25

Issue #248, opened 2021-05-03
What's a good value for RSpec example length?


mokagio:
When working on https://github.com/wordpress-mobile/release-toolkit/pull/242, I added some tests and RuboCop warned me they should have been 5 lines or less.

```
RSpec/ExampleLength: Example has too many lines [7/5].
```

I addressed the violation with [a cheap trick](https://github.com/wordpress-mobile/release-toolkit/pull/242/commits/b63a2a30ee5de849c3146f29fd9692b0d28913c5).

I'm all in for having short methods actually, I do really like the [Sandi Metz rules](https://thoughtbot.com/blog/sandi-metz-rules-for-developers). But, in the context of that commit, I feel like I didn't make the test more readable.

I can see two options:

- Relax the rule to acknowledge the current state of this codebase, where there's a lot of setup and mocking required for testing
- Stick with the rule, and do our best to work within it in a way that makes the codebase better.

mokagio:
>  I feel like I didn't make the test more readable.

As an aside, something else I could have done was extracting that setup logic in a method so that it would have been a one liner all in the example and I would have satisfied the RuboCop requirement 🤔

PR #254, opened 2021-05-10
Fix rubocop offenses

AliSoftware:
While working on https://github.com/wordpress-mobile/release-toolkit/pull/253 I found out that our codebase had new rubocop offenses – that were introduced by some recent changes made before we made CI run rubocop on PRs via #246 .

This PR fixes the recently introduced offenses

### To test

```
bundle exec rake rubocop
bundle exec rake test
```


.rubocop_todo.yml
```diff
@@ -121,26 +121,20 @@ RSpec/DescribedClass:
...
     - 'spec/android_merge_translators_strings_spec.rb'
+    - 'spec/android_version_helper_spec.rb'
     - 'spec/ci_helper_spec.rb'
     - 'spec/configuration_spec.rb'
+    - 'spec/configure_helper_spec.rb'
     - 'spec/encryption_helper_spec.rb'
     - 'spec/file_reference_spec.rb'
     - 'spec/git_helper_spec.rb'
+    - 'spec/github_helper_spec.rb'
+    - 'spec/ios_git_helper_spec.rb'
```
AliSoftware, line 137: This rule is about having the spec files in the right directory hierarchy that matches the classes they test and their module hierarchy. It would indeed be nice to re-organize our file hierarchy in the `spec/` folder to follow the conventions here, but that's a larger refactoring, so let's keep that for later.

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
```diff
@@ -54,44 +54,42 @@ def self.prepare_repository
       ### Check with the user whether we should overwrite the file, if it exists
       ###
       def self.apply_file(file_reference, force)
-        begin
```
AliSoftware, line 57: The rest of that diff is indentation-related only; so despite the diff being large, the only change is the removal of the `begin` and `end` keywords which wrapped the whole function body ([method bodies are already implicit exception blocks](https://docs.rubocop.org/rubocop/1.0/cops_style.html#styleredundantbegin)).

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -95,14 +95,14 @@ def self.create_release(repository, version, release_notes, assets, prerelease)
...
-        tag = tag.concat('/')
+        repository = repository.delete_prefix('/').chomp('/')
+        file_path = file_path.delete_prefix('/').chomp('/')
         file_name = File.basename(file_path)
         download_path = File.join(download_folder, file_name)
 
         begin
-          open(URI.join('https://raw.githubusercontent.com/', repository, tag, file_path).to_s.chomp('/')) do |remote_file|
+          uri = URI.parse("https://raw.githubusercontent.com/#{repository}/#{tag}/#{file_path}")
+          uri.open do |remote_file|
```
AliSoftware, line 105: ` Security/Open: The use of Kernel#open is a serious security risk.` – see [here](https://docs.rubocop.org/rubocop/cops_security.html#securityopen).

.rubocop.yml
```diff
@@ -44,6 +44,10 @@ Style/FrozenStringLiteralComment:
 Style/AsciiComments:
   Enabled: false
 
+# We are not strict to the point that we want to enforce super-short `it '…' do … end` blocks in our specs.
+RSpec/ExampleLength:
+  Enabled: false
```
mokagio, line 49: At the same time, it's useful to have some kind of warning when you end up with an example that is _too_ long.

Now, how long is "_too_ long"? Well... that's a good question 😅

Maybe it's fair to say that we have bigger fishes to fry right now and that if we'll get to the point where "_too_ long" specs really pop out as an issue, then it means we'll have achieve a lot of our codebase code quality improvement goals 😄 👍

mokagio: See also https://github.com/wordpress-mobile/release-toolkit/issues/248.

As per my previous comment, I'm happy to got with this change and close the issue for the time being.

If you're reading this and that sounds good to you, feel free to close the issue.

.rubocop_todo.yml
mokagio: 👍 

Took the liberty of opening https://github.com/wordpress-mobile/release-toolkit/issues/255 to track this.

lib/fastlane/plugin/wpmreleasetoolkit/actions/configure/configure_apply_action.rb
mokagio: This is one of the reasons I love tools like RuboCop and SwiftLint: they actually teach you how to write _better_ code.

mokagio (approved):
image

Thank you 🙇

.rubocop_todo.yml
AliSoftware: Thanks!

Though there are _a lot_ of other violations to fix as a whole on this ruby project first before we get to this 😛  (basically the more general issue would be "fix _all_ the TODO violations that have been temporarily disabled in `.rubocop_todo.yml`" 😉 )

.rubocop.yml
AliSoftware: Yeah I almost just raised the max value instead of entirely disabling it.

I could also have set this rule in `.rubocop_todo.yml` instead of `.rubocop.yml` to signal the real intent, which really was "yeah that's a good rule, but we have other fishes to fry first so we don't want to burn too much time on this one in priority while there are other more important violations to fix first" 😄 

My plan is to fix all the violations incrementally, starting with what we deem the most "important" rules/conventions to respect first (the ones that really make the code consistent, avoid security issues, and make the code easier to navigate by having e.g. file names matching the name of the class they implement, etc), and slowly and incrementally go through the exceptions in `.rubocop_todo.yml` one by one until the TODO file is empty and we got rid of all those "we'll deal with that later" exceptions. Once we get to that point, and have a real clean codebase, then it can be time to revisit the rules we decided to enable/disable back then and get the finishing touches to iron things out 😉

mokagio: > My plan is to fix all the violations incrementally, starting with what we deem the most "important" rules/conventions to respect first

>  Once we get to that point, and have a real clean codebase, then it can be time to revisit the rules we decided to enable/disable back then and get the finishing touches to iron things out

👍

PR #256, opened 2021-05-11
Add ios_get_build_version action

AliSoftware:
## Why

When creating a GitHub release (after a release build is triggered via `new_beta_release` or `finalize_release` in the host apps), iOS currently always uses `ios_get_app_version()`  in the host Fastfiles as the version to use for the GitHub release title and tag name.

This means that even GH releases created as a result of `new_beta_release` end up with a title and git tag of `x.y` instead of using `x.y.0.z` for betas.

To solve this, the host app's `Fastfiles` will need to be able to fetch the build version (`VERSION_LONG` in `xcconfig` files) in order to pass it at call site to `create_gh_release` and for it to use the correct title and tag for betas. This is what this PR adds.

## Target branch

This PR sits on top of `gh-releases/fix-target` from https://github.com/wordpress-mobile/release-toolkit/pull/253 because I needed both fixes at once to continue working on the fixes needed in the host apps' `Fastfile` and the overall issues with GH releases in our process. https://github.com/wordpress-mobile/release-toolkit/pull/253 Should thus ideally be merged first – which should update this PR's branch to `develop` before we can merge that one next.

## To test

 - Edit the `fastlane/Pluginfile` of an host app to point the release toolkit to this branch, and run `bundle install`
 - Run `bundle exec fastlane run ios_get_build_version`. Notice it fails telling you an env var is missing
 - Run `PUBLIC_CONFIG_FILE=config/Version.public.xcconfig bundle exec fastlane run ios_get_build_version` and check that it returns the public build version (e.g. `17.3.0.1`)
 - Run `PUBLIC_CONFIG_FILE=config/Version.public.xcconfig bundle exec fastlane run ios_get_build_version internal:true`, Notice it fails telling you an env var is missing.
 - Run `INTERNAL_CONFIG_FILE=config/Version.internal.xcconfig PUBLIC_CONFIG_FILE=config/Version.public.xcconfig bundle exec fastlane run ios_get_build_version internal:true` and check that it returns the internal build version (e.g. `17.3.0.20210507`)

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_build_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_build_version.rb
@@ -1,0 +1,57 @@
+module Fastlane
+  module Actions
+    class IosGetBuildVersionAction < Action
+      def self.run(params)
+        require_relative '../../helper/ios/ios_version_helper.rb'
+
+        if params[:internal]
+          Fastlane::Helper::Ios::VersionHelper.get_internal_version
+        else
+          Fastlane::Helper::Ios::VersionHelper.get_build_version
+        end
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        'Gets the build version of the app'
+      end
+
+      def self.details
+        'Gets the build version (`VERSION_LONG`) of the app from the xcconfig file'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :internal,
+            env_name: 'FL_IOS_BUILD_VERSION_INTERNAL',
+            description: 'If true, returns the internal build version, otherwise returns the public one',
+            is_string: false, # Boolean
+            default_value: false
+          ),
+        ]
+      end
+
+      def self.output
+        # Define the shared values you are going to provide
+      end
+
+      def self.return_value
+        # If you method provides a return value, you can describe here what it does
+        'Return the public or internal build version of the app'
+      end
+
+      def self.authors
+        # So no one will ever forget your contribution to fastlane :) You are awesome btw!
+        ['AliSoftware']
+      end
+
+      def self.is_supported?(platform)
+        platform == :ios
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_build_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_build_version.rb
@@ -4,7 +4,10 @@ module Fastlane
       def self.run(params)
         require_relative '../../helper/ios/ios_version_helper.rb'
 
+        UI.user_error!('You need to set at least the PUBLIC_CONFIG_FILE env var to the path to the public xcconfig file') unless ENV['PUBLIC_CONFIG_FILE']
+
         if params[:internal]
+          UI.user_error!('You need to set the INTERNAL_CONFIG_FILE env var to the path to the internal xcconfig file') unless ENV['INTERNAL_CONFIG_FILE']
           Fastlane::Helper::Ios::VersionHelper.get_internal_version
         else
           Fastlane::Helper::Ios::VersionHelper.get_build_version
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_build_version.rb
```diff
@@ -0,0 +1,60 @@
+module Fastlane
+  module Actions
+    class IosGetBuildVersionAction < Action
+      def self.run(params)
+        require_relative '../../helper/ios/ios_version_helper.rb'
+
+        UI.user_error!('You need to set at least the PUBLIC_CONFIG_FILE env var to the path to the public xcconfig file') unless ENV['PUBLIC_CONFIG_FILE']
```
AliSoftware, line 7: Due to an implementation details of the `ios_version_helper` – basically both `get_internal_version` and `get_build_version` rely on the private `get_version_strings` method which require `PUBLIC_CONFIG_FILE` to be set and returns the pair of public+internal, even if we don't end up using the public value from the returned tuple – the `PUBLIC_CONFIG_FILE` env var has to be set _even_ if we only want the internal version.

I considered this was not worth refactoring the whole version helper just to fix this, especially since `PUBLIC_CONFIG_FILE` is _always_ set in our use case and the context of all our host apps' `Fastfile` anyway. We might get that cleaned up and that useless requirement lifted when we get to implement #203 later.

jkmassel: This _may_ not be compatible with Jetpack, where it needs to share a repo (and fastfile) with WPiOS.

Could we accept it as a parameter, but fall back to `ENV['PUBLIC_CONFIG_FILE']` if it's not present?

jkmassel (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_build_version.rb
AliSoftware: I'll get on that. Though note that _many_ actions rely on such an env var, not just `get_*_version` but also the various `bump_*_version` ones, and same story for Android. So I wonder if that should be part of this PR vs having a dedicated PR to support Jetpack for _all_ the iOS and Android actions related to version wrangling?

I'll make the change for `ios_get_build_version` and take the occasion to replicate it on `ios_get_app_version` but I'll let any other changes required in other actions and for all the platforms for a separate PR dedicated to Jetpack.

AliSoftware: Ok, after trying to play around this, even trying to refactor just around `ios_get_build_version` and nothing else quickly leads me down the rabbit hole, having to change `get_versions()` which both methods rely on, but also having to update all the other methods relying on that private method and using those `ENV` vars in the helper, including the ones that deal with bumping the version… And all that ends up being way more than just a couple of lines to change after all, and would require a dedicated pass of testing all the changes.

After chatting in Slack, the consensus is to keep that for a future and separate PR that would address all things related to Jetpack versioning, which means the whole family of `{ios,android}_get_{app,build,alpha,release}_version` + `{ios,android}_{codefreeze,betabuild,hotfix,finalize}_prechecks` + `{ios,android}_bump_version_{beta,final,hotfix,release}` actions and all the helper methods they rely on.

Maybe when we get to address that – in a separate PR – that would be a good occasion to jump back on implementing #203 and continue the work I started in the `203-version` branch, which would help a lot in our goal to support multiple apps in a single repo (like WP+Jetpack) by stopping relying on ENV vars, global constants and methods, and same parameters having to be passed over and again to each class method calls, replacing all that with the passing of those parameters only to the constructor, then deal with the instance transparently from there.

---

TL;DR I won't implement that change in this PR after all, and we should have a separate and dedicated PR to add Jetpack support to all version methods, which extends way further than just amending this new action.

AliSoftware: @jkmassel given ☝️ , can you dismiss your "requested-changes" and give me a 👍  on the PR? 🥺

jkmassel (approved):

PR #257, opened 2021-05-12
Fix the drawText binary

AliSoftware:
Trying to fix the `Thread 1: EXC_BAD_ACCESS (code=1, address=0x11094c0e4)` error on `drawGlyphs`

image

Seems that passing around a `CGContext` messes things with reference counting (despite Zombies and Address Sanitizer not telling us anything). Passing `NSGraphicsContext` instead seems to fix the crash.

## To Test

- Point WPiOS to this branch of the toolkit, run bundle install
- Ensure you have the promo strings in `fastlane/appstoreres/metadata/*`. If not, run `bundle exec fastlane download_promo_strings` first.
- Run the following command:
```
 bundle exec drawText html="fastlane/appstoreres/metadata/en-US/app_store_screenshot_1.txt" maxWidth=1242 maxHeight=420 output=drawText.out.png fontSize=110 stylesheet="./fastlane/appstoreres/assets/style.css" alignment="center"
```
- Check it didn't crash, and that it generated the `drawText.out.png` file exists, and that it indeed has a size of 1242x480 (at 72ppi) when you open it in Preview

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
 
       - run:
           name: Install Homebrew dependencies, if neeeded
-          command: xargs brew install --verbose < .circleci/.brewfile
+          command: brew update && xargs brew install --verbose < .circleci/.brewfile
 
       - run:
           name: Install Ruby dependencies, if neeeded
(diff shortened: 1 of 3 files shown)


jkmassel (approved):
Tested this with WordPress iOS screenshots and it worked great 🎉 

:shipit:

Issue #205, opened 2021-02-03
Remove the need for PROJECT_ROOT_DIR env var


AliSoftware:
We should stop relying on this env var and requiring it to be set as we could instead be able to detect that project root automatically in the actions that needs it.

At first I thought we could make use of something like `git rev-parse --show-toplevel` to detect that project root path automatically. But then I realized that I think that won't even be necessary, because [as documented in the _fastlane_ docs](https://docs.fastlane.tools/advanced/fastlane/#directory-behavior), the working directory is already guaranteed to be the project root itself for code run from actions defined in plugin like the `release-toolkit`. So we can probably just replace it with `.` and use relative paths, as the `Dir.pwd` should already be the project root folder in our context!

jkmassel:
The environment variable is `PROJECT_ROOT_FOLDER`, for any future searchers who get very confused why they can't grep for `PROJECT_ROOT_DIR` 😜

PR #258, opened 2021-05-12
Jetpack App: Update release toolkit to support Jetpack

JavonDavis:
This PR introduces changes that were primarily needed to support the new Jetpack app in the WPAndroid repo but are more focused on updating how versions are managed in our current and future Android apps.  The proposals here are: 

* The use of a new app parameter/environment variable to add an additional context when doing versioning. This is a new parameter that can be passed as an option to the commands related to versioning, this value can can also be set as an environment variable for projects that don't have multiple 'apps' within them. 
* It also proposes the use of a new properties file in the project repos to manage versioning, a version.properties file that follows the following key-value format

```
appName.versionName=<version name>
appName.versionCode=<version code>

appName.alpha.versionName=<version name>
appName.alpha.versionCode=<version code>
```

Example here: https://github.com/wordpress-mobile/WordPress-Android/blob/fe4352dbdf2d8b56f760ceeb6a0bd591883b23e4/version.properties

With accompanying changes in the `build.gradle` file to read the versions from this file instead of being defined in the `build.gradle`. Something like so -> 

```
def versionProperties = new Properties()
    file("../version.properties").withInputStream { versionProperties.load(it) }

    compileSdkVersion rootProject.compileSdkVersion

    defaultConfig {
	@@ -60,9 +63,10 @@ android {
        if (project.hasProperty("versionName")) {
            versionName project.property("versionName")
        } else {
            versionName versionProperties.getProperty("wordpress.alpha.versionName")
        }

        versionCode versionProperties.getProperty("wordpress.alpha.versionCode").toInteger()
```

* This PR also proposes moving the logic for updating the version into a gradle task which can be a bit more stable than our previous approach, see proposed gradle task currently used in https://github.com/wordpress-mobile/WordPress-Android/pull/14613... The PR assumes this task exists and relies on it to update the `version.properties` file 

```
task "updateVersionProperties" {
    doLast {
        ant.propertyfile(
                file: "../version.properties") {
            entry( key: "$key", value: "$value")
        }
    }
}
```


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -11,11 +11,13 @@ def self.run(params)
         # Checkout develop and update
         Fastlane::Helper::GitHelper.checkout_and_pull('develop')
 
+        @app = ENV['APP'].nil? ? params[:app] : ENV['APP']
```
AliSoftware: `self.run` is a static method not an instance method, so using `@app` aka an instance variable in that context seems incorrect to me 🤔  (wonder if rubocop will warn about this)

[EDIT] I see that other similar actions already uses instance variables and `@` in their `create_config` and `show_config` helpers – which is also bad practice. I'm guessing that the existing code is probably what led you to use the same constructs here; but might be worth fixing both at once and take the occasion of this PR to clean up those (just defining local variables from `create_config` directly inside `self.run` + get rid of `create_config` and `show_config` should do it)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -80,6 +82,11 @@ def self.available_options
                                        description: 'Skips confirmation',
                                        is_string: false, # true: verifies the input is a string, false: every kind of value
                                        default_value: false), # the default value if the user didn't provide one
+          FastlaneCore::ConfigItem.new(key: :app,
+                                       env_name: 'APP',
```
AliSoftware: That's quite a short name for an env var, given that env vars are all sharing a global space. The convention is usually to use some prefix like `FASTLANE_APP_NAME` or similar

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -80,6 +82,11 @@ def self.available_options
...
+                                       description: 'The app to get the release version for',
```
AliSoftware:
```suggestion
                                       description: 'The name of the app to get the release version for',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
AliSoftware: Also, worth noting that introducing this new key will be a breaking change in the toolkit.
It will default to `wordpress` so this means it won't break WordPress, but it will break woocommerce-android.

Don't we already have an ENV var in our current fastfiles (`PROJECT_NAME` or something like that iirc?) containing this value? If so, using this existing env var name as `env_name` here would make this portable and non-breaking.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_build_prechecks.rb
```diff
@@ -62,6 +64,11 @@ def self.available_options
                                        description: 'True if this is for a final build',
                                        is_string: false,
                                        default_value: false),
+          FastlaneCore::ConfigItem.new(key: :app,
+                                       env_name: 'APP',
+                                       description: 'The app to get the release version for',
+                                       is_string: true, # true: verifies the input is a string, false: every kind of value
+                                       default_value: 'wordpress'), # the default value if the user didn't provide one
```
AliSoftware: Same feedbacks as for `betabuild_prechecks`

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
```diff
@@ -14,7 +14,7 @@ module GitHelper
         def self.commit_version_bump
           Fastlane::Helper::GitHelper.commit(
             message: 'Bump version number',
-            files: File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'build.gradle'),
+            files: File.join(ENV['PROJECT_ROOT_FOLDER'], 'version.properties'),
```
AliSoftware: This suggests that all host/app projects migrate to this new `version.properties` convention.

I'm all for this convention and separate `version.properties` file, seems cleaner 👍 but that also means that this will be a breaking change that will require SNAndroid, WCAndroid and WPAndroid to be updated accordingly before they can adopt the new toolkit.

I suggest that we:
 - Either make this change so that it's compatible with both versions, to ease the transition from the old way to the new way. Maybe with a flag to pass to the actions to do one or the other. Drawback of this is that it will complexify the code and likely be the kind of "patch" that will stay for a while before we clean things up, so maybe not the ideal path
 - Or create new actions, which use the new way, and deprecate the old actions
 - And/Or wait for us to do our first stable 1.0 – to which all repos can point to, then immediately prepare for a 2.0 – with those breaking changes in – with semver telling us that this will be a breaking change and will require all host app / clients to update their fastfiles in parallel of updating the release-toolkit
 - Lastly we could do one last `0.x.y` version before that one, and make 1.0 be that the version that holds the breaking change. There are still a lot couple of non-breaking changes pending PR approval that I'd want us to release first before introducing new breaking changes and a new major version bump tho, so that apps that need the pending fixes (GH releases fixes mainly) but are not yet ready to migrate to this new `version.properties` on Android can still have a release to point to.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -30,42 +30,54 @@ module VersionHelper
...
         #
         # @return [Hash] A hash with 2 keys "name" and "code" containing the extracted version name and code, respectively
         #
-        def self.get_release_version
-          section = ENV['HAS_ALPHA_VERSION'].nil? ? 'defaultConfig' : 'vanilla {'
-          gradle_path = self.gradle_path
-          name = get_version_name_from_gradle_file(gradle_path, section)
-          code = get_version_build_from_gradle_file(gradle_path, section)
-          return { VERSION_NAME => name, VERSION_CODE => code }
+        def self.get_version_from_properties(product_name, is_alpha)
```
AliSoftware: nit: When reading the call sites, it was really unclear to me what the 2nd parameter was supposed to be, until I looked at the implementation line here. What do you think about using named parameters for that method instead, so that the call site end up being more readable? Would also allow you to provide a default value for `is_alpha` as a bonus.

```suggestion
        def self.get_version_from_properties(product_name:, is_alpha: false)
```

AliSoftware (commented):
I like the suggestion to introduce a `version.properties` 👍 
But that also means that this PR is quite a breaking change, as it will require the apps to update:
 - Their `Fastfile` to update all the calls to the various actions you modified, to provide the app name now (especially WooCommerce Android as you otherwise made the app name default to `wordpress`).
 - Their `build.gradle` file to load the new `version.properties` and make the versionName/versionCode references point to it.
 - Create their `version.properties`.

That is quite the breaking change, so this should not be merged before we do an intermediate release of the release-toolkit before this PR, which first contains all the pending but non-breaking fixes that are waiting for review, and only could we merge this PR and immediately create a new major version (since breaking changes)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -80,6 +82,11 @@ def self.available_options
...
+                                       is_string: true, # true: verifies the input is a string, false: every kind of value
+                                       default_value: 'wordpress'), # the default value if the user didn't provide one
```
AliSoftware: I don't think this is the right choice here to provide a default value of `wordpress`. This will only work on the WordPress-Android repo, and won't make sense in all other repos (WCAndroid, SNAndroid). Default values should only be used imho if they are the common value used in _most_ of our client repos, which isn't the case here 😉

JavonDavis: I was debating on this, I agree let's not use wordpress as the default here 👍

CHANGELOG.md
```diff
@@ -10,11 +10,11 @@ _None_
 
 ### New Features
 
-_None_
+* Support for a version.properties to manage app versioning - all existing paths remain intact and new paths are only used when a version.properties file is present
```
AliSoftware, line 15:
```suggestion
* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -11,11 +11,13 @@ def self.run(params)
         # Checkout develop and update
         Fastlane::Helper::GitHelper.checkout_and_pull('develop')
 
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
```
AliSoftware: Ruby offers a common syntax for this:
```suggestion
        app = params[:app] || ENV['PROJECT_NAME']
```

Though to be honest you could just go with this, because the fallback to read the ENV var is already handled by Fastlane's `ConfigItem`:
```suggestion
        app = params[:app]
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_build_prechecks.rb
```diff
@@ -3,20 +3,22 @@ module Actions
     class AndroidBuildPrechecksAction < Action
       def self.run(params)
         require_relative '../../helper/android/android_version_helper'
+        require_relative '../../helper/git_helper'
 
         UI.user_error!("Can't build beta and final at the same time!") if params[:final] && params[:beta]
 
         Fastlane::Helper::GitHelper.ensure_on_branch!('release') unless other_action.is_ci()
 
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
```
AliSoftware: ditto here… and in all other actions where you used that logic of course 😉 
```suggestion
        app = params[:app]
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_beta.rb
```diff
@@ -46,24 +64,6 @@ def self.authors
...
-      end
-
-      def self.show_config
-        vname = Fastlane::Helper::Android::VersionHelper::VERSION_NAME
-        vcode = Fastlane::Helper::Android::VersionHelper::VERSION_CODE
-        UI.message("Current version: #{@current_version[vname]}(#{@current_version[vcode]})")
-        UI.message("Current alpha version: #{@current_version_alpha[vname]}(#{@current_version_alpha[vcode]})") unless ENV['HAS_ALPHA_VERSION'].nil?
-        UI.message("New beta version: #{@new_version_beta[vname]}(#{@new_version_beta[vcode]})")
-        UI.message("New alpha version: #{@new_version_alpha[vname]}(#{@new_version_alpha[vcode]})") unless ENV['HAS_ALPHA_VERSION'].nil?
-      end
```
AliSoftware, line 66: 👌  Nice cleanup, especially since that was using bad practices (use of instance variables in a class method context). ✨

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_current_branch_is_hotfix.rb
```diff
@@ -7,7 +7,8 @@ module SharedValues
     class AndroidCurrentBranchIsHotfixAction < Action
       def self.run(params)
         require_relative '../../helper/android/android_version_helper'
-        Fastlane::Helper::Android::VersionHelper.is_hotfix?(Fastlane::Helper::Android::VersionHelper.get_release_version)
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
+        Fastlane::Helper::Android::VersionHelper.is_hotfix?(Fastlane::Helper::Android::VersionHelper.get_release_version(app))
```
AliSoftware: 💄 nitpick, given that the test about `ENV['PROJECT_NAME']` is not necessary after all, this below might end up being a nicer split of the code statements:
```suggestion
        version = Fastlane::Helper::Android::VersionHelper.get_release_version(params[:app])
        Fastlane::Helper::Android::VersionHelper.is_hotfix?(version)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
```diff
@@ -12,11 +12,19 @@ module GitHelper
         # @env PROJECT_NAME The name of the directory containing the project code (especially containing the `build.gradle` file)
         #
         def self.commit_version_bump
-          Fastlane::Helper::GitHelper.commit(
-            message: 'Bump version number',
-            files: File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'build.gradle'),
-            push: true
-          )
+          if ENV['HAS_VERSION_PROPERTIES'].nil?
```
AliSoftware: Nice of you to support both modes 👏

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -30,42 +30,77 @@ module VersionHelper
...
+        def self.get_release_version(app)
+          if properties_file_exists
+            section = ENV['HAS_ALPHA_VERSION'].nil? ? 'defaultConfig' : 'vanilla {'
+            gradle_path = self.gradle_path
+            name = get_version_name_from_gradle_file(gradle_path, section)
+            code = get_version_build_from_gradle_file(gradle_path, section)
+            return { VERSION_NAME => name, VERSION_CODE => code }
+          end
+
+          return get_version_from_properties(product_name: app)
```
AliSoftware: 🚨  Wait isn't the logic inverted there? If I'm reading that properly, the logic in that code is "if the `version.properties` file exists, read from gradle file, and if it doesn't exist… reads from it"?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
AliSoftware: What about using `properties_file_exists` for that test instead, and getting rid of the need for the `HAS_VERSION_PROPERTIES` env var altogether, relying on `properties_file_exists` to auto-detect the right case?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -30,42 +30,77 @@ module VersionHelper
...
+          version_code_key = "#{product_name}.#{is_alpha ? 'alpha.' : ''}versionCode"
+
+          properties_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'version.properties')
+
+          return nil unless File.exist?(properties_file_path)
+
+          File.open(properties_file_path, 'r') do |f|
+            text = f.read
+            name = text.match(/#{version_name_key}=(\S*)/m)&.captures&.first
+            code = text.match(/#{version_code_key}=(\S*)/m)&.captures&.first
```
AliSoftware, line 83: nice use of regex and `.captures` 👌

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -30,42 +30,77 @@ module VersionHelper
...
-        # Extract the version name and code from the `defaultConfig` of the `$PROJECT_NAME/build.gradle` file
+        # Extract the version name and code from the `version.properties` file in the project root
         #
         # @return [Hash] A hash with 2 keys `"name"` and `"code"` containing the extracted version name and code, respectively,
         #                or `nil` if `$HAS_ALPHA_VERSION` is not defined.
         #
-        def self.get_alpha_version
-          return nil if ENV['HAS_ALPHA_VERSION'].nil?
+        def self.get_alpha_version(app)
+          if properties_file_exists
```
AliSoftware: Same here, `if properties_file_exists` is true we call `get_version_name_from_gradle_file` et al…

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -267,15 +302,26 @@ def self.bump_version_release
...
-        # @param [Hash] new_version_beta The version hash for the beta (vanilla flavor), containing values for keys "name" and "code"
-        # @param [Hash] new_version_alpha The version hash for the alpha (defaultConfig), containing values for keys "name" and "code"
-        # @env HAS_ALPHA_VERSION If set (with any value), indicates that the project uses `vanilla` flavor.
+        # @param [Hash] new_version_beta The version hash for the beta, containing values for keys "name" and "code"
+        # @param [Hash] new_version_alpha The version hash for the alpha , containing values for keys "name" and "code"
         #
-        def self.update_versions(new_version_beta, new_version_alpha)
-          self.update_version(new_version_beta, ENV['HAS_ALPHA_VERSION'].nil? ? 'defaultConfig' : 'vanilla {')
-          self.update_version(new_version_alpha, 'defaultConfig') unless new_version_alpha.nil?
+        def self.update_versions(app, new_version_beta, new_version_alpha)
```
AliSoftware, line 318: Missing doc for `@param [String] app`

TIP: you can use `bundle exec rake docstats` to find missing documentation, and `bundle exec rake 'docstats[subdir/path]'` to run the stats only on a given sub-directory (like the ones where you made changes in that PR, to avoid noise about existing code you didn't touch)

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -267,15 +302,26 @@ def self.bump_version_release
...
+          if properties_file_exists
+            self.update_version(new_version_beta, ENV['HAS_ALPHA_VERSION'].nil? ? 'defaultConfig' : 'vanilla {')
+            self.update_version(new_version_alpha, 'defaultConfig') unless new_version_alpha.nil?
+            return
+          end
```
AliSoftware: Once again logic seems inverted to me here? (at that point, I'm guessing, probably at every place where this is used too)

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -337,6 +383,37 @@ def self.get_version_parts(version)
...
+        # @param [String] string The string to test
+        #
+        # @return [Bool] true if the string is representing an integer value, false if not
+        #
+        def self.is_int? string
+          true if Integer(string) rescue false
+        end
+
+        #########
+        # Functions to support versioning through build.gradle - can be removed once all projects adopt version.properties
```
AliSoftware, line 422: 👍  for the reminder note about future removal

spec/android_version_helper_spec.rb
```diff
@@ -1,6 +1,45 @@
...
+        wordpress.alpha.versionName=alpha-222
+        wordpress.alpha.versionCode=1234
+      CONTENT
+
+      allow(File).to receive(:exist?).and_return(true)
+      allow(File).to receive(:open).with('./version.properties', 'r').and_yield(StringIO.new(test_file_content))
+      expect(subject.get_version_from_properties(product_name: 'wordpress', is_alpha: true)).to eq('name' => 'alpha-222', 'code' => 1234)
+    end
+
+    it 'returns nil when alpha version name and code when not present' do
```
AliSoftware:
```suggestion
    it 'returns nil when alpha version name and code not present' do
```

AliSoftware (changes requested):
The inverted logic for every place where `properties_file_exist` is used needs fixing

For the rest, I mainly read the code from GitHub, i.e. didn't try to run any command nor the tests on my Mac, but apart from the inverted logic ☝️ rest looks good to me.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
JavonDavis: Ahhhh good catch!

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
JavonDavis: TIL

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
JavonDavis: Yep you're right on this one

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
JavonDavis: Resolved in https://github.com/wordpress-mobile/release-toolkit/pull/258/commits/97b8b032190c097de9b5807e4f9c09c254b1a2f3

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_get_release_version.rb
```diff
@@ -3,7 +3,8 @@ module Actions
     class AndroidGetReleaseVersionAction < Action
       def self.run(params)
         require_relative '../../helper/android/android_version_helper'
-        Fastlane::Helper::Android::VersionHelper.get_release_version()
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
```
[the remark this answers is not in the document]
JavonDavis: fixed in https://github.com/wordpress-mobile/release-toolkit/pull/258/commits/cb939d6f00ed0483eb4baf107361b7281bf9feac

JavonDavis:
@AliSoftware whenever you have a chance again this is ready for another look, thanks for all the feedback so far!

CHANGELOG.md
```diff
@@ -10,11 +10,11 @@ _None_
 
 ### New Features
 
-_None_
-
+* Support for a version.properties to manage app versioning - all existing paths remain intact and new paths are only used when a version.properties file is present
+* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
 ### Bug Fixes
```
AliSoftware, line 16: nit: empty line between sections (I think my release script which processes the changelog supports if you omit it? but better be sure)
```suggestion

### Bug Fixes
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_final_release.rb
```diff
@@ -8,11 +8,20 @@ def self.run(params)
...
+        current_version_alpha = Fastlane::Helper::Android::VersionHelper.get_alpha_version(app)
+        final_version = Fastlane::Helper::Android::VersionHelper.calc_final_release_version(current_version, current_version_alpha)
+
+        vname = Fastlane::Helper::Android::VersionHelper::VERSION_NAME
+        vcode = Fastlane::Helper::Android::VersionHelper::VERSION_CODE
+        UI.message("Current version[#{app}]: #{current_version[vname]}(#{current_version[vcode]})")
+        UI.message("Current alpha version[#{app}]: #{current_version_alpha[vname]}(#{current_version_alpha[vcode]})") unless current_version_alpha.nil?
+        UI.message("New release version[#{app}]: #{final_version[vname]}(#{final_version[vcode]})")
+
+        UI.message 'Updating version.properties...'
```
AliSoftware: To be fair it only updates `version.properties` if there is one, otherwise it updates `build.gradle` 😛

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_hotfix.rb
```diff
@@ -6,11 +6,21 @@ def self.run(params)
...
+        current_version_alpha = Fastlane::Helper::Android::VersionHelper.get_alpha_version(app)
+        new_version = Fastlane::Helper::Android::VersionHelper.calc_next_hotfix_version(params[:version_name], params[:version_code])
+        new_short_version = new_version_name
+        new_release_branch = "release/#{new_short_version}"
+
+        UI.message("Current version[#{app}]: #{current_version[Fastlane::Helper::Android::VersionHelper::VERSION_NAME]}(#{current_version[Fastlane::Helper::Android::VersionHelper::VERSION_CODE]})")
+        UI.message("New hotfix version[#{app}]: #{new_version[Fastlane::Helper::Android::VersionHelper::VERSION_NAME]}(#{new_version[Fastlane::Helper::Android::VersionHelper::VERSION_CODE]})")
+        UI.message("Release branch: #{new_release_branch}")
+
+        UI.message 'Updating version.properties...'
```
AliSoftware: ditto

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
```diff
@@ -9,8 +9,10 @@ def self.run(params)
 
         UI.user_error!('This is not a release branch. Abort.') unless other_action.git_branch.start_with?('release/')
 
-        version = Fastlane::Helper::Android::VersionHelper.get_public_version
-        message = "Completing code freeze for: #{version}\n"
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
```
AliSoftware:
```suggestion
        app = params[:app]
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_codefreeze_prechecks.rb
```diff
@@ -14,18 +14,21 @@ def self.run(params)
         # Checkout develop and update
         Fastlane::Helper::GitHelper.checkout_and_pull('develop')
 
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
```
AliSoftware:
```suggestion
        app = params[:app]
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
```diff
@@ -11,17 +11,31 @@ def self.run(params)
         other_action.ensure_git_branch(branch: 'develop')
 
         # Create new configuration
-        @new_short_version = Fastlane::Helper::Android::VersionHelper.bump_version_release()
-        create_config()
-        show_config()
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
```
AliSoftware:
```suggestion
        app = params[:app]
```
The fallback is already handled by fastlane given how that param is defined in `ConfigItem` below

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_finalize_prechecks.rb
```diff
@@ -13,7 +13,8 @@ def self.run(params)
 
         UI.user_error!('This is not a release branch. Abort.') unless other_action.git_branch.start_with?('release/')
 
-        version = Fastlane::Helper::Android::VersionHelper.get_public_version
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
```
AliSoftware: same for all instances of this everywhere in this PR

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_get_alpha_version.rb
```diff
@@ -3,7 +3,8 @@ module Actions
     class AndroidGetAlphaVersionAction < Action
       def self.run(params)
         require_relative '../../helper/android/android_version_helper'
-        Fastlane::Helper::Android::VersionHelper.get_alpha_version()
+        app = ENV['PROJECT_NAME'].nil? ? params[:app] : ENV['PROJECT_NAME']
+        Fastlane::Helper::Android::VersionHelper.get_alpha_version(app)
```
AliSoftware:
```suggestion
        Fastlane::Helper::Android::VersionHelper.get_alpha_version(params[:app])
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
AliSoftware: @JavonDavis reiterating my comment here since you don't seem to have addressed this since my last review?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -26,39 +26,84 @@ module VersionHelper
...
-        # @env HAS_ALPHA_VERSION If set (with any value), indicates that the project uses `vanilla` flavor.
+        # @param [String] app The name of the app to be used for beta and alpha version update
         #
         # @return [Hash] A hash with 2 keys "name" and "code" containing the extracted version name and code, respectively
         #
-        def self.get_release_version
+        def self.get_release_version(app)
+          if properties_file_exists
+            return get_version_from_properties(product_name: app)
+          end
```
AliSoftware: nit: 
```suggestion          
          return get_version_from_properties(product_name: app) if properties_file_exists
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -26,39 +26,84 @@ module VersionHelper
...
+        # @param [String] app The name of the app to be used for beta and alpha version update
         #
         # @return [Hash] A hash with 2 keys `"name"` and `"code"` containing the extracted version name and code, respectively,
         #                or `nil` if `$HAS_ALPHA_VERSION` is not defined.
         #
-        def self.get_alpha_version
+        def self.get_alpha_version(app)
+          if properties_file_exists
+            return get_version_from_properties(product_name: app, is_alpha: true)
+          end
```
AliSoftware:
```suggestion
          return get_version_from_properties(product_name: app, is_alpha: true) if properties_file_exists
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -267,13 +313,25 @@ def self.bump_version_release
...
+        def self.update_versions(app, new_version_beta, new_version_alpha)
+          if properties_file_exists
+            new_version_name_beta_key = "#{app}.versionName"
+            new_version_code_beta_key = "#{app}.versionCode"
+            new_version_name_alpha_key = "#{app}.alpha.versionName"
+            new_version_code_alpha_key = "#{app}.alpha.versionCode"
+            Action.sh('./gradlew', 'updateVersionProperties', "-Pkey=#{new_version_name_beta_key}", "-Pvalue=#{new_version_beta[VERSION_NAME]}")
+            Action.sh('./gradlew', 'updateVersionProperties', "-Pkey=#{new_version_code_beta_key}", "-Pvalue=#{new_version_beta[VERSION_CODE]}")
+            Action.sh('./gradlew', 'updateVersionProperties', "-Pkey=#{new_version_name_alpha_key}", "-Pvalue=#{new_version_alpha[VERSION_NAME]}") unless new_version_alpha.nil?
+            Action.sh('./gradlew', 'updateVersionProperties', "-Pkey=#{new_version_code_alpha_key}", "-Pvalue=#{new_version_alpha[VERSION_CODE]}") unless new_version_alpha.nil?
```
AliSoftware, line 327: We should document that this task is assumed to exist in the host app repo.

This should be documented at least:
 - in the `CHANGELOG`, to help people know how to migrate to the new system when they will migrate their host app and look at what's new
 - in each actions' documentation in which this is relevant, so that any future adopter (even one that just migrates right now, but will decide to use this new system in the future) has a reference documentation

As a bonus, it would also be nice to make the release toolkit print a message if that task doesn't exist, but I'm not sure that is gonna be easy (maybe catch the potential error of `sh` to print additional help?)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_codefreeze_prechecks.rb
```diff
@@ -14,18 +14,21 @@ def self.run(params)
...
+
         # Create versions
-        current_version = Fastlane::Helper::Android::VersionHelper.get_release_version
-        current_alpha_version = Fastlane::Helper::Android::VersionHelper.get_alpha_version
+        current_version = Fastlane::Helper::Android::VersionHelper.get_release_version(app)
+        current_alpha_version = Fastlane::Helper::Android::VersionHelper.get_alpha_version(app)
         next_version = Fastlane::Helper::Android::VersionHelper.calc_next_release_version(current_version, current_alpha_version)
-        next_alpha_version = ENV['HAS_ALPHA_VERSION'].nil? ? nil : Fastlane::Helper::Android::VersionHelper.calc_next_alpha_version(next_version, current_alpha_version)
+        next_alpha_version = current_alpha_version.nil? ? nil : Fastlane::Helper::Android::VersionHelper.calc_next_alpha_version(next_version, current_alpha_version)
```
AliSoftware, line 23: It seems that after your changes, there are only 3 instances of `HAS_ALPHA_VERSION` remaining in the codebase
```
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
54:          section = ENV['HAS_ALPHA_VERSION'].nil? ? 'defaultConfig' : 'vanilla {'
100:        #                or `nil` if `$HAS_ALPHA_VERSION` is not defined.
107:          return nil if ENV['HAS_ALPHA_VERSION'].nil?
335:          self.update_version(new_version_beta, ENV['HAS_ALPHA_VERSION'].nil? ? 'defaultConfig' : 'vanilla {')
```

It seems worth eliminating the remaining ones and add a mention in the CHANGELOG that host apps adopting this new version of the toolkit could get rid of it in their Fastfile

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
JavonDavis: @AliSoftware Wouldn't applying this suggestion break clients that don't conform to these changes -> I left this in to allow for the environment variable to still be used

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_final_release.rb
JavonDavis: 😅

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
JavonDavis: Ahhh Right!

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_completecodefreeze_prechecks.rb
JavonDavis: Oof! Nevermind I missed that Fastlane handles this for me 👍

AliSoftware: yeah that's the whole point of the `env_name` attribute of the `ConfigItem` you define a couple of lines below in the action :)

CHANGELOG.md
JavonDavis: Added along with other Changelog updates

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
JavonDavis: I've both updated the Changelog and the action documentation here! Hmm I'll think a bit more about how best to fail here if the gradle task isn't present but might handle that in a smaller PR

AliSoftware: Yeah the failure detection might be tricky enough, totally ok to handle that in a separate PR… or not at all and consider this to be ok to fail as long as it's documented.

CHANGELOG.md
```diff
@@ -6,15 +6,16 @@
 
 ### Breaking Changes
 
-_None_
+* Clients adopting the new version.properties will need to implement a gradle task to update the version.properties file named `updateVersionProperties`
```
AliSoftware: I'm unsure if that's a breaking change, given that you made all the efforts to make this release backwards compatible with projects not updating this new feature… So I think this not can just go in the New Features section as part of the bullet point about new `version.properties` support.

Otherwise if you put this in Breaking Changes, this will mean a major version bump to `2.0.0`, and thus client apps not auto-updating to the release toolkit automatically because of the major version bump won't be semantically compatible with `~> 1.0`… even if in practice nothing will break if they update because the change is backwards compatible…

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_codefreeze_prechecks.rb
JavonDavis: I'll make mention of it in the Changelog that this file can be removed but at the moment it looks like the remaining ones are currently only used for allowing adoption without conforming right away. It allows for updating to this version and later versions without necessarily supporting the version.properties changes right away. Once all our clients have supported it we can definitely go ahead and remove all the instances.

AliSoftware: Ok make sense, we can totally handle this later (could be worth opening an issue, similar to [the one we have to get rid of `PROJECT_ROOT_FOLDER`)](https://github.com/wordpress-mobile/release-toolkit/issues/205), to not forget later)

CHANGELOG.md
JavonDavis: ok sounds good!

JavonDavis: I had somewhat considered it a breaking change because it's somewhat a requirement for completely adopting the approach but I think your reasoning here is better

CHANGELOG.md
```diff
@@ -6,13 +6,12 @@
...
 
-* Clients adopting the new version.properties will need to implement a gradle task to update the version.properties file named `updateVersionProperties`
-
 ### New Features
 
 * Support for a version.properties to manage app versioning - all existing paths remain intact and new paths are only used when a version.properties file is present
 * Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
-* Supporting the new version.properties file also allows for the HAS_ALPHA_VERSION variable to be removed as the alpha reference in the properties file will be used going forward. 
+* Supporting the new version.properties file also allows for the HAS_ALPHA_VERSION variable to be removed as the alpha reference in the properties file will be used going forward.
+* Clients adopting the new version.properties will need to implement a gradle task to update the version.properties file named `updateVersionProperties`
```
AliSoftware: 👌

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_beta.rb
```diff
@@ -23,14 +34,21 @@ def self.run(params)
       #####################################################
 
       def self.description
-        'Bumps the version of the app for a new beta'
+        'Bumps the version of the app for a new beta. Depends on a gradle task to update the keys in a version.properties file.'
```
AliSoftware: We need to think about the people who will adopt this version of the release toolkit and have never seen your PR nor code. They should not have to guess what gradle task they need to add to make this working.

```suggestion
        'Bumps the version of the app for a new beta. Requires the `updateVersionProperties` gradle task to update the keys if you are using a `version.properties` file.'
```

(And same suggestion (search/replace all) on all the places and other actions where you mentioned this)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_final_release.rb
```diff
@@ -23,11 +32,21 @@ def self.run(params)
...
+        'Bumps the version of the app for a new beta. Depends on a gradle task to update the keys in a version.properties file.'
+      end
+
+      def self.available_options
+        # Define all options your action supports.
+        [
+          FastlaneCore::ConfigItem.new(key: :app,
+                                       env_name: 'PROJECT_NAME',
+                                       description: 'The name of the app to get the release version for',
+                                       is_string: true), # true: verifies the input is a string, false: every kind of value
```
AliSoftware, line 48: Nit: This is legacy api of `ConfigItem` constructor. The more modern version is:
```suggestion
                                       type: String),
```

This is not a blocker and can be fixed in a subsequent PR; but could be nice to take the occasion to fix while you're at it; this suggestion applies in every places and actions where you added this of course.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb
```diff
@@ -12,11 +12,20 @@ module GitHelper
...
+              files: File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'build.gradle'),
+              push: true
+            )
+          else
+            Fastlane::Helper::GitHelper.commit(
+              message: 'Bump version number',
+              files: File.join(ENV['PROJECT_ROOT_FOLDER'], 'version.properties'),
+              push: true
+            )
+          end
```
AliSoftware: 🚨  This one still seems inverted, commiting `build.gradle` if `version.properties` exists…

CHANGELOG.md
```diff
@@ -6,15 +6,16 @@
...
 
-_None_
-
 ### New Features
 
-_None_
+* Support for a version.properties to manage app versioning - all existing paths remain intact and new paths are only used when a version.properties file is present
+* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
+* Supporting the new version.properties file also allows for the HAS_ALPHA_VERSION variable to be removed as the alpha reference in the properties file will be used going forward.
+* Clients adopting the new version.properties will need to implement a gradle task to update the version.properties file named `updateVersionProperties`
```
AliSoftware: 💄  nit:
```suggestion
* Clients adopting the new `version.properties` will need to implement a gradle task named `updateVersionProperties` to update the `version.properties` file.
```

CHANGELOG.md
```diff
@@ -6,15 +6,16 @@
...
 ### Breaking Changes
 
-_None_
-
 ### New Features
 
-_None_
+* Support for a version.properties to manage app versioning - all existing paths remain intact and new paths are only used when a version.properties file is present
+* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
+* Supporting the new version.properties file also allows for the HAS_ALPHA_VERSION variable to be removed as the alpha reference in the properties file will be used going forward.
```
AliSoftware: 💄  small nit:
```suggestion
* Supporting the new `version.properties` file also allows for the `HAS_ALPHA_VERSION` variable to be removed as the alpha reference in the properties file will be used going forward.
```

CHANGELOG.md
```diff
@@ -6,15 +6,16 @@
 
 ### Breaking Changes
 
-_None_
-
 ### New Features
 
-_None_
+* Support for a version.properties to manage app versioning - all existing paths remain intact and new paths are only used when a version.properties file is present
```
AliSoftware: 💄  small nitpick:
```suggestion
* Support for a `version.properties` to manage app versioning - all existing paths remain intact and new paths are only used when a `version.properties` file is present.
```

AliSoftware (changes requested):
Still spotted an inverted logic remaining…

Once this is fixed and the other nits are addressed, should hopefully be finally good to merge tho. (Don't forget to also create the issue about alpha env var removal to handle it later)

AliSoftware:
Also there's a CI failure on tests, we should fix it before merging, in order to make CI actually run the tests and ensure those changes didn't break any existing test nor introduce regressions.

oguzkocer:
@JavonDavis About your review request, since @AliSoftware is already reviewing this, and he's our expert :) I don't know if I can add any value to this PR by reviewing it. I have quite a few things to get through in the next couple days, so I wanted to check if you'd still like me to review it or not. If you do, I'll try to get to it before my EOD.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_beta.rb
JavonDavis: That's true! Agree with being more explicit here

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_codefreeze_prechecks.rb
JavonDavis: Created https://github.com/wordpress-mobile/release-toolkit/issues/277

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_final_release.rb
JavonDavis: Ahhh I think I'd rather tackle it in another PR, opened https://github.com/wordpress-mobile/release-toolkit/issues/278 to track it.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
JavonDavis: Do you think it also might be useful to log another user message here? something like `attempting to update properties file with './gradlew updateVersionProperties' `

JavonDavis:
>  If you do, I'll try to get to it before my EOD.

No worries @oguzkocer, though I'd definitely appreciate your feedback on the changes on the Android side once I open it for review for this so feel free to ignore this one!

AliSoftware (approved):
Ok, all looks good now 🎉

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -46,8 +46,8 @@ def self.get_public_version(app)
         #
         # @return [Hash] A hash with 2 keys "name" and "code" containing the extracted version name and code, respectively
         #
-        def self.get_release_version(app)
-          return get_version_from_properties(product_name: app) if properties_file_exists
+        def self.get_release_version(product_name:)
```
AliSoftware, line 49: You should update the comment on line 45 too, to match the param name 😉 

```
        # @param [String] product_name The name of the app to be used for beta and alpha version update (i.e the prefix in version.properties file)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
AliSoftware: Nah iinm `Action.sh` already logs the commands it runs in stdout so that should be enough

PR #259, opened 2021-05-13
Manually inspect the file system to evaluate whether path is in Git repo

mokagio:
As opposed to shelling out to `git rev-parse`.

See https://github.com/wordpress-mobile/release-toolkit/pull/252#discussion_r631451151.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -23,11 +23,14 @@ module Fastlane
         return current_dir.root? == false
       end
 
-      # Travels back the hierarchy of the given path until it finds an existing ancestor, or it reaches the root of the file system.
+      # Travels back the hierarchy of the given path until it finds an existing
+      # ancestor, or it reaches the root of the file system.
       #
       # @param [String] path The path to inspect
       #
-      # @return [Pathname] The first existing ancestor, or `path` itself if it exists
+      # @return [Pathname] The first existing ancestor, or `path` itself if it
+      #         exists
+      #
       def self.first_existing_ancestor_of(path:)
         p = Pathname(path).expand_path
         p = p.parent until p.exist? || p.root?
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -15,12 +15,19 @@ module Fastlane
       #         (i.e. a local working copy) or a subdirectory of one.
       #
       def self.is_git_repo?(path: Dir.pwd)
-        working_path = first_existing_ancestor_of(path: path)
-        current_dir = working_path.directory? ? working_path : working_path.dirname
+        # If the path doesn't exist, find its first ancestor.
+        path = first_existing_ancestor_of(path: path)
+        # Get the path's directory, so we can look in it for the Git folder
+        dir = path.directory? ? path : path.dirname
 
-        current_dir = current_dir.parent until Dir.entries(current_dir).include?('.git') || current_dir.root?
+        # Recursively look for the Git folder until it's found or we read the
+        # the file system root
+        dir = dir.parent until Dir.entries(dir).include?('.git') || dir.root?
 
-        return current_dir.root? == false
+        # If we reached the root, we haven't found a repo. (Technically, there
+        # could be a repo in the root of the system, but that's a usecase that
+        # we don't need to support at this time)
+        return dir.root? == false
       end
 
       # Travels back the hierarchy of the given path until it finds an existing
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -15,22 +15,20 @@ module GitHelper
...
+        full_path = Pathname(current_dir).expand_path
 
-        Action.sh(args, print_command_output: false) do |status, _, _|
-          status.success?
+        if full_path.children.map { |f| File.basename(f) }.include?('.git')
+          return true
+        elsif full_path == ROOT_FOLDER
+          return false
+        else
+          return is_git_repo?(path: full_path.parent)
```
mokagio: I'm not too happy with how this code _looks_.

Keen to hear if you have any suggestions 😄 

Also notice that I added two extra tests for the recursively traversing scenario in the base branch: 6cf529fa2c45ff87051da79c1d40a206e55a0d73

jkmassel: Feels a little dense in the first two lines – I don't think I'd expand out the ternaries, but maybe a well-placed comment or two would help it flow a bit more nicely? 🤔 

IMHO not a huge deal though – grokking it only took a few seconds 🤷‍♂️

jkmassel (approved):
I like this approach due to the lack of shelling out, personally!

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -15,22 +15,20 @@ module GitHelper
...
 
-        args += ['rev-parse']
+        # Calling `expand_path` to get an absolute path and avoid and endless
+        # loop of adding ".." to the path without ever triggerting the "/"
+        # safeguard.
+        full_path = Pathname(current_dir).expand_path
 
-        Action.sh(args, print_command_output: false) do |status, _, _|
-          status.success?
+        if full_path.children.map { |f| File.basename(f) }.include?('.git')
```
AliSoftware: 1. I thought `children` already only returned the relative path (ie only the file name) not the full path? Which means `File.basename(f)` is probably unnecessary here?
2. Also, isn't there a more appropriate method in `File` or `Dir` to check if a dir contains a file? What about testing `if File.exist?(File.join(full_path, '.git'))` instead? Or even better, `if File.directory?(File.join(full_path, '.git'))` to both check existence AND that it is a directory not a file? Feels less consuming than asking to list all the children of a dir only to check if a single specific child (`.git`) exists 😉

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: One thing that also feels strange with this is that there are 2 loops / recursion levels: one within `first_existing_ancestor_of` to go up until it finds an existing ancestor, and another one here to go up until it finds a `.git` or the root folder. Couldn't we avoid going thru the `first_existing_ancestor_of` loop on subsequent recursion loops, to avoid unnecessarily repeating some steps there?

Personally I think it's not worth using recursion here as opposed to using a `while` or `until` loop, something like:

```ruby
until (full_path + '.git').exist? || full_path.root? do
  full_path = full_path.parent
end
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
AliSoftware: Actually I see that you're using `Pathname` to play with the paths (which is great, I love that class instead of using `File.xxx` and `Dir.xxx` class methods! 😍 ), but then if we have a `Pathname`, maybe better (1) create the `Pathname` earlier in order to take advantage of the nicer API of that type early on in the code (2) use the `Pathname` instance methods everywhere from that point in this implementation, instead of having a mix of `Pathname` instance methods and `File.x(f)` global static methods?

This is also true about the test for root folder on line 28 below: `Pathname` has a `root?` instance method that feels better suited than a comparison to a custom `ROOT_FOLDER` constant 😉

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -14,34 +14,34 @@ module GitHelper
...
-          path = first_existing_ancestor_of(path: path)
-          args += ['-C', File.file?(path) ? File.dirname(path) : path]
-        end
-
-        args += ['rev-parse']
-
-        Action.sh(args, print_command_output: false) do |status, _, _|
-          status.success?
-        end
+      def self.is_git_repo?(path: Dir.pwd)
```
mokagio, line 17: 🤦 

Why did I have a if `nil?` default to `Dir.pwd` check in the original version where I could have defaulted to `Dir.pwd` in the first place?!

mokagio:
Thank you @jkmassel and @AliSoftware for your help.

I've incorporated some/all of your suggestions in what seems to me like a clearer and more concise implementation now 🙇 

The tests still pass so I'm gonna go ahead and merge this. Happy to keep iterating if you have more suggestions ⚒️

PR #263, opened 2021-05-21
Feature/check translations status

loremattei:
This PR adds a new action which checks the state of the translations on GlotPress. 
It can be used to verify that the expected translations are available on GlotPress before finalising a release. 

Since there's no API on GlotPress (AFAIK) to get this information, but it's pretty easy to parse the HTTP code, the action extract the important data from the GlotPress project page. 

By default the action expects all the languages to be translated at 100% and raises a blocking error if it finds something below the threshold, but it can has options to modify the threshold and to ask user confirmation instead of failing directly. It has also an option to print the error report without blocking the flow at all. 

### To Test
- Verify CI is green. 
- Test on a live GlotPress repository (e.g. https://translate.wordpress.org/projects/apps/android/dev/) and verify the following options: 
a. _min_acceptable_translation_percentage_: pick a value that should trigger the error _and_ one which shouldn't (you can check the actual state opening the GlotPress project page on a web browser. 
b. _abort_on_violations = false_: set proper values of _min_acceptable_translation_percentage_ and verify that the output report is coherent and that the flow gets aborted when the user answers with **No** to the confirmation request. 
c. _abort_on_violations = false_ && _skip_confirm=true_: set proper values of _min_acceptable_translation_percentage_ and verify that the output report is coherent and that the flow is not interrupted.


.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - restore_cache:
           name: Restore Homebrew + Ruby Dependencies
           keys:
-            - &cache_key brew-dependencies-{{ checksum ".circleci/.brewfile" }}-{{ checksum "Gemfile.lock" }}
+            - &cache_key brew-dependencies-{{ checksum ".circleci/cache-version" }}-{{ checksum ".circleci/.brewfile" }}-{{ checksum "Gemfile.lock" }}
```
loremattei, line 18: It looks like `imagemagick@6` formulae got updated very recently and it broke the installation on CI because the previous version was in the cache and `homebrew` didn't know which one to link. Interestingly, `homebrew` defaults to returning error code 1 in this case. 

I'm pretty sure there are smarter things we can do with how we manage the CircleCI cache, but, since we're planning to switch away from CircleCI and, so, spending more time on it may not be worth it, I opted for implementing the solution we've used on WPiOS for a few years now.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+* Added a `check_translation_progress` action which checks the status of the translations on GlotPress. [#263]
```
loremattei, line 14: I'm not totally sure this is the way this file is intended to be used :-)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb
```diff
@@ -0,0 +1,135 @@
...
+      def self.check_translations(glotpress_url:, language_codes:, abort_on_violations:, threshold:)
+        under_threshold_langs = []
+
+        data = Fastlane::Helper::GlotPressHelper.get_translation_status_data(glotpress_url: glotpress_url) rescue nil
+        UI.abort_with_message!("Can't retrieve data from #{glotpress_url}") if data.nil? || data.length == 0
+
+        language_codes.each do | language_code |
+          progress = Fastlane::Helper::GlotPressHelper.get_translation_status(
+                                      data: data,
+                                      language_code: language_code) rescue -1
```
loremattei: There are few ways this can fail, but since we are trying to parse a HTML file, I think that the safest solution is to default to fail the job every time things are not exactly as expected.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb
```diff
@@ -0,0 +1,135 @@
...
+        end
+
+        if skip_confirm 
+          UI.important(message) 
+        else
+          if UI.interactive? 
+            UI.abort_with_message!("Aborted by user!") unless UI.confirm("#{message}Do you want to continue?") 
+          else 
+            UI.abort_with_message!(message)
+          end
```
loremattei: This is not supposed to be run on CI and with the current options there is some control. As a first implementation, I think it's not worth it to add complexity to handle cases which have low caches to happen and so the action raises an error if there are warnings when run on CI.

spec/check_localization_progress_spec.rb
```diff
@@ -0,0 +1,337 @@
+require 'spec_helper.rb'
+require 'webmock/rspec'
+
+describe Fastlane::Actions::CheckTranslationProgressAction do
+  before(:each) do
+    allow(FastlaneCore::UI).to receive(:'message')
+    allow(FastlaneCore::UI).to receive(:'interactive?').and_return(true)
+  end
```
loremattei, line 8: The behaviour is slightly different on CI and local, so this forces the tests to always run as in a local environment. This is the most likely scenario for this action and the logic on our side is pretty simple, so I opted for this rather than adding tests for all the cases which would basically end up testing Fastlane rather than our logic.

.circleci/config.yml
mokagio: Makes sense 👍

mokagio: **Out of scope**

I was just looking at #267, where there's [a comment](https://github.com/wordpress-mobile/release-toolkit/pull/267#discussion_r639242521) about how `drawText` (which I'm pretty sure under the hood depends on ImageMagick) affects the CI flow.

I think we might benefit from moving the screenshots composition tool to its own gem. It changes at a far slower pace that the release toolkit but it adds a bunch of overhead due to ImageMagick. It's also detached enough in what it does that it would feel odd if it wasn't part of this repo, IMHO.

CHANGELOG.md
mokagio: This seems good to me.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb
```diff
@@ -0,0 +1,135 @@
+module Fastlane
+  module Actions
+    class CheckTranslationProgressAction < Action
+      def self.run(params)
+        require_relative '../../helper/glotpress_helper.rb'
+
+        UI.message('Check translations status...')
```
mokagio: Nitpick, what do you think of "checking" to explain the action is happening right now? "check" makes me think the tool is telling the user to go and check.

 ```suggestion
        UI.message('Checking translations status...')
```

File: lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb [hunk 7]
```diff
@@ -0,0 +1,135 @@
...
+                                  abort_on_violations: params[:abort_on_violations],
+                                  threshold: params[:min_acceptable_translation_percentage])
+
+        check_results(
+          under_threshold_langs: under_threshold_langs, 
+          threshold: params[:min_acceptable_translation_percentage], 
+          skip_confirm: params[:skip_confirm]
+        ) unless under_threshold_langs.length == 0
+
+        UI.message('Done')
```
mokagio: Nitpick, what do you think of using the `success` method to print a green (?) message to the user?

```suggestion
        UI.success('Done')
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb
```diff
@@ -0,0 +1,135 @@
...
+        UI.abort_with_message!("Can't retrieve data from #{glotpress_url}") if data.nil? || data.length == 0
+
+        language_codes.each do | language_code |
+          progress = Fastlane::Helper::GlotPressHelper.get_translation_status(
+                                      data: data,
+                                      language_code: language_code) rescue -1
+          
+          if (abort_on_violations)
+            UI.abort_with_message!("Can't get data for language #{language_code}") if (progress == -1) 
+            UI.abort_with_message!("#{language_code} is translated #{progress}% which is under the required #{threshold}%.") if (progress < threshold)
```
mokagio: Nitpick. How do you think this reads?

```suggestion
            UI.abort_with_message!("#{language_code} is #{progress}% translated, which is under the required #{threshold}% threshold.") if (progress < threshold) 
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb
```diff
@@ -0,0 +1,135 @@
...
+          FastlaneCore::ConfigItem.new(key: :glotpress_url,
+                                       env_name: 'FL_CHECK_TRANSLATION_PROGRESS_GLOTPRESS_URL',
+                                       description: 'URL to the GlotPress project',
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :language_codes,
+                                       env_name: 'FL_CHECK_TRANSLATION_PROGRESS_LANGUAGE_CODES',
+                                       description: 'The list of the codes of the languages to check',
+                                       type: Array,
+                                       optional: true,
+                                       # Default to Mag16.
```
mokagio: What is Mag16? I [Googled it](https://www.google.com/search?sxsrf=ALeKk03OAvPnIceRXZSbg2zBv3sbfmBUbg:1622003918773&q=mag+16&spell=1&sa=X&ved=2ahUKEwjAhJy7w-bwAhUzxjgGHQMYAu0QBSgAegQIARAx&cshid=1622003949568059&biw=1164&bih=877) and only found results about the Marine Airforce Group 16 🤔

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+      # @param [Array] Data from GlotPress project overview page.
+      # @param [String] The code of the language to get information about (GlotPress format).
+      # @param [String] The status which data should be extracted for.  
+      #
+      # @return [Integer] The percentage of the translated strings.
+      #
+      def self.extract_value_from_translation_info_data(data:, language_code:, status:)
+        regex = "\/#{language_code}\/.*#{status}.*>([0-9,]+)"
+
+        # 1. Grep the line with contains the required info.
```
mokagio, line 48: Should this be

```suggestion
        # 1. Grep the line which contains the required info.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+        # 2. Match the info and extract the value in group 1. 
+        # 3. Values use comma as thousands separator, so remove it.
```
mokagio: TIL "thousands separator" is the name for those 😅

spec/check_localization_progress_spec.rb
```diff
@@ -0,0 +1,337 @@
...
+      .to_return(
+        status: 200,
+        body: generate_glotpress_response_body(languages: langs)
+      )
+
+    expect(FastlaneCore::UI).to receive(:'message').with('Done')
+
+    Fastlane::Actions::CheckTranslationProgressAction.run(
+      glotpress_url:'https://translate.wordpress.org/projects/apps/my-test-project/dev',
+      language_codes: 'ar de es'.split(),
```
mokagio: I recently learned about [Ruby's percent strings](https://ruby-doc.org/core-2.7.3/doc/syntax/literals_rdoc.html#label-Percent+Strings) which is an IMHO simpler / more idiomatic way to achieve this. 


```suggestion
      language_codes: %w[ar de es],
```

spec/check_localization_progress_spec.rb
```diff
@@ -0,0 +1,337 @@
...
+
+    stub = stub_request(
+      :get,
+      'https://translate.wordpress.org/projects/apps/my-test-project/dev')
+      .to_return(
+        status: 200,
+        body: generate_glotpress_response_body(languages: langs)
+      )
+
+    expect(FastlaneCore::UI).to receive(:'abort_with_message!').with('Can\'t get data for language es')
```
mokagio: Nitpick: 

```suggestion
    expect(FastlaneCore::UI).to receive(:'abort_with_message!').with("Can't get data for language es")
```

spec/check_localization_progress_spec.rb
```diff
@@ -0,0 +1,337 @@
...
+  languages.each do | language |
+    response << generate_glotpress_response_for_language(
+      lang: language[:lang_name],
+      lang_code: language[:lang_code],
+      current: language[:current],
+      fuzzy: language[:fuzzy],
+      waiting: language[:waiting],
+      untranslated: language[:untranslated]
+    )
+  end
```
mokagio: This does the job well and I don't think there's great ROI in changing it, but it might be a good use case for using ERB, the Ruby templating language.

spec/check_localization_progress_spec.rb
```diff
@@ -0,0 +1,337 @@
+require 'spec_helper.rb'
+require 'webmock/rspec'
+
+describe Fastlane::Actions::CheckTranslationProgressAction do
+  before(:each) do
+    allow(FastlaneCore::UI).to receive(:'message')
```
mokagio: Not sure why Danger is not commenting here, but Rubocop locally gives me this error

```
Style/SymbolLiteral: Do not use strings for word-like symbol literals.
```

In fact, there are a few violations that I get when running `bundle exec rubocop spec/check_localization_progress_spec.rb` locally (I had to rebase on `develop` and run `bundle install` first to workaround an configuration error). 🤔

mokagio (approved):
LGTM. I think it makes sense to focus on having this working locally and only put in the effort to support a non interactive mode (with the unit testing setup it involves) only once we have a use case for running this in CI.

I only left a bunch of nitpicks, nothing blocking a merge.

Here's the tests I run

```
check_translation_progress \
  glotpress_url:"https://translate.wordpress.org/projects/apps/ios/dev" \
  min_acceptable_translation_percentage:99

=> [!] zh-tw is translated 97% which is under the required 99%.
```

```
check_translation_progress \
  glotpress_url:"https://translate.wordpress.org/projects/apps/ios/dev" \
  min_acceptable_translation_percentage:1

=> Done
   Result: true
```

```
check_translation_progress \
  glotpress_url:"https://translate.wordpress.org/projects/apps/ios/dev" \
  min_acceptable_translation_percentage:99 \
  abort_on_violations:false

=> The translations for the following languages are below the 99% threshold:
   - zh-tw is at 97%.
   Do you want to continue? (y/n)
```

```
check_translation_progress \
  glotpress_url:"https://translate.wordpress.org/projects/apps/ios/dev" \
  min_acceptable_translation_percentage:99 \
  abort_on_violations:false \
  skip_confirm:true

=> The translations for the following languages are below the 99% threshold:
   - zh-tw is at 97%.

   Done
   Result: true
```

I'm not sure where "Result: true" comes from? 🤔 

---

During testing, I found it useful to add a `UI.message` reporting the status of each translation. YMMV 🤷‍♂️ 

image

spec/check_localization_progress_spec.rb
mokagio: Looking [in CI](https://github.com/wordpress-mobile/release-toolkit/pull/263/checks?check_run_id=2656810727), I think the reason is because these are new files?

This is the output of the Danger run, which also includes the error I mentioned:

```
Updated 0 paths from the index
.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
Error: The `Naming/UncommunicativeMethodParamName` cop has been renamed to `Naming/MethodParameterName`.
(obsolete configuration found in .rubocop_todo.yml, please update it)
```

Weird that Danger didn't fail the build because of that 🤔 Although it might be due to the inline comment configuration. cc @AliSoftware.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb
AliSoftware: @mokagio Mag16 is the "Magnificent 16", what we nicknamed the main 16 locales that our apps have official support for (many apps like WP support more locales than the Mag16 but the Mag16 are the ones for which we have official translators for and want to have 100% localization coverage, while translations for other locales are provided thanks to our external contributors, which can't guarantee 100% coverage)

TL;DR: You should search MGS and the a8c glossary before Google 😅

mokagio: > TL;DR: You should search MGS and the a8c glossary before Google 😅

😮 😅

spec/check_localization_progress_spec.rb
AliSoftware: The bug with inline comment configuration not always posting comments for all violations is because of a GitHub API limitation (apparently), which likely limits the number of comments per second a bot can post. So when we hit this limitation:
 - We generally at least have N (=2? 3?) Danger comments and only the additional Danger comments that the Danger job should have posted after those first N get dropped. While here we don't have any Danger comment at all
 - To double-check, you can open [the log of the Danger GitHub Action](https://github.com/wordpress-mobile/release-toolkit/pull/263/checks?check_run_id=2656810727) and look at the "Run Danger" step. That step will list all the violations it found, which includes both the first N violations that it would have managed to post an inline comment for and the subsequent violations which the GH limit would have potentially blocked

In our present case this is none of that; instead, Danger got interrupted when rubocop exited with exit code 1 due to a configuration error ("Error: The `Naming/UncommunicativeMethodParamName` cop has been renamed to `Naming/MethodParameterName`.") and didn't do the linting. This was fixed in my other PR (which has now been merged) so rebasing this PR on top of develop should fix that.

PR #266, opened 2021-05-25
Fix CI setup to push releases (-> develop)

AliSoftware:
This is the same PR as https://github.com/wordpress-mobile/release-toolkit/pull/267 but targeting `develop` (while #267 was for `trunk`). It has already been validated and tested during the release of `1.0.1`.

Now merging this back into `develop` just to get the branches aligned and sync'd up.

### Why?

Since the initial CircleCI setup steps failed to push `1.0.1`, I fixed the `.circleci/config.yml` in a dedicated branch (cut from `trunk`) and deleted the old `1.0.1` tag to recreate it on top of this fix.

### How?

This fix:

 - Avoids installing `brew` dependencies that are not needed for `gem build` and `gem push`, avoiding to waste CI time unnecessarily
 - Test the install of the gem using `--user-install` to avoid permission issues of trying to install it system-wide
 - Setup the RubyGems credentials file before calling `gem push` (the [RubyGems doc](https://guides.rubygems.org/command-reference/#gem-push) made me wrongly believe having the `GEM_HOST_API_KEY` env var set would be enough, but apparently not)


AliSoftware:
Note: as with previous PRs, the CI failure on the Danger check is related to a glitch with GitHub Actions not finding `nokogiri` (which CircleCI apparently has no issue with) and is unrelated to that PR, so can be ignored for now (we'll investigate this unrelated issue later, in a separate PR)

mokagio (approved):
:shipit:

PR #267, opened 2021-05-25
Fix CI setup to create releases (-> trunk)

AliSoftware:
Since the initial CircleCI setup steps failed to push `1.0.1`, I fixed the `.circleci/config.yml` in a dedicated branch (cut from `trunk`) and deleted the old `1.0.1` tag to recreate it on top of this fix.

The fix:
 - Avoids installing `brew` dependencies that are not needed for `gem build` and `gem push`, avoiding to waste CI time unnecessarily
 - Test the install of the gem using `--user-install` to avoid permission issues of trying to install it system-wide
 - Setup the RubyGems credentials file before calling `gem push` (the [RubyGems doc](https://guides.rubygems.org/command-reference/#gem-push) made me wrongly believe having the `GEM_HOST_API_KEY` env var set would be enough, but apparently not)

Now merging this fix back into `trunk`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,20 +7,24 @@ orbs:
 commands:
   setup_tools:
     description: "Install Homebrew and Ruby dependencies"
+    parameters:
+      skip_brew_install:
+        type: boolean
+        default: false
     steps:
       - restore_cache:
           name: Restore Homebrew + Ruby Dependencies
           keys:
             - &cache_key brew-dependencies-{{ checksum ".circleci/.brewfile" }}-{{ checksum "Gemfile.lock" }}
-
-      - run:
-          name: Install Homebrew dependencies, if neeeded
-          command: brew update && xargs brew install --verbose < .circleci/.brewfile
-
+      - unless:
+          condition: <<parameters.skip_brew_install>>
+          steps:
+            - run:
+                name: Install Homebrew dependencies, if neeeded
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -7,20 +7,24 @@ orbs:
...
-      - run:
-          name: Install Homebrew dependencies, if neeeded
-          command: brew update && xargs brew install --verbose < .circleci/.brewfile
-
+      - unless:
+          condition: <<parameters.skip_brew_install>>
+          steps:
+            - run:
+                name: Install Homebrew dependencies, if neeeded
+                command: brew update && xargs brew install --verbose < .circleci/.brewfile
```
jkmassel, line 24: This is nice – ideally we could get rid of the need for `brew` entirely before long?

jkmassel (approved):

.circleci/config.yml
AliSoftware: Yeah I'm not entirely sure about this because I have assumed that the `test` job needs imagemagick to at least run the tests / `rspec` which probably include some tests about running `drawText` or testing some screenshot-related actions, which is why it would need those… So I think we might still need that step for the `test` job

AliSoftware:
Once again, Danger failure check on GitHub Actions is unrelated, so keeping this to investigate for later, and merging this despite that red check.

PR #268, opened 2021-05-27
Fix: Typography dashes substitution in Android

loremattei:
The Android linter includes some typography checks and, in particular, it checks for the use of en-dashes in numeric ranges (see `Typography Dashes` [here](http://tools.android.com/tips/lint-checks)).

This rule is not always applied by translators with the result that the build fails because of the related error raised by the linter. 
The old bash script that was used to download the translations had a check for this case and some logic to substitute the dash on the fly during the download, but a similar behaviour hasn't been implemented in the new Ruby script. 

This PR fixes the issue by implementing the substitution logic. The detection code follows the implementation in the Android linter (see `mCheckDashes` https://android.googlesource.com/platform/tools/base/+/master/lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/TypographyDetector.java#262).

## To Test
1. Verify CI is green.
2. The following step can be performed on any client, though I'm using WPAndroid as a reference to make things clearer: 
2.a Checkout a new branch from `develop`.
2.b [Optional] Edit `Fastfile` to add `skip_commit: true` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/a6fad4f9869d221b51ddcc643761b88aaba65458/fastlane/Fastfile#L679) to avoid unintentional commits. 
2.c Run `bundle exec fastlane download_translations`.
2.d Verify that the lane fails because of the `TypographyDashes` error from the linter (it does as of today because some strings on GlotPress contains the wrong dash).
2.e [Optional]You may want to commit the state of the branch now, to make it easier to spot the differences after the fix.
2.f Update `Pluginfile` to use the HEAD commit of this branch. 
2.g Run `bundle exec fastlane download_translations` again.
2.h Verify that the lane succeeds.
2.i Verify that the only diff is related to the en-dash in numeric range.


spec/android_localize_helper_spec.rb
```diff
@@ -117,12 +117,52 @@ def generated_file(code)
...
+          include_examples 'en-dash substitutions', "/resources/string[@name='threat_fix_description']", '0-1', '0-1', '0–1'
+          include_examples 'en-dash substitutions', "/resources/string[@name='threat_fix_description_large']", '0 - 1', '0 - 1', '0 – 1'
+
+          # en-dash: don't substitute negative numbers
+          include_examples 'en-dash substitutions', "/resources/string[@name='field_allowed_values']", '1 -1 -2', '1 –1 –2', '1 -1 -2'
+
+          # en-dash: don't substitute checklists
+          include_examples 'en-dash substitutions', "/resources/string[@name='checklist_one']", '- 1.', '– 1.', '- 1.'
+          include_examples 'en-dash substitutions', "/resources/string[@name='checklist_two']", '- o', '– o', '- o'
         end
```
loremattei, line 166: There may be a lot more cases which can be tested, but I think these are the most relevant ones.

spec/test-data/translations/glotpress-download/expected/values/strings.xml
```diff
@@ -45,6 +45,29 @@
...
+        <item>To specify a range, enter the low value and high value separated by a dash. Example: 12.12.12.1–12.12.12.100</item>
+    </string-array>
+    <string name="threat_fix_description_large">Pricing ranges from $70 – 120/hour</string>
+    <string-array name="settings_jetpackdescription_large">
+        <item>You may mark an IP address (or series of addresses) as \"Always allowed\"</item>
+        <item>preventing them from ever being blocked by Jetpack. IPv4 and IPv6 are acceptable.</item>
+        <item>To specify a range, enter the low value and high value separated by a dash. Example: 12.12.12.1 – 12.12.12.100</item>
+    </string-array>
+    <string name="field_allowed_values">The field can assume these values: 4 3 2 1 -1 -2</string>
+    <string-array name="settings_number_series">
```
loremattei, line 63: I think all the previous strings in these files are strings which are actually used in some apps, but I haven't find good fits for some cases that I wanted to test, so I went with synthetic ones.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -299,6 +299,29 @@ def self.merge_xml_documents(all_xmls)
         #
         def self.apply_substitutions(tag)
           tag.content = tag.content.gsub('...', '…')
+
+          # Typography en-dash
+          if tag.content.include?('-')
+            matcher = tag.content.match(/.*(\d+\s*)-(\s*\d+).*/)
```
AliSoftware: `.match` will only match one occurrence even if the `tag.content` contains multiple ranges. Since the regex is not anchored (`^` and `$`) and also bc you used `.*` around the regex, this will greedily eat as much as it can via the first `.*` and return the _last_ range found in the string.

For example, notice how this only reports a single `MatchData`, and the single match found is for `3-4`, missing `1-2` entirely

```ruby
pry(main)> 'abcd 1-2 def 3-4 ghi'.match(/.*(\d+\s*)-(\s*\d+).*/)
=> #<MatchData "abcd 1-2 def 3-4 ghi" 1:"3" 2:"4">
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -299,6 +299,29 @@ def self.merge_xml_documents(all_xmls)
...
+            index = 1
+            while index < matcher.length do
+              # Make sure that if there is no space before digit there isn't
+              # one on the left either -- since we don't want to consider
+              # "1 2 -3" as a range from 2 to 3
+              isNegativeNumber = (matcher[index + 1][0] != ' ') && (matcher[index][matcher[index].length - 1] == ' ')
```
AliSoftware: ruby supports negative indexes, so `some_string[-1]` returns the last character already (and `some_string[-2]` the one before that, etc). No need to use `.length` to do the math yourself.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -299,6 +299,29 @@ def self.merge_xml_documents(all_xmls)
...
+              puts "Original: #{tag.content} - #{isNegativeNumber}"
+              tag.content = tag.content.gsub(
+                "#{matcher[index]}-#{matcher[index + 1]}",
+                "#{matcher[index]}\u2013#{matcher[index + 1]}"
+                ) unless isNegativeNumber
+              puts "Updated: #{tag.content}"
+
+              index = index + 2
+            end unless matcher.nil?
+          end
```
AliSoftware: Here `matcher.length` does not do what you think it does.
Instead `matcher` is a `MatchData` (the **first and only** match that `.match` returned in previous line) and `matcher.length` contains the number of capture groups. So on the string `'abc 1-2 def'`, `matcher[0]` is always the whole matched string, and `matcher[1]` is the `"1"` while `matcher[2]` is the `"2"` – because those are the substrings captured by the parentheses in your regex.

I think all this could be replaced by either using `scan` (to return all occurrences found using a regex instead of the first one like with `.match`), or looking at ruby doc to see if there is not a variant of `.match` that would return all the matches not the first one.

Another, maybe better, way shorter, and more elegant solution would be to use the block variant of `gsub`. Not tested but I think a code like this one could be enough to replace you whole `while` loop:

```ruby
tag.content.gsub(/(\d+\s*)-(\s*\d+)/) do |str|
  match = Regexp.last_match # of type `MatchData`. match[0] == str == whole match, match[1] = 1st capture group (left part of the range), match[2] = second capture group (right part of the range)
  isNegativeNumber = match[2][0] != ' ' && match[1][-1] == ' ' # if right part of range does not start with a space (e.g. `-3`), but left part of range does end with space, it's not a range after all but more likely a list containing negative numbers in it (e.g. `2 -3`)
  isNegativeNumber ? str : "#{match[1]}\u2013#{match[2]}"
end
```

spec/android_localize_helper_spec.rb
```diff
@@ -117,12 +117,52 @@ def generated_file(code)
...
+            end
+          end
+        end
+
         context 'with //string tags' do
-          include_examples 'substitutions', "/resources/string[@name='shipping_label_payments_saving_dialog_message']"
+          include_examples 'ellipsis substitutions', "/resources/string[@name='shipping_label_payments_saving_dialog_message']"
+
+          # en-dash: substitute ranges
+          include_examples 'en-dash substitutions', "/resources/string[@name='threat_fix_description']", '0-1', '0-1', '0–1'
```
AliSoftware: the `shared_examples 'en-dash substitutions'` seems to take 4 parameters, but here you pass 5?

spec/android_localize_helper_spec.rb
AliSoftware: I think it's worth adding cases where:
 - we have multiple ranges in a single string (e.g. `abc 1-2 def 3-4 ghi $5-$6 jkl'`)
 - we have ranges that are not just numbers, but numbers with currency symbols too (e.g. `$70-$120` – which is the case that we stubbled upon in WP iinm – and `70€-120€`)

spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
```diff
@@ -30,4 +30,27 @@ Language: fr
...
+    <string-array name="settings_jetpackdescription">
+        <item>Vous pouvez marquer une adresse IP (ou une série d’adresses IP) comme « Toujours autorisé »</item>
+        <item>pour empêcher leur blocage par Jetpack. IPv4 et IPv6 sont acceptables.</item>
+        <item>Pour spécifier une plage, entrez la valeur basse et la valeur haute en les séparant par un tiret. Exemple : 12.12.12.0–12.12.12.100</item>
+    </string-array>
+    <string name="threat_fix_description_large">Les prix vont de 70 – 120 $ par heure</string>
+    <string-array name="settings_jetpackdescription_large">
+        <item>Vous pouvez marquer une adresse IP (ou une série d’adresses IP) comme « Toujours autorisé »</item>
+        <item>pour empêcher leur blocage par Jetpack. IPv4 et IPv6 sont acceptables.</item>
+        <item>Pour spécifier une plage, entrez la valeur basse et la valeur haute en les séparant par un tiret. Exemple : 12.12.12.0 – 12.12.12.100</item>
```
AliSoftware, line 43: 👌  nice case having an example with IP address ranges 👍

spec/test-data/translations/glotpress-download/expected/values/strings.xml
```diff
@@ -45,6 +45,29 @@
...
-    
+
     <string name="shipping_label_payments_cant_edit_warning">Only the site owner can manage the shipping label payment methods. Please contact Store Owner %1$s (%2$s) to manage payment methods.</string>
+    <string name="threat_fix_description">Pricing ranges from $70–120/hour</string>
+    <string-array name="settings_jetpackdescription">
+        <item>You may mark an IP address (or series of addresses) as \"Always allowed\"</item>
+        <item>preventing them from ever being blocked by Jetpack. IPv4 and IPv6 are acceptable.</item>
+        <item>To specify a range, enter the low value and high value separated by a dash. Example: 12.12.12.1–12.12.12.100</item>
+    </string-array>
+    <string name="threat_fix_description_large">Pricing ranges from $70 – 120/hour</string>
```
AliSoftware, line 56: What about if the string is `$70 - $120` instead of `$70 - 120`? worth adding a test case for that imho

spec/test-data/translations/glotpress-download/expected/values/strings.xml
```diff
@@ -45,6 +45,29 @@
...
+    <string-array name="settings_number_series">
+        <item>The field can assume these values: 4 3 2 1 -1 -2</item>
+        <item>Please select your choice</item>
+    </string-array>
+    <string name="checklist_one"> - 1. First point</string>
+    <string name="checklist_two"> - o Two. Second point</string>
+    <string-array name="checklist_array">
+        <item> - 1. First point</item>
+        <item> - o Two. Second point</item>
+    </string-array>
```
AliSoftware, line 72: Nice to have thought about the use cases of checklists 👍

AliSoftware (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
loremattei: Ah! Good catch! It only works when there's one dash in the string.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
loremattei: TIL! Cool!

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
loremattei: > Instead matcher is a MatchData (the first and only match that .match returned in previous line) and matcher.length contains the number of capture groups. So on the string 'abc 1-2 def', matcher[0] is always the whole matched string, and matcher[1] is the "1" while matcher[2] is the "2" – because those are the substrings captured by the parentheses in your regex.

Yep, this is exactly what I expected 🤔 👍

I love your shorter and more elegant solution, but unluckily it doesn't work. I think that it doesn't catch the last dash in the string, so it basically always fails when there's only one dash (probably the most frequent case).

spec/android_localize_helper_spec.rb
loremattei: Sorry, I don't understand 🤔 Are you sure?

AliSoftware: Or right, sorry, I counted the name of the example (`en-dash substitutions`) as if it was one of the parameters passed to `|xpath, fixture_block, gp_block, final_block|` param list 🤦  nvm

AliSoftware: Though, it looks (at least from GitHub) that the last 3 parameters always have the same value each… Is that because from GitHub (and the font they use) the difference between a hyphen and an em-dash is not super visible? If so, would it make sense to use `\u2013` in the parameters that use an em-dash in those examples, to make those test cases more readable?

e.g. something like this (untested)?
```
include_examples 'en-dash substitutions', "/resources/string[@name='threat_fix_description']", '0-1', '0-1', "0\u{2013}1"
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
loremattei: Sorry, I was wrong. The problem was with the assignment. I thought `gsub!` would do the substitutions in the original `tag.content`, though it doesn't seem to be the case.

AliSoftware: > Yep, this is exactly what I expected 🤔 👍

Ah? Then it seems a bit confusing because you start your look with `index = 1` on line 306 and then, given the Regexp you use, `MatchData#length` will always be 2 (number of capture groups), so basically your loop is reading `while index < 2` with an index starting at 1, meaning it will only ever loop once… so not sure I see what the intent was for the loop to be there then 😅 

> but unluckily it doesn't work. I think that it doesn't catch the last dash in the string, so it basically always fails when there's only one dash

Are you sure? I tested it with `pry`/`irb` and it worked as expected. (In the below example I've replaced `\u2013` with `…` during my test to make it more visible/readable when a substitution occurred, because a hyphen and an em dash can look pretty similar in the terminal and making it harder to see, but the logic still applies)

```
$ pry

[1] pry(main)> def checksubst(tag_content)
[1] pry(main)*   tag_content.gsub(/(\d+\s*)-(\s*\d+)/) do |str|
[1] pry(main)*     match = Regexp.last_match
[1] pry(main)*     isNegativeNumber = match[2][0] != ' ' && match[1][-1] == ' '
[1] pry(main)*     isNegativeNumber ? str : "#{match[1]}…#{match[2]}"
[1] pry(main)*   end  
[1] pry(main)* end  
=> :checksubst
[2] pry(main)> checksubst('Hello world')
=> "Hello world"
[3] pry(main)> checksubst('Hello 1-2 world')
=> "Hello 1…2 world"
[4] pry(main)> checksubst('Hello 1-2')
=> "Hello 1…2"
[5] pry(main)> checksubst('Price is $70-120')
=> "Price is $70…120"
[6] pry(main)> checksubst('Values can be 0 -1 -2')
=> "Values can be 0 -1 -2"
```
So seems to match even cases with one range, and even if that range is at the end…?

loremattei: Yep, see my previous comment :-)

AliSoftware: Ah, right! 👌 
So `tag.content = tag.content.gsub(…)` should work then :)

loremattei: Yes, it works. But isn't `gsub!` supposed to substitute in the original variable? 🤔

AliSoftware: Yep, it should indeed 🧐  not sure why it doesn't 🤔

spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
AliSoftware: 🤯  I just realized that this specific string might be a case where, even if the em-dash would indeed be the correct one to use typographically speaking, and the right replacement to do here according to the linter blind rules… for a case like this, **the original intent of using a hyphen might be intentional** here.

This string basically says: `"To specify a range, enter a low value and high value by separating them with a hyphen. Example: 12.12.12.0 - 12.12.12.100"`.

 * The instructions that this string gives to the user explicitly uses "un tiret" (a hyphen) and not "an em-dash" in their copy/wording
 * This looks like some help text for an input field, and the user is likely expected to use a simple hyphen (especially when typing from their phone when most people don't know how to type an em-dash from their Android or iOS keyboard) when filling that input field – value that is probably then sent to a server, which might itself not accept values with em-dash instead of hyphens when parsing the IP range…)
 * If that's the case, that might also be even more confusing for the user if they try to copy/paste the example range provided (which uses an em-dash after our substitution) and have that example be rejected by the server response expecting a hyphen… 😓 

All that to say: should we introduce an attribute like `typography_autofix='false'` that could be set on `<item>` and `<string>` (like we have for `translatable='false'` and friends) to allow developer to opt out of this automatic substitutions for the strings for which it might cause trouble?

_We could then  `return if tag['typography_autofix'] == 'true'` at the start of `apply_substitutions` to exit early, and add that `typography_autofix` attribute to the `attributes_to_copy` list so it gets reported to translations downloaded from GlotPress when present on the original XML._

Not sure then how to tell the Android Linter to ignore the typography rule on that particular entry when linting though? 🤔

spec/android_localize_helper_spec.rb
```diff
@@ -117,12 +117,66 @@ def generated_file(code)
...
+              expect(gp_node.content).to include(fixture_block)
+            end
+          end
+
+          TEST_LOCALES_MAP.each do |h|
+            it "has the text substituted in #{h[:android]}" do
+              final_xml = File.open(generated_file(h[:android])) { |f| Nokogiri::XML(f, nil, Encoding::UTF_8.to_s) }
+              final_node = final_xml.xpath(xpath).first
+              if gp_block.is_a?(Array)
+                gp_block.each_with_index do | gp_val, index |
```
loremattei: Ideally, we should check that both gp_block and final_block are arrays and they have the same size... but since we can only fail in any case, I opted to keep it simple.

spec/android_localize_helper_spec.rb
loremattei: I added a test case for the first situation. 
I don't think the second one works at all with the current regex and since it's the same rule used by the linter, I'd opt to ship this version and iterate in the future if the need will arise.

spec/test-data/translations/glotpress-download/expected/values-fr/strings.xml
loremattei: Yep, good point. Since the Android linter doesn't cover this case though, I think we can live with this version and iterate in the future if required. 
I mean, while your suggestion about the attribute is good and, on the linter side, we could add the strings to the baseline (though, IIRC, this should be done for every language, so it's not a great solution), it's pretty easy to get side tracked and end up doing something very complex :-) For example: then, how do we handle the case where there are two ranges in a string and only one should be substituted? 😅

loremattei:
@AliSoftware this should be ready for another pass! 🙇‍♂️

AliSoftware (approved):

PR #269, opened 2021-05-27
Add toolkit-autoupdate action

AliSoftware:
## Why?

Now that the release toolkit uses semver, I started to [introduce an auto-update mechanism for it in the woocommerce repositories](https://github.com/woocommerce/woocommerce-ios/pull/4288/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R20-R32).

As the check logic implemented in WC's Fastfile still took a couple of lines, @oguzkocer [rightfully pointed out](https://github.com/woocommerce/woocommerce-android/pull/4089#discussion_r639923075) that we would benefit from making this into an action, for other apps to be able to benefit it. That's what this PR is all about.

## How

Inspired by [fastlane's own action to auto-update fastlane itself](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/update_fastlane.rb), it checks for the locally-installed version and if there is a more recent update was available.

If there is one – be it a breaking change or not – it warns you about the latest version available, then propose to run `bundle update` for you. If you say yes, it will run `bundle update` then stop the lane – both in order to let you commit the changes and to force you to restart your fastlane invocation after the update so that fastlane picks up the updated plugin.

## To Test

This is gonna be tricky to test, because this is gonna be the first version of the toolkit that is gonna introduce this action, so it's not like we could point to an older version and test that it detects a new one…

The best thing I was able to do is to temporarily tweak the action to check for another gem instead:

 * Temporarily modify the `TOOLKIT_SPEC_NAME` constant in that branch, to be another random gem we use in our app repos (e.g. `fastlane`, `nokogiri`, `fastlane-plugin-appcenter`, …)
 * Commit and push the changes on that branch
 * In an app repo, edit the `Pluginfile` to point the toolkit to `git: …, branch: 'action/autoupdate'`, and run `bundle install`
 * Add a call to `check_for_toolkit_updates unless is_ci` in the `before_all` block of your `Fastfile`
 * Ensure that the dummy gem we chose on step 1 for this test is not yet on the latest version in that app repo. If necessary, update your `Gemfile` to temporarily force installing an older version of that dummy gem instead.
 * Run an arbitrary (and ideally side-effects-free) lane from the app repo (e.g. `bundle exec fastlane get_pullrequests_list` or `bundle exec fastlane download_promo_strings`), and check that the `check_for_toolkit_updates` action gets called.
 * Check that it warns you about the dummy gem being out of date. Refuse to update when prompted, and see it continue with the lane
 * Run the same lane again, and this time accept to run bundle update when prompted. See it update your `Gemfile.lock` and stop the lane, telling you to commit and restart fastlane.
 * Run the same lane one more time, and ensure that this time there is no new update found (providing the version constraint for the dummy gem you used in `Gemfile` wasn't too restrictive to get that latest version)
 * Revert the commit you made on step 1-2 (the temporary change to `TOOLKIT_SPEC_NAME`)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
```diff
@@ -0,0 +1,77 @@
...
+          UI.success('Your release toolkit is up-to-date! ✅')
+          return nil
+        end
+
+        # Return type of which_to_update differs before/after RubyGems 3.1.0, so normalize to always use Gem::NameTuple
+        updates_needed = Gem::NameTuple.from_list(updates_needed)
+        latest_version = updates_needed.find { |gem_info| gem_info.name == TOOLKIT_SPEC_NAME }.version
+
+        UI.message(['There is a new version '.yellow, latest_version.to_s.red, ' of the release toolkit!'.yellow].join)
+        non_breaking_update = Gem::Requirement.new(local_version.approximate_recommendation).satisfied_by?(latest_version)
```
AliSoftware: `Gem::Version#approximate_recommendation` will return something like `"~> 1.0"` (i.e. `~> <major>.<minor>`) for a `local_version` with a value of `1.0.1`.

This thus allows us to check if the `Gem::Requirement` built from that `~> x.y` will be satisfied by the latest version. If `latest_version` doesn't satisfy the current `~> x.y` constraint, that means that it is a major version bump, aka breaking change. SemVer magic! ✨

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
```diff
@@ -0,0 +1,77 @@
...
+        UI.important('The latest version introduces breaking changes!') unless non_breaking_update
```
AliSoftware: Note that even if the `latest_version` is incompatible with the `local_version` (breaking changes), we will still suggest to run `bundle update` – to at least get the latest non-breaking update, aka latest version taht is still compatible with the `~> x.y` requirement we typically use in our `Pluginfile` and that is still safe to update to.

In those cases, after `bundle update` has run, we would still not be on the very _latest_ version (so running the check again will still warn that we're not on the latest). This is expected, because we don't want to accidentally auto-update to next major and pull breaking-changes in during that auto-update.

Instead, if the latest version is a breaking change, it's up to you to manually go modify your `Pluginfile` to willingly and explicitly update the `~> 1.0` to `~> 2.0`, then make the appropriate changes in your `Fastfile` according to the breaking changes described in the toolkit's CHANGELOG to fix anything that would otherwise break, then run `bundle install` and commit all the changes associated with that major version bump.

That doesn't mean we shouldn't run `bundle update` to at least auto-update to the latest still-compatible version; which is why we don't exit after this line, even if breaking update is detected by this test.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
```diff
@@ -0,0 +1,77 @@
...
+require 'rubygems/command_manager'
+
+module Fastlane
+  module Actions
+    class CheckForToolkitUpdatesAction < Action
+      TOOLKIT_SPEC_NAME = 'fastlane-plugin-wpmreleasetoolkit'.freeze
+
+      def self.run(params)
+        updater = Gem::CommandManager.instance[:update]
+        installed_gems = updater.highest_installed_gems.select { |spec| spec == TOOLKIT_SPEC_NAME }
```
AliSoftware, line 11: Note that when run in the context of `bundle exec`, this will already rightfully use the rubygems environment provided by Bundler and specific to the current app's repo. Especially it will only check for the latest installed gems in `vendor/bundler/…` (the location provided by `bundle exec gem env gemdir`, so the gem dir provided by Bundler's context, not including any system-wide installed gems), which is exactly what we want.

oguzkocer:
Just wanted to say that I've skimmed through this, but it doesn't look like I'll have time to properly review it until after the Simplenote code freezes, which I have no idea how long will take.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
```diff
@@ -0,0 +1,77 @@
+require 'fastlane/action'
+require 'rubygems/command_manager'
+
+module Fastlane
+  module Actions
+    class CheckForToolkitUpdatesAction < Action
+      TOOLKIT_SPEC_NAME = 'fastlane-plugin-wpmreleasetoolkit'.freeze
+
+      def self.run(params)
```
oguzkocer, line 9: Would we also want a `no_update` option similar to [fastlane's own update](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/update_fastlane.rb#L11)? Even if we wouldn't want to pass this in most cases, maybe it'll help when/if there is a bug in this updater and we don't want to change the release toolkit during a code freeze or release, what do you think?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
```diff
@@ -0,0 +1,77 @@
...
+      def self.description
+        'Check that we are on the latest version of the release toolkit, and propose to update if not'
+      end
+
+      def self.authors
+        ['Olivier Halligon']
+      end
+
+      def self.return_value
+        'Returns the latest version of the toolkit available if your toolkit is not up-to-date, or nil if you are already up-to-date.'
```
oguzkocer, line 67: Unless I am misunderstaning the [return statement](https://github.com/wordpress-mobile/release-toolkit/pull/269/files?file-filters%5B%5D=.md&file-filters%5B%5D=.rb#diff-3a08f262f19672eb0b3baa7c9336cb4ec676e8762a6f2905e6f9d8e831b9115aR30) it doesn't look like it's returning anything if `skip_update_suggestion` param is `false`. It also looks like it's returning the latest version value even if toolkit is up to date. Not sure it's worth it, but should we clarify those and maybe say:

```suggestion
        'Returns the latest version of the toolkit available if update is skipped, otherwise returns nil'
```

Or a better phrasing of this ^

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
oguzkocer: @AliSoftware I am a little confused by what the `latest_version` value will be if the installed version is `1.xx` and local version is `2.yy`. Will it be `1.aa` where `aa > xx`, so a non-breaking update, or will it be `2.yy`? Your comments suggest it'll be a minor bump, but I can't figure out why that's the case since [which_to_update seems to be finding the `highest_remote_version`](https://ruby-doc.org/stdlib-2.6.6/libdoc/rubygems/rdoc/Gem/Commands/UpdateCommand.html#method-i-which_to_update). (no idea what that actually means in this context)

I'd appreciate it if you can help me understand that. Thanks!

oguzkocer (commented):
@AliSoftware This looks great. I've left couple minor suggestions and a question for you. I think this is good to go, but I just want to make sure about that question first.

oguzkocer:
The testing steps are ... 😨 - @AliSoftware if you've gone through those, I think that's good enough for this PR because I am more likely to break something during those steps than your PR breaking the release toolkit 😂 Let me know if you disagree and I'll give it a shot.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
AliSoftware: So this action does 2 things:
 - Check if there's an update available, and shows an informative message of the outcome (new version found or not)
 - Suggest to update, prompt us if we want to, and if we accept, run `bundle update`

The second step is gated by the `skip_update_suggestion` parameter.
If we want to add a param to gate the first step (and thus also the second step since it doesn't make sense to do the second step without the first, it's either 1+2, or just 1, or nothing), it's easier to just not call the action at all.

So for these cases, I suggest to use something like that in the client app's Fastfile:
```ruby
check_for_toolkit_updates(…) unless ENV['TOOLKIT_SKIP_UPDATE_CHECK']
```

Which is actually how Fastlane does it via their `FASTLANE_SKIP_UPDATE_CHECK` env var, like [here](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/update_checker/update_checker.rb#L14)

AliSoftware: As for the `params[:no_update]` that Fastlane's own code is using, it's a little bit different. If you look at the documentation for that param and how it's used in their action, this param is not intended to be used by the users directly, but instead is used internally so that when Fastlane auto-restart itself after a successful update – by calling `exec(…)` [here](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/update_fastlane.rb#L82) to spawn the same process again, they don't redo a check on the second run that gets automatically triggered there.

In my action I chose not to go the `exec` route to auto-restart `fastlane` after the update, because I preferred to let us check the changes to the `Gemfile.lock` – and also give the possibility to make additional changes to our `Fastfile` in case we wanted to bump to a next major version with breaking changes – before restarting manually.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
AliSoftware: There's a `return nil` earlier in the code [on line 19](https://github.com/wordpress-mobile/release-toolkit/pull/269/files#diff-3a08f262f19672eb0b3baa7c9336cb4ec676e8762a6f2905e6f9d8e831b9115aR19) if there was no update found tho.

So:
 - If NO update is found (i.e. we are up to date), line 19 returns nil.
 - If an update is found but we don't update (either because we have skip_update_suggestion set, or because we were suggested but replied no to the `UI.confirm` prompt), line 30 returns the new version
 - If an update is found but we DO update, we end fastlane – with an `abort_with_message!` – after the `bundle update` is done and the abort message suggest the user to restart fastlane. Since we abort, there's no concept of returned value there.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
oguzkocer: That all makes a lot of sense, thanks a lot for the detailed explanation!

I am sorry for my lack of knowledge & experience on Ruby/fastlane front and I hope you don't mind these questions (or possibly incorrect) suggestions.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
oguzkocer: 🤦

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
AliSoftware: No no you understood it correctly, `which_to_update` returns the highest version available on RubyGems – regardless of semver, even if that's a major version. So just the very latest version of the toolkit we published.

So if we have currently `1.xx` installed but the latest version we released of the toolkit is `2.yy`, then `latest_version` is gonna be `2.yy`, not `1.aa`.

This is how I detect if there are breaking changes between the currently `local_version` installed version and the `latest_version`, because line 27 will see that `local_version` being `1.xx` is not compatible with `latest_version` which is `2.yy` (or more especially, that `local_version.approximate_recommendation`, aka `"~> 1.xx"` requirement, won't be satisfied by a `2.yy` version since that's not semver-compatible with `~>1.x`)

What might be confusing tho is the wording of the message maybe?

To clarify, here is what is supposed to be printed in the terminal if we are on `1.2.3` and the latest version available is `2.1.4`:
```
Currently using release toolkit version 1.2.3
Checking for updates now...
There is a new version 2.1.4 of the release toolkit!
The latest version introduces breaking changes!
Do you want to run bundle update now? y
$ bundle update fastlane-plugin-wpmreleasetoolkit
[…long log of bundle update there...]
fastlane-plugin-wpmreleasetoolkit have been updated. Please check and commit the changes in your Gemfile.lock file, then restart your previous invocation of fastlane to use the new toolkit.
```
BUT after those messages, the toolkit would have only been updated to something like `1.3.7` – the latest `1.x` version before the first `2.0.0`, because `bundle update` will respect the version constraint you put in your `Gemfile` (well, in the `Pluginfile` that is loaded from the `Gemfile`), which will supposedly be `'~> 1.0'` if we want to respect semver (using `~> x.y` in your `Gemfile` is the typical way to avoid accidentally bumping to a breaking version during `bundle update`, as it allows any version x.z where z>y but won't allow changing the x major number)

AliSoftware: I guess one thing that I could maybe try is to make the last message include the version it has been updating to:
```
fastlane-plugin-wpmreleasetoolkit have been updated to 1.3.7. Please check and commit the changes in your Gemfile.lock file, then restart your previous invocation of fastlane to use the new toolkit.
```
That version that `bundle update` would update to would depend on what you have in your `Pluginfile`, but most likely (since you're supposed to have `"~> x.y"` type of requirement there) it will be either the latest version if that latest version was `x.z` and not requiring a major version jump, or to an earlier version `x.z` if the latest version `u.v` was crossing the major version bump boundary (u>x).

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
AliSoftware: I don't mind at all, and please keep asking questions, that's how we learn!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
oguzkocer: Thanks a lot for the explanation, it matches my expectations of what the code does. As you pointed out what is confusing to me is the messaging around it. Specifically, from your example:
```
Currently using release toolkit version 1.2.3
Checking for updates now...
There is a new version 2.1.4 of the release toolkit!
The latest version introduces breaking changes!
Do you want to run bundle update now?
```

When I am answering this question, it feels like I'll be updating to `2.14` because that's the information I've been presented so far. Especially for experienced Ruby developers, this is probably not an issue, but I think it'd help to clarify the messaging a bit for people who have less experience with bundler. It'll also not require the user to know whether the dependency is setup as `~> x.y` or not. How about something like this:

```
There is a new version 2.1.4 of the release toolkit!
The latest version introduces breaking changes!
Running bundle update will bump the version to 1.3.7, would you like to do it now?
```

I assume that this will make it obvious that a major bump needs to be manually handled, but if we want to include that in the messaging, we can. It's hard to balance between being too verbose and not providing enough info.

I'll go ahead and approve this and will leave the decision of how best to move forward to you. Thanks again!

oguzkocer (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb
AliSoftware: Yep and tbh that's also the message I wanted to print, but unfortunately there's no easy way to know what bundler will update the version to (1.3.7) before actually running it.

I mean there surely are ways to figure it out but that will not be as trivial as reading a property that would give us that value, but instead require us to run `bundle outdated` and have bundler fetch all the available versions from the server, then solve the dependency graph according to what is set in the Gemfile and the Pluginfile to determine which version is compatible with the constraint set there, to finally get that value. All the things (contacting server, resolving dev graph…) that `bundle update` will then redo when we finally invoke it to do the _actual_ update.

I'll see if we can get a better guess using the Updater class build in fastlane's codebase to get a better guess at which version it might update to — eg by assuming the Pluginfile will always be "~> x.y" and by assuming the dependency graph will never have another gem requirement that is a dependency of the toolkit blocking the the update, etc, so will still be a guess — or if not, if I can improve the message to make it clearer that bundle update might not update to the very latest if breaking changes are involved.

AliSoftware: Turns out even `bundle outdated` doesn't tell you which version it is going to update to, but instead tells you the latest available version regardless of the dependency graph constraints too.

For example if I run `bundle outdated` on WPiOS now, I get amongst other things:
```
$ bundle outdated --parseable

activesupport (newest 6.1.3.2, installed 5.2.6)
bigdecimal (newest 3.0.2, installed 1.4.4)
fastimage (newest 2.2.4, installed 2.2.3)
…
```
And you can see that `bigdecimal` is listed to have a newest version of 3.0.2. But if I run `bundle update bigdecimal`, it doesn't update it at all – because in our context, `bigdecimal` comes into our bundle by being a dependency on the `fastlane-plugin-wpmreleasetoolkit`, and that dependency is constrained to be `~> 1.4` in the toolkit's `gemspec`, so `bundle update` won't update to anything greated than `1.x` with `x>4`, and indeed `1.4.4` is the latest of the `1.x` version for that gem:
```
$ bundle update bigdecimal
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
…
Bundler attempted to update bigdecimal but its version stayed the same
Bundle updated!
```

---

So even `bundle outdated` doesn't know which version it is gonna update to until you actually run `bundle update` and it does the whole "resolving dependency graph" dance (and resolving a dependency graph is not always trivial indeed!). Meaning that there is apparently no easy way to determine what version will be installed before running bundle update.

AliSoftware:
@oguzkocer I've improved the wording for the case of latest version being a breaking change, to give lengthy details in the message instead.
Here's what it looks like in case of new major version – after faking current version to be 1.4.4 and latest released version to be 3.0.2:

```
[13:50:17]: ---------------------------------------
[13:50:17]: --- Step: check_for_toolkit_updates ---
[13:50:17]: ---------------------------------------
[13:50:17]: Currently using release toolkit version 1.4.4.
[13:50:17]: Checking for updates now...
[13:50:29]: There is a newest version 3.0.2 of the release toolkit!
[13:50:29]: The latest version available (3.0.2) introduces breaking changes compared to the 1.4.4 you are currently using.

 - To update to 3.0.2, first edit your `Pluginfile` to use '~> 3.0', run `bundle update`,
   then be sure to make all the necessary changes to your `Fastfile` (see the toolkit's CHANGELOG)
   to take those breaking changes into account.

 - If you are not ready to make the major version bump, you can still try to update to the latest compatible,
   non-breaking version by running `bundle update` now. This will not update to the latest 3.0.2, but
   might still update to a newer version compatible with '~> 1.4' if one exists; which is still valuable
   to at least get bugfixes, until you are ready to jump to the next major version later.

[13:50:29]: Do you want to run bundle update now? (y/n)
```

Better?

oguzkocer:
@AliSoftware Thanks a lot for the changes - that messaging looks very(!) clear, really appreciate you doing that 🙇

PR #270, opened 2021-06-02
Fix new rubocop violations

AliSoftware:
### Why?

We recently fixed the issue with Danger not running as part of the GitHub Action (due to nokogiri version issue); part of that fix involved updating `rubocop` from `0.75.0` to `1.15.0`.

But that means that (1) we have introduced some violations while our code wasn't linted by Danger, and (2) new cops were introduced by the newer version of `rubocop`, detecting new violations. Both those cases are what this PR is fixing.

### Scope

**This PR does not intend to fix _all_ the existing violations of the release-toolkit.**

Instead, the scope of that PR only **focuses on making running `bundle exec rubocop`** on the current state of the codebase **not generate any violation** – ignoring some known existing violations for the moment if appropriate (via the `.rubocop_todo.yml` file) – so that any _future_ changes we do on the codebase going forward will be free of _new_ violations.

There are still a lot of _existing_ violations that we will want to get fixed at some point in the future, but don't have the bandwidth to address yet – because fixing them would require some refactoring or breaking changes that will take some time to get it right. Those violations are listed in `.rubocop_todo.yml`, with the intent to get to fix them later when we have the bandwith for it; while this PR only focus on fixing the less risky/involved ones.

### Review

I suggest you to review commit-per-commit – instead of looking at the whole diff directly, since I splits commits to fix one rule per commit, which should make it easier to reason about the changes involved.

Note that most of those changes were made by `bundle exec rubocop --auto-correct --only <RuleName>`.
The only changes that were made manually are:
 - Changes to the `.rubocop.yml` config in d13a848752b6699b0fba759f3b5dde82370c8ca3, 2e1887ec5df513db9e860c04f460d147b3f83a5a, 61064fd4731916e7c82365ca391795173446fea9
 - ba2e1985d8a24271279b33beefff46796194660a (done via auto-correct but I didn't commit all the suggested changes as some were incorrect when involving `Pathname` concatenation; hence also disabling the rule – see comment in `.rubocop.yml`)
 - 4ac9ec72592b792fc8cb7ebb82d715660c98a77d (not really related to rubocop, but since I ran the tests between rubocop fixes to ensure they didn't break, figured it was worth using the occasion to improve this)
 - 444e4a63a898706832e087d49ff7f70f6bc6a231 and 3e0c434725d01d8d4a1f181272f81cc9b43ad0a2 (non-auto-correctable, manual fix)


oguzkocer (approved):
Thanks for these fixes @AliSoftware! I've gone through the PR review commit by commit as you suggested, checking out the manual changes first in more detail and skimming through every commit afterwards. I've also ran `bundle exec rubocop`:

> 119 files inspected, no offenses detected

Although I am not a Ruby developer, all the changes - combined with your comments and commit messages - make sense to me. It also helps that this is a self verifying type of PR, so it's a 👍 from me.

PR #275, opened 2021-06-10
Update how Android app versions are managed

JavonDavis:
Supercedes https://github.com/wordpress-mobile/release-toolkit/pull/258 due to issue with Github actions not re kicking off 
----- 

This PR introduces changes that were primarily needed to support the new Jetpack app in the WPAndroid repo but are more focused on updating how versions are managed in our current and future Android apps. The proposals here are: 

* The use of a new app parameter/environment variable to add an additional context when doing versioning. This is a new parameter that can be passed as an option to the commands related to versioning, this value can can also be set as an environment variable for projects that don't have multiple 'apps' within them. 
* It also proposes the use of a new properties file in the project repos to manage versioning, a version.properties file that follows the following key-value format

```
appName.versionName=<version name>
appName.versionCode=<version code>

appName.alpha.versionName=<version name>
appName.alpha.versionCode=<version code>
```

Example here: https://github.com/wordpress-mobile/WordPress-Android/blob/fe4352dbdf2d8b56f760ceeb6a0bd591883b23e4/version.properties

With accompanying changes in the `build.gradle` file to read the versions from this file instead of being defined in the `build.gradle`. Something like so -> 

```
def versionProperties = new Properties()
    file("../version.properties").withInputStream { versionProperties.load(it) }

    compileSdkVersion rootProject.compileSdkVersion

    defaultConfig {
	@@ -60,9 +63,10 @@ android {
        if (project.hasProperty("versionName")) {
            versionName project.property("versionName")
        } else {
            versionName versionProperties.getProperty("wordpress.alpha.versionName")
        }

        versionCode versionProperties.getProperty("wordpress.alpha.versionCode").toInteger()
```

* This PR also proposes moving the logic for updating the version into a gradle task which can be a bit more stable than our previous approach, see proposed gradle task currently used in https://github.com/wordpress-mobile/WordPress-Android/pull/14613... The PR assumes this task exists and relies on it to update the `version.properties` file 

```
task "updateVersionProperties" {
    doLast {
        ant.propertyfile(
                file: "../version.properties") {
            entry( key: "$key", value: "$value")
        }
    }
}
```


JavonDavis:
Closing since Github actions issue still present here

PR #276, opened 2021-06-16
Fix a bug in the `plugin_root` method

jkmassel:
When generating promo screenshots, the error `Unable to determine the plugin root directory` would come up – this fixes it by better determining the location of the plugin root in production environments (where the .gemspec file would not be present in `fastlane-plugin-wpmreleasetoolkit`.

**To Test:**
Run https://github.com/wordpress-mobile/WordPress-iOS/pull/16571 before and after this PR (don't bother with any of the setup steps – just checking out the branch, updating the reference to the release toolkit to point at this branch, and running `bundle exec fastlane create_jetpack_promo_screenshots` will expose the error)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,12 @@ commands:
           steps:
             - run:
                 name: Install Homebrew dependencies, if neeeded
-                command: brew update && xargs brew install --verbose < .circleci/.brewfile
+                command: |
+                  # Because the CircleCI image uses shallow clones, we need to unshallow them first. See https://bit.ly/3vx6EAL
+                  git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core fetch --unshallow
+                  git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask fetch --unshallow
+
+                  brew update && xargs brew install --verbose < .circleci/.brewfile
       - run:
           name: Install Ruby dependencies, if neeeded
           command: bundle check --path vendor/bundle || bundle install --with screenshots
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb
@@ -36,7 +36,8 @@ module Fastlane
         while continue
           child_filenames = dir.children.map! { |x| File.basename(x) }
 
-          if child_filenames.include? 'fastlane-plugin-wpmreleasetoolkit.gemspec'
+          # The first option is for development (where the .gemspec is present in the project root), the second is for production (where it is not)
+          if child_filenames.include?('fastlane-plugin-wpmreleasetoolkit.gemspec') || File.basename(dir).include?('fastlane-plugin-wpmreleasetoolkit')
             continue = false
           else
             dir = dir.parent
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb
@@ -38,7 +38,7 @@ module Fastlane
 
           # The first case is for development – where the `.gemspec` is present in the project root
           # The second case is for production – where there's no `.gemspec`, but the root dir of the plugin is named `fastlane-plugin-wpmreleasetoolkit-{version}`.
-          if child_filenames.include?('fastlane-plugin-wpmreleasetoolkit.gemspec') || File.basename(dir).start_with?('fastlane-plugin-wpmreleasetoolkit')
+          if child_filenames.include?('fastlane-plugin-wpmreleasetoolkit.gemspec') || File.basename(dir).start_with?('fastlane-plugin-wpmreleasetoolkit-')
             continue = false
           else
             dir = dir.parent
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb
```diff
@@ -36,7 +36,8 @@ def self.plugin_root
         while continue
           child_filenames = dir.children.map! { |x| File.basename(x) }
 
-          if child_filenames.include? 'fastlane-plugin-wpmreleasetoolkit.gemspec'
+          # The first option is for development (where the .gemspec is present in the project root), the second is for production (where it is not)
+          if child_filenames.include?('fastlane-plugin-wpmreleasetoolkit.gemspec') || File.basename(dir).include?('fastlane-plugin-wpmreleasetoolkit')
```
AliSoftware: Alternatively, you could make `File.basename` on line 37 remove the file extension by using `dir.children.map { |x| File.basename(x, '.gemspec') }`, then this line would simplify to just this:
```suggestion
          # The first option is for development (where the .gemspec is present in the project root), the second is for production (where it is not)
          if child_filenames.include?('fastlane-plugin-wpmreleasetoolkit')
```


[EDIT] Actually that might lead to different results, as the first one exists if the dir contains the gemspec, while the other exits if the current dir is `fastlane-plugin-wpmreleasetoolkit`… while my suggested modification makes it return the parent of that dir instead, changing the behavior. So maybe not a good suggestion after all.

AliSoftware: This made me realize that this code as well as the one for `project_root` both use `map!` (which is the version that mutates-in-place) instead of `map`. It doesn't really makes sense to mutate `dir.children` in place especially because this is the result of a function / property, and not a variable, so I'm actually surprised that Ruby doesn't crash telling you you're trying to mutate a constant value there on lines 37 and 18…

AliSoftware (commented):
LGTM, but left a comment that I think would simplify the code even more, so I'll let you test it before approving ;)

.circleci/config.yml
```diff
@@ -21,7 +21,12 @@ commands:
           steps:
             - run:
                 name: Install Homebrew dependencies, if neeeded
-                command: brew update && xargs brew install --verbose < .circleci/.brewfile
+                command: |
+                  # Because the CircleCI image uses shallow clones, we need to unshallow them first. See https://bit.ly/3vx6EAL
+                  git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core fetch --unshallow
```
mokagio: Totally unrelated: _I just read `Taps` as "tapas"! Gosh, I must be hungry_!

lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb
AliSoftware: Ok I think what confused me during last review (see EDIT in comment above) is that you use `include?` to compare the `File.basename(dir)` and the expected dir name. Like if `File.basename(dir)` was returning an array or list of files, while it's returning a `String`. So your current code would also yield true if the dir was named `fastlane-plugin-wpmreleasetoolkit-backup` as `include?` would match that name. Probably better to use a simple `==` here.
```suggestion
          # The first option is for development (where the .gemspec is present in the project root), the second is for production (where it is not)
          if child_filenames.include?('fastlane-plugin-wpmreleasetoolkit.gemspec') || File.basename(dir) == 'fastlane-plugin-wpmreleasetoolkit'
```

AliSoftware:
Nudge @jkmassel: could we get that merged before next week?

I'm planning to do a 1.3.1 hotfix soon, and would love to include this fix in it, especially since @zwarm  also needs this fix in https://github.com/wordpress-mobile/WordPress-Android/pull/14744#issuecomment-862839216 to move the Jetpack Screenshots PR forward 😅 🙇

lib/fastlane/plugin/wpmreleasetoolkit/helper/filesystem_helper.rb
jkmassel: Sorry for the slow followup – so your edit was correct! Different behaviour in production vs development.

In development, we'll hit the `.gemspec` path, but in production the path looks like `vendor/bundle/ruby/2.6.0/gems/fastlane-plugin-wpmreleasetoolkit-1.3.0`, hence the `include` (though `starts_with` might be _technically_ safer?)

WDYT?

AliSoftware: Ahhh yeah makes more sense now. Yeah I think `start_with?` would make more sense then (and maybe even `start_with?('fastlane-plugin-wpmreleasetoolkit-')`? 🤷 )

I think it's worth improving the comment on that line a bit to explain that better?

```suggestion
          # The first case is for development – where the `.gemspec` is present in the project root
          # The second case is for production – where there's no `.gemspec`, but the root dir of the plugin is named `fastlane-plugin-wpmreleasetoolkit-{version}`.
          if child_filenames.include?('fastlane-plugin-wpmreleasetoolkit.gemspec') || File.basename(dir).start_with?('fastlane-plugin-wpmreleasetoolkit')
```

AliSoftware (approved):
Code LGTM 👍 

I'll double-check that there's no issue likely next week when I have the occasion to go back to work on Jetpack Screenshots automation (https://github.com/wordpress-mobile/WordPress-Android/pull/15067/files) – which needed that fix to work – but I think we can merge this in the meantime

PR #280, opened 2021-06-22
Merge Release 1.3.0 into trunk

JavonDavis:
Merge New version 1.3.0. into trunk


JavonDavis:
Not sure why Danger is complaining about nokogiri here @AliSoftware , any ideas?

AliSoftware:
It happened in the past too, apparently while CircleCI has no trouble fetching the nokogiri version that we use, GitHub action fails to do so sometimes.

I believe this might be related to the odd way nokogiri release multiple platform-specific versions of each of their releases, eg one precompiled for darwin (Mac), another precompiled for other architectures, and one not precompiled (Ruby), etc.

Running `bundle update nokogiri` to fetch the latest, or removing the platforms other that `ruby` in the Gemfile.lock might help forcing it to use the ruby, non-platform-specific version of nokogiri and make it work on all CPU architectures, including on GitHubActions?

CHANGELOG.md
```diff
@@ -20,6 +20,19 @@ _None_
...
+### New Features
+
+* Support for a `version.properties` to manage app versioning - all existing paths remain intact and new paths are only used when a `version.properties` file is present.
+* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
+* Supporting the new `version.properties` file also allows for the `HAS_ALPHA_VERSION` variable to be removed as the alpha reference in the properties file will be used going forward.
+* Clients adopting the new `version.properties` will need to implement a gradle task named `updateVersionProperties` to update the `version.properties` file.
+
+### Bug Fixes
+
+* Some cleanup to how we scope variables in some of our actions
```
AliSoftware, line 34: I just realized that this item  should, by definition, have been in "internal changes" not "bug fixes" (since it doesn't fix any bug we had but instead is only an internal refactoring with no impact on users of the plugin)

Gemfile.lock
```diff
@@ -214,21 +214,19 @@ GEM
...
     minitest (5.14.4)
     multi_json (1.15.0)
     multipart-post (2.0.0)
     nanaimo (0.3.0)
     nap (1.1.0)
     naturally (2.2.1)
     no_proxy_fix (0.1.2)
-    nokogiri (1.11.7)
-      mini_portile2 (~> 2.5.0)
+    nokogiri (1.11.7-x86_64-darwin)
```
AliSoftware: This (the use of the `x86_64-darwin` variant) is what makes the build on GitHubAction (which runs Danger on the code) fail to install nokogiri since the GHA image/machine we use is likely a Linux machine not a Mac/Darwin CPU.

I didn't check but I think we need to remove darwin from the list of platforms (or maybe instead add the platform used by GHA to the list of platforms) that is likely listed down below in that Gemfile.lock? 🤔
Or maybe there's some flag to use during `bundle install` to ensure it's not gonna pull anything platform-specific? 🧐

AliSoftware (commented):
Didn't review it yet, because I'm on mobile so only skimmed the code to leave quick comments on the Danger failure. (Will re-add release managers as reviewers after those initial skimming comments)

Gemfile.lock
JavonDavis: Ahhh looked right over that!

CHANGELOG.md
JavonDavis: Yeah this makes sense, I'll change the heading here to Internal Changes

JavonDavis: Updated in https://github.com/wordpress-mobile/release-toolkit/pull/280/commits/16a2d27fe0620b704b4211bc17a209ab70c0fd8f

mokagio: GitHub doesn't mark this conversation as outdated, but it is as one can see by looking at the content and the commits history.

I'm going to mark it as resolved.

mokagio (approved):
The changes relevant to the release publishing match what's on #281 and the commits history checks out<sup>*</sup> 👍 

---

`*` I was initially surprised to see the "merge release into develop" commit, but it obviously makes sense, since the release branch is cut from `develop`.

CHANGELOG.md
AliSoftware: @mokagio Thanks. FYI, GitHub didn't auto-flag this as outdated because my comment was on line 34 only, but what was changed to address this was on line 32, so GitHub couldn't realize it was related or guess that the change made on line 32 would outdate the comment on line 34 😉 (It would have helped if I made my initial comment on lines 32-34 instead of just 34 😅 )

PR #282, opened 2021-06-23
Update dependencies

jkmassel:
Updates the `bigdecimal` dependency to the latest version (along with active support).

**To Test**
- Ensure tests pass

There's a good argument to be made for this being a breaking change (given the fact that the underlying dependencies have major version changes).

I don't feel strongly about it either way, but I don't think we use either ActiveSupport or BigDecimal directly, so it _shouldn't_ matter?

### DIFF
--- a/fastlane-plugin-wpmreleasetoolkit.gemspec
+++ b/fastlane-plugin-wpmreleasetoolkit.gemspec
@@ -43,13 +43,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'progress_bar', '~> 1.3'
   spec.add_dependency 'parallel', '~> 1.14'
   spec.add_dependency 'chroma', '0.2.0'
-  spec.add_dependency 'activesupport', '~> 5'
-  # Some of the upstream code uses `BigDecimal.new` which version 2.0 of the
-  # `bigdecimal` gem removed. Until we'll find the time to identify the
-  # dependencies and see if we can move them to something compatible with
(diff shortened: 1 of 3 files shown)


oguzkocer (approved):
I think we give it a go and revert it if it's actually a breaking change? I am not merging the PR in case others feel differently.

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -43,13 +43,8 @@ Gem::Specification.new do |spec|
...
   spec.add_dependency 'chroma', '0.2.0'
-  spec.add_dependency 'activesupport', '~> 5'
-  # Some of the upstream code uses `BigDecimal.new` which version 2.0 of the
-  # `bigdecimal` gem removed. Until we'll find the time to identify the
-  # dependencies and see if we can move them to something compatible with
-  # modern `bigdecimal`, let's constrain the gem to a version still supporting
-  # `.new` but which warns about it deprecation.
-  spec.add_dependency 'bigdecimal', '~> 1.4'
+  spec.add_dependency 'activesupport', '~> 6.0'
+  spec.add_dependency 'bigdecimal', '~> 2.0'
```
mokagio, line 47: Thank you for following up on my quick workaround @jkmassel 🙇

mokagio (approved):
> I think we give it a go and revert it if it's actually a breaking change?

Agreeing with @oguzkocer

PR #283, opened 2021-06-23
Fix crashes after testing beta lanes post 1.3.0

AliSoftware:
As I started using the new `1.3.0` release-toolkit to generate WordPress 17.6-rc-4, Jetpack 17.6-rc-4, I encountered some crashes, that this PR intends to fix.

Depending if I run into issues during the final release on Friday, I might need to ad more fixes to this PR before we consider a new release, but the changes that are in so far are already improving the current state and ready to be reviewed.

\cc @JavonDavis for your information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Fix crashes introduced in `1.3.0` – incorrect parameters in calls to `get_release_version`. [#283]
+* Fix the way versioning is handled for alphas – i.e. `version.properties` is indexed by flavor name, defaulting to `zalpha` for alphas. [#283]
 
 ### Internal Changes
 
@@ -25,7 +26,7 @@ _None_
 ### New Features
 
 * Support for a `version.properties` to manage app versioning - all existing paths remain intact and new paths are only used when a `version.properties` file is present.
-* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
+* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo.
 * Supporting the new `version.properties` file also allows for the `HAS_ALPHA_VERSION` variable to be removed as the alpha reference in the properties file will be used going forward.
 * Clients adopting the new `version.properties` will need to implement a gradle task named `updateVersionProperties` to update the `version.properties` file.
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
@@ -14,18 +14,18 @@ module Fastlane
         app = params[:app]
 
         # Check versions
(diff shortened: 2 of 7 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -49,11 +49,11 @@ def self.run(params)
...
       end
 
-      def self.get_user_build_version(version, message)
+      def self.get_user_build_version(product_name:, version:, message:)
         UI.user_error!("[#{app}] Release branch for version #{version} doesn't exist. Abort.") unless Fastlane::Helper::GitHelper.checkout_and_pull(release: version)
-        release_version = Fastlane::Helper::Android::VersionHelper.get_release_version
+        release_version = Fastlane::Helper::Android::VersionHelper.get_release_version(product_name: product_name)
         message << "#{app}] Looking at branch release/#{version} as requested by user. Detected version: #{release_version[Fastlane::Helper::Android::VersionHelper::VERSION_NAME]}.\n"
-        alpha_release_version = Fastlane::Helper::Android::VersionHelper.get_alpha_version
+        alpha_release_version = Fastlane::Helper::Android::VersionHelper.get_alpha_version(product_name)
```
AliSoftware, line 56: Side note: we need to harmonize the parameter names for all `get_*_version` helper methods at some point, so that we're consistent with using `product_name:` named parameter for all of those, instead of having a mix like here.

But that's something for another day, as the goal of this PR is just to make things work and stop crashing, limiting any refactoring to a minimum for now; and fixing consistency and doing the clean up only later.

CHANGELOG.md
```diff
@@ -25,7 +26,7 @@ _None_
 ### New Features
 
 * Support for a `version.properties` to manage app versioning - all existing paths remain intact and new paths are only used when a `version.properties` file is present.
-* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo
+* Add support for providing an `app:` parameter to most versioning-related actions to allow support for multiple apps hosted in a monorepo.
```
jkmassel, line 30: I'm still not convinced we need / should maintain this given the fact that we're using a common version between apps sharing a repo right now. I feel like this will be a future maintenance burden for something we _might_ need in the future. WDYT?

jkmassel (changes requested):

CHANGELOG.md
AliSoftware: @jkmassel I agree, but this change was already shipped in 1.3.0 (this entry is from `## 1.3.0`, the only change I made on this line is to fix a missing `.`)

My take is that we should at least avoid having the toolkit crash in its current 1.3.0 version, so we should still land this `1.3.1` hotfix as is, to at least get back to a non-crashing version.
And only after that lands, revert the changes as we discussed (to get back to using a common version for both WP and JP, and remove that `app:` parameter) for a future 1.4.0 – well actually a 2.0.0 as this would be a breaking change I think. It would have maybe be better to not land 1.3.0 at all, but here we are, so I think that's the best course of action.

Does this make sense? If so, can you approve the PR so it can be merged in the meantime before I work on reverting those changes later?

jkmassel (approved):

PR #284, opened 2021-06-24
Fix the translation progress action

loremattei:
In https://github.com/wordpress-mobile/release-toolkit/pull/263 we introduced a new action which checks the state of the translations on GlotPress.
Since there's no API on GlotPress to get this information, the action extracts the important data from the GlotPress project page. 

While the information about the percentage of translated strings is available in that page, it's a bit difficult to parse because of the layout of the page itself. Extracting the number of `Translated`, `Fuzzy`, `Untranslated` and `Waiting` strings is far easier and the first implementation of the lane used those value to calculate the percentage of the translated strings. 

This is generally acceptable, but it turned out there are edge cases where it may result in pessimistic estimations. 
In particular, it's possible that a string has a current, valid translation, but it also has a pending proposal for a different one. In this case, the number of `Waiting` strings can be higher than 0 even when all the strings are translated, which causes a slightly wrong evaluation of the progress percentage. 
This is especially annoying in the projects where we require translations at 100%, because a single `Waiting` string can lower this number under the threshold. 

This PR updates the action to use a different approach and parse the percentage directly from the HTML page. This requires a regex which analyses text on three or four lines (depending on the case), so it's a bit more complex. A detailed explanation of the new regex is in the code. 

### To test
- Verify CI is green.
- Test on a live GlotPress repository (e.g. https://translate.wordpress.org/projects/apps/android/dev/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Fixed an issue in `check_translation_progress` where a wrong evaluation of the progress is possible when there are Waiting string in Glotpress.
 
 ### Internal Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-* Fixed an issue in `check_translation_progress` where a wrong evaluation of the progress is possible when there are Waiting string in Glotpress.
+* Fixed an issue in `check_translation_progress` where a wrong evaluation of the progress is possible when there are Waiting string in GlotPress.
 
 ### Internal Changes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
@@ -26,7 +26,7 @@ module Fastlane
       # @return [Integer] The percentage of the translated strings.
       #
       def self.get_translation_status(data:, language_code:)
-        # The status is parsed from Glotpress project page.
+        # The status is parsed from the GlotPress project page.
         # The row can be identified by the language code and the progress is in the column identified by the class "stats percent".
         # When the progress is above 90%, a special badge is added.
         # Because of the way the HTML is organized, this regex matches content spawned on three or four lines
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
@@ -26,7 +26,7 @@ module Fastlane
       # @return [Integer] The percentage of the translated strings.
       #
       def self.get_translation_status(data:, language_code:)
-        regex = "<strong><a href=\".*\\/#{language_code}\\/default\\/\">.*<\\/strong>\\n.*<span.*>([0-9]+)%<\\/span>"
+        regex = "$\\s*<strong><a href=\".*\\/#{language_code}\\/default\\/\">.*<\\/strong>\\n.*\\n*.*\\n*.*<td class=\"stats percent\">([0-9]+)%<\/td>$"
 
         # 1. Grep the line with contains the required info.
         # 2. Match the info and extract the value in group 1.
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
@@ -26,12 +26,13 @@ module Fastlane
       # @return [Integer] The percentage of the translated strings.
       #
       def self.get_translation_status(data:, language_code:)
-        current = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'current')
-        fuzzy = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'fuzzy')
-        untranslated = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'untranslated')
-        waiting = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'waiting')
+        regex = "<strong><a href=\".*\/#{language_code}\/default\/\">.*<\/strong>\n<span.*>([0-9]+)%<\/span>"
 
-        (current * 100 / (current + fuzzy + untranslated + waiting)).round
+        # 1. Grep the line with contains the required info.
+        # 2. Match the info and extract the value in group 1.
+        # 3. Convert to integer.
+        puts data
+        data.grep(/#{regex}/)[0].match(/#{regex}/)[1].to_i
       end
 
       # Extract the number of strings which are in the given status.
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
```diff
@@ -26,12 +26,35 @@ def self.get_translation_status_data(glotpress_url:)
...
+        #       \d\d\d?%    : 2 or 3 digits and the percentage char
+        #       <\/span>\n  : Special badge closure and new line
+        # )?          : end of the "morethan90" special badge section. Expect this zero or one times.
+        # \s*<\/td>\n : column closure tag. Any space before of it are ok. Expect new line after it.
+        # \s*         : any space
+        # <td class="stats percent"> : This is the tag which can be used to extract the progress
+        # ([0-9]+)    : progress is the first group
+        # %<\/td>     : tag closure
+        regex = "^\\s*<strong><a href=\".*\\/#{language_code}\\/default\\/\">.*<\\/strong>\\n"
+        regex += '(?:\s*<span class="bubble morethan90">\d\d\d?%<\/span>\n)?\s*<\/td>\n\s*<td class="stats percent">([0-9]+)%<\/td>$'
```
AliSoftware, line 52: Seeing that `morethan90` CSS class here, I'm assuming/guessing it might only be set to that `span` if the percentage is indeed >90 for that `tr` row? Which could mean that this regex could fail to catch cases where the stats percent <90 for which that CSS class is not included in the HTML?

AliSoftware (commented):
> This is generally acceptable, but it turned out there are edge cases where it may result in pessimistic estimations.
In particular, it's possible that a string has a current, valid translation, but it also has a pending proposal for a different one. In this case, the number of Waiting strings can be higher than 0 even when all the strings are translated, which causes a slightly wrong evaluation of the progress percentage.

Daniel suggested recently (p1624537244164400/1624533980.163600-slack-C02AED43D) to monitor the "Untranslated" count instead of "Translated" %, ie we'd consider the translations ready not when Translated = 100% and Waiting = 0 (which is subject to the edge case you explain here), but instead when Untranslated = 0.
Do you think his suggestion could be a better approach to try here?

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
```diff
@@ -26,12 +26,35 @@ def self.get_translation_status_data(glotpress_url:)
...
+        #       \s*         : any space
+        #       <span class="bubble morethan90"> : Start of the special badge
+        #       \d\d\d?%    : 2 or 3 digits and the percentage char
+        #       <\/span>\n  : Special badge closure and new line
+        # )?          : end of the "morethan90" special badge section. Expect this zero or one times.
+        # \s*<\/td>\n : column closure tag. Any space before of it are ok. Expect new line after it.
+        # \s*         : any space
+        # <td class="stats percent"> : This is the tag which can be used to extract the progress
+        # ([0-9]+)    : progress is the first group
+        # %<\/td>     : tag closure
```
AliSoftware, line 50: ❤️ 💯  for explaining the regex step by step in comments 🤗

loremattei:
> Daniel suggested recently (p1624537244164400/1624533980.163600-slack-C02AED43D) to monitor the "Untranslated" count instead of "Translated" %, ie we'd consider the translations ready not when Translated = 100% and Waiting = 0 (which is subject to the edge case you explain here), but instead when Untranslated = 0.
Do you think his suggestion could be a better approach to try here?

Hmm... This is totally doable, but I'm not sure it's going to work. IIRC; we don't download the current translation for fuzzied strings 🤔 But we often get fuzzied strings, like when we do small changes. Also, sometimes untranslated doesn't seem to tell the whole story, for example here (hoping that this doesn't change before you can look at it): https://translate.wordpress.org/projects/apps/android/dev/he/default/. There are 2849 total strings, but only 2 untranslated, the other 3 are fuzzy. 

Anyway, I'm not removing the parser for categories in this PR, so we could also run both of them and compare to learn more in the future.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
loremattei: Yeah! Initially I thought the same and implemented a slighter simpler regex thinking that I could always get the value in that span, but it's not the case. When the percentage is lower than 90, all the span is missing. You can check the relevant GlotPress code here:
https://github.com/GlotPress/GlotPress-WP/blob/0c395a7a8f37ab3b5ebafd2b239c74392e5177f9/gp-templates/project.php#L100

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Fixed an issue in `check_translation_progress` where a wrong evaluation of the progress is possible when there are Waiting string in Glotpress.
```
mokagio: Nitpick:

```suggestion
* Fixed an issue in `check_translation_progress` where a wrong evaluation of the progress is possible when there are Waiting string in GlotPress.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
```diff
@@ -26,12 +26,35 @@ def self.get_translation_status_data(glotpress_url:)
       # @return [Integer] The percentage of the translated strings.
       #
       def self.get_translation_status(data:, language_code:)
-        current = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'current')
-        fuzzy = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'fuzzy')
-        untranslated = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'untranslated')
-        waiting = extract_value_from_translation_info_data(data: data, language_code: language_code, status: 'waiting')
+        # The status is parsed from Glotpress project page.
```
mokagio: Nitpick:

```suggestion
        # The status is parsed from the GlotPress project page.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_helper.rb
mokagio: Alos ❤️ 💯 for 🤯 regex-ing in this PR!

mokagio:
@loremattei: If I read your the previous comment correctly, the concern with monitoring only the "Untranslated" count is that we might get the wrong picture because when don't download translations for "Fuzzy" strings.

So, if I'm getting this right, in a case like the one in the screenshot below for the Korean translation, monitoring "Untranslated" would result in 100% progress because there are 0 untranslated strings, but we wouldn't be downloading _all_ translations because of the 22 waiting and 23 fuzzy ones.

Screen Shot 2021-06-25 at 12 48 48 pm

mokagio (approved):
> Anyway, I'm not removing the parser for categories in this PR, so we could also run both of them and compare to learn more in the future.

If we wanted to really experiment with this, we could print the output of both evaluations and compare them.

PR #287, opened 2021-07-16
Add option to skip updating Deliverfile to ios_bump_version_hotfix

mokagio:
In https://github.com/wordpress-mobile/WordPress-iOS/pull/16805, made the WordPress and Jetpack iOS automation read the version value to apps to `deliver` from the `.xcconfig`. This makes updating that key in the `Deliverfile` unnecessary. In fact, that PR entirely removes the `Deliverfile`.

When it came time to code freeze WordPress/Jetpack iOS version 17.8, the lane failed because of the missing `Deliverfile`. I thought I'd fixed the issue, but I clearly missed the hotfix use case.

This PR adds a `skip_deliver` option to `ios_bump_version_hotfix` to mimic the one in `ios_bump_version_release`. I tested these changes "live" when creating the https://github.com/wordpress-mobile/WordPress-iOS/tree/release/17.7.1 hotfix.

## To test

1. Open WordPress iOS
3. Run `bundle exec new_hotfix_release version:17.7.1` (or whichever version you prefer)
4. Verify the process fails because `Deliverfile` can't be found

image

1. Delete the new `release/*` branch that the `new_hotfix_release` lane created, both locally and on GitHub
1. [Point the release toolkit to this branch](https://github.com/wordpress-mobile/WordPress-iOS/commit/56ea967f52ccb1f63a2dc914f6774233d7f65111)
2. [Update the `ios_bump_version_hotfix` call to include `skip_deliver: true`](https://github.com/wordpress-mobile/WordPress-iOS/commit/d6a842f8e3a851176d7e1fa7058880b21f6ccc4f)
3. Run `bundle exec new_hotfix_release version:17.7.1` (or whichever version you prefer)
4. Verify the process succeeds

image

1. Delete the new `release/*` branch that the `new_hotfix_release` lane created, both locally and on GitHub

---

Based on top of `revert-activesearch-v6-bump` branch for ease of review. Either I or GitHub will update the base appropriately before merging into `trunk`.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_hotfix.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_hotfix.rb
@@ -9,14 +9,18 @@ module Fastlane
         create_config(params[:previous_version], params[:version])
         show_config()
 
-        UI.message 'Updating Fastlane deliver file...'
-        Fastlane::Helper::Ios::VersionHelper.update_fastlane_deliver(@new_short_version)
-        UI.message 'Done!'
+        update_deliverfile = params[:skip_deliver] == false
+        if update_deliverfile
+          UI.message 'Updating Fastlane deliver file...'
+          Fastlane::Helper::Ios::VersionHelper.update_fastlane_deliver(@new_short_version)
+          UI.message 'Done!'
+        end
+
         UI.message 'Updating XcConfig...'
         Fastlane::Helper::Ios::VersionHelper.update_xc_configs(@new_version, @new_short_version, @new_version_internal)
         UI.message 'Done!'
 
-        Fastlane::Helper::Ios::GitHelper.commit_version_bump(include_deliverfile: true, include_metadata: false)
+        Fastlane::Helper::Ios::GitHelper.commit_version_bump(include_deliverfile: update_deliverfile, include_metadata: false)
 
         UI.message 'Done.'
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_hotfix.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_hotfix.rb
@@ -47,6 +47,16 @@ module Fastlane
             description: 'The version to branch from',
             is_string: true
           ),
+          FastlaneCore::ConfigItem.new(
+            key: :skip_deliver,
+            env_name: 'FL_IOS_BUMP_VERSION_HOTFIX_SKIP_DELIVER',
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_hotfix.rb
```diff
@@ -9,14 +9,18 @@ def self.run(params)
         create_config(params[:previous_version], params[:version])
         show_config()
 
-        UI.message 'Updating Fastlane deliver file...'
-        Fastlane::Helper::Ios::VersionHelper.update_fastlane_deliver(@new_short_version)
-        UI.message 'Done!'
+        update_deliverfile = params[:skip_deliver] == false
```
mokagio, line 12: As per comment in the code below, I decided to call this parameter "skip" instead of "include" because the majority of the consumers still have `Deliverfile`.

When they'll all have moved to not use `Deliverfile`, we shall remove the parameter altogether.

mokagio (commented):
I wanted to add tests here, but I felt it would require too much work for the scope of this PR.

My idea to add tests was to create a new object, `IosBumpVersionHotfixActionRunner`, or simply `IosBumpVersionHotfixRunner`, to delegate all the logic from `self.run` to.

It would have a `run` method with named parameters for each of the parameters read from the action. Doing so would help making the input space for the tests clear, because the code wouldn't be allowed to read any kind of value (hidden dependency) from a dictionary (like the action does with `params`).

What do you think of this approach?

AliSoftware:
> My idea to add tests was to create a new object, IosBumpVersionHotfixActionRunner, or simply IosBumpVersionHotfixRunner, to delegate all the logic from self.run to.

@mokagio From looking at how Fastlane itself writes the tests for actions, they seem to already have some logic to help with that, including providing a dummy minimalist `Fastfile` as a string to be interpreted, then invoking the runner on it.

https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/deliver_action_spec.rb#L19-L21
```ruby
        expect do
          Fastlane::FastFile.new.parse("lane :test do
            deliver
          end").runner.execute(:test)
        end.to raise_error(/Could not find ipa file at path '/)
```

I think it could make sense to borrow their approach here.

AliSoftware (approved):
:shipit: 

As for tests, given I want to finally be able to merge all the hotfix PRs in all the repos that are open in parallel and don't want to hold off on a hotfix release for too long, I think I'm gonna merge this as is to do a new release of the toolkit now but feel free to add those tests in a subsequent PR even after the new version has shipped.

mokagio:
> feel free to add those tests in a subsequent PR even after the new version has shipped.

👍 yeah, it would have been to much work and would have prevented this sort-of important fix from getting in

Issue #289, opened 2021-08-02
Version clash for the `activesupport` & `bigdecimal` gems between toolkit and cocoapods-core


mokagio:
While working on testing https://github.com/wordpress-mobile/release-toolkit/pull/287, I run into a dependency resolution error when trying to point WordPress iOS to use my branch (which I had just rebased on top of [my latest `develop`](https://github.com/wordpress-mobile/release-toolkit/commit/ab3d6dba976929956f3924ca85e6c463cadafb60):

```
Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    cocoapods (~> 1.10) was resolved to 1.10.2, which depends on
      cocoapods-core (= 1.10.2) was resolved to 1.10.2, which depends on
        activesupport (> 5.0, < 6)

    fastlane-plugin-wpmreleasetoolkit was resolved to 1.3.1, which depends on
      activesupport (~> 6.0)
```

As you can see, `cocoapods-core` requires `activesupport` `> 5.0, < 6`, while we require it `~> 6.0`.

I checked, and 1.10.2 is the latest version. Worth noting that `cocoapod-core` [main branch supports version 6, `> 5.0, < 7`](https://github.com/CocoaPods/Core/blob/397fa25c069afc7c75666b74ffc39b2d61e8130c/cocoapods-core.gemspec#L22).

We bumped `activesupport` [here](https://github.com/wordpress-mobile/release-toolkit/pull/282) to resolve a workaround for `bigdecimal` (an `activesupport` dependency) introduced [here](https://github.com/wordpress-mobile/release-toolkit/pull/161/commits/6c11d29292e612ae772291c21dad89a400414f46), with motivation:

> Some of the upstream code uses `BigDecimal.new` which version 2.0 of the `bigdecimal` gem removed. Until we'll find the time to identify the dependencies and see if we can move them to something compatible with modern `bigdecimal`, let's constrain the gem to a version still supporting `.new` but which warns about it deprecation.

---

We merged the [PR](https://github.com/wordpress-mobile/release-toolkit/pull/282) that introduced the `activesupport` bump by saying:

> give it a go and revert it if it's actually a breaking change

I suggest we revert it because it turns out to be a breaking change 😳

mokagio:
I opened #290 to "address" this by reverting. I think it's the best course of action right now.

When CocoaPods will ship a new version, the new `activesupport` requirement will become available and we should be able to upgrade again.

PR #290, opened 2021-08-02
Revert activesupport version bump

mokagio:
This removes the explicit version requirement for `activesupport` to prevent dependency resolution issues with CoacoaPods.

See #289.

It also adds CocoaPods as a development dependency to prevent this issue from happening again.


fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -56,4 +61,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'simplecov', '~> 0.16.1'
   spec.add_development_dependency 'fastlane', '~> 2'
   spec.add_development_dependency 'rmagick', '~> 4.1'
+  spec.add_development_dependency 'cocoapods', '~> 1.10'
```
mokagio, line 64: My rationale is that, by adding this as a development dependency only, we won't be able to bump dependencies to versions that are not compatible with CocoaPods locally. At the same time, by being only a _development_ dependency, users of this gem won't need to install CocoaPods, which I think is valuable for the Android repos which are still using the Ruby toolchain.

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -43,8 +43,13 @@ Gem::Specification.new do |spec|
...
   spec.add_dependency 'chroma', '0.2.0'
-  spec.add_dependency 'activesupport', '~> 6.0'
-  spec.add_dependency 'bigdecimal', '~> 2.0'
+  spec.add_dependency 'activesupport', '~> 5'
+  # Some of the upstream code uses `BigDecimal.new` which version 2.0 of the
+  # `bigdecimal` gem removed. Until we'll find the time to identify the
+  # dependencies and see if we can move them to something compatible with
+  # modern `bigdecimal`, let's constrain the gem to a version still supporting
+  # `.new` but which warns about it deprecation.
+  spec.add_dependency 'bigdecimal', '~> 1.4'
```
AliSoftware, line 52: I agree with the rationale of reverting our change since it's breaking. But I'm wondering if we couldn't just remove those dependencies from release toolkit completely – tbh I'm not really sure we use activesupport and bigdecimal directly from within the toolkit? If we do it would be worth identifying where and why.

I think the right course of action will be to:
 - Approve and ship that PR that reverts the changes
 - Ship a new version of the release toolkit to include our other PRs (hotfix lanes, skip deliverfile)
 - Revisit that dependency removal after that, by doing a quick investigation on where it's used in the RT and why

AliSoftware (approved):
Agree with the rationale to revert since it's breaking. Left a strategy/plan for the future, but first let's release a new release-toolkit version with that bigdecimal revert, and hopefully we'll find some time investigate that bigdecimal case in more details later 🤞

fastlane-plugin-wpmreleasetoolkit.gemspec
mokagio: > But I'm wondering if we couldn't just remove those dependencies from release toolkit completely

That would be idea. Best way to solve a problem is to not have the problem in the first place!

PR #292, opened 2021-08-02
Release 1.4.0 into develop

AliSoftware:
New version 1.4.0

PR #293, opened 2021-08-04
Opt-out form platform-specific gems installation in Bundler

mokagio:
See discussion [here](https://github.com/wordpress-mobile/release-toolkit/pull/292#issuecomment-891758279)

To ensure the setting is persisted for every user of the repository, I decided to track the Bundler config file. This seems like a good idea anyway, even though Bundler currently has a [bug](https://github.com/rubygems/rubygems/issues/3708) resulting in that file being changed sometimes, which can be confusing for developers.

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,0 +1,3 @@
+---
+BUNDLE_SPECIFIC_PLATFORM: "false"
+BUNDLE_WITH: "test"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ commands:
                   brew update && xargs brew install --verbose < .circleci/.brewfile
       - run:
           name: Install Ruby dependencies, if neeeded
-          command: bundle check --path vendor/bundle || bundle install --with screenshots
+          command: bundle check --path vendor/bundle || bundle install
       - save_cache:
           name: Cache Homebrew + Ruby Dependencies
           key: *cache_key
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -260,6 +260,7 @@ GEM
     method_source (0.9.2)
     mini_magick (4.11.0)
     mini_mime (1.1.0)
+    mini_portile2 (2.5.3)
     minitest (5.14.4)
     molinillo (0.6.6)
     multi_json (1.15.0)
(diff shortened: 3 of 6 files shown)


.bundle/config
```diff
@@ -0,0 +1,3 @@
+---
+BUNDLE_SPECIFIC_PLATFORM: "false"
+BUNDLE_WITH: "test"
```
mokagio, line 3: I added comments for each of those options, but the disappear after every `bundle install`, so I decided to leave them out to avoid noise in the repo locally. 😞 

The PR description already explains why we have `BUNDLE_SPECIFIC_PLATFORM: "false"`. As for `BUNDLE_WITH: "test"`, my rationale is that it makes local development easier –if you work on this repo you run the tests, right? Right?! 😝– so I left it up to CI to opt-out when appropriate.

.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ commands:
                   brew update && xargs brew install --verbose < .circleci/.brewfile
       - run:
           name: Install Ruby dependencies, if neeeded
-          command: bundle check --path vendor/bundle || bundle install --with screenshots
+          command: bundle check --path vendor/bundle || bundle install
```
mokagio, line 31: There's no "screenshots" group in the Gemfile.

Gemfile.lock
```diff
@@ -415,4 +417,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   2.2.24
+   2.2.25
```
mokagio, line 420: While at it...

AliSoftware (approved):
Thanks!

Solution makes sense to me — as discussed in Slack — and GitHub Action checks are now green (which means GHA was able to install `nokogiri` on their Ubuntu image without encountering the issue of being tied to darwin-specific in lockfile) so I think we're good.

~~Only question that remains @mokagio: should we remove `.bundle/config` from the `.gitignore` or was it left here intentionally (so that the bundler bug and any potential accidental changes to it didn't get suggested to be commited)?~~

Never mind, I missed the removal of `.bundle` for the `.gitignore` during my first read of the PR diff, I see you _did_ remove it already, so we're all good 👍

PR #295, opened 2021-08-10
Relax Rake version requirement to > 12.3, < 14.0

mokagio:
I was working with @rustyshelf on [this PR](https://github.com/shiftyjelly/pocketcasts-ios/pull/3420) today.

For some reason, the version of `rake` in the Pocket Casts lockfile [is at `13.0.6`](https://github.com/shiftyjelly/pocketcasts-ios/blob/develop/Gemfile.lock#L261`) and we couldn't resolve the release plugin dependency:

```
Bundler could not find compatible versions for gem "rake":
  In snapshot (Gemfile.lock):
    rake (= 13.0.6)

  In Gemfile:
    danger-swiftlint was resolved to 0.26.0, which depends on
      rake (> 10)

    fastlane was resolved to 2.189.0, which depends on
      google-cloud-storage (~> 1.31) was resolved to 1.34.1, which depends on
        digest-crc (~> 0.4) was resolved to 0.6.4, which depends on
          rake (>= 12.0.0, < 14.0.0)

    fastlane-plugin-wpmreleasetoolkit was resolved to 1.4.0, which depends on
      rake (~> 12.3)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which
```

Rake 13 was released in September 2019. @jkmassel added the `~> 12.3` constraint (there was none before) [in June 2019](https://github.com/wordpress-mobile/release-toolkit/commit/2f8f3a36330e0a9d73144c877e4368277e76807d).

I don't know whether we should straight bump to `~> 13.0`, or do like `digest-crc` does and allow for versions 12 and 13.  According to [Rake's changelog](https://github.com/ruby/rake/blob/master/History.rdoc#1300-) the major version bump was due to dropping support for Ruby 2.2, so it should be safe to bump.

On the other hand, I wouldn't want to trigger a cascade of Rake updates in the clients.
That's why, of the two options, I went to the more flexible one.

What do you think?


AliSoftware (approved):
LGTM 👍 

I'm ok in allowing both 12.3+ and 13.x instead of imposing 13.x, rationale makes sense and it's indeed more flexible to avoid forcing an update when it's likely not strictly mandatory.

I'll let you add the "Internal Changes" `CHANGELOG` entry before merging 😉

jkmassel:
>I'm ok in allowing both 12.3+ and 13.x instead of imposing 13.x, rationale makes sense and it's indeed more flexible to avoid forcing an update when it's likely not strictly mandatory.

+1

loremattei:
I agree with that as well and I'd suggest that we start to bump clients to 13+ if there are no blockers so that we don't fall too behind over the time and get stuck in the future.

PR #296, opened 2021-08-10
Introduce Locale helper models

AliSoftware:
This PR introduces the `Locales` model class in the release toolkit. Its intent is to work way more easily with locale codes in our `Fastfiles`, by replacing huge constant lists like [those 65 lines in WPAndroid's `Fastfile`](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/fastlane/Fastfile#L10-L75) with way shorter and nicer lines like:

```ruby
RELEASE_NOTES_LOCALES  = Fastlane::Locales.mag16
```
This will return an `Array<Locale>` where `Locale` is a struct with 5 fields containing the locale codes for each space (glotpress, android, google_play, ios, app_store).

A `Locale` can also be converted into a `Hash` for retro-compatibility with the existing toolkit actions that expect such a `Hash` (even though in the future we might want to support both a `Hash` and a `Locale` for those actions, but that's for a separate story)
This means we will be able to replace [those WCAndroid 18 lines](https://github.com/woocommerce/woocommerce-android/blob/develop/fastlane/Fastfile#L25-L42) with just this – using `.map(&:to_h)` to convert it from an `Array<Locale>` to an `Array<Hash>` for compatibility with existing actions:
```ruby
SUPPORTED_LOCALES = Fastlane::Locales[%w[ar de es fr he id it ja ko nl pt-br ru sv tr zh-cn zh-tw]].map(&:to_h)
```

## Draft State

We still need to
 - [ ] Complete the list of `ALL_KNOWN_LOCALES`: add the missing values for `ios` and `app_store` keys, then ensure all the codes are correct for all the various keys, and that we are not missing any known locale.
 - [x] Enable `from_ios` and `from_app_store` family of tests once we have the values in `ALL_KNOWN_LOCALES`

@mokagio I've added you as preliminary reviewer, feel free to take a first look (especially at the `Locales` class API and at the tests) and provide initial feedback. I'll add the whole Owl team as final reviewers once ☝️  is completed and I undraft the PR.


lib/fastlane/plugin/wpmreleasetoolkit/models/locales.rb
```diff
@@ -0,0 +1,145 @@
...
+    end
+  end
+
+  # A class with static methods to manipulate lists of locales.
+  # Exposes various `Array<Locale>` lists like all known locales, the Mag16,
+  # and convenience methods to turn list of Strings into list of Locales.
+  class Locales
+    ###################
+    ## Constants
+    ALL_KNOWN_LOCALES = [
```
AliSoftware, line 12: WIP: To be completed with `ios:` and `app_store:` values for each locale; we'll also need to double-check all the tuples to ensure we defined the correct ones.

spec/locale_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  it 'raises if no locale was found for a given code' do
+    expect do
+      described_class['invalidcode']
+    end.to raise_error(RuntimeError, "Unknown locale for glotpress code 'invalidcode'")
+  end
+
+  it 'can convert a Locale to a hash' do
+    h = described_class['fr'].to_h
+    expect(h).to eq({ glotpress: 'fr', android: 'fr', google_play: 'fr-FR', ios: nil, app_store: nil })
```
AliSoftware: To be updated with actual ios and app_store codes [when we'll get to define them](https://github.com/wordpress-mobile/release-toolkit/pull/296/files#r686194520)

spec/locales_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  describe 'from_android' do
+    include_examples 'from_xxx', :android, 'fr', 'pt-rBR'
+  end
+
+  describe 'from_google_play' do
+    include_examples 'from_xxx', :google_play, 'fr-FR', 'pt-BR'
+  end
+
+  # @TODO: from_ios, from_app_store
```
AliSoftware: Add same tests for `from_ios` and `from_app_store` once [we complete the list](https://github.com/wordpress-mobile/release-toolkit/pull/296/files#r686194520)

spec/locales_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      expect(locales[0]).to be_instance_of(Fastlane::Locale)
+      expect(locales[0].glotpress).to eq('fr')
+
+      expect(locales[1]).to be_instance_of(Fastlane::Locale)
+      expect(locales[1].glotpress).to eq('pt-br')
+    end
+
+    it 'raises if one of the locale codes passed was not found' do
+      expect do
+        described_class.send(method_sym, [fr_code, 'invalidcode', 'pt-br'])
```
AliSoftware:
```suggestion
        described_class.send(method_sym, [fr_code, 'invalidcode', pt_code])
```

lib/fastlane/plugin/wpmreleasetoolkit/models/locales.rb
```diff
@@ -0,0 +1,145 @@
...
+      #
+      def all
+        ALL_KNOWN_LOCALES
+      end
+
+      # Define from_glotpress(code_or_list), from_android(code_or_list) … methods.
+      #
+      # Those can be used in the rare cases where you need to find locales via codes other than the glotpress ones,
+      # like searching by android locale code(s) or google_play locale code(s).
+      # In most cases, prefer using the `Locales[…]` method instead ()with glotpress locale codes).
```
AliSoftware:
```suggestion
      # In most cases, prefer using the `Locales[…]` method instead (with glotpress locale codes).
```

AliSoftware:
I'm actually unsure if we use/need list of locale codes in the iOS Fastfiles like we do in the Android ones 🤔  at least yet?

A quick look seem to indicate that we don't really rely on any hardcoded list for iOS, since when post-processing locales we just iterate over folder names in `fastlane/metadata`, and the only other place where we need the list of locales is when we download them from GlotPress, for which, as of today, we still do it via the `download_metadata.swift` Swift script, not via a ruby action where we could use that new `Fastlane::Locales[…]` magic.

But I still think it would be nice to have those ios/app store locale codes in that list even if we don't yet really use them in our Fastfile, so that when we get the opportunity work on the "Improve Localization tooling" project (paaHJt-sd-p2) and likely migrate that Swift script into a fastlane action by then, we'd have those locale codes ready to be used.

---

Another point: unlike android's `values-<code>` vs PlayStore locale codes being different, for iOS it seems (at least at first, from the couple ones I quickly checked) that the locale codes used in iOS (`<code>.lproj`) and in AppStore are the same (some consistency at least, not like Android 😅 ), which means I'm unsure if there's a point keeping both `:ios` and `:app_store` keys in the `Locale` struct (allows to get a nice mirroring with Android) vs only have the `:ios` key (would break the symmetry, but avoids repetition). A middle ground would be to only use `:ios` in the initializer and list of Struct fields, and declare `alias_method :app_store, :ios` as an alias for `ios`.

@mokagio Thoughts?

mokagio:
Sorry about closing the PR I went to click the text area but hit the close button 😮‍💨

mokagio:
Sorry for taking so long to reply!

> But I still think it would be nice to have those ios/app store locale codes in that list even if we don't yet really use them in our Fastfile

+1 there's value in having them there for sure.

>  A middle ground would be to only use `:ios` in the initializer and list of Struct fields, and declare `alias_method :app_store, :ios` as an alias for `ios`.

I like this idea very much.

PR #297, opened 2021-08-24
Strip trailing new lines in single line `msgid` when generating `.po[t]` file

mokagio:
At the moment, this is just a quick fix. After about an hour of `puts`-driven debugging against https://github.com/Automattic/simplenote-ios/pull/1406, I identified the culprit and wrote a fix for the specific issue.

There's more work to be done in this are though. For a start, [`an_metadata_update_helper.rb`](https://github.com/wordpress-mobile/release-toolkit/blob/07b3152b2b22874fee0c373457668546c9424955/lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb) and [`metadata_update_helper.rb`](https://github.com/wordpress-mobile/release-toolkit/blob/07b3152b2b22874fee0c373457668546c9424955/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_update_helper.rb) define the same types. E.g.:

https://github.com/wordpress-mobile/release-toolkit/blob/07b3152b2b22874fee0c373457668546c9424955/lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb#L1-L4

https://github.com/wordpress-mobile/release-toolkit/blob/07b3152b2b22874fee0c373457668546c9424955/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_update_helper.rb#L1-L4

It took me a while to understand why my "breakpoints" in `metadata_helper.rb` ([imported](https://github.com/wordpress-mobile/release-toolkit/blob/07b3152b2b22874fee0c373457668546c9424955/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_update_metadata_source.rb#L2) by `gp_update_metadata_source.rb`, [used](https://github.com/wordpress-mobile/release-toolkit/blob/07b3152b2b22874fee0c373457668546c9424955/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_update_metadata_source.rb#L8-L10) by `ios_update_metadata_source`, [used](https://github.com/Automattic/simplenote-ios/blob/8ef355d4133ce0db9acde631d91e466685ab5218/fastlane/Fastfile#L119-L121) by the lane that generates the `.pot`) weren't hit. The reason is that for some reason the Ruby interpreter loaded the ones defined in `an_metadata_helper.rb`, possibly due to the `an_` prefix which would make that file load first? Although, I'm not sure why that would be the case since there's no import for that file.

## To test

I added a dedicated unit test here. I ought to add more for other facets of the behavior, but I just wanted to fix the issue to begin with.

Also see steps from https://github.com/Automattic/simplenote-ios/pull/1407.

## Next step

I see a few options:

1. Merge this and ship a new version, hoping this is enough to cover all the paths that generate the issues we've encountered so far
2. Spend a bit more time on this PR, replicating the `.strip` on single line fix for all the other block types, then ship a new version
3. Invest the time right now to get to the bottom of the `an_metadata_` vs `metadata_` issues and improve the `.pot` generation logic, then ship a new version

I'm leaning towards option 2. My current implementation doesn't seem future proof, while option 3 sounds like a lot of work that could easily explode in scope.

Keen to hear what the rest of @wordpress-mobile/owl-team thinks, which is why I asked for your review even though this is still a draft.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
@@ -50,7 +50,7 @@ module Fastlane
 
         if line_count <= 1
           # Single line output
-          fw.puts("msgid \"#{File.open(@content_file_path, 'r').read}\"")
+          fw.puts("msgid \"#{File.open(@content_file_path, 'r').read.strip}\"")
         else
           # Multiple line output
           fw.puts('msgid ""')
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
@@ -50,7 +50,7 @@ module Fastlane
 
         if line_count <= 1
           # Single line output
-          fw.puts("msgid \"#{File.open(@content_file_path, 'r').read.strip}\"")
+          fw.puts("msgid \"#{File.open(@content_file_path, 'r').read.rstrip}\"")
         else
           # Multiple line output
           fw.puts('msgid ""')
--- a/spec/an_metadata_update_helper_spec.rb
+++ b/spec/an_metadata_update_helper_spec.rb
@@ -1,0 +1,34 @@
+require 'tmpdir'
+require_relative './spec_helper'
+
+describe Fastlane::Helper::StandardMetadataBlock do
+  it 'strips any trailing newline when generating the block for a single-line input' do
+    Dir.mktmpdir do |dir|
+      input = "Single line message with new line\n"
+
+      # Generate the input file to convert to .pot block
+      input_path = File.join(dir, 'input')
+      File.write(input_path, input)
+      # Ensure the input has only one line
+      expect(File.read(input_path).lines.count).to eq 1
+
+      # Write the .pot block in an output file
+      output_path = File.join(dir, 'output')
+      File.open(output_path, 'w') do |file|
+        described_class.new('any-key', input_path).generate_block(file)
+      end
+
+      # Ensure the output matches the expectation: the trailing new line has
+      # been stripped.
+      #
+      # Note that the final new line is intentional. It's part of the
+      # formatting at the time of writing.
+      expect(File.read(output_path)).to eq <<~EXP
+        msgctxt "any-key"
+        msgid "Single line message with new line"
+        msgstr ""
+
+      EXP
+    end
+  end
+end
--- a/spec/an_metadata_update_helper_spec.rb
+++ b/spec/an_metadata_update_helper_spec.rb
@@ -12,18 +12,17 @@ describe Fastlane::Helper::StandardMetadataBlock do
       # Ensure the input has only one line
       expect(File.read(input_path).lines.count).to eq 1
 
-      # Write the .pot block in an output file
-      output_path = File.join(dir, 'output')
-      File.open(output_path, 'w') do |file|
-        described_class.new('any-key', input_path).generate_block(file)
-      end
+      # Write the .pot block in a StringIO to bypass the filesystem and have
+      # faster tests
+      output_io = StringIO.new
+      described_class.new('any-key', input_path).generate_block(output_io)
 
       # Ensure the output matches the expectation: the trailing new line has
       # been stripped.
       #
       # Note that the final new line is intentional. It's part of the
       # formatting at the time of writing.
-      expect(File.read(output_path)).to eq <<~EXP
+      expect(output_io.string).to eq <<~EXP
         msgctxt "any-key"
         msgid "Single line message with new line"
         msgstr ""
--- a/spec/an_metadata_update_helper_spec.rb
+++ b/spec/an_metadata_update_helper_spec.rb
@@ -30,4 +30,31 @@ describe Fastlane::Helper::StandardMetadataBlock do
       EXP
     end
   end
+
+  it 'does not strip a trailing new line when generating the block for a multi-line input' do
+    Dir.mktmpdir do |dir|
+      input = "Multi-line\nmessage\nwith\ntrailing new line\n"
+
+      # Generate the input file to convert to .pot block
+      input_path = File.join(dir, 'input')
+      File.write(input_path, input)
+
+      # Write the .pot block in a StringIO to bypass the filesystem and have
+      # faster tests
+      output_io = StringIO.new
+      described_class.new('any-key', input_path).generate_block(output_io)
+
+      # Note that the new line after `msgstr` is intentional. It's part of the
+      # formatting at the time of writing.
+      expect(output_io.string).to eq %q(msgctxt "any-key"
+msgid ""
+"Multi-line\n"
+"message\n"
+"with\n"
+"trailing new line\n"
+msgstr ""
+
+)
+    end
+  end
 end
--- a/spec/an_metadata_update_helper_spec.rb
+++ b/spec/an_metadata_update_helper_spec.rb
@@ -46,15 +46,16 @@ describe Fastlane::Helper::StandardMetadataBlock do
 
       # Note that the new line after `msgstr` is intentional. It's part of the
       # formatting at the time of writing.
-      expect(output_io.string).to eq %q(msgctxt "any-key"
-msgid ""
-"Multi-line\n"
-"message\n"
-"with\n"
-"trailing new line\n"
-msgstr ""
+      expect(output_io.string).to eq <<~'EXP'
+        msgctxt "any-key"
+        msgid ""
+        "Multi-line\n"
+        "message\n"
+        "with\n"
+        "trailing new line\n"
+        msgstr ""
 
-)
+      EXP
     end
   end
 end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -131,6 +131,7 @@ RSpec/FilePath:
     - 'spec/android_localize_helper_spec.rb'
     - 'spec/android_merge_translators_strings_spec.rb'
     - 'spec/android_version_helper_spec.rb'
+    - 'spec/an_metadata_update_helper_spec.rb'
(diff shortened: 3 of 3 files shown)


mokagio:
🤔 Buildkite is failing on this PR because there's no pipeline to upload... But I don't see a PR to configure Buildkite in this repo.

**Update**: There's actually a WIP branch with no PR attached, [this](https://github.com/wordpress-mobile/release-toolkit/commit/5fa9325358174cf0adcc71d24bff8fe2f4f36084) is the latest commit. 👌 

That's all to say, **we can safely ignore the Buildkite failure**.

AliSoftware:
> For a start, an_metadata_update_helper.rb and metadata_update_helper.rb define the same types.

If only they were the only duplications 😓  There are also [remainders of legacy code in some repos themselves (like WPAndroid)](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/fastlane/actions/android_update_metadata_source.rb) from before those actions were moved into the toolkit… We really need to remove those too at some point and clean up everything, but we need to check if the implementations of the actions in the repos (which take precedence over toolkit iinm) match the ones in the toolkit before blindly removing them first… so maybe let's keep that for later?

> The reason is that for some reason the Ruby interpreter loaded the ones defined in an_metadata_helper.rb, possibly due to the an_ prefix which would make that file load first? Although, I'm not sure why that would be the case since there's no import for that file.

There is [a mechanism in the plugin that imports/require every action+helper file automagially though](https://github.com/wordpress-mobile/release-toolkit/blob/develop/lib/fastlane/plugin/wpmreleasetoolkit.rb#L3-L16) – and I think that's the one that explains the behavior you saw.

lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
```diff
@@ -50,7 +50,7 @@ def generate_block(fw)
 
         if line_count <= 1
           # Single line output
-          fw.puts("msgid \"#{File.open(@content_file_path, 'r').read}\"")
+          fw.puts("msgid \"#{File.open(@content_file_path, 'r').read.strip}\"")
```
AliSoftware: 🚶  Sidetrack'd comment: looking at the existing code on lines 49-62 I'm tempted to refactor all that 😅 (tho probably for a later PR)

I see we're using an optimized/uneager `foreach` method to consule the file line by line (great)… but only to consume the whole file anyway once… and then re-consume it again line-by-line in the `else` block.

At that point we might as well read the whole file into an array of lines just once (I doubt we'll have `.txt` files with more than, say, 100 lines, so memory consumption is unlikely to be an issue), and then test `.count <= 1` directly (rather than using `inject`) and iterate on the array…

spec/an_metadata_update_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'tmpdir'
+require_relative './spec_helper'
+
+describe Fastlane::Helper::StandardMetadataBlock do
```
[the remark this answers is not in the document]
AliSoftware: Given there's a huge amount of work to rename our files and fix our directory structure in the plugin in order to follow the ruby dir structure conventions like we were supposed to do from the start, we can put this one in the `.rubocop_todo.yml` file for now until we have time(*) to do proper cleanup of all those

(*) I know, I know… not sure we'll even have that time… 😒

spec/an_metadata_update_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      input_path = File.join(dir, 'input')
+      File.write(input_path, input)
+      # Ensure the input has only one line
+      expect(File.read(input_path).lines.count).to eq 1
+
+      # Write the .pot block in an output file
+      output_path = File.join(dir, 'output')
+      File.open(output_path, 'w') do |file|
+        described_class.new('any-key', input_path).generate_block(file)
+      end
```
AliSoftware: 💄 nit: I think we couldn't use a StringIO instead here to avoid having to write to disk there.
(`StringIO` is a ruby built-in class that is quite useful to mock `IO` operations, as it's a `String` that acts like and has the same methods as an `IO` (e.g. a `File`), but do it all in-memory and not on-disk. We should try to use it more often in our specs imho)

I think something like this should work then (not tested):

```suggestion
      output_io = StringIO.new
      described_class.new('any-key', input_path).generate_block(output_io)
```

Then use `expect(output_io.string).to eq …` below.

AliSoftware (commented):
Thanks for debugging and tackling this!

I left some food for thought in the spec implementation, but not a blocker and I think the fix is worth landing ASAP even if we come back to this code later to clean up things – and especially get rid of the other, legacy and duplicate implementations of the same helper/action.

While you're add it and added a spec, I'd probably suggest to also add another test which covers the multi-line case, for parity. But once this is done (and maybe you take a stab at using `StringIO`) it should be good to merge 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
AliSoftware: @mokagio  Note that `strip` removes both whitespaces and CRLFs, and both from start and end of the string. I'm wondering if we shouldn't use `chomp` instead – which only removes trailing CR/LF but doesn't touches whitespaces, neither leading nor trailing – or `rstrip` – which removes trailing CR/LF _and_ whitespaces, but not leading ones.

```ruby
" - Hello. \n".strip # "- Hello."; notice the leading whitespace being removed too, in addition to trailing \n and whitespace
" - Hello. \n".chomp # " - Hello. "; notice only the trailing \n got removed but we keep whitespaces, both leading and trailing
" - Hello. \n".rstrip # " - Hello."; notice both trailing \n and whitespace got removed, but we keep leading whitespace
```

So TL;DR I think we should use `rstrip` both here and for the multiline case on line 60, and also adapt the test/spec to cover those cases too.

spec/an_metadata_update_helper_spec.rb
mokagio: :TIL: ✨ 

That's neat, thanks. Addressed in 28dba6d.

lib/fastlane/plugin/wpmreleasetoolkit/helper/an_metadata_update_helper.rb
mokagio: Good point. Better to be precise, just in case. Addressed in f7ecdfd.

spec/an_metadata_update_helper_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(output_io.string).to eq <<~EXP
+        msgctxt "any-key"
+        msgid "Single line message with new line"
+        msgstr ""
+
+      EXP
+    end
+  end
+
+  it 'does not strip a trailing new line when generating the block for a multi-line input' do
```
mokagio, line 34: Notice "_does not_ strip..." and the trailing `\n` in the last line. This is the current behavior (e.g.: [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/2386060a1b9990a9a5d170b8e0877f3aa4f98b3a/WordPress/Resources/AppStoreStrings.po#L23-L34)) and, unlike the one this PR aims to fix in the single-line instance, it doesn't generate a validation error. 

Personally, I think we should strip that final `\n`, but I don't want to do it in the context of this PR and without discussing it.

spec/an_metadata_update_helper_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      # formatting at the time of writing.
+      expect(output_io.string).to eq %q(msgctxt "any-key"
+msgid ""
+"Multi-line\n"
+"message\n"
+"with\n"
+"trailing new line\n"
+msgstr ""
+
+)
```
mokagio: I would have liked to use the heredoc syntax like in the other test to avoid the difference in indentation, but that syntax escapes characters, and the comparison against the expected `\n` fails.

mokagio:
@AliSoftware I think I addressed all your comments and suggestions. Let me know what you think.

I got distracted by my support rotation last week, but it would be great to ship a new version with this before the iOS apps need to update their `.po[t]` when the new release notes are available. Thanks 🙇‍♂️ 

cc @oguzkocer if you want to point to this branch during your Simplenote iOS code freeze to test / avoid the manual fix of the `.pot` file.

oguzkocer:
> cc @oguzkocer if you want to point to this branch during your Simplenote iOS code freeze to test / avoid the manual fix of the .pot file.

I think I'll just do it manually once more (hopefully I won't forget it).

spec/an_metadata_update_helper_spec.rb
AliSoftware: FYI, since Ruby 2.3 you can use `<<~'EXPECTED'` – with single-quotes around the heredoc identifier – to prevent interpolation of the content. (and the `~` is to allow indentation of the content to be ignored.

```suggestion
      expect(output_io.string).to eq <<~'EXPECTED'
        msgctxt "any-key"
        msgid ""
        "Multi-line\n"
        "message\n"
        "with\n"
        "trailing new line\n"
        msgstr ""
      EXPECTED
```

See https://rubyreferences.github.io/rubyref/language/literals.html#here-documents-heredocs

This is just a nitpick so not a blocker for this to be merged, but figured you'd be curious to know 😉

spec/an_metadata_update_helper_spec.rb
AliSoftware: Yeah it might be fine to remove the trailing `\n` at the end of the _last_ line but I'm not sure either of the consequences there indeed, let's keep it for now as it doesn't do any harm.

AliSoftware (approved):
Apart from the heredoc tip/suggestion, which isn't a blocker, LGTM!

spec/an_metadata_update_helper_spec.rb
mokagio: Oh YES! I got close to that syntax in my research for a solution yesterday, but not quite there.

PR #299, opened 2021-09-08
Add Buildkite

jkmassel:
This PR adds Buildkite and does some housekeeping, including:
- Adding rubocop in CI
- Requiring that development gems be installed in the `vendor` directory
- Adding `git-lfs` installation (IMHO we should deprecate LFS stuff in the future...)
- Removes the `test` group from the `Gemfile` – if you're using that instead of the `gemspec` you are sort of by definition out to do development work on the toolkit, so you should have all the tooling present.


CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ _None_
 
 ### Internal Changes
 
-_None_
+Adopts Buildkite
```
```suggestion
- Adopt Buildkite
```
I think our past wording was more either infinitive tense or using the 1st person "[We now] adopt" rather than speaking for the tool as "[It] adopts" 🤷‍♂️

Also adding bullet point for consistency with those sections being lists, even if only one item in it so far.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,47 @@
+# Nodes with values to reuse in the pipeline.
+common_params:
+  plugins: &common_plugins
+  - &bash_cache automattic/bash-cache#v1.3.2: ~
+  # Common environment values to use with the `env` key.
```
AliSoftware: Nit:
```suggestion
```
Feels kind of redundant with the previous comment on line 1, and with the intent for those sub-keys to be self-documenting? 🤷‍♂️

Also we don't have such extra comment for plugins so why have it for env? 🙃

Gemfile
```diff
@@ -5,12 +5,9 @@ gemspec
...
 gem 'danger-rubocop', '~> 0.6'
 
-group :test do
-  gem 'codecov', require: false
-  gem 'rspec'
-  gem 'webmock', require: false, group: :test
-  gem 'yard'
-end
+gem 'codecov', require: false
+gem 'webmock', require: false, group: :test
```
AliSoftware:
```suggestion
gem 'webmock', require: false
```
Not sure why the group was repeated in the original code, likely an oversight.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,47 @@
...
+  #################
+  # Lint
+  #################
+  - label: "🧹 Lint"
+    command: |
+      echo "--- :rubygems: Setting up Gems"
+      install_gems
+
+      echo "--- :rubocop: Run Rubocop"
+      bundle exec rubocop
```
AliSoftware, line 43: This is already [handled by GitHub Actions calling Danger calling rubocop — and this doing inline comments on the PR for violations found](https://github.com/wordpress-mobile/release-toolkit/blob/develop/Dangerfile).

Personally despite inline comments being a nice feature I feel this setup has been a bit too noisy in the past (especially when we work on a new feature and add lots of code and totally forget to run rubocop before opening the PR and have lots of small, and often similar, violations, but also because running Danger via GitHub Action doesn't seem to delete past comments once we fix the violations, only marking them as solved, which is nice but still noisy in the case of those typical use cases).

So I'd be totally ok to have Builkite handle rubocop like you did here (with just a check failure and no inline comment on the PR) but then we should get rid of it being run by Danger with inline comments.

Not sure if you planned to get rid of GitHub Actions and Danger in a separate PR but I'd be up for us to do it in this one?

AliSoftware (commented):
Looks good overall but I think we should also get rid of GitHub Actions and Danger in this PR to avoid the double rubocop reporting.

We might also want to get rid of the CircleCI config too on that PR, except if you planned to do that in a separate one as a secondary step?

.buildkite/pipeline.yml
jkmassel: Yeah, my plan was to remove all of that separately, but happy to do it as part of this one 😃

AliSoftware: Ok 👍  whatever you prefer (separate or same PR), as long as it's already in your plans 🙂

mokagio (dismissed):
:shipit:

AliSoftware (dismissed):
The `gem-push` job from CircleCI is missing from the `.buildkite/pipeline.yml`, we need to add it before being ready to remove CircleCI.

AliSoftware:
Note that I noticed while doing the 2.0.0 that even today the `gem-push` job from CircleCI fails because it doesn't install the packages from the `.brewfile` like it does for the `test` job.

Not sure why we set `skip_brew_install: true` on the `gem-push` job (maybe because we expected the `test` job to run beforehand and install it for us? But it doesn't run on tags…) but when we'll port this to BuildKite in this PR we need to ensure the `brew install` step is done during gem push too.

AliSoftware:
### PR update: gem-push step

@jkmassel @mokagio I've updated the PR to add the `gem-push` step to the Buildkite pipeline:

 - Added the `gem-push` step via [these set of commits](https://github.com/wordpress-mobile/release-toolkit/pull/299/files/3f0d80c073c50340671be20a5eb7d7c295b1f15d..ac143b1965a5e4906cb9ed516b27309e934c8f04), with an `if: build.tag != null` condition
 - Added `RUBYGEMS_API_KEY` to the `shared-secrets.json` in mobile-secrets, and updated the `CI/secrets/release-toolkit/env` file to export it, then ran `./build` to upload new secrets to S3. (I used the same API Key that we have been using on CircleCI and which is in our Secret Store – didn't create a new one)

You can check the changes I made to mobile-secrets by running `git -C ~/.mobile-secrets show c20d2d5c6d0302e7d477284a628dd25572fc9174` on your Mac.

### CircleCI Deploy Key re-added (and PR check failure)

I've incidentally [re-added a public key for CircleCI during the release of 2.0.0](https://github.com/wordpress-mobile/release-toolkit/settings/keys) while trying to `gem push` the release `2.0.0` to RubyGems via the existing CircleCI job; which is why we have the `ci/circleci: Build Error` failure in this PR's checks. We'll need to re-delete that deploy key and remove the project from CircleCI once this Buildkite migration is merged to stop it from trying to run anymore.

### Reviews

 - @jkmassel I can't add you as reviewer of this PR because you're also its author, but please review the changes I made for `gem-push` before we can consider merging it.
 - @mokagio I've dismissed your approval of this PR so that we don't accidentally merge it before someone re-reviews the new changes with new `gem-push` step.

AliSoftware:
### Testing the gem-push step

I tested the new `gem-push` step by temporarily disabling the `if build.tag != null` condition via 4f86bece4ca3387fb0e256c09d465ee0d8c3960c to test the command (I expected it to fail on push saying that 2.0.0 already existed, but passing all the previous steps otherwise).

Had to do a lot of back and forth 😓 to test and make `gem push` be able to use authentication as expected 😓 .

 - First I tried to use the same method as what we used on CircleCI, i.e. writing the credentials in `$HOME/.gem/credentials`
 - But then it appeared that [we don't have write access to `$HOME` in our Buildkite images](https://buildkite.com/wordpress-mobile/release-toolkit/builds/92#44304825-da23-437a-b2ae-fa63d6280aa0).
 - Then I tried to use the `GEM_HOST_API_KEY` env var instead of the `~/.gem/credentials` file, [as suggested by the `gem push` help here](https://guides.rubygems.org/command-reference/#gem-push), but apparently it wasn't picked up by CI (while it seemed to work on my Mac… may it depends on the version of `rubygems`?).
 - So after a lot of reading in the docs and finally looking at [the rubygems source code](https://github.com/rubygems/rubygems/blob/5b910cc4a78827c0571252dd93776b559cef608a/lib/rubygems/config_file.rb#L271-L287) on any alternatives to set the API keys (the official docs only mentioned `~/.gem/credentials` as an acceptable file…) I finally found out that using `--config-file` would apparently not only allow to provide default flags for gem commands (original documented purpose, similar to what we can put into a `gemrc` file)… but also allows us to provide the API key via this file.

The last solution [finally "succeeded"](https://buildkite.com/wordpress-mobile/release-toolkit/builds/96#7bc17e00-80f5-49e0-8e99-aced747210cf) (well the step failed but on the error that "version 2.0.0 has already been published" which is the error I expected – as opposed to being blocked on an interactive prompt asking us for our credentials, or failing to authenticate with RubyGems), so I reverted the test commit via ac143b1965a5e4906cb9ed516b27309e934c8f04 to only do the `gem push` on tags.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,74 @@
...
+    agents:
+      queue: "mac"
+
+  #################
+  # Push to RubyGems
+  #################
+  - label: ":rubygems: Publish to RubyGems"
+    key: "gem-push"
+    if: build.tag != null
+    command: .buildkite/gem-push.sh
```
AliSoftware, line 73: Note: it's important for this step to call a separate `.sh` file (as opposed to having all the commands written inline in the `pipeline.yml` file here) because if we were to have [this line from the `.sh` script in particular](https://github.com/wordpress-mobile/release-toolkit/pull/299/files#diff-61d3706214f2ca063dd08939a472f67386bdb9f24a8a31d9698ca96aedc777baR16) be inlined in the `pipeline.yml`, then it would leak the key in clear in the `BUILDKITE_COMMAND` environment variable set by Buildkite (and be visible in the "Environment" tab of the build log)

.buildkite/gem-push.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
```
mokagio: Nitpick

```suggestion
#!/bin/bash -eu
```

or a comment on why we're not using the standard syntax (we should put on a post somewhere, maybe mobile.blog ? on why we use that)

.buildkite/pipeline.yml
mokagio: 6a0e52b resolves this. I'm going to resolve this conversation accordingly to avoid confusion when seeing the wall of text 😄

.buildkite/pipeline.yml
mokagio: Nitpick: it would be good to have this as an inline comment for future us


```suggestion
    # Note: We intentionally call a separate `.sh` script here (as opposed to having all the 
    # commands written inline) to avoid leaking a key used in the process in clear in the
    # BUILDKITE_COMMAND environment variable.
    command: .buildkite/gem-push.sh
```

CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ _None_
 
 ### Internal Changes
 
-_None_
+- Adopt Buildkite
```
mokagio: Nitpick:

```suggestion
- Replace CircleCI and GitHub Actions with Buildkite
```

Gemfile
```diff
@@ -5,12 +5,9 @@ gemspec
...
 
-group :test do
-  gem 'codecov', require: false
-  gem 'rspec'
-  gem 'webmock', require: false, group: :test
-  gem 'yard'
-end
+gem 'codecov', require: false
+gem 'webmock', require: false
+gem 'yard'
```
mokagio, line 10: This LGTM. Just curious, though, did you decide to remove the `test` group because:

- Everyone running `bundle install` on this project is expected to run the tests; and
- Consumer of the gem get the dependencies via what's specified in the `.gemspec`, so there's no need to add groups here because, if we don't add these to the `.gemspec` they won't be part of the gem dependency resolution chain

?

mokagio (approved):
This looks ready to ship. I only left nitpicks 👍 

I assume you intent to keep the test and revert commit pair, as opposed interactive rebase and `drop` them, because there's no mention of removing them. Correct?

Part of me would like to get rid of them to keep the Git history neat, but I can see the value in having them here for the extra context. You did a great job at making it clear they are test commit, so I'm sure if someone landed on them (e.g. during a bisect) they'd understand the state of the codebase is intentionally "incorrect". 👌  

The CircleCI failure is obviously due to this PR removing CircleCI 😄 

image

.buildkite/gem-push.sh
oguzkocer: @mokagio I assume you meant `/bin/bash`?

Also for the `e` flag, I understand that it makes sense in most cases, but there are cases where it's OK & necessary to let the script handle the errors and decide whether to continue or not. I am OK with us defaulting to `eu` flags as a team, but if this was a personal project, I'd only use them when necessary and go with the defaults. 🤷

mokagio: > @mokagio I assume you meant /bin/bash?

Yep. Thank you. I updated the snipped.

Regarding the rest of your comment. My understanding is the same. There are some scripts in which we want to handle errors manually, therefore `#!/bin/bash -e` or `set -e` would not be appropriate. The same goes for `-u` for unassigned variables: we might want to prompt the user about them instead of failing.

I think my suggestion is appropriate in the context of this script, but I'll let @AliSoftware weight in.

Also, maybe my representation of what's our desired standard was incorrect. Should we simply go with "use `bash` instead of `sh`?"

oguzkocer: For scripts, my preference between `sh` & `bash` is always on the `bash` side, unless there is a reason for going with `sh`. (such as bash not being available)

Gemfile
AliSoftware: Yep, exactly. As Jeremy said in the description of this PR:

> Removes the test group from the Gemfile – if you're using that instead of the gemspec you are sort of by definition out to do development work on the toolkit, so you should have all the tooling present.

.buildkite/gem-push.sh
AliSoftware: Fixed via https://github.com/wordpress-mobile/release-toolkit/pull/299/commits/6a1631416397a952bd46d410207993193552f717 by applying your GitHub suggestion (`bash -eu`)

---

I agree that in some cases we might want to handle the errors ourselves and thus `-e` is not the always right choice indeed. But in this case I think it is, as I have no custom error handling in this script, so better for it to stop at the first command that fails, rather than eat the error silently.

---

As for `-u`, it's opinionated but I think this one almost always makes sense, because even with that option, if you do want to handle cases where a variable is not set, or provide default values when a var is not provided by the user, there are safer ways to handle those (I love [this cheatsheet](https://devhints.io/bash) for good tips).
So IMHO it's better to use `-u` – which helped me catch typos in env vars in the past and find bugs many times – and use `MYVAR=${1:-default}` if you want to provide default values, rather than not use `-u` and risk issues (we've had a few in our Build Script phases of iOS projects in the past due to that `-u` missing and typos slipping). But that's just my own opinion 🤷

AliSoftware:
> I assume you intent to keep the test and revert commit pair, as opposed interactive rebase and drop them, because there's no mention of removing them. Correct?

Yep.

I'm not strong about this and I'm ok either case, and I thought about rebasing to eliminate those 2 commits too instead of keeping them. But as you say it also helps keep context – especially if someone comes back to this PR and the conversations in it even after the PR got merged – also because I do mention those commits' SHA1 in some of my comments, so removing them would make the comment outdated and odd to read in the future.

Overall even if I do like a clean and nice git graph and history, I don't think it's that a big deal to include those, both to keep context for this PR but also to show in the git history that we did test this scenario, and maybe show future us how we test similar stuff if we need more similar changes to be tested later.

AliSoftware:
> The CircleCI failure is obviously due to this PR removing CircleCI 😄

I've just hit the "Stop Building" red button on the [project settings in CircleCI](https://app.circleci.com/settings/project/github/wordpress-mobile/release-toolkit) which has [removed the CircleCI Deploy Key from this repo's settings in GitHub](https://github.com/wordpress-mobile/release-toolkit/settings/keys), so CircleCI will no longer try to build this project (and fail).

PR #303, opened 2021-10-19
Release 2.0.0 into develop

AliSoftware:
New version 2.0.0

PR #304, opened 2021-10-19
Release 2.0.0 into trunk

AliSoftware:
New version 2.0.0. 

Next Steps once this PR gets merged (ref: paaHJt-2d8-p2):
 - Create a GitHub Release and tag (pointing to `trunk`)
 - Announce the new tagged version to our build-and-ship channel

---

Side note: updating the client apps to this 2.0.0 once released should be straightforward as, even if it's a major version (with breaking changes), the only changes that this breaks were only used in WPAndroid (which is already ready for this version), not any other client app.


mokagio (approved):
:shipit: 

See also https://github.com/wordpress-mobile/release-toolkit/pull/303#pullrequestreview-783986229 regarding the Buildkite "failures".

AliSoftware:
I had to release the gem (`gem push`) manually, by restarting the CircleCI job with SSH then connecting to the image with SSH and re-running the commands from the `gem-push` job manually, after installing the brewfile.

This is because the job is currently configure to skip the `brew install` during `gem-push`, while we need those brew formulae to be installed so we can `gem build` and `gem install` the gem before `gem push`. So [the compilation failed during installation of `rmagick` due to `pkg-config` not being installed](https://app.circleci.com/pipelines/github/wordpress-mobile/release-toolkit/1097/workflows/e37683bf-b106-4a9a-a5aa-2211176d118a/jobs/1081).
I am not sure why we configured the `gem-push` job to skip `brew install` back then and how it has been working in the past.

We'll have to fix this at some point, likely during the migration to BuildKite – See https://github.com/wordpress-mobile/release-toolkit/pull/299#issuecomment-947686070 for additional details.

PR #305, opened 2021-10-22
Release 2.1.0 into develop

AliSoftware:
New version 2.1.0

There's not much new compared to `2.0.0` but it still felt like a good idea to do a new release so soon because:

 - I'm planning to start work on Localization toolchain (ref: paaHJt-2Ib-p2), which means there might be multiple big (potentially breaking?) changes coming soon; so I figured better do a release before starting that large work.
 - Since we just migrated our CI to Buildkite, this was a good occasion to test the `gem push` job on Buildkite early, and fix any issue sooner rather than later if there's any, instead of encountering them while I'll be working on bigger changes.

[EDIT] In fact after opening this PR, the CI failed because we forgot to install some dependencies. Not sure why it worked in #299 but not here now that I look at the pipeline 🤔 . So I took the occasion to fix CI via fd1a15d too, hence the extra changes in this diff.

### DIFF
--- a/.buildkite/install-dependencies.sh
+++ b/.buildkite/install-dependencies.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -eu
+
+echo "--- :beer: Installing Dependencies"
+brew bundle --file .buildkite/brewfile
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
--- a/.buildkite/gem-push.sh
+++ b/.buildkite/gem-push.sh
@@ -1,10 +1,6 @@
 #!/bin/bash -eu
 
-echo "--- :beer: Installing Dependencies"
-brew bundle --file .buildkite/brewfile
-
-echo "--- :rubygems: Setting up Gems"
-install_gems
+./install-dependencies.sh
 
 echo "--- :hammer: Build Gemspec"
 gem build fastlane-plugin-wpmreleasetoolkit.gemspec
--- a/.buildkite/gem-push.sh
+++ b/.buildkite/gem-push.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -eu
 
-./install-dependencies.sh
+.buildkite/install-dependencies.sh
 
 echo "--- :hammer: Build Gemspec"
 gem build fastlane-plugin-wpmreleasetoolkit.gemspec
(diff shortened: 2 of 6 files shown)


AliSoftware:
> An error occurred while installing rmagick (4.2.3), and Bundler cannot continue.

😡 _(at least I'm glad I'm testing this before starting working on bigger changes…)_

[EDIT] Fixed via fd1a15d

.buildkite/install-dependencies.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -eu
+
+echo "--- :beer: Installing Dependencies"
+brew bundle --file .buildkite/brewfile
```
AliSoftware, line 4: Note that this takes a bit of time to run on CI – especially since it's done by each separate step, as each one runs in a separate VM, so each step installs brew dependencies individually – and would benefit some caching.

That being said, caching brew installs, unlike other dependencies like gems or pods, has proven quite tricky and not always trivial, especially since it installs stuff system-wide, and not always all in a single location or folder we could cache. My past experience at trying to cache brew installs have often led to quirks, issues and instabilities (depending on the formulae installed and what they install and where on the system), which is why I didn't spend time on this yet.

AliSoftware:
2.1.0 has now been [released](https://github.com/wordpress-mobile/release-toolkit/releases/tag/2.1.0) and [pushed to RubyGems](https://buildkite.com/wordpress-mobile/release-toolkit/builds/111) 🎉 .

Changes have already been [merged into `trunk`](https://github.com/wordpress-mobile/release-toolkit/pull/306) where the 2.1.0 tag was created, so merging into `develop` (this PR) should ideally be done soon to sync everything.

oguzkocer (approved):
Admittedly I still don't have much familiarity with the release-toolkit, so I am always reluctant to review these PRs. However, since the tag is already created and everything in this PR seems reasonable, I am going to approve it and leave it to @AliSoftware to decide whether to merge it or wait for another review.

.buildkite/install-dependencies.sh
mokagio: I had trouble with that in the past, too.

I don't think it's a big issue for the moment, albeit it's admittedly "wasteful" to do that every time.

I wonder if it would be too much of a stretch to provision our VMs with those tools, since all the steps here run on those?

mokagio (approved):

PR #309, opened 2021-11-18
Add ios_generate_strings_file_from_code action & specs

AliSoftware:
[ Part of project ref paaHJt-2Ib-p2 ]

## What

Adds a new `ios_generate_strings_file_from_code` action and its specs & test data.

This is aimed to replace our `localize.py` script duplicated in all our app repos

## Draft… but Ready for first review

This PR is still a Draft because once this new action is validated, I will deprecate the `IosLocalizeProjectAction` (and the `self.localize_project` method from `ios_git_helper.rb` it calls) which currently end up calling the `localize.py` script in the client app repo. Before doing so I wanted to give us the opportunity to test this new action's basics.

### Reviewing the Unit Tests and covered cases

To test at a high level, verify that:
 - The Unit Tests Pass on CI, and rubocop is happy
 - The Unit Tests I wrote cover all the use cases you can think of, especially the use cases that we could have already encountered when running `localize.py` vs if we ran this action on the apps' codebase.
 - The documentation is clear enough.

### Testing on an app repo

 - Create a test branch, point your `fastlane/Pluginfile` to this branch of the toolkit
 - Run `Scripts/localize.py` from your app repo, commit the changes in the `Localizable.strings` generated
 - Add a lane to your `Fastfile` to call this new action, and make sure you not only pass `WordPress` but also the paths to `Pods/` as the list of files to scan – ideally only explicitly listing the a8c Pods you want to include (instead of everything in Pods)
 - Compare the generated `.strings` file from this action with the ones generated by `localize.py` and ensure they match (†)

ℹ️  PS: I haven't personally run those steps above (aka testing my action on an app repo) yet, which is also why I'm opening this Draft PR early so that you can help me test that part while I wrap things up (i.e. deprecate the old action, etc)

### Merge dance in `Localize.py` not replicated in the new action (by design)

See paaHJt-2KK-p2 (search `merge_with` in the doc): the `Localize.py` script does this merge dance to keep any English translation already present in the local `.strings` that would differ from the generated one; this new fastlane action doesn't, because we expect to get rid of the need for this by fixing the apps codebase to use `genstrings` as Apple intended.

(†) I know of 2 keys in WPiOS which currently don't have the same key vs value in the existing `.strings` file, and thus for which `localize.py` handles the case by merging the generated `.strings` file with the existing one (giving values from the latter the priority) – while my new action does not handle such merging logic and would override the translation from the existing `.strings` file.

This is by design, as I expect to fix this not via the tooling (aka by implementing the merge logic from `localize.py` in this new action) but instead by fixing the app codebase, using `NSLocalizedString("key", value: "value")` when we need key ≠ value, avoiding to put the burden and complexity on the tooling and instead using build-in mechanisms already provided by iOS and `genstrings`.

This is why you should expect a diff on those 2 strings when comparing the output of running `localize.py` vs this action on WPiOS, which should disappear once we fix the WPiOS codebase appropriately.

### Parsing Pods option now more explicit (by design)

 - In `Localize.py`, the main method has an `include_pods_and_frameworks` boolean parameter.
 - In the new action, instead of a boolean, I've opted to let the caller provide an arbitrary list of paths to parse. It's now up to the caller to provide e.g. `['WordPress', 'Pods/`]` as the list of files – or even better, `['WordPress', 'Pods/WordPressLogin', 'Pods/WordPressUI', …]` to be more explicit about which Pods we want to translate the strings of via the app translation pipeline

### Generation of Widgets' base keys for reference not replicated (by design)

See also paaHJt-2KK-p2 – especially notes about 4️⃣ in the details of each script's role

In addition to generating the `.strings` file in `WordPress/Resources` for GlotPress to import, the `Localize.py` script also [generates the `Base.lproj/Localizable.strings` files for 2 of the Widget targets](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/Scripts/localize.py#L156-L157).
Those 2 additional files were not imported by GlotPress, but were instead user later in the process, when we download the translations from GlotPress, to compare which keys from those translations we need to "redistribute" to the widgets, by comparing the keys in those 2 generated `.strings` files from `localize.py` with the keys pulled from GlotPress

This behaviour is not reproduced in this action, because we plan to address this once again by a change in the codebase – especially by making the widgets instead use the strings from the app target at runtime, using `NSLocalizedString(…, bundle: theHostAppBundle, …)` in the widgets' codebase instead of defaulting to looking the strings in their own bundle.

Especially see Step 2 of paaHJt-2J8-p2 for explanations on that plan.
I'll have to explicitly test that use case when I get to implementing Step 2 of the plan though, I'm not there yet.

### No file exclusion option in the new action (bug; will add later)

While writing this PRs description and list of differences I just realised that this action doesn't not allow for providing exceptions to the list of files to be parsed, while in `localize.py`, [we do a bunch of `grep -v` pipes after the `find` in oder to exclude some files](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/Scripts/localize.py#L137). This possibility is missing from the current implementation of my new action as of today. I'll see to add that feature in a subsequent commit on this PR tomorrow.


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
+require 'open3'
+
+module Fastlane
+  module Actions
+    class IosGenerateStringsFileFromCode < Action
+      def self.run(params)
+        globbed_paths = params[:paths].map { |p| File.file?(p) ? p : "#{p}/**/*.{m,swift}" }
```
AliSoftware: 💡 I should probably use `File.join` here, especially to handle cases where `p` could already end in `/`

```suggestion
        globbed_paths = params[:paths].map { |p| File.file?(p) ? p : File.join(p, '**', '*.{m,swift}') }
```

I should also support more cases of user input for this param, like if they pass values like `Foo/**/` or `Bar/**/*.m` (which would return `false` to `File.file?` because of it containing `**`)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+        'Use genstrings to generate the .strings files from your ObjC and Swift code.'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :paths,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_PATHS',
+                                       description: 'Array of directory paths to scan for `.m` and `.swift` files. The entries can contain glob patterns too.',
+                                       type: Array,
+                                       default_value: ['.']),
```
AliSoftware: I just realised that I'll need to add an `exclude_paths` options to replicate the various `grep -v` we do in `localize.py`. Will push a commit to fix that later (hopefully tomorrow).

spec/ios_generate_strings_file_from_code_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    Dir.mktmpdir('a8c-wpmrt-ios_generate_strings_file_from_code-') do |tmp_dir|
+      # Act
+      genstrings_logs = described_class.run(paths: paths_to_scan, quiet: quiet, swiftui: swiftui, output_dir: tmp_dir)
+
+      output_files = Dir[File.join(tmp_dir, '*.strings')]
+      expected_files = Dir[File.join(test_data_dir, expected_dir_name, '*.strings')]
+
+      # Assert: same list of generated files
+      expect(genstrings_logs).to eq(expected_logs) unless expected_logs.nil?
+      expect(output_files.map { |f| File.basename(f) }.sort).to eq(expected_files.map { |f| File.basename(f) }.sort)
```
AliSoftware: Move comment next to the appropriate line it is commenting:
```suggestion
      # Assert: stdout/stderr warning messages from genstrings
      expect(genstrings_logs).to eq(expected_logs) unless expected_logs.nil?
      # Assert: same list of generated files
      expect(output_files.map { |f| File.basename(f) }.sort).to eq(expected_files.map { |f| File.basename(f) }.sort)
```

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Sources/AppSampleView.swift
```diff
@@ -0,0 +1,23 @@
+import SwiftUI
+
+struct AppSampleView: View {
+    var body: some View {
+      VStack {
+        Text("app.swiftui.key1") // You should always provide a comment, not providing one like here is bad practice
```
mokagio: Nitpick.

I got confused by a comment say that the code is not providing a comment 🤔 Then I realized, looking at the line below, it refers to the `comment` argument.

How about this?

```suggestion
        // You should always provide a `comment` argument to `Text`. Not providing one like here is bad practice, we 
        // are doing it only to verify the localization implementation.
        Text("app.swiftui.key1")
```

spec/ios_generate_strings_file_from_code_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'spec_helper'
+require 'tmpdir'
+
+describe Fastlane::Actions::IosGenerateStringsFileFromCode do
+  let(:test_data_dir) { File.join(File.dirname(__FILE__), 'test-data', 'translations', 'ios_generate_strings_file_from_code') }
+  let(:app_src_dir) { File.join(test_data_dir, 'sample-project', 'Sources') }
+  let(:pod_src_dir) { File.join(test_data_dir, 'sample-project', 'Pods') }
```
mokagio: Nitpick: 

```suggestion
  let(:pods_src_dir) { File.join(test_data_dir, 'sample-project', 'Pods') }
```

spec/ios_generate_strings_file_from_code_spec.rb
mokagio: One advantage of doing this check explicitly, instead of simply checking the `length` of the two arrays and relying on the enumeration below to ensure each `output_file` has a match, is that if there's a failure, we'll get a list of the base names to compare 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+        # Return the warnings as an Array
+        out.split("\n")
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        'Generate the .strings files from your ObjC and Swift code'
```
mokagio: Nitpick
```suggestion
        'Generate the .strings files from your Objective-C and Swift code'
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      def self.details
+        'Use genstrings to generate the .strings files from your ObjC and Swift code.'
```
mokagio: Same nitpick:

```suggestion
        'Use genstrings to generate the .strings files from your Objective-C and Swift code.'
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+                                       type: Array,
+                                       default_value: ['.']),
+          FastlaneCore::ConfigItem.new(key: :quiet,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_QUIET',
+                                       description: 'In quiet mode, genstrings will log warnings about duplicate values, but not about duplicate comments',
+                                       is_string: false, # Boolean
+                                       default_value: true),
+          FastlaneCore::ConfigItem.new(key: :swiftui,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_SWIFTUI',
+                                       description: 'Should we include SwiftUI\'s `Text()` when parsing code with `genstrings`?',
```
mokagio: Why did you chose to use single quotes as opposed to 

```suggestion
                                       description: "Should we include SwiftUI's `Text()` when parsing code with `genstrings`?",
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+      def self.output
+        'Generates .strings files (especially Localizable.strings but could generate more if the code uses custom tables)'
+      end
+
+      def self.return_value
+        'List of warnings generated by genstrings on stdout'
+      end
+
+      def self.authors
+        ['AliSoftware']
```
mokagio: Nitpick. I remember we discussed this but I'm not sure if we ever came to a conclusion. Should we use something like `"Automattic"` here?

FWIW, I'm more than fine for you to be the official author, given you're the one working on this. 🤷  In the long run, if we see this as an action that we could share stand-alone with the community, "Automattic" or "WordPress" might be better.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      def self.is_supported?(platform)
+        platform == :ios
```
mokagio: Should `:mac` be in here, too?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+module Fastlane
+  module Actions
+    class IosGenerateStringsFileFromCode < Action
+      def self.run(params)
+        globbed_paths = params[:paths].map { |p| File.file?(p) ? p : "#{p}/**/*.{m,swift}" }
+        files = Dir.glob(globbed_paths)
+
+        flags = [('-q' if params[:quiet]), ('-SwiftUI' if params[:swiftui])].compact
+
+        out, status = Open3.capture2e('genstrings', '-o', params[:output_dir], *flags, *files)
```
mokagio: This LGTM, is simple enough, and code that is easy to follow. Well done.

How do you feel about having the core of the action logic, which in this case would be the lines I highlighted, extracted in a dedicated object? 

I'm more interested in the approach in general, as it's possibly overkill here. When I worked on the `configure_*` actions, I wished the logic was extracted because it does so much in a single `run` call and it was hard to test in isolation.

Or, I guess we can just be happy with higher-abstraction, functional-like tests like the ones from this PR 🤔

mokagio: Another reason I'm wary of testing `Action` subclasses directly is that `UI.` calls can leak into the tests results, unless we explicitly account for it in the spec, e.g. with `allow(UI).to receive...`

```
➜ be rspec spec/ios_generate_strings_file_from_code_spec.rb

Fastlane::Actions::IosGenerateStringsFileFromCode
  when including pods
    Generates the expected .strings files with SwiftUI support
    Generates the expected .strings files without SwiftUI support
  when not including pods
    Generates the expected .strings files with SwiftUI support
    Generates the expected .strings files without SwiftUI support
  when genstrings finds warnings
    only logs warnings about multiple values in quiet mode
Logging disabled while running tests. Force them by setting the DEBUG environment variable 👈👈👈👈👈
    logs warnings about both duplicate values and duplicate comments if not in quiet mode
```

```
➜ DEBUG=1 be rspec spec/ios_generate_strings_file_from_code_spec.rb

Fastlane::Actions::IosGenerateStringsFileFromCode
  when including pods
    Generates the expected .strings files with SwiftUI support
    Generates the expected .strings files without SwiftUI support
  when not including pods
    Generates the expected .strings files with SwiftUI support
    Generates the expected .strings files without SwiftUI support
  when genstrings finds warnings
    only logs warnings about multiple values in quiet mode
👇👇👇👇
[15:24:16]: ▸ Key "app.key5" used with multiple values. Value "app value 5\nwith multiple lines." kept. Value "app value 5\nwith multiple lines, and different value than in Swift" ignored.
[15:24:16]: ▸ genstrings: warning: Key "app.key5" used with multiple comments "App key 5, with value, custom table and placeholder." & "Duplicate declaration of App key 5 between ObjC and Swift,and with a comment even spanning multiple lines!"
[15:24:16]: ▸ genstrings: warning: Key "pod.key5" used with multiple comments "Duplicate declaration of Pod key 5 between ObjC and Swift,and with a comment even spanning multiple lines!" & "Pod key 5, with value, custom table and placeholder."
👆👆👆👆
    logs warnings about both duplicate values and duplicate comments if not in quiet mode
```

Granted, that's only a minor annoyance, but it can make for noisy test outputs at times.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+                                       default_value: true),
+          FastlaneCore::ConfigItem.new(key: :output_dir,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_OUTPUT_DIR',
+                                       description: 'The path to the directory where the generated .strings files should be created.',
+                                       type: String),
+        ]
+      end
+
+      def self.output
+        'Generates .strings files (especially Localizable.strings but could generate more if the code uses custom tables)'
```
mokagio: One question: Some of the strings here use backticks around code, file extensions, and `genstrings` while other don't. Is there a rationale for when to use them and when not to?

spec/ios_generate_strings_file_from_code_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  context 'when genstrings finds warnings' do
+    it 'only logs warnings about multiple values in quiet mode' do
+      expected_logs = [
+        %(Key "app.key5" used with multiple values. Value "app value 5\\nwith multiple lines." kept. Value "app value 5\\nwith multiple lines, and different value than in Swift" ignored.),
+      ]
+      test_genstrings(paths_to_scan: [app_src_dir, pod_src_dir], quiet: true, swiftui: true, expected_dir_name: 'expected-pods-swiftui', expected_logs: expected_logs)
+    end
+
+    it 'logs warnings about both duplicate values and duplicate comments if not in quiet mode' do
```
mokagio: Question. Have you considered using another `context` nesting?

```ruby
  context 'when genstrings finds warnings' do
    context 'when in quiet mode' do
      it 'only logs warnings about duplicate values' do
      end
    end

    context 'when not in quiet mode' do
      it 'logs warnings about both duplicate values and duplicate comments' do
```

I'm not sure how I feel about multiple `context` either. It does make the test structure clearer, but contributes to the pyramid of doom ⚖️ 🤷

mokagio: Also, nitpick. If I read this correctly, `genstring` will warn us always using "multiple", not "duplicated":

```suggestion
    it 'logs warnings about both multiple values and multiple comments if not in quiet mode' do
```

This would also make the example description consistent with the one above.

mokagio (commented):
Structure and code LGTM. Posting my observations as a comment review before moving on to running the manual integration tests

mokagio:
@AliSoftware 

> In the new action, instead of a boolean, I've opted to let the caller provide an arbitrary list of paths to parse. It's now up to the caller to provide e.g. `['WordPress', 'Pods/]` as the list of files – or even better, `['WordPress', 'Pods/WordPressLogin', 'Pods/WordPressUI', …]` to be more explicit about which Pods we want to translate the strings of via the app translation pipeline

+1 from me. Being explicit costs more on setup and maintenance time, but makes the call-site easier to understand and makes the code safer under the hood. The setup and edit vs reading and running of the code ratio is heavily skewed towards reading and running, making this is a good trade 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      def self.details
+        'Use genstrings to generate the .strings files from your ObjC and Swift code.'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :paths,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_PATHS',
+                                       description: 'Array of directory paths to scan for `.m` and `.swift` files. The entries can contain glob patterns too.',
```
mokagio: When I run `bundle exec fastlane action ios_generate_strings_file_from_code`, I got this error:

```
[!] Do not let descriptions end with a '.', since it's used for user inputs as well for key :paths
```

Screen Shot 2021-11-19 at 3 46 22 pm

The command return a non 0 exit code.


```suggestion
                                       description: 'Array of directory paths to scan for `.m` and `.swift` files. The entries can contain glob patterns too',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+                                       is_string: false, # Boolean
+                                       default_value: true),
+          FastlaneCore::ConfigItem.new(key: :swiftui,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_SWIFTUI',
+                                       description: 'Should we include SwiftUI\'s `Text()` when parsing code with `genstrings`?',
+                                       is_string: false, # Boolean
+                                       default_value: true),
+          FastlaneCore::ConfigItem.new(key: :output_dir,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_OUTPUT_DIR',
+                                       description: 'The path to the directory where the generated .strings files should be created.',
```
mokagio:
```suggestion
                                       description: 'The path to the directory where the generated .strings files should be created',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
mokagio: Interestingly, Fastlane doesn't complain about the final `?` in the `swiftui` parameter 😄

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
mokagio: I had to make this into an array to make `bundle exec fastlane action ios_generate_strings_file_from_code` work locally.

```suggestion
        ['Generates .strings files (especially Localizable.strings but could generate more if the code uses custom tables)']
```

Otherwise, I'd get this error

```
/Users/gio/Developer/a8c/wpios/vendor/bundle/ruby/2.7.0/gems/fastlane-2.195.0/fastlane/lib/fastlane/documentation/actions_list.rb:155:in `print_output_variables': \e[0;31;49m[!] undefined method `map' for #<String:0x00007fdab597b6b0> (NoMethodError)
Did you mean?  tap\e[0m
Did you mean?  tap
```

mokagio (commented):
🤔 even after fixing the documentation generation errors, I'm still unable to run the lane...

```
[16:00:29]: ------------------------------
[16:00:29]: --- Step: default_platform ---
[16:00:29]: ------------------------------
[16:00:30]: ------------------------------
[16:00:30]: --- Step: default_platform ---
[16:00:30]: ------------------------------
[16:00:30]: Driving the lane 'test_309' 🚀
[16:00:30]: -------------------
[16:00:30]: --- Step: is_ci ---
[16:00:30]: -------------------
[16:00:30]: ---------------------------------------
[16:00:30]: --- Step: check_for_toolkit_updates ---
[16:00:30]: ---------------------------------------
[16:00:30]: Currently using release toolkit version 2.1.0.
[16:00:30]: Checking for updates now...
[16:00:32]: Your release toolkit is up-to-date! ✅
[16:00:32]: -------------------
[16:00:32]: --- Step: is_ci ---
[16:00:32]: -------------------
[16:00:32]: -----------------------------
[16:00:32]: --- Step: setup_circle_ci ---
[16:00:32]: -----------------------------
[16:00:32]: Not running on CI, skipping CI setup
+------------------+----------+
|        Lane Context         |
+------------------+----------+
| DEFAULT_PLATFORM | ios      |
| PLATFORM_NAME    |          |
| LANE_NAME        | test_309 |
+------------------+----------+
[16:00:32]: Could not find action, lane or variable 'ios_generate_strings_file_from_code'. Check out the documentation for more details: https://docs.fastlane.tools/actions
```

I pushed my test branch [here](https://github.com/wordpress-mobile/WordPress-iOS/tree/test/release-toolkit-309). I might be missing something,but I'm not sure... it's a very basic setup.

What's even more strange, is that Fastlane does find the action when generating the index of extra actions the plugin offers:

```
+-----------------------------------+---------+-------------------------------------------+
|                                      Used plugins                                       |
+-----------------------------------+---------+-------------------------------------------+
| Plugin                            | Version | Action                                    |
+-----------------------------------+---------+-------------------------------------------+
| fastlane-plugin-wpmreleasetoolkit | 2.1.0   | configure_download, configure_setup,      |
|                                   |         | configure_apply, configure_update,        |
|                                   |         | configure_add_files_to_copy,              |
|                                   |         | configure_validate,                       |
|                                   |         | ios_generate_strings_file_from_code,      |
                                                👆
```

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Sources/AppClass2.m
```diff
@@ -0,0 +1,27 @@
+#import <Foundation/Foundation.h>
+
+@interface AppClass2: NSObject
+@end
```
AliSoftware: Move this to a separate `.h` (and add an import of the `.h` at the top of the `.m`) then add some `// NSLocalizedString` in the `.h` as a way to ensure that `genstrings` won't pick up occurrences from `.h` files but only `.m` and `.swift`. Also add an `.md` file with a similar trap.

AliSoftware:
> +1 from me. Being explicit costs more on setup and maintenance time, but makes the call-site easier to understand and makes the code safer under the hood. The setup and edit vs reading and running of the code ratio is heavily skewed towards reading and running, making this is a good trade 👍

Also note that the long-term plan will be to stop translating the Pods (aka a8c libraries) via the app translation pipeline, and instead have each a8c library (which contains localizable strings) have their own GlotPress project, translation pipeline, and `.strings` file to be independant, only translated once, and most importantly usable in OSS without depending on the client app integrating them.

For more info about this long-term plan, see pxLjZ-6Fc-p2 ; this work will not start until phase 2 of the project in 2022 though, so it's important to keep the ability to add Pods to the scanned source code for now, but that also participated in the rationale for moving away from an `include_pods_and_frameworks` boolean 🙂

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Yeah I'm balanced about this.

On one end I totally agree that it's better to test this in isolation.
On the other end, I thought about creating a helper but then realized it felt kinda pointless to have a brand new helper file just for one method with 4 lines of code.

And also, it might actually be interesting to test the action itself, especially to test those `UI.message` outputs.
For example, I just realized that there was an unbalance in the `:quiet` parameter, as it both acts on the `-q` flag on `genstrings` (which either logs only multiple values, or both values and comments, but never deactivate all warnings) and on the `UI.command_output`. So `quiet: true` will set `-q` but also not print anything in stdout (which kinda renders the use of `-q` pointless)… Something I'd have caught with a dedicated test.
(I'll probably add a commit to change the code to provide a `log_level:{0,1,2}` instead of `quiet:{true,false}`)

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Sources/AppSampleView.swift
AliSoftware: Addressed in 87e97cefdef1d0ad2026aee6facd8bb6d1c86a2a (with even slightly more descriptive wording)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: I think `rubocop` chose that for me 😄  I'll change the wording to get rid of the `?` anyway while also addressing https://github.com/wordpress-mobile/release-toolkit/pull/309/files#r752860056.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: > Is there a rationale

Yep, and a pretty simple one at that. The reasoning is that if I wrote that code before my morning coffee or late after a busy day, then I don't add backticks, and if I'm caffeinated enough then I add them. :trollface: 

I'll fix the consistency 🙂

spec/ios_generate_strings_file_from_code_spec.rb
AliSoftware: Yeah I went back and forth in how to structure that.

Similarly for the context nesting I used for Pods On/Off + SwiftUI On/Off, I am not sure how to split that nicely, given it's a 2x2 matrix of options to test, but we can only choose one axis when nesting in one `context` – or if we nest in 2 `context` levels, one per axis, we'll have to privilege one axis (e.g. Pods) over the other (e.g. SwiftUI).

AliSoftware (commented):
Forgot to submit my own code-review with notes for self. here goes.

spec/ios_generate_strings_file_from_code_spec.rb
AliSoftware: Addressed in ade6f1ecd52fe8d10243ec17e03dbda3a28e1530

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Fixed in 2b4d317230f1383dcec12361441a8f0c52ce5265

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Fixed in 2b4d317230f1383dcec12361441a8f0c52ce5265

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,74 @@
...
+          FastlaneCore::ConfigItem.new(key: :output_dir,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_OUTPUT_DIR',
+                                       description: 'The path to the directory where the generated .strings files should be created',
+                                       type: String),
+        ]
+      end
+
+      def self.output
+        ['Generates .strings files (especially Localizable.strings but could generate more if the code uses custom tables)']
+      end
```
AliSoftware, line 78: Ok after looking at the source, it seems that `self.output` is expected to `# Return the keys you provide on the shared area` (aka `SharedValues::…` iiuc), so it's not expected to return readable text, but key names. Which is not relevant to this action, so I've deleted it in b3d0ce3aec3053ff322e83aaeff4c127c21d9c99.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Fixed in c96a2eb8ee20e7abb36fd5b36fb127865ffec0e1

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Fixed in bf6177a6075749d76d85bec4d456fdb3d40660e3

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Ok so with the latest commits I pushed, I went with:
 - Switching from `Open3.capture2e` to `sh` – which is the more idiomatic way `fastlane` suggest to call shell commands anyway, and has the benefit to handle the output verbosity for us, etc
 - Which is why I kept the `:quiet` parameter as-is after all (and not `:log_level`) as it now only affects the `-q` more of `genstrings`, while the output verbosity is managed by fastlane itself.
 - I decided to still keep testing the action itself (as opposed to separate the logic in a dedicated object. Even if I moved the code into a dedicated object, it would still need to call `sh` to invoke the `genstrings` command anyway, which would try printing stuff to stdout anyway even if invoked from an helper object, so that wouldn't help with that

I might still be swayed into moving the code from the action into a dedicated helper object now that the code has changed (and that added helper methods `glob_pattern` and `files_matching` in the action) if you really feel strongly about it… but since that would not change the fact that `UI.command_output` will still be called internally by `sh` (and the "Logging disabled" warnings will still appear), as the test still has to actually call `genstrings` anyway. So given that, I feel like it might not be worth creating a helper in `helpers/` and having to switch back and forth between the 2 files while maintaining this code, just for so few lines of code, as opposed to keeping it as helper methods in the Action. WDYT?

spec/ios_generate_strings_file_from_code_spec.rb
AliSoftware: I've reorganized the test examples to be grouped differently and adapted the nesting accordingly (10b57fb2fef5e555306a7bf238d83510ec53710b); I feel it now makes more sense that way, but curious to hear what you think of the new spec structure.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Addressed in b266bbec7332feca1cfbdb79b1046a6f6d5d0ceb

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: Addressed in fc4cecf4944f4c3edc3acc5bc3fa5fd3f9d6b51d

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Sources/AppClass2.m
AliSoftware: Addressed in a6d3a652602cf6afe4e22dc97ddaa98d11a864bd

AliSoftware:
> I pushed my test branch here. I might be missing something,but I'm not sure... it's a very basic setup.
> 
> What's even more strange, is that Fastlane does find the action when generating the index of extra actions the plugin offers:

I tested on my end (by doing the steps myself, not by looking at your test branch) and I experienced the exact same error, with fastlane not finding the lane despite it being listed in the ones loaded and found in the plugin… I'm quite puzzled as to why this is.

I'll call it a day (and a week) and let the weekend rest my brain and try to debug it in fastlane's own code on Monday.

AliSoftware:
Got an epiphany after coming back from my evening out, and fixed what was making it crash in 03faa895a99e0445fa8a0f6573c194ccbb4eabab.

Also fixed a couple of other things, like:
 - don't print the whole `genstrings` command unless in `--verbose` mode – because with the huge list of files we parse the command is huge
 - Disable `-SwiftUI` by default, because this mode of `genstrings` super buggy after all (see 5e72f8aae8a7f079f05545555f002344465383c4 and its commit message for details)

---

After those changes, I tested the action in WPiOS using this lane:
```ruby
lane :genstrings do
    ios_generate_strings_file_from_code(
      paths: ['WordPress/', 'Pods/WordPress*/', 'Pods/WPMediaPicker/', 'WordPressShared/WordPressShared/', 'Pods/Gutenberg/'],
      exclude: ['*Vendor*', 'WordPress/WordPressTest/I18n.swift', 'WordPress/WordPressStatsWidgets/Views/Localization/LocalizedStringKey+extension.swift', '*/Secrets.swift'],
      output_dir: 'WordPress/Resources/en.lproj'
    )
  end
```

After running `Scripts/localize.py`, commiting, then running the lane after that, the diff in the `Localizable.strings` generated by the new action only concerned the 2 keys I already mentioned in the PR description:
```diff
-  "Siri Reset Confirmation" => "Successfully cleared Siri Shortcut Suggestions"
-  "Siri Reset Prompt" => "Clear Siri Shortcut Suggestions"
+  "Siri Reset Confirmation" => "Siri Reset Confirmation"
+  "Siri Reset Prompt" => "Siri Reset Prompt"
```

After fixing the source code in `WordPress/Classes/ViewRelated/Me/App Settings/AppSettingsViewController.swift` to provide an explicit `value: "…"` parameter to the `NSLocalizedString` calls for those 2, and re-running `be fastlane genstrings` once more, the diff was finally empty again, showing that the result matches what `Script/localize.py` was producing.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
mokagio: I agree with your point that this code is quite minimal and therefore the maintenance overhead of having the `gentstrings`-wrapping code extracted in its own object would be more than the value added by isolated testing. 👍 

Were this action to grow in functionality, or say have a two step process, then we might revisit the decision 👌

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Sources/Doc.md
```diff
@@ -0,0 +1,14 @@
...
+```swift
+    NSLocalizedString("key", comment: "translator context")
+    NSLocalizedString("key", value: "English copy", comment: "translator context")
+    NSLocalizedString("key", bundle: Bundle(for: BundleToken.self), value: "English copy", comment: "translator context")
+    NSLocalizedString("key", tableName: "FileName", value: "English copy", comment: "translator context")
+```
+
+And also similar calls in your Objective-C code, and `Text(…)` calls in SwiftUI.
+
+Note that it should not detect the instances of those in `.md` files like this one though.
```
mokagio, line 14: I'm pretty sure this is in `sample-project/Sources` rather than `sample-project/` because the tests uses the `Sources` folder. Correct?

Did you consider calling this `README.md`?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -0,0 +1,93 @@
+module Fastlane
+  module Actions
+    class IosGenerateStringsFileFromCodeAction < Action
+      def self.run(params)
+        files = files_matching(paths: params[:paths], exclude: params[:exclude])
+        flags = [('-q' if params[:quiet]), ('-SwiftUI' if params[:swiftui])].compact
+        cmd = ['genstrings', '-o', params[:output_dir], *flags, *files]
+        out = Actions.sh_control_output(*cmd, print_command: FastlaneCore::Globals.verbose?, print_command_output: true)
+        out.scrub.strip.split("\n")
```
mokagio, line 9: This updated version is even better than the previous, IMHO. More compact and with no manual calls to `UI` because the `sh_control_output` and `genstrings` parameters combo does it all for us.

👏

mokagio (approved):
My results match yours. :shipit: ?

I'm going to approve this in case you're ready to merge it and simply haven't switch the PR to reflect it because you went into the weekend. If you have more work planned, feel free to request a re-review. 👍

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Sources/Doc.md
AliSoftware: Yes, I've added 2 different `.md` files in the test data:
 - One under `Pods/SomePod/README.md`
 - This one under `Sources/Doc.md`

The idea was to ensure that none of those 2 `md` files were scanned by `genstrings` (and incorrectly consider the `NSLocalizedString` text in them to be valid keys to add to the generated `.strings` file) in both of the with/without Pods cases. I could have only added one of the 2 and consider that if it's not parsed in the "with Pods" case, then there was no reason why the md would be parsed in the "without Pods" either, but having 2 files to cover both cases didn't hurt 🤷 

As for naming this `README.md`, since I already named the one in `Pods/SomePod/` as `README.md`, I wanted to use a different name here on purpose in order to ensure that any exclusion we'd do in our logic to ensure we only scan `.m` and `.swift` files would not just exclude the `README.md` common name but any `*.md` file whatever its name. It's a little thing, but I prefer to vary my red herring in my test data to cover even more edge cases that way 😛

AliSoftware:
> I'm going to approve this in case you're ready to merge it and simply haven't switch the PR to reflect it because you went into the weekend. If you have more work planned, feel free to request a re-review. 👍

Yep, I indeed mostly forgot to undraft it 😅 Though I also wanted to deprecate the old lane (and comment with deprecation notices all the related code in actions and helpers calling the `localize.py` script)  in favor of this new one before finally merging.

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Sources/Doc.md
mokagio: > As for naming this `README.md`, since I already named the one in `Pods/SomePod/` as `README.md`, I wanted to use a different name here on purpose in order to ensure that any exclusion we'd do in our logic to ensure we only scan `.m` and `.swift` files would not just exclude the `README.md` common name but any `*.md` file whatever its name. 

Nice idea 👍

PR #310, opened 2021-11-22
Track result of running `bundle install` on M1 chip

mokagio:
Run on my M1 MacMini test device.

Either I didn't run `bundle install` at the time of `f509e58`, or we had to run it on an M1 first in order for the `nokogiri` platform-agnostic value to be picked up, or there was some bug in Bundler itself that
prevent that from happening. I'm mentioning the chance of a bug in Bundler because, when I run `bundle install` with Bundler 2.2.11, it remained stuck with the Intel-based version of the gem, unable to understand it had to build from native extensions. Upgrading to the latest version solved the issue. I decided not to investigate further.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -267,6 +267,7 @@ GEM
     method_source (0.9.2)
     mini_magick (4.11.0)
     mini_mime (1.1.2)
+    mini_portile2 (2.6.1)
     minitest (5.14.4)
     molinillo (0.8.0)
     multi_json (1.15.0)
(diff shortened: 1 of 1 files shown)


AliSoftware (approved):
It's even more surprising given that `.bundle/config` contains `BUNDLE_SPECIFIC_PLATFORM: "false"` since f509e58 as you mention, and that `PLATFORMS` only contains `ruby` and no `darwin-…` variant… 🤷

PR #311, opened 2021-11-22
L10n: Add support for routines other than NSLocalisedStrings (for widgets)

AliSoftware:
Part of project ref paaHJt-2Ib-p2
Follow-up on #309

## What

This adds support to provide custom alternate routines in addition to the default `NSLocalizedString(…)` macros parsed by `genstrings`.

This will be quite useful in order to make Widgets look up their localized strings in the main app bundle instead of their own bundle.
Instead of having to pass `Bundle(identifier: the_app_bundle)` or similar in the widgets' code _every time_ we need an `NSLocalizedString` from the app in the widgets, we'll now be able to create a custom `func AppLocalizedString(…)` helper function in the widgets' codebase, and use that in their codebase, DRYing the widget code significantly.

## How

 - Added new `ConfigItem` which ends up adding the appropriate `-s <routine>` flags to the invocation of `genstrings`
 - Updated the unit tests accordingly

## To test

- Verify CI is green
- Review the `.strings` files I used as "expected" output for the unit tests and ensure they make sense according to the code being parsed.


spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Pods/SomePod/Sources/PodClass1.swift
```diff
@@ -1,5 +1,9 @@
 import Foundation
 
+func PodLocalizedString(_ key: String, value: String? = nil, comment: StaticString) -> String {
+    Bundle(for: BundleToken.self).localizedString(forKey: key, value: value, table: nil)
+}
```
AliSoftware, line 5: Note that the real use case in client apps will most likely be **Widgets** having a `AppLocalizedString` helper method to reference strings from the host app, rather than us parsing methods like this from **Pods**.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-* Added a new `ios_generate_strings_file_from_code` action to replace the soon-to-be-obsoleted `ios_localize_project` action (and `Scripts/localize.py` script in app repos). [#309]
+* Added a new `ios_generate_strings_file_from_code` action to replace the soon-to-be-obsoleted `ios_localize_project` action (and `Scripts/localize.py` script in app repos). [#309, #311]
```
```suggestion
* Added a new `ios_generate_strings_file_from_code` action to replace the now-deprecated `ios_localize_project` action (and `Scripts/localize.py` script in app repos). [#309, #311]
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -4,6 +4,7 @@ class IosGenerateStringsFileFromCodeAction < Action
       def self.run(params)
         files = files_matching(paths: params[:paths], exclude: params[:exclude])
         flags = [('-q' if params[:quiet]), ('-SwiftUI' if params[:swiftui])].compact
+        flags += Array(params[:routines]).flat_map { |s| ['-s', s] }
```
mokagio: I found that sequence of `s`s a bit confusing. How would you feel about a more verbose `routine`?

```suggestion
        flags += Array(params[:routines]).flat_map { |routine| ['-s', routine] }
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -54,6 +55,11 @@ def self.available_options
                                        description: 'Array of paths or glob patterns to exclude from scanning',
                                        type: Array,
                                        default_value: []),
+          FastlaneCore::ConfigItem.new(key: :routines,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_ROUTINES',
+                                       description: 'Base name of the alternate methods to be parsed in addition to the standard NSLocalizedString() one',
```
mokagio: Consistency nitpick:

```suggestion
                                       description: 'Base name of the alternate methods to be parsed in addition to the standard `NSLocalizedString()` one',
```

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Pods/SomePod/Sources/PodClass1.swift
```diff
@@ -12,10 +16,18 @@ enum PodClass1 {
...
                                               value: "pod value 5\n"
                                                 + "with multiple lines.",
                                               comment: "Pod key 5, with value, custom table and placeholder.")
+    static let podStringX = PodLocalizedString("pod.custom_l10n_function",
+                                               value: "Using custom macro \n"
+                                                    + "with multiple lines.",
+                                               comment: """
+                                               A string processed via a custom macro
+                                               instead of the standard NSLocalizedString one.
+                                               """)
```
mokagio, line 30: I didn't know about routines before looking at this PR. Do you think it would be useful to add a few lines here (or in the action `.rb` source, maybe) to explain them? Or is it safe to assume that a future developer would do as I did, "oh, I don't know about this thing, let me look it up: `man genstrings` `/ routine`."

mokagio (approved):
Code LGTM, only left nitpicks.

I pocked around at the unit test and by running `genstrings` manually and verified the behavior is correct.

I had another `genstrings: error:` output after what I wrote in https://github.com/wordpress-mobile/WordPress-iOS/pull/17519#pullrequestreview-813180246.

As you might have noticed, when running `genstrings` for the new code with the routine, we get this error:

```
genstrings: error: bad entry in file /Users/gio/Developer/a8c/release-toolkit/spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Pods/SomePod/Sources/PodClass1.swift (line = 3): Argument is not a literal string.
```

Technically, I think that's due to `genstrings` trying to parse the routine definition (`func PodLocalizedString(_ key: String, ...`) and complaining about it not being a valid localizable string call syntax.

TBH, I'm surprised `genstrings` doesn't fail when it gets an error, even though in this case it works out well in our favour or we'd have to find a way to exclude the portion of code where the routine is defined 🤔 

I looked at the `man` page but couldn't find any rationale for that behavior. Do you happen to know why it does that?

In the WP iOS PR, I [wrote](https://github.com/wordpress-mobile/WordPress-iOS/pull/17519#pullrequestreview-813180246) that I was surprised to see an error in the logs when the action succeeded. I still find that surprising, but I guess if that's the standard behavior for `genstrings` we might want to roll with it? 🤔

AliSoftware:
> As you might have noticed, when running `genstrings` for the new code with the routine, we get this error:
> 
> ```
> genstrings: error: bad entry in file /Users/gio/Developer/a8c/release-toolkit/spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Pods/SomePod/Sources/PodClass1.swift (line = 3): Argument is not a literal string.
> ```
> 
> Technically, I think that's due to `genstrings` trying to parse the routine definition (`func PodLocalizedString(_ key: String, ...`) and complaining about it not being a valid localizable string call syntax.

Wow, I didn't realize this, thanks for catching it. That's quite unfortunate that `genstrings` is not smart enough to detect that this is a function declaration…

> TBH, I'm surprised `genstrings` doesn't fail when it gets an error, even though in this case it works out well in our favour or we'd have to find a way to exclude the portion of code where the routine is defined 🤔

Indeed. The easiest way for this would be to declare that helper method in a dedicated file, provide that file path to the `exclude` parameter of the action (like we did for [the `LocalizedStringKey+extension.swift` file](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressStatsWidgets/Views/Localization/LocalizedStringKey+extension.swift)).

> I looked at the `man` page but couldn't find any rationale for that behavior. Do you happen to know why it does that?

I think the rationale is just that `genstrings` is a little dumb and is mostly driven by regular expressions under the hood. It does not has the ability to parse the AST and understand the code context, all it does is detect text patterns, so that why it falls into this trap.

This can sadly be shown even more clearly if we enable SwiftUI mode (`genstrings -SwiftUI`) and run it on the WPiOS codebase: it then incorrectly interprets a lot of non-SwiftUI code like `Aztec.TextView(…)` as if it were `Text(…)` or `TextView(…)` calls from SwiftUI – due to, I assume, their internal regex being too loose – and prints a lot of errors related to those mis-detections. (Which is why I've disabled the SwiftUI option by default on the action btw, see 5e72f8aae8a7f079f05545555f002344465383c4's commit message)

> In the WP iOS PR, I [wrote](https://github.com/wordpress-mobile/WordPress-iOS/pull/17519#pullrequestreview-813180246) that I was surprised to see an error in the logs when the action succeeded. I still find that surprising, but I guess if that's the standard behavior for `genstrings` we might want to roll with it? 🤔

As mentioned in the other PR, I still think it would be nice to fail on error, by detecting any `error:` in the stdout, and even if genstrings returns a 0 exit code. We can find a workaround for the declaration of the helper routine (define in separate file + exclude), but it's important that other failures (like if the developers use intermediate variables, which `genstrings` don't support) are reported. I'll add the option then.

spec/test-data/translations/ios_generate_strings_file_from_code/sample-project/Pods/SomePod/Sources/PodClass1.swift
AliSoftware: Addressed in d828ea11d5305b69395bf5d6377ac23134015f3f

AliSoftware:
@mokagio I've added a `fail_on_error` parameter to the action, defaulting to `true`, which will call `UI.user_error!` when it sees `genstrings: error:` messages in the command's `stdout`.

I've also addressed your suggestion to describe the concept of routines in more details (in the action's `self.details` documentation)

Let me know what you think.

AliSoftware:
Ok I'll now merge this so that I can pick that "fail on error" behavior on WPiOS and merge https://github.com/wordpress-mobile/WordPress-iOS/pull/17519 and start working on Step 2 in the WPiOS repo.
Let me know if you have some feedback or things that didn't get fix in that PR and I'll integrate your feedback in the next PR on release-toolkit if needed.

mokagio:
@AliSoftware 

> added a `fail_on_error` parameter to the action, defaulting to `true`, which will call `UI.user_error!` when it sees `genstrings: error:` messages in the command's `stdout`.

Excellent! Thank you 🙇 

+1 for the exclude path idea. It's a bit annoying having to _know_ about the caveat that `genstrings` can't tell the `routine` definition from its usages apart and work around it that way, but it's a small one-time setup cost 👌 

If we'll ever extract this action in a standalone "localization" plug-in, we could compensate for this gotcha in the documentation.

PR #312, opened 2021-11-23
Remove drawtext as direct dependency

jkmassel:
This PR removes all `drawText` (and thus swift code) from this project.

This has a few benefits, namely:
- Compatibility with non-macOS machines.
- No requirement for a Swift compiler on macOS machines.
- A simpler installation process.

This is _technically_ a breaking change, so IMHO it's worth discussion about how to land it (I'll update the release notes based on what we decide).


AliSoftware:
From looking at the code the changes look good to me (I also appreciate the helpful message when we detect that drawText is not installed 👍 ), but I've yet to properly test this in practice by running it to ensure nothing breaks. I'll try to find some time to do that later this week (I'm in the middle of multiple changes on my own on the toolkit rn).

As for the level change, I'm not sure tbh. Technically it is not a breaking change in the sense that it won't require any changes in the adopting client repos (like having to edit our app's `Fastfile` to update call sites to use new API) nor changes the behavior of the commands and output, … 
Te only change that will be needed is for adopters to now run `brew install drawText`… and once they've done that on their Mac once for the first repo, that will work for all other repos (`brew` does system-wide install after all)… so maybe not that breaking?

The only argument I see for considering it worth a major version bump would be about CI config: we might need to update our pipelines to ensure we `brew install drawText` before running the lane, while it wasn't necessary before. Though arguably we could also include `drawText` directly into our VM images so that it comes pre-installed 😛

jkmassel:
> The only argument I see for considering it worth a major version bump would be about CI config: we might need to update our pipelines to ensure we `brew install drawText` before running the lane, while it wasn't necessary before.

Because this update isn't safe to simply apply without any other considerations, I'm marking it as a breaking change. We can hold this PR back until other work lands though, in order to avoid a breaking change in the midst of the localization updates.

WDYT @AliSoftware?

AliSoftware:
That works for me too.

.buildkite/pipeline.yml
```diff
@@ -32,28 +29,33 @@ steps:
   #################
   - label: "🧹 Lint"
     command: |
-      .buildkite/install-dependencies.sh
-
+      bundle install
```
AliSoftware, line 50: Is this change intentional? That means it will install gems but not install `brew` dependencies like `libxml` or `imagemagick`. I mean maybe those are indeed not necessary for linting, but just wanted to check. And in any case, if that's intentional, maybe worth a `# comment` to state it explicitly 🙂

.buildkite/pipeline.yml
```diff
@@ -32,28 +29,33 @@ steps:
...
       echo "--- :rubocop: Run Rubocop"
       bundle exec rubocop
     env: *common_env
-    plugins: *common_plugins
+    plugins:
+      - docker#v3.8.0:
+          image: "ruby:2.7.4"
```
AliSoftware: I'm guessing you didn't create a YAML anchor for this one because the only other place you currently use it is on line 51 but you need to pass the API token on that other one so they end up differing?

I'd still suggest to:

 - Make line 71 (for the `gem-push` step) use the same `ruby:2.7.4` docker image than the one we use for linting and testing
 - Then get rid of the `*common_plugins` and `bash_cache` anchors as they would not be used anymore (we won't rely on the bash-cache docker image for release-toolkit, only on `ruby:2.7.4`)
 - Make new anchors for the ruby image:

```yaml
common_params:
  plugins:
  - &docker_plugin
    docker#v3.8.0:
      image: &ruby_version "ruby:2.7.4"
  - &docker_plugin_with_danger_token
    docker#v3.8.0:
      image: *ruby_version
      propagate-environment: true
      environment:
        - "DANGER_GITHUB_API_TOKEN"
```
That way, the docker plugin version as well as the ruby version we use are all centralized on top of the file, and we have less risk of forgetting to update everything at once next time we update those.

Then of course for the lint step you'd use `plugins: [*docker_plugin]` while for the danger step you'd use 
`plugins: [*docker_plugin_with_danger_token]`


PS: There are more advanced solutions in YAML to avoid repeating the `docker#3.8.0` key here (i.e ways to create an anchor just on the `docker#v3.8.0` key name, not the whole dict, etc), but I don't think it's worth getting into those. As long as all the versions we might want to update are grouped at the top in a centralized place this is good enough.

.buildkite/pipeline.yml
```diff
@@ -32,28 +29,33 @@ steps:
...
+      queue: "default"
 
   #################
   # Danger
   #################
   - label: "⛔️ Danger"
     command: |
-      .buildkite/install-dependencies.sh
-
+      bundle install
```
AliSoftware, line 63: Same question as to wether this is intentional not to `brew install` during this step. If it is, worth a quick comment to indicate it was intentional and not an accidental overlook.

Rakefile
```diff
@@ -30,12 +30,10 @@ task :docstats, [:path] do |_, args|
   sh('yard', 'stats', '--list-undoc', path)
 end
 
-Rake::ExtensionTask.new('drawText')
-
 GEM_NAME = 'fastlane-plugin-wpmreleasetoolkit'.freeze
 VERSION_FILE = File.join('lib', 'fastlane', 'plugin', 'wpmreleasetoolkit', 'version.rb')
 
-desc 'Try to build and install the gem to ensure it can be installed properly (with the native extension and all)'
+desc 'Try to build and install the gem to ensure it can be installed properly'
```
AliSoftware: Took the liberty to push that change myself here, as this rake task was added mostly to ensure that the drawText native extensions was able to build.

We might even consider removing this task? Or at least removing the `check_build` step that is made during `new_release` task, as now that the CI can check that the gem builds for us, no need to do the same check locally before creating the PR… even less so now that `drawText` won't risk creating native extensions build issues anymore 🤷

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -28,6 +28,8 @@ def initialize
           message << 'Aborting.'
           UI.user_error!(message)
         end
+
+        UI.user_error!('`drawText` not found – install it using `brew install automattic/build-tools/drawText`.') unless system('which drawText')
```
AliSoftware: `shellcheck` taught me recently that `command -v drawText` was more recommended than `which drawText` 🤷 

See https://github.com/koalaman/shellcheck/wiki/SC2230

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -233,7 +235,7 @@ def draw_text_to_canvas(canvas, text, width, height, x_position, y_position, fon
         begin
           tempTextFile = Tempfile.new()
 
-          command = "bundle exec drawText html=\"#{text}\" maxWidth=#{width} maxHeight=#{height} output=#{tempTextFile.path} fontSize=#{font_size} stylesheet=\"#{stylesheet_path}\" alignment=\"#{position}\""
+          command = "drawText html=\"#{text}\" maxWidth=#{width} maxHeight=#{height} output=#{tempTextFile.path} fontSize=#{font_size} stylesheet=\"#{stylesheet_path}\" alignment=\"#{position}\""
 
           UI.crash!('Unable to draw text') unless system(command)
```
AliSoftware: Also could be worth converting this to `sh('drawText', 'html="#{text}"', 'maxWidth=#{width}', …)` at some point, as it would prevent both parsing and security issues (via injection) and also be more idiomatic and also allow fastlane's build-in handling of output and error for shell commands to take over.

This can probably be for a separate PR though (but worth keeping in mind).

AliSoftware (commented):
Did another pass and LGTM except for some thins in the Buildkite pipeline.
I'm not formally approving the PR anyway in order to avoid accidentally having someone merging it before we're ready for a breaking change version.

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
jkmassel: Ah yes, this is even better!

Addressed in 5e5bf9b636f1b70c877ebc18d6d163772bc89ddd

Rakefile
jkmassel: Yep I'm good to get rid of it since it's replicated in CI!

Removed in f727f3b3097d27088ad72cf517f8ea7c6106231c

.buildkite/pipeline.yml
jkmassel: Yep – [the Docker image includes libxml](https://github.com/docker-library/ruby/blob/928ce974a8356c21238af67b50680cc42fc489fe/2.7/alpine3.14/Dockerfile#L52), and IIRC `imagemagick` isn't needed unless we use `--with screenshots`.

IMHO the goal should be to require no special native dependencies, right now the need for them is an aberration (because IIRC `libxml` is a requirement of ruby itself), which is why I didn't think to comment on it. WDYT?

AliSoftware: If the docker image really includes libxml now then we should remove it from the brewfile.

For the rest, I totally agree with now installing the other brew dependencies if they're not needed for the step; it's gonna speed up the build after all 🙂

But I feel like it gives an odd asymmetry/imbalance to have `bundle install` (aka direct call) in some steps and `install-dependencies.sh` (aka external script) in others. At that stage I'd probably just get rid of the `install-dependencies.sh` script completely and call the brew install command directly from the `pipeline.yml` on the rare steps where it's still necessary, to make it more consistent.

Rakefile
AliSoftware: You seem to have forgotten to also remove the lines 69-71 ie when this is called from the `new_release` task

Rakefile
```diff
@@ -101,10 +93,3 @@ def update_version_constant(version_file, new_version)
 
   sh('bundle', 'install', '--quiet') # To update Gemfile.lock with new wpmreleasetoolkit version
 end
-
-def check_install(version)
```
AliSoftware, line 104: Call site needs to also be removed on lines 69-71 otherwise `new_release` task will crash.

AliSoftware (changes requested):

.buildkite/pipeline.yml
jkmassel: Addressed in 0e7a33ed21e0c538c9e6eff89e50cd8b09447fa9

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
jkmassel: Good call – addressed in 36f4d6a04fdb7b2cad7928da4823a6487051a9bb.

Rakefile
jkmassel: Thanks! Addressed in d04fe6f73a87d002eca17a2d9d96c9cdb23628bc.

.buildkite/pipeline.yml
jkmassel: I left a note where we use `brew install` directly to hopefully help clarify when it is / is not required.

Rakefile
jkmassel: Thanks for pointing out the updated lines below – addressed in https://github.com/wordpress-mobile/release-toolkit/commit/d04fe6f73a87d002eca17a2d9d96c9cdb23628bc.

.buildkite/pipeline.yml
jkmassel: Addressed in 538430e84ebdf3f4daa70ae8639bc665d9ba5f19 (followed by some noodling around to move the `gem push` stuff over to docker as well).

jkmassel:
@AliSoftware, thanks for the thorough review – I think this one is ready for another look!

.buildkite/pipeline.yml
```diff
@@ -12,18 +21,21 @@ steps:
...
 
+      echo "--- :rubygems: Setting up Gems"
+      install_gems
+
       echo "--- :rspec: Run Rspec"
       bundle exec rspec --profile 10 --format progress
-    env: *common_env
-    plugins: *common_plugins
+    env: 
+      IMAGE_ID: xcode-12.5.1
```
AliSoftware: nit: I know that this value is only appearing once in the pipeline now, so would not technically benefit any DRY-ing, but I wonder if it wouldn't still be nice to move that at the top of the YAML just so every version-dependent value (like the Xcode version used here) would all be grouped at the top, making it easier to do maintenance and updates.

```yaml
common_params:
  env: &xcode_image
    IMAGE_ID: xcode-12.5.1
  …
```

This not a big deal nor a blocker but just a suggestion / personal preference, so up to you to do this or not; but I like the idea of having all versions grouped at the top and easily accessible for future updates 🙂

.buildkite/pipeline.yml
```diff
@@ -12,18 +21,21 @@ steps:
...
+    plugins:
+      automattic/bash-cache#2.0.0: ~
```
AliSoftware, line 40: nit: similar to the comment above about xcode image, I think it could be nice to move this one in the `common_params`, even if it's only used once, just in order to have all of those referenced at the top of the YAML for easier maintenance and update later. wdyt?

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -233,9 +235,7 @@ def draw_text_to_canvas(canvas, text, width, height, x_position, y_position, fon
         begin
           tempTextFile = Tempfile.new()
 
-          command = "bundle exec drawText html=\"#{text}\" maxWidth=#{width} maxHeight=#{height} output=#{tempTextFile.path} fontSize=#{font_size} stylesheet=\"#{stylesheet_path}\" alignment=\"#{position}\""
-
-          UI.crash!('Unable to draw text') unless system(command)
+          sh('drawText', "html=\"#{text}\"", "maxWidth=#{width}", "maxHeight=#{height}", "output=#{tempTextFile.path}", "fontSize=#{font_size}", "stylesheet=\"#{stylesheet_path}\"", "alignment=\"#{position}\"")
```
AliSoftware: Even if this is unrelated to this PR per se (as the issue was apparently already present before that diff), would be nice to ensure the output path is quoted to avoid issues with spaces in path. I know it's unlikely because temp files usually reside in `/tmp` and don't usually have spaces in their path, but still better safe than sorry 😃 

```suggestion
          sh('drawText', "html=\"#{text}\"", "maxWidth=#{width}", "maxHeight=#{height}", "output=\"#{tempTextFile.path}\"", "fontSize=#{font_size}", "stylesheet=\"#{stylesheet_path}\"", "alignment=\"#{position}\"")
```

AliSoftware (approved):
Overall LGTM, so I'm approving to both remove my previous "Request changes" review and unblock you.

I still left a couple of nit / suggestions on the pipeline (non blockers, just nice to have) as well as a suggested fix that's unrelated to this PR but still worth fixing while we're at it imho. I'll let you address those or not before you decide to merge the PR.

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
AliSoftware: Actually, that was an incorrect suggestion here, because we now already pass parameters separately to `sh` — as opposed to passing a single, long string with spaces and leave to the shell interpreter to split parameters based on spaces and quotes — so `sh(x, y, z)` already passes the separate `y` and`z` parameters to the `x` executable directly without any shell interpreter step to split them, and thus without the need to even quote/escape them.

The escaping/quoting is only needed if we pass a single string to `sh`/`system` and let the shell interpreter to the split of the parameters from that long, single string. So here the right solution would have been to not only keep the `tempTextFile.path` unquoted as before, but also to remove the various `\"` quotes from other parameters like `html=`, `stylesheet=` and `alignment=`.

Kudos to @rachelmcr for spotting this! [ref: pdnsEh-hU-p2#comment-699]

PR #313, opened 2021-11-28
Add support for commenting on GitHub PRs

jkmassel:
Adds an action that makes it easier to add/update PR comments (particularly for installable builds).

This will nicely simplify the process for installable builds, in particular.

**To Test:**
- Includes a new `rspec` test to cover the changes to the GitHub token lookup (to ensure backward compatibility)
- To test locally, you can run: `bundle exec fastlane run comment_on_pr project:wordpress-mobile/release-toolkit pr_number:313 reuse_identifier:my-test`.
- Enter a comment and note that it shows up on this PR.
- Run the command again and enter different text.
- Note that the comment is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ _None_
 
 * Added a new `ios_generate_strings_file_from_code` action to replace the now-deprecated `ios_localize_project` action (and `Scripts/localize.py` script in app repos). [#309, #311]
 
+* Added a `comment_on_pr` action to allow commenting on (and updating comments on) PRs.
+
+* Added the ability to use the `GITHUB_TOKEN` environment variable for GitHub operations. `GHHELPER_ACCESS` will be deprecated in a future version.
+
 ### Bug Fixes
 
 _None_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,9 @@ _None_
 
 * Added a new `ios_generate_strings_file_from_code` action to replace the now-deprecated `ios_localize_project` action (and `Scripts/localize.py` script in app repos). [#309, #311]
 
-* Added a `comment_on_pr` action to allow commenting on (and updating comments on) PRs.
+* Added a `comment_on_pr` action to allow commenting on (and updating comments on) PRs. [#313]
 
-* Added the ability to use the `GITHUB_TOKEN` environment variable for GitHub operations. `GHHELPER_ACCESS` will be deprecated in a future version.
+* Added the ability to use the `GITHUB_TOKEN` environment variable for GitHub operations. `GHHELPER_ACCESS` will be deprecated in a future version. [#313]
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
@@ -52,13 +52,13 @@ module Fastlane
           FastlaneCore::ConfigItem.new(
             key: :reuse_identifier,
             description: 'If provided, the reuse identifier can identify an existing comment to overwrite',
-            is_string: true,
+            type: String,
             default_value: nil
           ),
           FastlaneCore::ConfigItem.new(
             key: :project,
             description: 'The project slug (ex: `rails/rails`)',
-            is_string: true
+            type: String
           ),
           FastlaneCore::ConfigItem.new(
             key: :pr_number,
@@ -68,7 +68,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(
             key: :body,
             description: 'The content of the comment',
-            is_string: true
+            type: String
           ),
         ]
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
@@ -4,16 +4,24 @@ require 'securerandom'
 
 module Fastlane
   module Actions
+    module SharedValues
+      PR_COMMENT_REUSE_IDENTIFIER = :PR_COMMENT_REUSE_IDENTIFIER
+    end
+
     class CommentOnPrAction < Action
       def self.run(params)
         require_relative '../../helper/github_helper'
 
-        Fastlane::Helper::GithubHelper.comment_on_pr(
+        reuse_identifier = Fastlane::Helper::GithubHelper.comment_on_pr(
           project_slug: params[:project],
           pr_number: params[:pr_number],
           body: params[:body],
           reuse_identifier: params[:reuse_identifier]
         )
+
+        Actions.lane_context[SharedValues::PR_COMMENT_REUSE_IDENTIFIER] = reuse_identifier
+
+        reuse_identifier
       end
 
       def self.description
@@ -63,6 +71,16 @@ module Fastlane
             is_string: true
           ),
         ]
+      end
+
+      def self.output
+        [
+          ['PR_COMMENT_REUSE_IDENTIFIER', 'The `reuse_identifier` for the most recently posted comment'],
+        ]
+      end
+
+      def self.return_value
+        'The `reuse_identifier` for the posted comment (useful for updating it later, if needed)'
       end
 
       def self.is_supported?(platform)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
@@ -1,0 +1,74 @@
+require 'fastlane/action'
+require 'rubygems/command_manager'
+require 'securerandom'
+
+module Fastlane
+  module Actions
+    class CommentOnPrAction < Action
+      def self.run(params)
+        require_relative '../../helper/github_helper'
+
+        Fastlane::Helper::GithubHelper.comment_on_pr(
+          project_slug: params[:project],
+          pr_number: params[:pr_number],
+          body: params[:body],
+          reuse_identifier: params[:reuse_identifier]
+        )
+      end
+
+      def self.description
+        'Post a comment on a given PR number (optionally updating an existing one)'
+      end
+
+      def self.authors
+        ['Jeremy Massel']
+      end
+
+      def self.details
+        <<~DETAILS
+          If used just once, this method makes it nice and easy to post a quick comment to a GitHub
+          PR. Subsequent runs will allow you to update an existing comment as many times as you need
+          to (for instance, across multiple CI runs).
+
+          The `:reuse_identifier` config item enables this.
+        DETAILS
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :access_token,
+            env_name: 'GITHUB_TOKEN',
+            description: 'The GitHub token to use for posting the comment',
+            is_string: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :reuse_identifier,
+            description: 'If provided, the reuse identifier can identify an existing comment to overwrite',
+            is_string: true,
+            default_value: SecureRandom.uuid
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :project,
+            description: 'The project slug (ex: `rails/rails`)',
+            is_string: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :pr_number,
+            description: 'The PR number',
+            is_string: false # integer
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :body,
+            description: 'The content of the comment',
+            is_string: true
+          ),
+        ]
+      end
+
+      def self.is_supported?(platform)
+        true
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
@@ -21,7 +21,7 @@ module Fastlane
       end
 
       def self.authors
-        ['Jeremy Massel']
+        ['Automattic']
       end
 
       def self.details
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
@@ -26,11 +26,10 @@ module Fastlane
 
       def self.details
         <<~DETAILS
-          If used just once, this method makes it nice and easy to post a quick comment to a GitHub
-          PR. Subsequent runs will allow you to update an existing comment as many times as you need
-          to (for instance, across multiple CI runs).
+          If used just once, this method makes it nice and easy to post a quick comment to a GitHub PR.
 
-          The `:reuse_identifier` config item enables this.
+          Subsequent runs will allow you to update an existing comment as many times as you need to
+          (e.g. across multiple CI runs), by using a `:reuse_identifier` to identify the comment to update.
         DETAILS
       end
 
@@ -40,13 +39,13 @@ module Fastlane
             key: :access_token,
             env_name: 'GITHUB_TOKEN',
             description: 'The GitHub token to use for posting the comment',
-            is_string: true
+            type: String
           ),
           FastlaneCore::ConfigItem.new(
             key: :reuse_identifier,
             description: 'If provided, the reuse identifier can identify an existing comment to overwrite',
             is_string: true,
-            default_value: SecureRandom.uuid
+            default_value: nil
           ),
           FastlaneCore::ConfigItem.new(
             key: :project,
@@ -56,7 +55,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(
             key: :pr_number,
             description: 'The PR number',
-            is_string: false # integer
+            type: Integer
           ),
           FastlaneCore::ConfigItem.new(
             key: :body,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
@@ -1,6 +1,4 @@
 require 'fastlane/action'
-require 'rubygems/command_manager'
-require 'securerandom'
 
 module Fastlane
   module Actions
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
@@ -55,7 +55,7 @@ module Fastlane
           ),
           FastlaneCore::ConfigItem.new(
             key: :project,
-            description: 'The project slug (ex: `rails/rails`)',
+            description: 'The project slug (ex: `wordpress-mobile/wordpress-ios`)',
             type: String
           ),
           FastlaneCore::ConfigItem.new(
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -7,8 +7,16 @@ module Fastlane
 
   module Helper
     class GithubHelper
+      def self.github_token
+        [
+          'GHHELPER_ACCESS', # For historical reasons / backward compatibility
+          'GITHUB_TOKEN',    # Used by the `gh` CLI tool
+        ].map { |key| ENV[key] }
+          .detect { |value| !value.nil? }
+      end
+
       def self.github_client
-        client = Octokit::Client.new(access_token: ENV['GHHELPER_ACCESS'])
+        client = Octokit::Client.new(access_token: github_token)
 
         # Fetch the current user
         user = client.user
@@ -127,6 +135,27 @@ module Fastlane
         end
 
         download_path
+      end
+
+      # Creates (or updates an existing) GitHub PR Comment
+      def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier:)
+        comments = github_client.issue_comments(project_slug, pr_number)
+
+        existing_comment = comments.detect do |comment|
+          comment.body.include?(reuse_identifier)
+        end
+
+        comment_body = prepare_comment_body(body, reuse_identifier)
+
+        if existing_comment.nil?
+          github_client.add_comment(project_slug, pr_number, comment_body)
+        else
+          github_client.update_comment(project_slug, existing_comment.id, comment_body)
+        end
+      end
+
+      def self.prepare_comment_body(body, reuse_identifier)
+        "<!-- REUSE_ID: #{reuse_identifier} -->" + body
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -142,23 +142,24 @@ module Fastlane
 
       # Creates (or updates an existing) GitHub PR Comment
       def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier: SecureRandom.uuid)
-        comments = github_client.issue_comments(project_slug, pr_number)
+        client = github_client
+        comments = client.issue_comments(project_slug, pr_number)
 
-        existing_comment = comments.find do |comment|
-          comment.body.include?(reuse_identifier)
-        end
+        complete_reuse_identifier = "<!-- REUSE_ID: #{reuse_identifier} -->"
 
-        comment_body = prepare_comment_body(body, reuse_identifier)
+        existing_comment = comments
+                           .select { |comment| comment.user.id == client.user.id } # Only match comments posted by the owner of the GitHub Token
+                           .find { |comment| comment.body.include?(complete_reuse_identifier) }
+
+        comment_body = complete_reuse_identifier + body
 
         if existing_comment.nil?
-          github_client.add_comment(project_slug, pr_number, comment_body)
+          client.add_comment(project_slug, pr_number, comment_body)
         else
-          github_client.update_comment(project_slug, existing_comment.id, comment_body)
+          client.update_comment(project_slug, existing_comment.id, comment_body)
         end
-      end
 
-      def self.prepare_comment_body(body, reuse_identifier)
-        "<!-- REUSE_ID: #{reuse_identifier} -->\n" + body
+        reuse_identifier
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -8,7 +8,7 @@ module Fastlane
 
   module Helper
     class GithubHelper
-      def self.github_token
+      def self.github_token!
         token = [
           'GHHELPER_ACCESS', # For historical reasons / backward compatibility
           'GITHUB_TOKEN',    # Used by the `gh` CLI tool
@@ -20,7 +20,7 @@ module Fastlane
       end
 
       def self.github_client
-        client = Octokit::Client.new(access_token: github_token)
+        client = Octokit::Client.new(access_token: github_token!)
 
         # Fetch the current user
         user = client.user
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -8,11 +8,14 @@ module Fastlane
   module Helper
     class GithubHelper
       def self.github_token
-        [
+        token = [
           'GHHELPER_ACCESS', # For historical reasons / backward compatibility
           'GITHUB_TOKEN',    # Used by the `gh` CLI tool
         ].map { |key| ENV[key] }
-          .compact.first
+        .compact
+        .first
+
+        token || UI.user_error!('Please provide a GitHub authentication token via the `GITHUB_TOKEN` environment variable')
       end
 
       def self.github_client
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -141,10 +141,10 @@ module Fastlane
       end
 
       # Creates (or updates an existing) GitHub PR Comment
-      def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier:)
+      def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier: SecureRandom.uuid)
         comments = github_client.issue_comments(project_slug, pr_number)
 
-        existing_comment = comments.detect do |comment|
+        existing_comment = comments.find do |comment|
           comment.body.include?(reuse_identifier)
         end
 
@@ -158,7 +158,7 @@ module Fastlane
       end
 
       def self.prepare_comment_body(body, reuse_identifier)
-        "<!-- REUSE_ID: #{reuse_identifier} -->" + body
+        "<!-- REUSE_ID: #{reuse_identifier} -->\n" + body
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -148,13 +148,13 @@ module Fastlane
         client = github_client
         comments = client.issue_comments(project_slug, pr_number)
 
-        complete_reuse_identifier = "<!-- REUSE_ID: #{reuse_identifier} -->"
+        reuse_marker = "<!-- REUSE_ID: #{reuse_identifier} -->"
 
         existing_comment = comments
                            .select { |comment| comment.user.id == client.user.id } # Only match comments posted by the owner of the GitHub Token
-                           .find { |comment| comment.body.include?(complete_reuse_identifier) }
+                           .find { |comment| comment.body.include?(reuse_marker) }
 
-        comment_body = complete_reuse_identifier + body
+        comment_body = reuse_marker + body
 
         if existing_comment.nil?
           client.add_comment(project_slug, pr_number, comment_body)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -30,8 +30,6 @@ module Fastlane
           # Auto-paginate to ensure we're not missing data
           client.auto_paginate = true
         end
-
-        @@client
       end
 
       def self.get_milestone(repository, release)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -20,16 +20,18 @@ module Fastlane
       end
 
       def self.github_client
-        client = Octokit::Client.new(access_token: github_token!)
+        @@client ||= begin
+          Octokit::Client.new(access_token: github_token!)
 
-        # Fetch the current user
-        user = client.user
-        UI.message("Logged in as: #{user.name}")
+          # Fetch the current user
+          user = client.user
+          UI.message("Logged in as: #{user.name}")
 
-        # Auto-paginate to ensure we're not missing data
-        client.auto_paginate = true
+          # Auto-paginate to ensure we're not missing data
+          client.auto_paginate = true
+        end
 
-        client
+        @@client
       end
 
       def self.get_milestone(repository, release)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -12,7 +12,7 @@ module Fastlane
           'GHHELPER_ACCESS', # For historical reasons / backward compatibility
           'GITHUB_TOKEN',    # Used by the `gh` CLI tool
         ].map { |key| ENV[key] }
-          .detect { |value| !value.nil? }
+          .compact.first
       end
 
       def self.github_client
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -1,6 +1,7 @@
 require 'fastlane_core/ui/ui'
 require 'octokit'
 require 'open-uri'
+require 'securerandom'
 
 module Fastlane
   UI = FastlaneCore::UI unless Fastlane.const_defined?('UI')
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -150,9 +150,10 @@ module Fastlane
 
         reuse_marker = "<!-- REUSE_ID: #{reuse_identifier} -->"
 
-        existing_comment = comments
-                           .select { |comment| comment.user.id == client.user.id } # Only match comments posted by the owner of the GitHub Token
-                           .find { |comment| comment.body.include?(reuse_marker) }
+        existing_comment = comments.find do |comment|
+          # Only match comments posted by the owner of the GitHub Token, and with the given reuse ID
+          comment.user.id == client.user.id and comment.body.include?(reuse_marker)
+        end
 
         comment_body = reuse_marker + body
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -12,8 +12,8 @@ module Fastlane
           'GHHELPER_ACCESS', # For historical reasons / backward compatibility
           'GITHUB_TOKEN',    # Used by the `gh` CLI tool
         ].map { |key| ENV[key] }
-        .compact
-        .first
+                .compact
+                .first
 
         token || UI.user_error!('Please provide a GitHub authentication token via the `GITHUB_TOKEN` environment variable')
       end
--- a/spec/github_helper_spec.rb
+++ b/spec/github_helper_spec.rb
@@ -46,4 +46,54 @@ describe Fastlane::Helper::GithubHelper do
       expect { described_class.github_token }.to raise_error(FastlaneCore::Interface::FastlaneError)
     end
   end
+
+  describe 'comment_on_pr' do
+    let(:client) { double(Octokit::Client, :issue_comments => [], :add_comment => nil, :update_comment => nil, :user => double('User', :id => 1234)) }
+
+    before do
+      allow(described_class).to receive(:github_client).and_return(client)
+    end
+
+    it 'will create a new comment if an existing one is not found' do
+      expect(client).to receive(:add_comment)
+      comment_on_pr
+    end
+
+    it 'will update an existing comment if one is found' do
+      allow(client).to receive(:issue_comments).and_return([mockComment])
+      expect(client).to receive(:update_comment)
+      comment_on_pr
+    end
+
+    it 'will not match text outside the reuseID tag' do
+      allow(client).to receive(:issue_comments).and_return([mockComment(body: 'test-id')])
+      expect(client).to receive(:add_comment)
+      comment_on_pr
+    end
+
+    it 'will not match comments belonging to other users' do
+      allow(client).to receive(:issue_comments).and_return([mockComment(user_id: 0)])
+      expect(client).to receive(:add_comment)
+      comment_on_pr
+    end
+
+    it 'will return the reuse identifier' do
+        expect(comment_on_pr).to eq 'test-id'
+    end
+
+    @private
+
+    def comment_on_pr
+      described_class.comment_on_pr(
+          project_slug:'test/test',
+          pr_number: 1234,
+          body: 'Test',
+          reuse_identifier: 'test-id'
+        )
+    end
+
+    def mockComment(body: '<!-- REUSE_ID: test-id --> Test', user_id: 1234)
+      double('Comment', :id => 1234, :body => body, :user => double('User', :id => user_id))
+    end
+  end
 end
--- a/spec/github_helper_spec.rb
+++ b/spec/github_helper_spec.rb
@@ -28,22 +28,22 @@ describe Fastlane::Helper::GithubHelper do
 
     it 'can use `GHHELPER_ACCESS`' do
       ENV['GHHELPER_ACCESS'] = 'GHHELPER_ACCESS'
-      expect(described_class.github_token).to eq('GHHELPER_ACCESS')
+      expect(described_class.github_token!).to eq('GHHELPER_ACCESS')
     end
 
     it 'can use `GITHUB_TOKEN`' do
       ENV['GITHUB_TOKEN'] = 'GITHUB_TOKEN'
-      expect(described_class.github_token).to eq('GITHUB_TOKEN')
+      expect(described_class.github_token!).to eq('GITHUB_TOKEN')
     end
 
     it 'prioritizes GHHELPER_ACCESS` over `GITHUB_TOKEN` if both are present' do
       ENV['GITHUB_TOKEN'] = 'GITHUB_TOKEN'
       ENV['GHHELPER_ACCESS'] = 'GHHELPER_ACCESS'
-      expect(described_class.github_token).to eq('GHHELPER_ACCESS')
+      expect(described_class.github_token!).to eq('GHHELPER_ACCESS')
     end
 
     it 'prints an error if no environment variable is present' do
-      expect { described_class.github_token }.to raise_error(FastlaneCore::Interface::FastlaneError)
+      expect { described_class.github_token! }.to raise_error(FastlaneCore::Interface::FastlaneError)
     end
   end
 
--- a/spec/github_helper_spec.rb
+++ b/spec/github_helper_spec.rb
@@ -48,7 +48,15 @@ describe Fastlane::Helper::GithubHelper do
   end
 
   describe 'comment_on_pr' do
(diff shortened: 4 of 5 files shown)


jkmassel:
This is a test comment

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,74 @@
...
+        DETAILS
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :access_token,
+            env_name: 'GITHUB_TOKEN',
+            description: 'The GitHub token to use for posting the comment',
+            is_string: true
```
AliSoftware:
```suggestion
            type: String
```
(`is_string` is legacy)

And same for all other `ConfigItem` down below of course

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,74 @@
...
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :reuse_identifier,
+            description: 'If provided, the reuse identifier can identify an existing comment to overwrite',
+            is_string: true,
+            default_value: SecureRandom.uuid
```
AliSoftware: I'd instead suggest a
```suggestion
            default_value: nil
```
And handle the generation of a default reuse_identifier at the Helper level

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,74 @@
...
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :project,
+            description: 'The project slug (ex: `rails/rails`)',
+            is_string: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :pr_number,
+            description: 'The PR number',
+            is_string: false # integer
```
AliSoftware:
```suggestion
            type: Integer
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -7,8 +7,16 @@ module Fastlane
 
   module Helper
     class GithubHelper
+      def self.github_token
+        [
+          'GHHELPER_ACCESS', # For historical reasons / backward compatibility
+          'GITHUB_TOKEN',    # Used by the `gh` CLI tool
+        ].map { |key| ENV[key] }
+          .detect { |value| !value.nil? }
```
AliSoftware: Alternatively:
```suggestion
          .compact.first
```
Which admissibility is O(n) and not O(1) like `detect`/`find` is, but imho more common and acceptable for small datasets like here?

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +136,27 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
 
         download_path
       end
+
+      # Creates (or updates an existing) GitHub PR Comment
+      def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier:)
+        comments = github_client.issue_comments(project_slug, pr_number)
+
+        existing_comment = comments.detect do |comment|
```
AliSoftware: Completely opinionated but I personally prefer `find` over `detect` (they are aliases of each other)
```suggestion
        existing_comment = comments.find do |comment|
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +136,27 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
...
+          comment.body.include?(reuse_identifier)
```
AliSoftware: This seems like a pretty fragile detection. Like if I use "installable" as `reuse_identifier` and it happens that someone commented with "Hey <someone> can you please test the installable build once ready" then it will pick up that comment instead of the expected one.

I'd instead suggest to detect the whole `<!-- REUSE_ID: #{reuse_identifier} -->` string, and also, as a bonus security measure, only search amongst comments made by the bot and not comments from other, human users.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +136,27 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
...
+
+        if existing_comment.nil?
+          github_client.add_comment(project_slug, pr_number, comment_body)
+        else
+          github_client.update_comment(project_slug, existing_comment.id, comment_body)
+        end
+      end
+
+      def self.prepare_comment_body(body, reuse_identifier)
+        "<!-- REUSE_ID: #{reuse_identifier} -->" + body
```
```suggestion
        "<!-- REUSE_ID: #{reuse_identifier} -->\n" + body
```

Though, given my comment about about the detection of existing comment, I'd instead refactor this into:
 - A `comment_marker(reuse_identifier)` method which simply returns `<!-- COMMENT_AUTOGENERATED_ID: #{reuse_identifier} -->`
 - Use that method on line 145 when searching for existing comment, using `comments.find { |comment| comment.body.include?(comment_marker(reuse_identifier)) }`
 - Then it's probably not worth having a dedicated `prepare_comment_body(body, reuse_identifier)` method anymore, and replace its call sites with `comment_marker(reuse_identifier) + body` instead.

spec/github_helper_spec.rb
```diff
@@ -19,4 +19,27 @@
       end
     end
   end
+
+  describe 'github_token' do
+    after do
+      ENV['GHHELPER_ACCESS'] = nil
+      ENV['GITHUB_TOKEN'] = nil
```
AliSoftware, line 26: ~I wonder if setting those env vars like that affect the terminal session running the tests? If so, that's not a great practice to have your personal environment modified because you ran a test suite.~

~Maybe that's not the case (if rspec creates a subshell in which it runs its tests instead of running them directly in the one from which it was invoked), so worth checking.~

~If it does affect the current shell/terminal session then it's worth finding a better solution, otherwise that means than any GITHUB_TOKEN I have set in my `~/.zshrc` (so I can use `gh`) would be clearer and force me to close and reopen a tab.~

EDIT: Ok I ended up checking and it seems it doesn't affect the calling shell (`rspec` is probably doing some magic for us to prevent side effects from leaking there) so should be all good after all, nvm.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +136,27 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
 
         download_path
       end
+
+      # Creates (or updates an existing) GitHub PR Comment
+      def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier:)
```
AliSoftware: This is where I'd make the reuse ID default to an uuid, instead of the call site:

```suggestion
      def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier: SecureRandom.uuid)
```

(Don't forget to remove the `require` in the action file and add it in that helper file ofc)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,74 @@
...
+          reuse_identifier: params[:reuse_identifier]
+        )
+      end
+
+      def self.description
+        'Post a comment on a given PR number (optionally updating an existing one)'
+      end
+
+      def self.authors
+        ['Jeremy Massel']
```
AliSoftware: Note: I thought we said that from now on we'd use "Automattic" as the author of actions?

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +136,27 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
...
+        end
+
+        comment_body = prepare_comment_body(body, reuse_identifier)
+
+        if existing_comment.nil?
+          github_client.add_comment(project_slug, pr_number, comment_body)
+        else
+          github_client.update_comment(project_slug, existing_comment.id, comment_body)
+        end
+      end
```
AliSoftware: This method should return the `reuse_identifier` used by the comment, so that the calling action also returns it, so that the caller on the Fastfile can access it, especially if it was an automatically-generated UUID.

You should then document in the action that it returns that reuse_id used (there's a method provided by `Fastlane::Action` for documenting that, I think it's called `self.return_value`?)

AliSoftware (changes requested):
I've only reviewed the code (from phone), didn't test running the test suite nor the actual action.

Overall idea looks good but comment detection seems way too fragile with high risk of false positives the way it's currently done.

~Also potential side effects on running the test suite? (I didn't run the code to check, on my phone, but worth checking)~

Finally I'd make the action return the reuse identifier (especially useful if it's an automatically generated UUID).

AliSoftware:
I think this will erroneously overwrite a user comment rather than a previous bot-generated one. Let's see.

AliSoftware:
☝️  My test above proves that https://github.com/wordpress-mobile/release-toolkit/pull/313#discussion_r757879594 is indeed a bug that can easily happen unless you implement my suggested fix.

Indeed, look at the edit history of [my comment above](https://github.com/wordpress-mobile/release-toolkit/pull/313#issuecomment-981086943) (by clicking on "Edited" and show the initial text) you can see that me running `bundle exec fastlane run comment_on_pr project:wordpress-mobile/release-toolkit pr_number:313 reuse_identifier:installable-build` did modify this comment while it shouldn't have considered it being a previously-bot-generated comment.

image

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -7,8 +7,16 @@ module Fastlane
 
   module Helper
     class GithubHelper
+      def self.github_token
+        [
+          'GHHELPER_ACCESS', # For historical reasons / backward compatibility
+          'GITHUB_TOKEN',    # Used by the `gh` CLI tool
+        ].map { |key| ENV[key] }
+          .detect { |value| !value.nil? }
+      end
```
AliSoftware: It would also be nice to make this fail gracefully if no token is defined.

Indeed, when testing this for the first time, I didn't have either of the env vars set in my shell, and I ended up with a crash showing a long and scary stacktrace making me think the release-toolkit code had an issue that needed to be fixed, instead of considering this was a user error of not providing a token.

<details><summary>💥 Stacktrace when we don't have env var</summary>

```
[14:37:35]: ---------------------------
[14:37:35]: --- Step: comment_on_pr ---
[14:37:35]: ---------------------------
[14:37:35]: To not be asked about this value, you can specify it using 'body'
[14:37:35]: The content of the comment: I think this will erroneously overwrite a user comment rather than a previous bot-generated one. Let's see.

Looking for related GitHub issues on fastlane/fastlane...

➡️  fastlane doesn't fetch version from ipa
    https://github.com/fastlane/fastlane/issues/401 [closed] 17 💬
    03 Oct 2016

🔗  You can ⌘ + double-click on links to open them directly in your browser.
bundler: failed to load command: fastlane (/Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/bin/fastlane)
Traceback (most recent call last):
	50: from /Users/olivier/.rbenv/versions/2.6.4/bin/bundle:23:in `<main>'
	49: from /Users/olivier/.rbenv/versions/2.6.4/bin/bundle:23:in `load'
	48: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/exe/bundle:37:in `<top (required)>'
	47: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/friendly_errors.rb:103:in `with_friendly_errors'
	46: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/exe/bundle:49:in `block in <top (required)>'
	45: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli.rb:25:in `start'
	44: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
	43: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli.rb:31:in `dispatch'
	42: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
	41: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
	40: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
	39: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli.rb:478:in `exec'
	38: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli/exec.rb:23:in `run'
	37: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli/exec.rb:58:in `kernel_load'
	36: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli/exec.rb:58:in `load'
	35: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/bin/fastlane:23:in `<top (required)>'
	34: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/bin/fastlane:23:in `load'
	33: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/bin/fastlane:23:in `<top (required)>'
	32: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/cli_tools_distributor.rb:122:in `take_off'
	31: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/commands_generator.rb:42:in `start'
	30: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/commands_generator.rb:353:in `run'
	29: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/delegates.rb:18:in `run!'
	28: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb:117:in `run!'
	27: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/runner.rb:444:in `run_active_command'
	26: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/command.rb:157:in `run'
	25: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/command.rb:187:in `call'
	24: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/commands_generator.rb:225:in `block (2 levels) in run'
	23: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/one_off.rb:22:in `execute'
	22: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/one_off.rb:42:in `run'
	21: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:229:in `execute_action'
	20: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:229:in `chdir'
	19: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:255:in `block in execute_action'
	18: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/actions/actions_helper.rb:69:in `execute_action'
	17: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:263:in `block (2 levels) in execute_action'
	16: from /Users/olivier/Documents/Dev/tools/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb:11:in `run'
	15: from /Users/olivier/Documents/Dev/tools/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb:142:in `comment_on_pr'
	14: from /Users/olivier/Documents/Dev/tools/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb:22:in `github_client'
	13: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/client/users.rb:34:in `user'
	12: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/connection.rb:19:in `get'
	11: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/connection.rb:156:in `request'
	10: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/sawyer-0.8.2/lib/sawyer/agent.rb:94:in `call'
	 9: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/connection.rb:200:in `get'
	 8: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/connection.rb:516:in `run_request'
	 7: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/rack_builder.rb:154:in `build_response'
	 6: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/request/retry.rb:148:in `call'
	 5: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/middleware/follow_redirects.rb:61:in `call'
	 4: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/middleware/follow_redirects.rb:73:in `perform_with_redirection'
	 3: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/middleware.rb:18:in `call'
	 2: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/response.rb:59:in `on_complete'
	 1: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/middleware.rb:19:in `block in call'
/Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/response/raise_error.rb:14:in `on_complete': GET https://api.github.com/user: 401 - Requires authentication // See: https://docs.github.com/rest/reference/users#get-the-authenticated-user (Octokit::Unauthorized)
	50: from /Users/olivier/.rbenv/versions/2.6.4/bin/bundle:23:in `<main>'
	49: from /Users/olivier/.rbenv/versions/2.6.4/bin/bundle:23:in `load'
	48: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/exe/bundle:37:in `<top (required)>'
	47: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/friendly_errors.rb:103:in `with_friendly_errors'
	46: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/exe/bundle:49:in `block in <top (required)>'
	45: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli.rb:25:in `start'
	44: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
	43: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli.rb:31:in `dispatch'
	42: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
	41: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
	40: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
	39: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli.rb:478:in `exec'
	38: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli/exec.rb:23:in `run'
	37: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli/exec.rb:58:in `kernel_load'
	36: from /Users/olivier/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/bundler-2.2.30/lib/bundler/cli/exec.rb:58:in `load'
	35: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/bin/fastlane:23:in `<top (required)>'
	34: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/bin/fastlane:23:in `load'
	33: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/bin/fastlane:23:in `<top (required)>'
	32: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/cli_tools_distributor.rb:122:in `take_off'
	31: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/commands_generator.rb:42:in `start'
	30: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/commands_generator.rb:353:in `run'
	29: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/delegates.rb:18:in `run!'
	28: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb:117:in `run!'
	27: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/runner.rb:444:in `run_active_command'
	26: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/command.rb:157:in `run'
	25: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/commander-4.6.0/lib/commander/command.rb:187:in `call'
	24: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/commands_generator.rb:225:in `block (2 levels) in run'
	23: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/one_off.rb:22:in `execute'
	22: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/one_off.rb:42:in `run'
	21: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:229:in `execute_action'
	20: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:229:in `chdir'
	19: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:255:in `block in execute_action'
	18: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/actions/actions_helper.rb:69:in `execute_action'
	17: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/fastlane-2.196.0/fastlane/lib/fastlane/runner.rb:263:in `block (2 levels) in execute_action'
	16: from /Users/olivier/Documents/Dev/tools/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb:11:in `run'
	15: from /Users/olivier/Documents/Dev/tools/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb:142:in `comment_on_pr'
	14: from /Users/olivier/Documents/Dev/tools/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb:22:in `github_client'
	13: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/client/users.rb:34:in `user'
	12: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/connection.rb:19:in `get'
	11: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/connection.rb:156:in `request'
	10: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/sawyer-0.8.2/lib/sawyer/agent.rb:94:in `call'
	 9: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/connection.rb:200:in `get'
	 8: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/connection.rb:516:in `run_request'
	 7: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/rack_builder.rb:154:in `build_response'
	 6: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/request/retry.rb:148:in `call'
	 5: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/middleware/follow_redirects.rb:61:in `call'
	 4: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/middleware/follow_redirects.rb:73:in `perform_with_redirection'
	 3: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/middleware.rb:18:in `call'
	 2: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/response.rb:59:in `on_complete'
	 1: from /Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/faraday-1.8.0/lib/faraday/middleware.rb:19:in `block in call'
/Users/olivier/Documents/Dev/tools/release-toolkit/vendor/bundle/ruby/2.6.0/gems/octokit-4.21.0/lib/octokit/response/raise_error.rb:14:in `on_complete': [!] GET https://api.github.com/user: 401 - Requires authentication // See: https://docs.github.com/rest/reference/users#get-the-authenticated-user (Octokit::Unauthorized)
```
</details>

I suggest instead to make this method return a `user_error!` when the token is not defined. As per ruby naming conventions, this method should thus be named with a `!` to indicate that it can raise (and of course you'd then update the call site on line 19 accordingly).

```suggestion
      def self.github_token!
        token = [
          'GHHELPER_ACCESS', # For historical reasons / backward compatibility
          'GITHUB_TOKEN',    # Used by the `gh` CLI tool
        ].map { |key| ENV[key] }.compact.first
        token || UI.user_error!('Please provide a GitHub authentication token via the `GITHUB_TOKEN` environment variable')
      end
```

spec/github_helper_spec.rb
```diff
@@ -19,4 +19,27 @@
...
+      ENV['GHHELPER_ACCESS'] = 'GHHELPER_ACCESS'
+      expect(described_class.github_token).to eq('GHHELPER_ACCESS')
+    end
+
+    it 'can use `GITHUB_TOKEN`' do
+      ENV['GITHUB_TOKEN'] = 'GITHUB_TOKEN'
+      expect(described_class.github_token).to eq('GITHUB_TOKEN')
+    end
+
+    it 'prioritizes GHHELPER_ACCESS` over `GITHUB_TOKEN` if both are present' do
```
AliSoftware, line 39: I wonder if we should not instead consider that `GHHELPER_ACCESS` is deprecated, meaning that while we'd still keeping support for it for a while like you did, `GITHUB_TOKEN` should be considered the "new way" and thus the one being prioritized instead?

spec/github_helper_spec.rb
```diff
@@ -19,4 +19,27 @@
...
+      ENV['GITHUB_TOKEN'] = 'GITHUB_TOKEN'
+      ENV['GHHELPER_ACCESS'] = 'GHHELPER_ACCESS'
+      expect(described_class.github_token).to eq('GHHELPER_ACCESS')
+    end
+  end
```
AliSoftware, line 104: Please add a unit test for when neither of the 2 env vars are defined.

As per [my suggestion above](https://github.com/wordpress-mobile/release-toolkit/pull/313/files#r757899837), the test should check that `UI.user_error!` is called, and with the expected message, when this happens.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      def self.details
+        <<~DETAILS
+          If used just once, this method makes it nice and easy to post a quick comment to a GitHub
+          PR. Subsequent runs will allow you to update an existing comment as many times as you need
+          to (for instance, across multiple CI runs).
+
+          The `:reuse_identifier` config item enables this.
+        DETAILS
```
AliSoftware: Nit/wording:

```suggestion
        <<~DETAILS
          If used just once, this method makes it nice and easy to post a quick comment to a GitHub PR.

          Subsequent runs will allow you to update an existing comment as many times as you need to
          (e.g. across multiple CI runs), by using a `:reuse_identifier` to identify the comment to update.
        DETAILS
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
mokagio: If I read this correctly, in case where the environment has both `GHHELPER_ACCESS` and `GITHUB_TOKEN`, the code will pick the former, legacy one. 

I think the desired resolution would be to pick the modern value, right?

I mean, it doesn't change much because both work, so maybe 🤷 

Update: oh, [from the tests](https://github.com/wordpress-mobile/release-toolkit/pull/313/files#r757900158), I see this was the desired behavior. 🤔 Let's keep discussing this [there](https://github.com/wordpress-mobile/release-toolkit/pull/313/files#r757900158).

mokagio (commented):
I like where this is going.

AliSoftware (commented):
Also it would be really nice to get Unit Tests for the `comment_on_pr` helper method.

e.g. you could use a test double for `Octokit::Client` and make it return a dummy list of `issue_comments` and expect then either `add_comment` or `update_comment` to be called depending on the mock comments used.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: Agreed – this isn't a breaking change since not having the environment variable would case the code to crash anyways. 

Added in 0374a55f0b81c2d5cec4e0a277991ffd7c6ae299 (with a test)

spec/github_helper_spec.rb
jkmassel: Yeah, I thought about this, but it would be a backward-incompatible change – there's no guarantee that we don't have both of these set somewhere, and I didn't want anything to break because of it.

In a major version bump I'd definitely be good to drop `GHHELPER_ACCESS` entirely though and scrub our codebases to make sure we no longer reference it.

spec/github_helper_spec.rb
jkmassel: Added with the check in 0374a55f0b81c2d5cec4e0a277991ffd7c6ae299

spec/github_helper_spec.rb
AliSoftware: Ah didn't think about the breaking change behavior, very good point!

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: @mokagio I asked the same question [here](https://github.com/wordpress-mobile/release-toolkit/pull/313#discussion_r758538199) and Jeremy made a good point that while this would be nice, this would be a breaking change to do so…

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
jkmassel: I copy/pasted from an old action and saw there was a name in it so I didn't know if people were doing this or not.

I've updated it to be `Automattic`, because I actually prefer it that way 😁

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: Addressed in 0db0d86a4cdf2e73695dba7224d44a5a58611fb2

jkmassel:
> Also it would be really nice to get Unit Tests for the `comment_on_pr` helper method.
> 
> e.g. you could use a test double for `Octokit::Client` and make it return a dummy list of `issue_comments` and expect then either `add_comment` or `update_comment` to be called depending on the mock comments used.

@AliSoftware done in ff03204cd48bfc9c956f3ad7a3c5cfd51a970bd8 (and 847b4eacd992e46ee7b4dc9a187b43fbfa6b3563 because rubocop)

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ _None_
 
 * Added a new `ios_generate_strings_file_from_code` action to replace the now-deprecated `ios_localize_project` action (and `Scripts/localize.py` script in app repos). [#309, #311]
 
+* Added a `comment_on_pr` action to allow commenting on (and updating comments on) PRs.
+
+* Added the ability to use the `GITHUB_TOKEN` environment variable for GitHub operations. `GHHELPER_ACCESS` will be deprecated in a future version.
+
```
AliSoftware: 💄  Small nit: adding PR # to the changelog

```suggestion
* Added a `comment_on_pr` action to allow commenting on (and updating comments on) PRs. [#313]

* Added the ability to use the `GITHUB_TOKEN` environment variable for GitHub operations. `GHHELPER_ACCESS` will be deprecated in a future version. [#313]

```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,91 @@
+require 'fastlane/action'
+require 'rubygems/command_manager'
+require 'securerandom'
```
AliSoftware: This `require` is not needed anymore in the action

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,91 @@
+require 'fastlane/action'
+require 'rubygems/command_manager'
```
AliSoftware: Why is this `require` needed? I don't see any code in the action where this would be used, did I miss something?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
```diff
@@ -0,0 +1,91 @@
...
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :reuse_identifier,
+            description: 'If provided, the reuse identifier can identify an existing comment to overwrite',
+            type: String,
+            default_value: nil
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :project,
+            description: 'The project slug (ex: `rails/rails`)',
```
AliSoftware: 💄 
```suggestion
            description: 'The project slug (ex: `wordpress-mobile/WordPress-iOS`)',
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -7,8 +7,19 @@ module Fastlane
 
   module Helper
     class GithubHelper
+      def self.github_token
```
AliSoftware: Ruby conventions suggest to end the name with a `!` to indicate the method is throwing
```suggestion
      def self.github_token!
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -7,8 +7,19 @@ module Fastlane
...
+        ].map { |key| ENV[key] }
+                .compact
+                .first
+
+        token || UI.user_error!('Please provide a GitHub authentication token via the `GITHUB_TOKEN` environment variable')
+      end
+
       def self.github_client
-        client = Octokit::Client.new(access_token: ENV['GHHELPER_ACCESS'])
+        client = Octokit::Client.new(access_token: github_token)
```
AliSoftware: Assuming you apply the previous suggestion on line 10, then:
```suggestion
        client = Octokit::Client.new(access_token: github_token!)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +139,28 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
 
         download_path
       end
+
+      # Creates (or updates an existing) GitHub PR Comment
+      def self.comment_on_pr(project_slug:, pr_number:, body:, reuse_identifier: SecureRandom.uuid)
```
AliSoftware, line 145: This requires that you add a `require 'securerandom'` at the top of that file (i.e. move the one you currently have in the action to instead do the `require` at the level of this helper file instead)

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +139,28 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
...
+        client = github_client
```
AliSoftware, line 146: 🤔 💭  The better practice in Ruby-world to do memoization and avoid the `github_client` to be reconstructed on every call would be something like this at the method definition:

```
def self.github_client
  @@client ||= begin
     … the rest of the already existing code
  end
end
```

_The `memoized_var ||= complex_value` construct has the effect of returning `memoized_var` if it already exists and is not `nil` – thus immediately returning the cached value – and if not yet cached, would evaluate the rhs `complex_value`, assign (`=`) it to the `memoized_var`, and return the value._

Usually for the `memoized_var` we use an **instance variable** (prefixed by `@`), so `@foo` or `@client` here. But since this method is a class method and this whole Helper class is not written in a way that expects us to instantiate it (as all its methods are `def self.xx` class methods), I _think_ we are expected to use `@@client` – aka a **class variable** rather than an instance variable – here? That being said I'm also remembering that class variables (`@@foo`) are not recommended in practice and might trigger a [rubocop violation](https://rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ClassVars)… so not sure what's the best way there.

---

I think the official proper solution would be to not use class methods for the Helper and instead expect us to instantiate a `GitHubHelper` instance (from the action, at call site) to be used. That would make more sense, make it easier to test as well, etc. Similar to how in Swift shared instances and singletons are usually not the best pattern and creating instances is more recommended. But then that would probably be a change outside of this PR's scope.

---

Do what you want with all that knowledge 😅 , either try the `@@client ||= …` solution I mention and see how `rubocop` reacts, or maybe SO has suggestions there as well (didn't check) for those cases… or just keep it for a future PR (though we all know how those kind of "TODOs" go…) 🤷

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +139,28 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
...
+        comments = client.issue_comments(project_slug, pr_number)
+
+        complete_reuse_identifier = "<!-- REUSE_ID: #{reuse_identifier} -->"
```
AliSoftware: 💄  nit: I'd name this `reuse_marker` or something like this instead.

`complete_…` reminds me of "completion", like if we were trying to do auto-completion of the reuse ID… while this is more like a "marker comment" here.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -128,6 +139,28 @@ def self.download_file_from_tag(repository:, tag:, file_path:, download_folder:)
...
+
+        existing_comment = comments
+                           .select { |comment| comment.user.id == client.user.id } # Only match comments posted by the owner of the GitHub Token
+                           .find { |comment| comment.body.include?(complete_reuse_identifier) }
```
AliSoftware: I'm ok with this… but as an alternative/suggestion, it could also be a bit better to loop only once, and put both conditions in the same block:

```suggestion
        existing_comment = comments.find do |comment|
           # Only match comments posted by the owner of the GitHub Token, and with the given reuse ID
          comment.user.id == client.user.id && comment.body.include?(complete_reuse_identifier)
        end
```

spec/github_helper_spec.rb
```diff
@@ -19,4 +19,87 @@
...
+      comment_on_pr
+    end
+
+    it 'will not match text outside the reuseID tag' do
+      allow(client).to receive(:issue_comments).and_return([mock_comment(body: 'test-id')])
+      expect(client).to receive(:add_comment)
+      comment_on_pr
+    end
+
+    it 'will not match comments belonging to other users' do
```
AliSoftware, line 82: 👌

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
jkmassel: Because I copy/pasted from another one and forgot to remove this 😅 

Addressed in bb71e2c967c7052ee5c60ab21db2212f78a63e91

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
jkmassel: Addressed in bb71e2c967c7052ee5c60ab21db2212f78a63e91

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: Done in 416fcc798988c2f3d0bec0209e5e95835b030f79

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: Rubocop doesn't love it, but it's the cleanest option for now, so I did that in d0ed11e6fb47410dfce650c129dbe1a62e5e3517 and added an exception.

In the future it'd be nice to not have quite so many static helpers floating around, but I think doing that here would kinda clutter up this PR.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: Makes sense – addressed in 76a7626cc548ad580390ea5687fef6027ba842f8

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: Yeah, I was back-and-forth on this – `O(n*2)` is IMHO ok for small datasets if it makes things easier to read, but since you mentioned a preference I'm fine to have my mind changed – addressed in 32fef0425dcdeb53d6b023ffbf2a4314a00bd964

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb
jkmassel: Addressed in 904b5540dbaffe729cf6bf80ad577668417265c7, but I kept the slug lowercase :)

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -1,23 +1,37 @@
...
+          UI.message("Logged in as: #{user.name}")
 
-        # Auto-paginate to ensure we're not missing data
-        client.auto_paginate = true
+          # Auto-paginate to ensure we're not missing data
+          client.auto_paginate = true
+        end
 
-        client
+        @@client
```
AliSoftware: In ruby, assignments like `a = b` or `a ||= b` already return the assigned value (just like in C where an expression like  `a = 1` itself evaluates to 1…).

That's the whole beauty of the `@foo ||= begin … end` construct as you only have to wrap the whole method in that without having to add anything else at the end after that wrapping to make the memoization work.

So you should remove this last `@@client` line and let the whole `@@client ||= begin … end` block be the return value itself.

_(I'm not creating a GitHub suggestion for it because you'll also need to remove the empty lines above it to not make `rubocop` complain, but GitHub does not support `suggestion` blocks on deleted lines…)_

AliSoftware (approved):
Left a small nit to conform to ruby conventions and simplify code on the `@@client ||= begin … end` memoization pattern usage, I'll let you fix it. After that last tweak, it should be good to merge. 🎉 

(I'm gonna approve the PR anyway so that you can merge as soon as you addressed that last tweak without having to wait for me)

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
jkmassel: TIL, thanks!

Addressed in https://github.com/wordpress-mobile/release-toolkit/pull/313/commits/f495914d48b63dd71437b65f35b85143c18f5941

PR #316, opened 2021-11-29
Sanitize GitHub links in Release Notes

AliSoftware:
## The issue

It seems like when we put full URLs to GitHub PRs in the body of a GitHub Release (especially if that URL is inside `[]` brackets), like is the case for our release notes, then GitHub renders those in an odd way.

For example, `[https://github.com/wordpress-mobile/WordPress-Android/pull/1234]`  would render like this:
image

## Changes made in this PR to fix it

The change in this toolkit PR replaces instances of text like `https://github.com/wordpress-mobile/WordPress-Android/pull/1234` with `wordpress-mobile/WordPress-Android#1234` – which is then properly rendered by the GitHub UI as a link to the appropriate PR.

It also handles:
 - `/issues/1234` in addition to `/pull/1234`
 - Common typo of `/issue/1234` instead of the official `/issues/1234` as well as `/pulls/1234` instead of the official `/pull/1234` (yep, GitHub is quite inconsistent here)
 - Links to PRs in other repos, like `https://github.com/wordpress-mobile/gutenberg-mobile/pull/4277` will become `wordpress-mobile/gutenberg-mobile#4277`.

I've tested this RegEx and this new line of code using the `pry` REPL, by running `release_notes = File.read('WordPress/metadata/release_notes.txt')` in the REPL, then pasting that new line of code afterwards, and validating that `release_notes` variable after that contained the expected fixed text.

## View the effect in action

You can see for example on [this past GitHub Release](https://github.com/wordpress-mobile/WordPress-Android/releases/tag/18.7-rc-1) how the body is oddly rendered around those links.

I've manually applied (via VSCode search/replace) the same substitution regex as in this code uses on [the latest GitHub Release](https://github.com/wordpress-mobile/WordPress-Android/releases/tag/18.8-rc-1). You can see how the links are rendered in a way nicer way, including the ones to other repos like Gutenberg-Mobile.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
@@ -11,6 +11,7 @@ module Fastlane
         version = params[:version]
         assets = params[:release_assets]
         release_notes = params[:release_notes_file_path].nil? ? '' : File.read(params[:release_notes_file_path])
+        release_notes.gsub!(%r{https://github.com/([^/]*/[^/]*)/(pulls?|issues?)/([0-9]*)}, '\1#\3')
         prerelease = params[:prerelease]
 
         UI.message("Creating draft release #{version} in #{repository}.")


mokagio (approved):
👍 thank you for this improvement

AliSoftware:
I just realized: why didn't Peril warn me that I forgot to add an entry in the CHANGELOG for this one?! 🤔

mokagio:
> I just realized: why didn't Peril warn me that I forgot to add an entry in the CHANGELOG for this one?! 🤔

Possibly because [there is no configuration for this repo](https://github.com/Automattic/peril-settings/search?q=release-toolkit) in our Peril instance?

AliSoftware:
@mokagio True, but we have [a local `Danger` rule for it](https://github.com/wordpress-mobile/release-toolkit/blob/develop/Dangerfile#L29-L31) though, that [Buildkite is supposed to run](https://github.com/wordpress-mobile/release-toolkit/blob/develop/.buildkite/pipeline.yml#L51-L52) 😉 

[The log for the Danger run on this PR (just before it got merged](https://buildkite.com/wordpress-mobile/release-toolkit/builds/159#4a7c611b-344c-47a1-aed4-9883342fc224) seems to have believed that this build wasn't for a PR. Maybe it's because I pushed the commit first – which triggered the Buildkite build but outside of a PR context, hence the skip – and only created the PR afterwards (well how could I have created a PR without pushing the sole commit first anyway? ^^) and at that point Buildkite decided not to re-run the checks since they were on the same commit than the last build was on? 🤔 

In contrast, [this other Buildkite run, for a different PR](https://buildkite.com/wordpress-mobile/release-toolkit/builds/152#58cdd555-bc71-43dd-9c6f-bd3e7f141ad8) clearly ran Danger and generated violations for it as expected, so at least it seems to work on other PRs. So maybe that only works (i.e. doesn't skip Danger) if you push new commits on an existing PR, but skips it for any commit that was pushed before the PR was opened?

I wonder if there's something we need to configure in Buildkite there (in order to avoid running the CI on branches that don't have an associated PR yet)? (Maybe @jkmassel have some insights there?)

PR #318, opened 2021-12-03
Fixes broken GitHub helper

jkmassel:
Fixes a bug introduced in #313 that causes any GitHub helper action to fail.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -21,7 +21,7 @@ module Fastlane
 
       def self.github_client
         @@client ||= begin
-          Octokit::Client.new(access_token: github_token!)
+          client = Octokit::Client.new(access_token: github_token!)
 
           # Fetch the current user
           user = client.user
@@ -29,6 +29,8 @@ module Fastlane
 
           # Auto-paginate to ensure we're not missing data
           client.auto_paginate = true
+
+          client
         end
       end
 
--- a/spec/github_helper_spec.rb
+++ b/spec/github_helper_spec.rb
@@ -49,10 +49,10 @@ describe Fastlane::Helper::GithubHelper do
 
   describe 'github_client' do
     let(:client) do
-      double(
+      instance_double(
         Octokit::Client,
         user: instance_double('User', name: 'test'),
-        'auto_paginate=': nil,
+        'auto_paginate=': nil
       )
     end
 
--- a/spec/github_helper_spec.rb
+++ b/spec/github_helper_spec.rb
@@ -47,6 +47,25 @@ describe Fastlane::Helper::GithubHelper do
     end
   end
 
+  describe 'github_client' do
+    let(:client) do
+      double(
+        Octokit::Client,
+        user: instance_double('User', name: 'test'),
+        'auto_paginate=': nil,
+      )
+    end
+
+    before do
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -21,14 +21,16 @@ def self.github_token!
 
       def self.github_client
         @@client ||= begin
-          Octokit::Client.new(access_token: github_token!)
+          client = Octokit::Client.new(access_token: github_token!)
```
jkmassel, line 24: Turns out if we don't define this, it's hard to pass it out of the block 🤦

spec/github_helper_spec.rb
```diff
@@ -47,6 +47,25 @@
     end
   end
 
+  describe 'github_client' do
```
jkmassel, line 50: Added a test to verify the fix so hopefully we can avoid this next time!

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: 🤦‍♂️ Sorry for misleading you on that one with my review on #313, reading the diff on GitHub I thought the client variable that I made you remove was out of the begin/end block, not inside of it… oops.

Nice touch adding a unit test for that 👍

spec/github_helper_spec.rb
```diff
@@ -47,6 +47,25 @@
...
+    let(:client) do
+      instance_double(
+        Octokit::Client,
+        user: instance_double('User', name: 'test'),
+        'auto_paginate=': nil
+      )
```
AliSoftware, line 56: Only potential issue I have with that is if we change the implementation detail in the future, and for example the implementation ends up calling more methods… won't this risk calling the real ones (and even potentially make real API/network calls)?

I'm not totally sure tbh as I don't remember how test doubles work in Ruby/rspec, if they fallback to the real object's implementation if a method is not mocked (which we'd want to avoid in this case), or do they fail the test (what we'd want here)?
_(Or is that the difference between normal doubles and verifying doubles? 🤔 I forgot)_

spec/github_helper_spec.rb
```diff
@@ -47,6 +47,25 @@
...
+
+    before do
+      allow(described_class).to receive(:github_token!).and_return('')
+      allow(Octokit::Client).to receive(:new).and_return(client)
+    end
+
+    it 'is not nil' do
+      expect(described_class.github_client).not_to be_nil
+    end
+  end
```
AliSoftware, line 67: I suggest to add another test checking that the memoization is working as expected. Especially than when calling the `github_client` multiple times, it will only create a single `Octokit::Client` instance and only authenticate once (instead of each time it's referenced)

spec/github_helper_spec.rb
jkmassel: Yep – with verifying doubles the test will fail if we call anything else. And if we try to make a network call it'll fail the test as well, so I feel like we're pretty covered there?

oguzkocer (approved):
@jkmassel mentioned that he'll address @AliSoftware's feedback in a subsequent PR due to the urgency of this PR. Since the feedback is for the test classes, I agree that it can wait an extra day or two.

spec/github_helper_spec.rb
jkmassel: This is probably a good idea – I'll take a look at it soon!

PR #319, opened 2021-12-04
Update default branch to trunk

oguzkocer:
This PR prepares the `release-toolkit` for the removal of `develop` branch for its clients as well as the library itself. I haven't worked with `release-toolkit` almost at all, so I basically searched for `develop` and replaced the ones that made sense to me with `trunk`. I considered using a `DEFAULT_BRANCH` constant, but I wasn't sure where to put it. 😅 Since we also refer to this branch in comments and it's not very likely to change again, I proceeded with a simple text replacement.

I am hoping reviewers can point out anything I might have missed. If you can take a moment to do a quick search, I'd really appreciate it.

Please review this PR but keep it as draft until the clients are ready for the update. Once we merge this, I think we should do a major update as this is a breaking change.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -74,7 +74,7 @@ task :new_release do
   Console.header 'Commit and push changes...'
   GitHelper.commit_files("Bumped to version #{new_version}", [VERSION_FILE, 'Gemfile.lock', 'CHANGELOG.md'])
 
-  Console.header 'Opening PR drafts in your default browser...'
+  Console.header 'Opening PR draft in your default browser...'
   GitHelper.prepare_github_pr("release/#{new_version}", 'trunk', "Release #{new_version} into trunk", "New version #{new_version}. Be sure to create a GitHub Release and tag once this PR gets merged.")
 
   Console.info <<~INSTRUCTIONS
--- a/Rakefile
+++ b/Rakefile
@@ -75,7 +75,6 @@ task :new_release do
   GitHelper.commit_files("Bumped to version #{new_version}", [VERSION_FILE, 'Gemfile.lock', 'CHANGELOG.md'])
 
   Console.header 'Opening PR drafts in your default browser...'
-  GitHelper.prepare_github_pr("release/#{new_version}", 'develop', "Release #{new_version} into develop", "New version #{new_version}")
   GitHelper.prepare_github_pr("release/#{new_version}", 'trunk', "Release #{new_version} into trunk", "New version #{new_version}. Be sure to create a GitHub Release and tag once this PR gets merged.")
 
   Console.info <<~INSTRUCTIONS
@@ -84,8 +83,8 @@ task :new_release do
 
     >>> WHAT'S NEXT
 
-    1. Create PRs to `develop` and `trunk`.
-    2. Once the PRs are merged, publish a GitHub release for \`#{new_version}\`, targeting \`trunk\`,
+    1. Create a PR against `trunk`.
+    2. Once the PR is merged, publish a GitHub release for \`#{new_version}\`, targeting \`trunk\`,
        creating a new \`#{new_version} tag in the process.
 
     The creation of the new tag will trigger a CI workflow that will take care of doing the
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
@@ -8,8 +8,9 @@ module Fastlane
         require_relative '../../helper/android/android_version_helper'
         require_relative '../../helper/android/android_git_helper'
 
-        # Checkout trunk and update
-        Fastlane::Helper::GitHelper.checkout_and_pull('trunk')
+        # Checkout default branch and update
+        default_branch = params[:default_branch] || 'develop'
+        Fastlane::Helper::GitHelper.checkout_and_pull(default_branch)
 
         # Check versions
         release_version = Fastlane::Helper::Android::VersionHelper.get_release_version
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
@@ -9,7 +9,7 @@ module Fastlane
         require_relative '../../helper/android/android_git_helper'
 
         # Checkout default branch and update
-        default_branch = params[:default_branch] || 'develop'
+        default_branch = params[:default_branch]
         Fastlane::Helper::GitHelper.checkout_and_pull(default_branch)
 
         # Check versions
@@ -81,6 +81,11 @@ module Fastlane
                                        description: 'Skips confirmation',
                                        is_string: false, # true: verifies the input is a string, false: every kind of value
                                        default_value: false), # the default value if the user didn't provide one
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH',
+                                       description: 'Default branch of the repository',
+                                       is_string: false, # true: verifies the input is a string, false: every kind of value
+                                       default_value: Fastlane::Helper::GitHelper::DEFAULT_GIT_BRANCH),
         ]
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
@@ -8,8 +8,8 @@ module Fastlane
         require_relative '../../helper/android/android_version_helper'
         require_relative '../../helper/android/android_git_helper'
 
-        # Checkout develop and update
-        Fastlane::Helper::GitHelper.checkout_and_pull('develop')
+        # Checkout trunk and update
+        Fastlane::Helper::GitHelper.checkout_and_pull('trunk')
 
         # Check versions
         release_version = Fastlane::Helper::Android::VersionHelper.get_release_version
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
@@ -84,7 +84,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :default_branch,
                                        env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH',
                                        description: 'Default branch of the repository',
-                                       is_string: false, # true: verifies the input is a string, false: every kind of value
+                                       type: String,
                                        default_value: Fastlane::Helper::GitHelper::DEFAULT_GIT_BRANCH),
         ]
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
@@ -8,7 +8,7 @@ module Fastlane
         require_relative '../../helper/android/android_version_helper'
         require_relative '../../helper/android/android_git_helper'
 
-        other_action.ensure_git_branch(branch: 'develop')
+        other_action.ensure_git_branch(branch: 'trunk')
 
         # Create new configuration
         new_short_version = Fastlane::Helper::Android::VersionHelper.bump_version_release
@@ -28,9 +28,9 @@ module Fastlane
         UI.message("New version: #{new_short_version}")
         UI.message("Release branch: #{new_release_branch}")
 
-        # Update local develop and branch
+        # Update local trunk and branch
         UI.message 'Creating new branch...'
-        Fastlane::Helper::GitHelper.create_branch(new_release_branch, from: 'develop')
+        Fastlane::Helper::GitHelper.create_branch(new_release_branch, from: 'trunk')
         UI.message 'Done!'
 
         UI.message 'Updating app version...'
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
@@ -8,7 +8,7 @@ module Fastlane
         require_relative '../../helper/android/android_version_helper'
         require_relative '../../helper/android/android_git_helper'
 
-        default_branch = params[:default_branch] || 'develop'
+        default_branch = params[:default_branch]
         other_action.ensure_git_branch(branch: default_branch)
 
         # Create new configuration
@@ -53,7 +53,13 @@ module Fastlane
       end
 
       def self.available_options
-        # Define all options your action supports.
+        [
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH',
+                                       description: 'Default branch of the repository',
+                                       is_string: false, # true: verifies the input is a string, false: every kind of value
+                                       default_value: Fastlane::Helper::GitHelper::DEFAULT_GIT_BRANCH),
+        ]
       end
 
       def self.output
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
@@ -8,7 +8,8 @@ module Fastlane
         require_relative '../../helper/android/android_version_helper'
         require_relative '../../helper/android/android_git_helper'
 
-        other_action.ensure_git_branch(branch: 'trunk')
+        default_branch = params[:default_branch] || 'develop'
+        other_action.ensure_git_branch(branch: default_branch)
 
         # Create new configuration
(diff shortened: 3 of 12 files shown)


jkmassel (commented):
This is an interesting approach for sure – every place I see `develop` I feel like we ought to treat as a bug where we should've passed in an argument.

We could presume the existence of a `DEFAULT_BRANCH` constant defined in the `fastfile` moving forward – it would still be a breaking change, but at least this would be configurable (because we'll likely want to use this for projects that don't use `trunk` as the default branch).

oguzkocer:
@jkmassel I was just about to ask for clarification for your review until I saw [your feedback on WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/15652#pullrequestreview-823295867).

The main thing I was confused about was that I specifically shared the reason for the text replacement in the PR description, so I initially found the "interesting approach" terminology a little bit unfair. I interpreted it as it's the "wrong approach" (not sure how else to interpret it really) because I'd argue that this PR itself does not have an approach, it just doesn't fix the one that's already in place. It also leaves room for that improvement in the PR description - with some help from the reviewers on how to handle the constant and feedback on whether it was a necessary change.

Anyhow, it's all clear to me now and I'll look into how constants are retrieved from the Fastfile and make this change on Monday. I thought I'd leave this feedback to you for your future PR reviews. Thanks for the review.

oguzkocer:
@jkmassel I've updated the implementation to accept a parameter for the default branch and used `develop` as the fallback option for backwards compatibility. I thought about using `default_branch` or `base_branch` for this parameter, and ended up going with `default_branch` because that's what Github calls it. However, I am open to suggestions if you think there is a clearer name. (or if you like `base_branch` better)

I am not sure how to test this without implementing it in one of the clients, so I was hoping you could review it first - since it's a small changeset.

I am hoping this addresses your feedback, so let me know if I am missing anything. Thanks!

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -8,8 +8,9 @@ def self.run(params)
         require_relative '../../helper/android/android_version_helper'
         require_relative '../../helper/android/android_git_helper'
 
-        # Checkout develop and update
-        Fastlane::Helper::GitHelper.checkout_and_pull('develop')
+        # Checkout default branch and update
+        default_branch = params[:default_branch] || 'develop'
```
mokagio: @oguzkocer you might have to explicitly define a new `FastlaneCore::ConfigItem` in the `available_options` below (line 72) in order to use this.

Fastlane has it's own mechanism to manage the parameters it sends to the actions, but Ruby's deceiving simplicity doesn't make it clear.

I verified this by adding a similar line to an action that has a test and got errors due to it:

```diff
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
@@ -2,6 +2,7 @@ module Fastlane
   module Actions
     class IosGenerateStringsFileFromCodeAction < Action
       def self.run(params)
+        foo = params[:foo] || 'foo'
         files = files_matching(paths: params[:paths], exclude: params[:exclude])
         flags = [('-q' if params[:quiet]), ('-SwiftUI' if params[:swiftui])].compact
         flags += Array(params[:routines]).flat_map { |routine| ['-s', routine] }
```

Errors when running the tests:

```
     Failure/Error: foo = params[:foo] || 'foo'

     NoMethodError:
       undefined method `fetch_env_value' for nil:NilClass
```

After adding the value, the tests passed as usual. 

```diff
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
@@ -2,6 +2,7 @@ module Fastlane
   module Actions
     class IosGenerateStringsFileFromCodeAction < Action
       def self.run(params)
+        foo = params[:foo]
         files = files_matching(paths: params[:paths], exclude: params[:exclude])
         flags = [('-q' if params[:quiet]), ('-SwiftUI' if params[:swiftui])].compact
         flags += Array(params[:routines]).flat_map { |routine| ['-s', routine] }
@@ -90,6 +91,10 @@ module Fastlane
                                        description: 'If true, will fail with user_error! if `genstrings` printed any error while parsing',
                                        is_string: false, # Boolean
                                        default_value: true),
+          FastlaneCore::ConfigItem.new(key: :foo,
+                                       env_name: 'FL_IOS_GENERATE_STRINGS_FILE_FROM_CODE_FOO',
+                                       description: 'foo',
+                                       default_value: 'foo'),
         ]
       end
 ```

This approach has the advantage of moving the default value, or rather fallback if the user doesn't set it, to the `ConfigItem` object, so the caller code becomes a bit clearer as one merely needs to run `params[:foo]`.

**Comment by mokagio** [comment]
@oguzkocer 

> I am not sure how to test this without implementing it in one of the clients...

Yeah 🤔 These lanes are massive and depends on a lot of local state, writing tests for them in their current implementation would prove to be quite time consuming.

---

+1 for using `default_branch`. `base_branch` would have made me think of the base branch for a PR.

---

Regarding using `develop` for backward compatibility, I think we're better off using the default we'd like to have in the future `trunk` and cut a new breaking change version.

The reason I'm saying this is because otherwise we'll always have to write `trunk` (no big deal) and also carry the legacy of `develop` in the codebase (even though it's limited to the default/fallback value definition). Given we're the only users of this library and we have SemVer on our side, I think it's okay to introduce a breaking change rather than making the usage (a tiny bit) more cumbersome. 

What do you think?

**Comment by AliSoftware** [comment]
> The reason I'm saying this is because otherwise we'll always have to write `trunk` (no big deal) and also carry the legacy of `develop` in the codebase (even though it's limited to the default/fallback value definition). Given we're the only users of this library and we have SemVer on our side, I think it's okay to introduce a breaking change rather than making the usage (a tiny bit) more cumbersome.

That's fair, but I think it also works with the backwards-compatible solution of using `develop` as default for now, and changing the default in a future release… especially if @oguzkocer designs all the `ConfigItem` that he adds to the various actions to all use the same `env_name` key (despite them being for different actions, I think that make sense that they use the same env var for all).

For example if we pick `ConfigItem.new(…, env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH', …)` for all actions, then all we'd have to do during the transition period would be to have that env var set in a single place (typically something like `ENV['FL_RELEASE_TOOLKIT_DEFAULT_BRANCH'] = 'trunk'` around the top of each `Fastfile` like we do for other similar constants) to apply it to all actions which depend on it.
Then once we do a breaking change and update the default to `trunk` in the toolkit, we could just remove it to use the new toolkit's default.

**Comment by oguzkocer** [comment]
@AliSoftware @mokagio Am I correct in understanding that with the `env_name` argument in the `ConfigItem` we are making it possible to use an environment variable instead of directly passing the value as a parameter?

@mokagio I was planning to reply to the breaking change comment after I got around to fixing the `ConfigItem` issue, but @AliSoftware beat me to it. The reason I went for `develop` was due to a discussion we had on Slack making the initial change non-breaking and then updating the value to `trunk` when we have other breaking changes. So, we'll basically arrive at the same situation you're suggesting, it'll just take an extra PR in the near future with some other possible breaking changes on the horizon.

**Comment by AliSoftware** [comment]
> @AliSoftware @mokagio Am I correct in understanding that with the env_name argument in the ConfigItem we are making it possible to use an environment variable instead of directly passing the value as a parameter?

Exactly. That's something specific to Fastlane and why they introduced those `ConfigItem` to configure the parameters an action would be allowed to take, so that they could attach a lot of metadata to each of these `ConfigItem`/parameters, like the param's own documentation, the default value, type… but also the env var to allow users to use to set that value via env var instead of parameter passing.

If the user does not provide a value for the corresponding parameter at call site when calling the action from their `Fastfile`, then fastlane will
 - Fallback to the value defined by the corresponding ENV var (defined by that `env_name` value here) if it exists and is defined
 - Fallback to the `default_value` of the `ConfigItem` otherwise.

_([Full details here in doc](https://docs.fastlane.tools/advanced/actions/#resolution-order))_

File: lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb (hunk 1 above)
Comment (oguzkocer) [reply lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb]: @mokagio Thanks a lot for pointing this out to me. I implemented the change in 8c510ea458ca3ccb176d4d85c3df969733de0e05, I hope that's correct. Let me know what you think! cc @AliSoftware (due to the other discussions we have had in the comments and on Slack)

File: lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb [hunk 2]
```diff
@@ -80,6 +81,11 @@ def self.available_options
                                        description: 'Skips confirmation',
                                        is_string: false, # true: verifies the input is a string, false: every kind of value
                                        default_value: false), # the default value if the user didn't provide one
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH',
+                                       description: 'Default branch of the repository',
+                                       is_string: false, # true: verifies the input is a string, false: every kind of value
+                                       default_value: Fastlane::Helper::GitHelper::DEFAULT_GIT_BRANCH),
```
AliSoftware: 👌

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
```diff
@@ -80,6 +81,11 @@ def self.available_options
                                        description: 'Skips confirmation',
                                        is_string: false, # true: verifies the input is a string, false: every kind of value
                                        default_value: false), # the default value if the user didn't provide one
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH',
+                                       description: 'Default branch of the repository',
+                                       is_string: false, # true: verifies the input is a string, false: every kind of value
```
AliSoftware: Well this parameter is supposed to be a String, and shouldn't be a boolean, integer or whatnot. So better use validation here. You could use `is_string: true`, but in practice that `is_string` flag has been obsoleted in Fastlane core, and is now replaced by a more versatile `type:`, so better use that from now on.

```suggestion
                                       type: String,
```

(PS: I'm thinking at some point in the future we should probably go thru all the existing actions in the toolkit to replace all the ConfigItems for which we used `is_string: true` to use `type: String` instead, not because it would really matter or fix anything behavior-wise, but because most of us use copy/pasting when we add new items or create new actions, so we might as well set a good precedent. Obviously this is not something for this PR to tackle, but something to keep in mind for later)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
```diff
@@ -52,7 +53,13 @@ def self.details
       end
 
       def self.available_options
-        # Define all options your action supports.
+        [
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH',
+                                       description: 'Default branch of the repository',
+                                       is_string: false, # true: verifies the input is a string, false: every kind of value
```
AliSoftware: Same as above:
```suggestion
                                       type: String,
```

(I'll let you fix all occurrences of that in all other actions rather than comment on every single one of them with a GitHub suggestion 😛 )

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -5,6 +5,11 @@ module Helper
     # Helper methods to execute git-related operations
     #
     module GitHelper
+      # Fallback default branch of the client repository. It's currently set to 'develop' for
+      # backwards compatibility.
+      # TODO: Set to 'trunk' for the next major release.
```
AliSoftware, line 10: 👍 

Side note about something I've been thinking about for a while now that we use SemVer is how to keep track of those cases of "let's deprecate X first and only remove it on next major" or, like here, "let's wait for next major to change this". We have had a few of those cases recently where we wrote some `TODO:` or `FIXME:` sometimes (not sure if every time…)… but it will be quite easy to forget to tackle every one of those next time we do a major version.
So if anyone has any suggestion on how to track those better, I'm all ears…
 - A long-lived GitHub issue for each next major we plan to do, and we'd regularly edit the issue description to add more and more checkboxes with links to lines/methods to change before we do a new release?
 - A note at the very top of the release notes, saying "this and that will have to be deleted on next major release"? (not as nice, but at least it'd be hard to miss when releasing versions)
 - A dedicated `.md` file in the repo – which could double-down as a "Migration Documentation" to tell people what they need to do or be prepared to for each major version
 - A convention on how to name our `TODO:`/`FIXME:` to annotated those, e.g. `FIXME: [NextMajor] blahblah"?

AliSoftware (commented):
This looks good, nice job!

Only nit I left is to make all your `ConfigItem` for the `default_branch` param enforce them to be a `String`. Once this is addressed, I think this should be good to ship 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
oguzkocer: 🤦

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_bump_version_release.rb
oguzkocer: But if you did, I'd just be able to apply your suggestions instead 😛

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
oguzkocer: I like the last option. If we use `TODO` sparingly (I have been using `TODO` because it's syntax highlighted for me, but maybe it works with `FIXME` as well) we could look at all `TODO`s before each major release cut and either address them (in cases like this) or move them into issues and clear them up before each release.

Dedicated `.md` is also 👍 from me.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_betabuild_prechecks.rb
oguzkocer: Addressed in dea6cb49f561f39b5ab05edb9041175aaddc6e3e.

oguzkocer:
@AliSoftware Thanks for the review, sorry about the silly mistake!

I am re-requesting a review from all of you who have reviewed it so far.

Rakefile
```diff
@@ -75,7 +75,6 @@ task :new_release do
   GitHelper.commit_files("Bumped to version #{new_version}", [VERSION_FILE, 'Gemfile.lock', 'CHANGELOG.md'])
 
   Console.header 'Opening PR drafts in your default browser...'
```
AliSoftware:
```suggestion
  Console.header 'Opening PR draft in your default browser...'
```

rakelib/git_helpers.rake
```diff
@@ -11,7 +11,7 @@ module GitHelper
     elsif branch_exists
       Rake.sh('git', 'checkout', release_branch)
     else # create it
-      abort('Aborted, as not run from develop nor release branch') unless current_branch == 'develop' || Console.confirm("You are not on 'develop', nor already on '#{release_branch}'. Do you really want to cut the release branch from #{current_branch}?")
+      abort('Aborted, as not run from trunk nor release branch') unless current_branch == 'trunk' || Console.confirm("You are not on 'trunk', nor already on '#{release_branch}'. Do you really want to cut the release branch from #{current_branch}?")
```
AliSoftware, line 14: Note that this is related to the branch of the `release-toolkit` **itself**, as this script is used when doing the releases of the gem to RubyGems.
So this assumes that the `release-toolkit` repo itself will migrate to `trunk` between when this PR is merged and when we do the next version of it.

Which I think is the plan anyway (including given the changes you also made to the `Rakefile` which are similarly related) so I'm considering this was intentional… but given that this change is a bit different than the other changes in this PR (which are otherwise related to the branch of the _client repos_ the toolkit is used in), I just wanted to double-check this was the intent and highlight it 🙂

AliSoftware (approved):
Assuming the fact that this touches both (1) the branches the toolkit references from the client repos (via the actions), and (2) the default branch of the toolkit itself (Rakefile), is intentional as I think it is (we want to migrate _all the things_ after all 😉) then LGTM 👌

rakelib/git_helpers.rake
oguzkocer: Yes, it was completely intentional. I'll not merge this PR right now as I want to do the `trunk` rename as I am merging this probably on Wednesday.

oguzkocer: And thanks again for the review @AliSoftware 🙇

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
jkmassel: I wonder if we could have a section in the release notes for this? That way we don't need to remember to check another file when cutting a breaking build?

oguzkocer:
The default branch for `release-toolkit` has been renamed to `trunk`.

PR #325, opened 2021-12-16
CI: Make gem push step depend on other checks

AliSoftware:
- Make the `gem-push` step depend on lint, test and danger
 - Update the `bash-cache` plugin version
 - ~Update the Xcode image to 13~ ➡️ Back-pedalled on this because xcode-13 image still [requires the `gem install bundler` FIXITs](https://github.com/wordpress-mobile/WordPress-iOS/blob/35819a3b8b150e7a8dad76a474950cace792cc17/.buildkite/commands/build-for-testing.sh#L3-L5) and `xcode-13.1` image (which fixes that) should be out soon, so better wait for it anyway

@jkmassel feel free to take over this PR to bump the `IMAGE_ID` to `xcode-13.1` if it's ready before this PR gets merged. (Otherwise nbd, we'll just migrate in a separate PR 🙃 )

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   plugins: &common_plugins
-  - &bash_cache automattic/bash-cache#v1.3.2: ~
+  - &bash_cache automattic/bash-cache#2.0.0: ~
   env: &common_env
     IMAGE_ID: xcode-12.5.1
 
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -10,7 +10,7 @@ steps:
   # Build and Test
   #################
   - label: "🧪 Build and Test"
-    key: "test"
+    key: test
     command: |
       .buildkite/install-dependencies.sh
 
@@ -30,7 +30,8 @@ steps:
   #################
   # Lint
   #################
-  - label: "🧹 Lint"
+  - label: "🧹 Lint (Rubocop)"
+    key: rubocop
     command: |
       .buildkite/install-dependencies.sh
 
@@ -45,6 +46,7 @@ steps:
   # Danger
   #################
   - label: "⛔️ Danger"
+    key: danger
     command: |
       .buildkite/install-dependencies.sh
 
@@ -61,6 +63,7 @@ steps:
   - label: ":rubygems: Publish to RubyGems"
     key: "gem-push"
     if: build.tag != null
+    depends_on: ["test", "rubocop", "danger"]
     # Note: We intentionally call a separate `.sh` script here (as opposed to having all the 
     # commands written inline) to avoid leaking a key used in the process in clear in the
     # BUILDKITE_COMMAND environment variable.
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,16 +1,16 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   plugins: &common_plugins
-  - &bash_cache automattic/bash-cache#v1.3.2: ~
+  - &bash_cache automattic/bash-cache#2.0.0: ~
   env: &common_env
-    IMAGE_ID: xcode-12.5.1
+    IMAGE_ID: xcode-13
```
AliSoftware: Could be worth waiting for the `xcode-13.1` image since it's supposed to be released soon by Jeremy (like, tomorrow)? 🤷

AliSoftware: Ok after seeing the CI fail with `xcode-13` image because of `bundler` version, I remembered that we'd need to also apply the temporary fixits for `gem install bundler` (like [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/35819a3b8b150e7a8dad76a474950cace792cc17/.buildkite/commands/build-for-testing.sh#L3-L5)) if we were to use that image, while this will be fixed in the soon-coming `xcode-13.1` image, so better wait for that new image anyway and do that bump only once it's available.

oguzkocer (approved):
I've been using a different syntax in Android projects for the `key` and `depends_on`, but I don't think that'd matter.

For `key`, [I've been using quotation marks](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/.buildkite/pipeline.yml#L30) and for `depends_on`, [I've been separating the items into new lines](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/.buildkite/pipeline.yml#L41-L44). I am not strongly attached to any syntax and would be happy to follow if we have a guideline for it.

AliSoftware:
⚠️  Just noticed: once this PR is merged, we'll need to update the GitHub Settings on the default branch (`trunk`) of this repo, since the (required) check name for lint have changed from `buildkite/release-toolkit/lint` to `buildkite/release-toolkit/lint-rubocop`

[EDIT: Done]

AliSoftware:
> I've been using a different syntax in Android projects for the `key` and `depends_on`, but I don't think that'd matter.
> 
> For `key`, [I've been using quotation marks](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/.buildkite/pipeline.yml#L30) and for `depends_on`, [I've been separating the items into new lines](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/.buildkite/pipeline.yml#L41-L44). I am not strongly attached to any syntax and would be happy to follow if we have a guideline for it.

That works for me 👍 

I don't usually use quotes for the key just because quotes in YAML are mostly necessary around strings if they contain special characters (like spaces, etc) while job keys should ideally not contain such chars, so that kinda acts as a double-check that I wouldn't be tempted in putting emojis or spaces for that field. But I'm really not strong about this at all, and quotes works for me too.

As for `depends_on` I agree that one-item-per-line is actually more readable, will change this 👍

AliSoftware:
@oguzkocer `depends_on` list format addressed in 7024ffd8aa8167ae1c29535c6de3f01c35be8eeb 🙂

PR #326, opened 2022-01-14
Add high-level test for the current `.po` generation behavior 

mokagio:
I'm starting to think of how to refactor (or deprecate and rewrite) the `.po[t]` file generation/update logic (`gp_update_metadata_source`).

My starting point was to understand how the action currently works and I did that with a simple high-level / functional test.

I think there's value in tracking it in the repo anyway, even if just to aid us during this migration stage and then be removed in favor of better ones for the new code.


spec/gp_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        msgid "previous version notes required to have current one added"
+        msgstr ""
+        msgctxt "key1"
+        msgid "this value should change"
+        msgstr ""
+        msgctxt "key2"
+        msgid "this value should change"
+        msgstr ""
+      PO
+      File.write(output_path, dummy_text)
```
mokagio: It was useful to have the file content inline in the test because I was jumping back and forth between it, the expected output, and the production code.

If we'll add more tests for new/modified code, it'll probably be best to have dedicated fixtures or methods doing this kind of work generation to keep the examples short and easy to glance

spec/gp_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      File.write(file_2_path, 'value 2')
+      file_3_path = File.join(dir, '3.txt')
+      File.write(file_3_path, 'value 3')
+
+      described_class.run(
+        po_file_path: output_path,
+        release_version: '1.23',
+        source_files: {
+          release_note: release_notes_path,
+          whats_new: whats_new_path,
```
AliSoftware: Wait what's the difference between `release_note` and `whats_new`? Or is it an intentional duplication you made here to be sure to cover both cases?

I mean I see that we indeed have both types of `ReleaseNoteMetadataBlock` and `WhatsNewMetadataBlock` in the current/existing codebase, to handle them separately, but I can't understand if there's supposed to be a difference in meaning between the two keys, or if, like I fear it is, it's just 2 names for the same thing — aka the Release Notes entry — and it's just that we've historically used different key names for those depending on the projects, in which case even if it's good to have unit tests that check the case for when both are present, in practice it will only be one or the other, depending on the key that the client project historically uses?

In that case in the future refactoring you plan to work on, let's only keep one and formalize its usage on all clients 🙂 

Related: see also [my recent comment on this WCiOS PR here](https://github.com/woocommerce/woocommerce-ios/pull/5906#discussion_r785287011) about this.

spec/gp_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          key1: file_1_path,
+          key2: file_2_path,
+          key3: file_3_path # This is not in the input .po and won't be added
```
AliSoftware: I wonder what we should do here to not forget to fix that.

I mean on one hand I'd want this test to go red, because our goal is ultimately to ensure all keys get included (see also the case of `release_notes_short` for Android which gets skipped because of this) – and the refactoring to fix that will probably re-generate the `.po` file from scratch instead of trying to modify it in place – so we'd want the tests to check what we want.

On the other hand, as you mention in your PR description you wrote those tests to kind of document the _current_, existing (buggy) behavior, not the future one…

One option that could be nice would be to duplicate that test with the expected result (the one that do include `key3` as a new key added in the `.po` file), but [use `xit` instead of `it` to make that test disabled for now](https://www.rubyguides.com/2018/07/rspec-tutorial/#How_to_Temporarily_Disable_a_Test). It would then be easy to turn the test for current behavior from `it` to `xit` and the new one from `xit` to `it` to switch between testing old and new behaviors.

spec/gp_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+describe Fastlane::Actions::GpUpdateMetadataSourceAction do
+  it 'updates any block in a given .po file with the values from the given sources' do
+    Dir.mktmpdir do |dir|
+      # 1: Create a dummy .po file to use as input.
+      output_path = File.join(dir, 'output.po')
+      dummy_text = <<~PO
+        msgctxt "v1.0-whats-new"
+        msgid "this value should change"
+        msgstr ""
+        msgctxt "release_note_0122"
```
AliSoftware: Would be nice to also have `release_note_0121` in here in addition to this one, so that your test also demonstrates/checks that the N-2 notes get **removed** from the `.po` once release notes for N get added in the _expected_ result.

AliSoftware (commented):
Left a few notes.

Also it would be worth adding similar tests for the `an_update_metadata_source_action`, which is a close-to-duplicate action of this `gp_update_metadata_source`.

Since one of the goals of this work on the `.po` generator action is to also **clean up** all the different versions of those scripts there is (to my knowledge there are at least 3: `an_update_metadata_source_action` and `gp_update_metadata_source` in the toolkit but also [this `android_update_metadata_source` in the WPAndroid repo](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/fastlane/actions/android_update_metadata_source.rb)), it would be very useful to have those different specs for each so we can compare them and see how they might differ, because as we all know, in the end…
there can be only one

spec/gp_update_metadata_source_spec.rb
AliSoftware: Today I just discovered some other alternatives to mark a test as "pending" in RSpec: https://relishapp.com/rspec/rspec-core/v/2-0/docs/pending/pending-examples

This makes the test be shown in yellow in the terminal (instead of green for passed or red for failed).

```
Pending: (Failures listed here are expected and do not affect your suite's status)

  1) FastlaneCore::Helper::Ios::L10nHelper#read_strings_file_as_hash 
     # Not yet implemented
     # ./spec/ios_l10n_helper_spec.rb:52
```

I thought it was a nice discovery and feature, and figured that you could take advantage of that here :)

spec/gp_update_metadata_source_spec.rb
mokagio: > Wait what's the difference between `release_note` and `whats_new`?

I think the code expects `release_note` for Android release notes and `whats_new` for the iOS ones. They end up matching different `msgctxt` values and have different handling, i.e. for Android it keeps the `n - 1` version while for iOS it always keeps only the latest.

> In that case in the future refactoring you plan to work on, let's only keep one and formalize its usage on all clients 🙂

I'd like to get to that place, yeah, even though it might require changing how the process works. I don't see the value in keeping the `n - 1` info in the Android `.po`, for example. If we dropped that, and allowed for either Android or iOS adopting the other's `msgctxt`, we should be able to consolidate.

spec/gp_update_metadata_source_spec.rb
mokagio: I like the pending idea. I'll add a dedicated test for the desired behavior and mark it as pending.

spec/gp_update_metadata_source_spec.rb
mokagio: Good point. Will do.

mokagio:
@AliSoftware FYI I only pushed some extra work I did today. This is _not_ ready for a new round of review. 😄

spec/an_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+          release_note: release_notes_path,
+          key1: file_1_path,
+          key2: file_2_path
+        }
+      )
+
+      expected = <<~PO
+        msgctxt "release_note_0123"
+        msgid ""
+        "1.23:\\n"
```
AliSoftware: Is this double-escape required by Ruby (meaning that in practice then expected file with actually contain a verbatim single literal `\` followed by a literal `n`) or by the output (meaning the expected file will contain a double `\\` literally in the text?

From my expectation of that action I'm guessing that's the latter, but it would be nice if we could avoid that double-escape there to avoid confusion. I can't remember what it is from the top of my head but I think Ruby has a syntax similar to `<<~` heredoc but which does not do interpolation (so would behave similar to `%q(…)` / `'…'` as opposed to `%Q(…)` / `"…"`, not replacing variables nor interpreting escapes)? If you can find such heredoc syntax marker in the Ruby doc it would be nice to use it here instead.

AliSoftware: Ah! I knew there was a way!

```ruby
expected = <<~'PO'
  …
PO
```

spec/an_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+        "- something else new\\n"
+        msgstr ""
+
+      PO
+      expect(File.read(output_path)).to eq(expected)
+    end
+  end
+
+  it 'adds entries passed as input even if not part of the original `.po` file' do
+    pending 'this currently fails and will be addressed as part of the upcoming refactor/rewrite of the functionality'
```
AliSoftware: Looool I started commenting on the line just above (L104) with this:

> I'm actually surprised that it already supports this case, I was kinda sure that, just like for release notes, our current actions were only able to update keys already in the `.po` but not add new ones, TIL this already works

Then right after writing that, I saw that `pending` line immediately after explaining exactly just that 😛  👌

spec/gp_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+        }
+      )
+
+      # 3: Assert given .po has been updated as expected
+      #
+      # Notice:
+      #
+      # - The new line after each block is added by the conversion
+      # - That there's no new line between release_note_0122 and key1, because
+      #   the notes are copied as they are with no extra manipulation
```
AliSoftware: Took me a moment to understand that sentence, until I guessed you probably meant this: 😅 
```suggestion
      # - But there's no new line between release_note_0122 and key1, because
      #   the notes are copied as they are with no extra manipulation
```

spec/gp_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,123 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::GpUpdateMetadataSourceAction do
+  it 'updates any block in a given .po file with the values from the given sources' do
```
AliSoftware: I'm wondering if there's a way (and if so, if it's worth it) to make sure we have the same test examples for both `gp_update_metadata_source_spec` and `an_update_metdata_source_spec` — even tho in some cases some may be marked pending in one action but not in the other if needed.

Given one of the goal of having those unit tests is also to check what might be the difference in behavior (if any) between the two actions (hinting at the possibility to plainly delete one of the two if there's no diff, or at what we can borrow from one to the other if not), that would help a lot if we were able to write this in a way that makes both test the exact same things.

 - One first step to enable that would be to move the input and expected texts into fixture files (I saw https://github.com/wordpress-mobile/release-toolkit/pull/326/files#r784521409 and get why you did it that way, but now that tests are written, might be time to move to files).
 - An even better thing (especially given both actions seem to expect the exact same parameters/`ConfigItems` and thus have exact same call sites I think?) would be to DRY the whole suite of examples (either using `def` or `shared_example` or whatever) and make it be parametrized only by the `described_class` and the `pending_message` (nil if not a _pending_ test). That way we'd run the _exact same_ `it` tests on both `AnUpdateMetadataSourceAction` and `GpUpdateMetadataSourceAction`, even if we need to mark so of those tests `pending` in one action but not in another.

☝️  The big benefit of trying this approach would be to make it easy to check if the actions have any difference in behavior and implementation, to make it easier to remove duplicate code with confidence during the refactoring.

spec/an_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      File.write(output_path, dummy_text)
+
+      release_notes_path = File.join(dir, 'release_notes.txt')
+      File.write(release_notes_path, "- release notes\n- more release notes")
+      file_1_path = File.join(dir, '1.txt')
+      File.write(file_1_path, 'value 1')
+      file_2_path = File.join(dir, '2.txt')
+      File.write(file_2_path, 'value 2')
+
+      described_class.run(
```
AliSoftware: I know that calling `run` directly is what we've always been doing in our specs that test a fastlane action directly (as opposed to testing just a helper), so it's fine;

But I recently got caught up in a case where my test crashed when calling `run` because I failed to pass a parameter that was supposed to be optional. The issue is that when you call `run` directly, this does not go thru the path in fastlane's code which evaluates the `ConfigItem` and fill `params` with the default values of missing parameters and do the param validation checks etc.

So instead I've recently started to do it differently in my new specs testing fastlane actions, inspired by how Fastlane's own specs do it in the fastlane repo. See https://github.com/wordpress-mobile/release-toolkit/pull/329#discussion_r792028262 for a peek at the technique.

Given those specific tests might (?) disappear soon™ alongside the corresponding actions, and be replaced by a new action and tests replacing them, and also given the test is not really about testing optional parameters or ConfigItem parsing/validation, I think it's fine to keep those like that (i.e. calling `run` directly) for those.
But just wanted to let you know about the alternate way (and the one used by fastlane itself in their test suite) for information and in case you saw a benefit in adopting that new technique in the future 🤷‍♂️

AliSoftware (commented):
Looking good, but I think we would benefit in DRYing the tests between the 2 actions.

Not just for the only sake of DRYing, but more because that will ensure that we can spot the behavior differences between those 2 legacy actions, and easily compare where they might differ, helping us getting rid of duplicate behavior and code with confidence during the upcoming refactoring of those.

spec/gp_update_metadata_source_spec.rb
mokagio: I thought about using the `shared_example`, too, but stopped once I realized the [`an_` version doesn't have the `whats_new` parameter](https://github.com/wordpress-mobile/release-toolkit/blob/8427fb85656145efbe3b8ff7c63ad43f55c645ea/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_update_metadata_source_action.rb#L126-L153). So, already the expectation on the happy path behavior would be different.

But, there's a bunch of shared edge case that could be moved into shared examples alright 👍

spec/an_update_metadata_source_spec.rb
mokagio: I think I'll have a go at using the approach you suggest, even if just to get comfortable with it and set it up for future work.

spec/gp_update_metadata_source_spec.rb
mokagio: I'll see what I can do with the shared examples

mokagio: Turned out there was I lot I could do... All the examples are in a shared one now 😄 [`d095cd2` (#326)](https://github.com/wordpress-mobile/release-toolkit/pull/326/commits/d095cd21470141dc2516080333f1d3e87da4f8cc)

AliSoftware: 👊
✋
' '
' '
🎤 💥

AliSoftware (approved):
👌

mokagio:
Ah! I should have disabled auto-merge, or turn this into a draft till I was done with the Fastfile parsing change. Oh well, I'll submit it as a dedicated PR later...

PR #329, opened 2022-01-25
Introduce ios_merge_strings_files action

AliSoftware:
This work is part of the L10n modernization project (refs: paaHJt-2Ib-p2 & paaHJt-2Px-p2)

### What

This PR builds on top of https://github.com/wordpress-mobile/release-toolkit/pull/328 and introduces

 - A new `ios_l10n_helper`, with helper methods to manipulate `.strings` files; especially:
   1. Checking the format of a `.strings` file (OpenStep, XML or Binary-Plist).
   2. Read an arbitrary `.strings` file (whatever its format) as a Hash of key/value pairs (ignoring file comments).
   3. Merge multiple OpenStep-formatted `.strings` files together.
   4. Generate an XML-formatted `.strings` file from a list of keys.
 - A new `ios_merge_strings_files` action, which is able to merge multiple OpenStep strings files into one, regardless of the encoding those input files are in, and report any duplicates found along the way
 - The relevant Unit Tests and test fixtures for each

### Why

These new helpers and new action are needed for the following cases in our new plan around Localization Tooling:

 - The `ios_merge_strings_files` action will be used during code-freeze, to merge the `Localizable.strings` — typically generated by the other `ios_generate_strings_file_from_code` action — with the other manually-maintained `.strings` files like `InfoPlist.strings`, or `Sites.strings` associated with the `Sites.intentdefinition` file together in a single `.strings` file that GlotPress will inport
 - The other helper methods are:
   1. `strings_file_type`: Used to validate that we only try to merge `.strings` files of the expected format as input
   2. `read_strings_file_as_hash`: Used to parse the keys used by each input file, to be able to find duplicates. But it will also be useful in a later PR and future action to be able to extract a subset of keys from the translated `Localizable.strings` file exported from GlotPress, in order to generate the `InfoPlist.strings` and `Sites.strings` files with the extracted translations.
   3. `merge_strings`: The main helper method called by the `ios_merge_strings_files` action, which makes use of (1) and (2)
   4. `generate_strings_file_from_hash`: I got a bit ahead of myself here as this one is not _yet_ used by anything introduced  by this PR, but will be useful in the future when we will need to generate `InfoPlist.strings` and `Sites.strings` files from the GlotPress exported translations (see mention in (2)).

_PS: I should probably have kept (4) for a future PR instead of this one, but by the time I decided to submit more incremental PRs rather than a big one, the method and its tests were already written, and it didn't feel worth the effort to remove and stash it to re-introduce later, so figured I'd keep it in there by now, even if it will only start to get used in the next PR after this one._

### What's Next

This is only part of the things I'm currently working on. I'll soon post separate PRs on top of this one to add:
 - A new action to port { paaHJt-2KK-p2 :: 2️⃣ + 3️⃣  } into Ruby (aka downloading + post-processing the translations from GlotPress)
 - Another action to port { paaHJt-2KK-p2 :: 4️⃣ } into Ruby (aka extract translations into `InfoPlist.strings` and `Sites.strings`, according to the plan in paaHJt-2Px-p2) — which will finally make use of (4) `generate_strings_file_from_hash`


spec/ios_merge_strings_files_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+          inputs.each { |f| FileUtils.cp(fixture(f), tmpdir) }
+
+          # Act
+          result = Dir.chdir(tmpdir) do
+            described_class.new.parse("lane :test do
+              ios_merge_strings_files(
+                paths: ['#{inputs[0]}', '#{inputs[1]}'],
+                destination: 'output.strings'
+              )
+            end").runner.execute(:test)
```
AliSoftware, line 28: This technique for testing Fastlane Actions was borrowed [from how fastlane itself does it](https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/ipa_spec.rb).

It seems like a better and more recommended approach compared to directly calling `Fastlane::Actions::IosMergeStringsFilesAction.run(…)` (like we've been doing on other action specs in our code in the past), because calling `Fastlane::FastFile.new.parse` like here ensures that the spec will go thru the whole `ConfigItem` parsing and interpretation (and use the default value from the action's `ConfigItem` if any is provided, etc)

spec/ios_l10n_helper_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      paths = [fixture('Localizable-utf16.strings'), fixture('InfoPlist-utf8.strings')]
+
+      # Just making sure we don't accidentally change the encoding of the fixture files in the future
+      expect(file_encoding(paths[0])).to eq(Encoding::UTF_16BE)
+      expect(file_encoding(paths[1])).to eq(Encoding::UTF_8)
+
+      Dir.mktmpdir('a8c-release-toolkit-l10n-helper-tests-') do |tmp_dir|
+        output_file = File.join(tmp_dir, 'output.strings')
+        described_class.merge_strings(paths: paths, output_path: output_file)
+        expect(File.read(output_file)).to eq(File.read(fixture('expected-merged.strings')))
```
mokagio, line 54: Worth adding a check also for the output file encoding?

```suggestion
        expect(File.read(output_file)).to eq(File.read(fixture('expected-merged.strings')))
        expect(file_encoding(output_file)).to eq(Encoding::UTF_8)
```

spec/ios_l10n_helper_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      end
+    end
+
+    it 'raises if one of the strings file is not in textual format' do
+      paths = [fixture('Localizable-utf16.strings'), fixture('xml-format.strings')]
+      Dir.mktmpdir('a8c-release-toolkit-l10n-helper-tests-') do |tmp_dir|
+        output_file = File.join(tmp_dir, 'output.strings')
+        expect do
+          described_class.merge_strings(paths: paths, output_path: output_file)
+        end.to raise_exception(RuntimeError, "The file `#{paths[1]}` is in xml format but we currently only support merging `.strings` files in text format.")
```
mokagio, line 85: Nitpick:
```suggestion
        end.to raise_exception(RuntimeError, "The file `#{paths[1]}` is in XML format but we currently only support merging `.strings` files in text format.")
```

spec/ios_l10n_helper_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      end
+    end
+
+    it 'raises if one of the strings file does not exist' do
```
mokagio: Nitpick:
```suggestion
    it 'raises if any of the strings file does not exist' do
```
Reading "one" made me wonder "what happens if more than one doesn't exist?" But... it could be just my contrived mind

spec/ios_l10n_helper_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      }
+      Dir.mktmpdir('a8c-release-toolkit-l10n-helper-tests-') do |tmp_dir|
+        output_file = File.join(tmp_dir, 'output.strings')
+        expected_file = fixture('expected-generated.strings')
+        described_class.generate_strings_file_from_hash(translations: hash, output_path: output_file)
+        expect(File.read(output_file)).to eq(File.read(expected_file))
+      end
+    end
+
+    # Reads non-latin content from UTF-16BE file and generating output file as UTF8
```
mokagio: Nitpick: verbs consistency

```suggestion
    # Reads non-latin content from UTF-16BE file and generates output file as UTF8
```

or 

```suggestion
    # Reading non-latin content from UTF-16BE file and generating output file as UTF8
```

spec/ios_l10n_helper_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+    it 'handles non-latin, Unicode content properly' do
```
mokagio: Other nitpick
```suggestion
    it 'handles non-Latin, Unicode content properly' do
```

spec/ios_l10n_helper_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      text_fixture = fixture('Localizable-utf16.strings')
+      expect(described_class.strings_file_type(path: text_fixture)).to eq(:text)
+    end
+
+    it 'returns nil on a non-existing path' do
+      expect(described_class.strings_file_type(path: '/invalid-path')).to be_nil
+    end
+
+    it 'returns nil on file that is not a strings file at all' do
+      png_fixture = fixture('not-a-strings-file.png')
```
mokagio: Is there a particular reason you used a PNG as opposed to a text file that is not a `.strings`?

Clearly, an image is quite different from a `.strings` file and therefore represents a good test input. I just "worry" about tracking images in the repo when we don't have to. Git is good at managing text files, not images/binaries.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
```diff
@@ -0,0 +1,75 @@
...
+
+          Especially useful to prepare a single `.strings` file merging strings from both `Localizable.strings` from
+          the app code — typically previously extracted from `ios_generate_strings_file_from_code` —
+          and string files like `InfoPlist.strings` — which values may not be generated from the codebase but
+          manually maintained by developers.
+
+          The action only supports merging files which are in the OpenStep (`"key" = "value";`) text format (which is
+          the most common format for `.strings` files, and the one generated by `genstrings`), but can handle the case
+          of different files using different encodings (UTF8 vs UTF16) and is able to detect and report duplicates.
+          It does not handle strings files in XML or binary-plist formats `.strings` files (which are valid but more rare)
```
mokagio: I tried looking for a webpage describing the OpenStep format but came up short. If you have one, it would be good to add it as a "see also" reference at the end of these details.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
```diff
@@ -0,0 +1,75 @@
...
+            type: Array,
+            optional: false
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :destination,
+            env_name: 'FL_IOS_MERGE_STRINGS_FILES_DESTINATION',
+            description: 'The path of the merged `.strings` file to generate. If nil, the merge will happen in-place in the first file in the `paths:` list',
+            type: String,
+            optional: true,
+            default_value: nil
```
mokagio, line 53: How do you feel about making this parameter required, so that we don't have to specify, implement, and manage the fallback logic to the in-place merge?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
```diff
@@ -0,0 +1,75 @@
...
+      def self.return_value
+        'The list of duplicate keys found while merging the various `.strings` files'
+      end
+
+      def self.authors
+        ['automattic']
+      end
+
+      def self.is_supported?(platform)
+        platform == :ios
```
mokagio: This should work on `:mac`, too, right?

```suggestion
        [:ios, :mac].include? platform
```

Or any equivalent syntax you prefer.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+            paths.each do |input_file|
+              fmt = strings_file_type(path: input_file)
+              raise "The file `#{input_file}` does not exist or is of unknown format." if fmt.nil?
+              raise "The file `#{input_file}` is in #{fmt} format but we currently only support merging `.strings` files in text format." unless fmt == :text
+
+              string_keys = read_strings_file_as_hash(path: input_file).keys
+              duplicates += (string_keys & all_keys_found) # Find duplicates using Array intersection, and add those to duplicates list
+              all_keys_found += string_keys
+
+              tmp_file.write("// MARK: - #{File.basename(input_file)}\n\n")
```
mokagio: Nitpick: The header comment pasted above uses `/* */`. Is it possible to be consistent with the comment format? I assume Xcode would be able to parse these `MARK: -` even if in a `/* */`-kind comment.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+              # Read line-by-line to reduce memory footprint during content copy; Be sure to guess file encoding using the Byte-Order-Mark.
+              File.readlines(input_file, mode: 'rb:BOM|UTF-8').each { |line| tmp_file.write(line) }
```
mokagio, line 67: Interesting. Out of curiosity, what kind of input files would result in memory footprint become a concern?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+              tmp_file.write("\n")
+            end
+            tmp_file.close # ensure we flush the content to disk
```
mokagio, line 70: Your use of the word ensure makes me wonder if it's worth wrapping this in a `begin - rescue - ensure - end` construct.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+        #
+        # @param [String] path The path to the `.strings` file to read
+        # @return [Hash<String,String>] A dictionary of key=>translation translations.
+        # @raise [RuntimeError] If the file is not a valid strings file or there was an error in parsing its content.
+        #
+        def self.read_strings_file_as_hash(path:)
+          output, status = Open3.capture2e('/usr/bin/plutil', '-convert', 'json', '-o', '-', path)
+          raise output unless status.success?
+
+          JSON.parse(output)
```
mokagio, line 86: Neat using this.

mokagio (commented):
Looking good. Love the tests coverage.

---

> PS: I should probably have kept (4) for a future PR instead of this one, but by the time I decided to submit more incremental PRs rather than a big one, the method and its tests were already written, and it didn't feel worth the effort to remove and stash it to re-introduce later, so figured I'd keep it in there by now, even if it will only start to get used in the next PR after this one.

The PR turned out to be of manageable size (~45 mins to review) anyway, so no dramas. Glad you didn't spend extra time on that.

mokagio (approved):
My previous review was meant to be an approval, but I got lured by the green background of the "Comment" button 🙃

spec/ios_l10n_helper_spec.rb
AliSoftware: @mokagio  This "XML" part of the message comes from the `:xml` symbol (returned from `strings_file_type` and turned into string when interpolated in the message) though; so not sure it's worth creating a mapping of {symbols => display names} just for uppercasing the "XML" case…

spec/ios_l10n_helper_spec.rb
AliSoftware: True. I couldn't really use a text file because `strings_file_type` would return `:text` in that case; this is because that method only checks the file general type but does not check it's actually a valid plist or strings file.

Another option could be to make `strings_file_type` _also_ check if the input is a valid strings/plist by running `plutil -lint` on it for example, but not sure if worth it; or maybe `file` does hint (in addition to format and encoding) if this _looks_ like an actual strings file… I'll check that and give it a try.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
AliSoftware: I actually implemented the in-place logic because I think that's how we'll likely use it in the `Fastfile`, merging `InfoPlist.strings` and `Sites.strings` directly with and into the `Localizable.strings` in-place, immediately after we generated it using `generate_strings_file_from_code`?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
AliSoftware: True. I think we do that mistake in a lot of other actions as well currently 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: Never tried using `MARK` within `/* */`, will double-check 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: Any large file, especially the Localizable.strings files that we have are quite large. Probably not to the MB order, so in practice it might not really matter for this use case… but I still as a general practice always prefer to copy small chunks at a time (and recycling memory between each line copied) rather than read the whole file in memory at once just to copy its content.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: It's not. It's not about a risk of an exception being raised at all (which what `rescue/ensure` is about), it's only about making sure the content has been flushed to disk, otherwise the file on disk risks being empty on disk as not yet flushed at the time I do the copy. It's a bit like having to commit/push a change in git or commit a transaction in SQL if you will. (Alternately to calling `.close` I think I could also have just used `tmp_file.flush` btw? But closing the file completely made more sense in that context)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
AliSoftware: The only links I could find (which are not really full docs about the format grammar but rather paragraphs in Apple's doc about some points about it) are:
 - [This page about strings file in the doc archive (so not even the current doc!)](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html#//apple_ref/doc/uid/10000051i-CH6-SW10)
 - On the same page, [this other paragraph near the end](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html#//apple_ref/doc/uid/10000051i-CH6-SW13)
 - [This paragraph in the wikipedia article about Property Lists](https://en.wikipedia.org/wiki/Property_list#NeXTSTEP), though this is more about the general NextStep PropertyList format (which supports more data types than just a `Dictionary<String,String>` root like `.strings` do) nor does it mention that `.strings` file use that format and can omit the surrounding `{ … }` of the root dictionary (and directly dump the key/value pairs that would otherwise typically be within those dictionary-scope-defining braces)… so not really complete either.

None of those are specifications nor even near being complete documentation and set of rules though. So not sure it's worth mentioning any of them here?

That's the whole issue and reason why I'm really reluctant in trying to generate strings files in this OpenStep/NextStep/ASCII format (well, except here because in this particular case it's basically just concatenating existing files already in that format, not creating entries ourselves and having to deal with all the escaping rules and all; and GlotPress only supports that format for `.strings` anyway, not binary nor XML), because I can't find an official specification for that format. It being deprecated and abandoned as an output format is probably the reason, but given it's still unofficially supported for reading and still used by Xcode for `.strings` and `.pbxproj` amongst other things, I'd still expect to find some legacy doc somewhere but still came up short.

AliSoftware:
```suggestion
          It does not handle `.strings` files in XML or binary-plist formats (which are valid but more rare)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: Seems it works just as well, so fixed this in 0354867 🙂

spec/ios_l10n_helper_spec.rb
AliSoftware: Fixed in 06d5a335b060f092cd8d2abdd636d1773733b497, by swapping the fixture to use a plain text file instead, and adding an additional check in `strings_file_type` via `plutil -lint` to ensure the file looks like a valid property-list before trying to determine its specifc format.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: For completeness: WPiOS's `en.lproj/Localizable.strings` is 848KB, so actually not that far away from the 1MB mark after all (I can easily imagine us reaching it in a near future as we add more and more features), so…

AliSoftware:
@mokagio want to give it a quick last check before we merge this, now that I've pushed more commits to address most of your feedback?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
AliSoftware: And to complete that answer, just wanted to note that even if we made this parameter mandatory, we would still want to be sure for the specs to cover the case where the output file would be one of files already listed in the array of input files.

I'm saying this because:
 - This will still likely be our typical use case at call site in our `Fastfiles` as mentioned above
 - But I've so often been bitten in the past by the case of running `cat foo.txt bar.txt >foo.txt` in bash and ending up with a totally empty `foo.txt` file in the end when `foo.txt` is both used as input and (redirect-)output of `cat`.

Given this action is a bit like a smart `cat` in a way (which adds support for files with different encodings and detecting duplicates), I wanted to make sure we didn't risk falling into the same trap.

That's also one of the main reasons(†) why the actual implementation [uses an intermediate file](https://github.com/wordpress-mobile/release-toolkit/pull/329/files#diff-6867f769ac6316c7b553d9e5ac93e7aa55d4f9b82bb867e2dc552c8500dd1dacR52) and [copies it back to the destination only once done](https://github.com/wordpress-mobile/release-toolkit/pull/329/files#diff-6867f769ac6316c7b553d9e5ac93e7aa55d4f9b82bb867e2dc552c8500dd1dacR71). Which is why I wanted to be sure to cover that case and test that this specific edge case still worked as expected and was accounted for and made clear in the API that this was supported 🙃 

_(†) another reason is for atomic writes, e.g. to not mess up the output file in case something crashes before we finish generating the whole content, similar to how [`write(toFile:atomically:)` does it in Foundation](https://developer.apple.com/documentation/foundation/nsdata/1408033-write)_ 😛

spec/test-data/translations/ios_l10n_helper/invalid-file.strings
```diff
@@ -0,0 +1,3 @@
+Contrary to what the extension of this file suggests, this is not a `.strings` file.
+Instead, this is just a regular, plain text file, which should be detected as invalid
+if provided as an input to `Ios::L10nHelper` methods.
```
mokagio, line 3: This turned out even better than my suggestion. Love the check against a fake file names `.strings` but that doesn't match the format 👏

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+* Introduce new `ios_merge_strings_files` action. [#329]
```
mokagio, line 13: 🎉

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
mokagio: Sounds good 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
mokagio: Thanks for the extra context.

mokagio (approved):
:shipit:

PR #330, opened 2022-01-28
Switch to running `an_` and `gp_` metadata actions via Fastfile parsing

mokagio:
Built on top of @AliSoftware's experiment to test actions by running them via a just-in-time `Fastfile` as opposed to via a `run` call to add an extra level of checks for the `ConfigItem`s parsing from https://github.com/wordpress-mobile/release-toolkit/pull/326#pullrequestreview-866044937.

This PR aims to extracts the boilerplate into a shared helper method to make calling the action in the tests straightforward.

### DIFF
--- a/spec/an_localize_libs_action_spec.rb
+++ b/spec/an_localize_libs_action_spec.rb
@@ -1,0 +1,54 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::AnLocalizeLibsAction do
+  # This test is more of a way of ensuring `run_described_action` handles array
+  # of hashes properly than a comprehensive test for the
+  # `an_localize_libs_action` action.
+  #
+  # Please consider expanding this test if you'll find yourself working on its
+  # action.
+  it 'merges the strings from the given array into the given main strings file' do
+    in_tmp_dir do |tmp_dir|
+      app_strings_path = File.join(tmp_dir, 'app.xml')
+      File.write(app_strings_path, android_xml_with_content('<string name="a_string">test from app</string>'))
+
+      lib1_strings_path = File.join(tmp_dir, 'lib1.xml')
+      File.write(lib1_strings_path, android_xml_with_content('<string name="a_lib1_string">test from lib 1</string>'))
+
+      lib2_strings_path = File.join(tmp_dir, 'lib2.xml')
+      File.write(lib2_strings_path, android_xml_with_content('<string name="a_lib2_string">test from lib 2</string>'))
+
+      run_described_action(
+        app_strings_path: app_strings_path,
+        libs_strings_path: [
+          { library: 'lib_1', strings_path: lib1_strings_path, exclusions: [] },
+          { library: 'lib_2', strings_path: lib2_strings_path, exclusions: [] },
+        ]
+      )
+
+      # Notice the extra indentation in the library strings. The action doesn't
+      # modify the app's strings content indentation, but it applies its own
+      # standard to the values read from the given library strings
+      expected = <<~XML
+        <string name="a_string">test from app</string>
+          <string name="a_lib1_string">test from lib 1</string>
+          <string name="a_lib2_string">test from lib 2</string>
+      XML
+      expect(File.read(app_strings_path)).to eq(android_xml_with_content(expected))
+    end
+  end
+end
+
+def android_xml_with_content(content)
+  # I couldn't find a way to interpolate a multiline string preserving its
+  # indentation in the heredoc below, so I hacked the following transformation
+  # of the input that adds the desired indentation to all lines.
+  indented_content = content.chomp.lines.map { |l| "  #{l}" }.join
+
+  return <<~XML
+    <?xml version="1.0" encoding="UTF-8"?>
+    <resources xmlns:tools="http://schemas.android.com/tools">
+    #{indented_content}
+    </resources>
+  XML
+end
--- a/spec/an_localize_libs_action_spec.rb
+++ b/spec/an_localize_libs_action_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe Fastlane::Actions::AnLocalizeLibsAction do
-  # This test is more of a way of ensuring `run_described_action` handles array
+  # This test is more of a way of ensuring `run_described_fastlane_action` handles array
   # of hashes properly than a comprehensive test for the
   # `an_localize_libs_action` action.
   #
@@ -18,7 +18,7 @@ describe Fastlane::Actions::AnLocalizeLibsAction do
       lib2_strings_path = File.join(tmp_dir, 'lib2.xml')
       File.write(lib2_strings_path, android_xml_with_content('<string name="a_lib2_string">test from lib 2</string>'))
 
-      run_described_action(
+      run_described_fastlane_action(
         app_strings_path: app_strings_path,
         libs_strings_path: [
           { library: 'lib_1', strings_path: lib1_strings_path, exclusions: [] },
--- a/spec/android_merge_translators_strings_spec.rb
+++ b/spec/android_merge_translators_strings_spec.rb
@@ -35,7 +35,7 @@ end
 def amts_run_test(script)
   test_script = @amtsTestUtils.get_test_from_file(script)
   @amtsTestUtils.create_test_data(test_script)
-  Fastlane::Actions::AndroidMergeTranslatorsStringsAction.run(strings_folder: @amtsTestUtils.test_folder_path)
+  run_described_fastlane_action(strings_folder: @amtsTestUtils.test_folder_path)
   expect(@amtsTestUtils.read_result_data(test_script)).to eq(test_script['result']['content'])
 end
 
--- a/spec/ios_merge_strings_files_spec.rb
+++ b/spec/ios_merge_strings_files_spec.rb
@@ -19,7 +19,7 @@ describe Fastlane::Actions::IosMergeStringsFilesAction do
 
         # Act
         result = Dir.chdir(tmpdir) do
-          run_described_action(
+          run_described_fastlane_action(
             paths: [inputs[0], inputs[1]],
             destination: 'output.strings'
           )
--- a/spec/ios_merge_strings_files_spec.rb
+++ b/spec/ios_merge_strings_files_spec.rb
@@ -1,41 +1,39 @@
-describe Fastlane do
-  describe Fastlane::FastFile do
-    let(:test_data_dir) { File.join(File.dirname(__FILE__), 'test-data', 'translations', 'ios_l10n_helper') }
+describe Fastlane::Actions::IosMergeStringsFilesAction do
+  let(:test_data_dir) { File.join(File.dirname(__FILE__), 'test-data', 'translations', 'ios_l10n_helper') }
 
-    def fixture(name)
-      File.join(test_data_dir, name)
-    end
+  def fixture(name)
+    File.join(test_data_dir, name)
+  end
 
-    describe '#ios_merge_strings_files' do
-      it 'calls the action with the proper parameters and warn and return duplicate keys' do
-        # Arrange
-        messages = []
-        allow(FastlaneCore::UI).to receive(:important) do |message|
-          messages.append(message)
+  describe '#ios_merge_strings_files' do
+    it 'calls the action with the proper parameters and warn and return duplicate keys' do
+      # Arrange
+      messages = []
+      allow(FastlaneCore::UI).to receive(:important) do |message|
+        messages.append(message)
+      end
+      inputs = ['Localizable-utf16.strings', 'non-latin-utf16.strings']
+
+      Dir.mktmpdir('a8c-release-toolkit-tests-') do |tmpdir|
+        inputs.each { |f| FileUtils.cp(fixture(f), tmpdir) }
+
+        # Act
+        result = Dir.chdir(tmpdir) do
+          run_described_action(
+            paths: [inputs[0], inputs[1]],
+            destination: 'output.strings'
+          )
         end
-        inputs = ['Localizable-utf16.strings', 'non-latin-utf16.strings']
 
-        Dir.mktmpdir('a8c-release-toolkit-tests-') do |tmpdir|
-          inputs.each { |f| FileUtils.cp(fixture(f), tmpdir) }
-
-          # Act
-          result = Dir.chdir(tmpdir) do
-            described_class.new.parse("lane :test do
-              ios_merge_strings_files(
-                paths: ['#{inputs[0]}', '#{inputs[1]}'],
-                destination: 'output.strings'
-              )
-            end").runner.execute(:test)
-          end
-
-          # Assert
-          expect(File).to exist(File.join(tmpdir, 'output.strings'))
-          expect(result).to eq(%w[key1 key2])
-          expect(messages).to eq([
-                                   'Duplicate key found while merging the `.strings` files: `key1`',
-                                   'Duplicate key found while merging the `.strings` files: `key2`',
-                                 ])
-        end
+        # Assert
+        expect(File).to exist(File.join(tmpdir, 'output.strings'))
+        expect(result).to eq(%w[key1 key2])
+        expect(messages).to eq(
+          [
+            'Duplicate key found while merging the `.strings` files: `key1`',
+            'Duplicate key found while merging the `.strings` files: `key2`',
+          ]
+        )
       end
     end
   end
--- a/spec/shared_examples_for_update_metadata_source_action.rb
+++ b/spec/shared_examples_for_update_metadata_source_action.rb
@@ -19,7 +19,7 @@ RSpec.shared_examples 'update_metadata_source_action' do |options|
       file_2_path = File.join(dir, '2.txt')
       File.write(file_2_path, 'value 2')
 
-      run_described_action(
+      run_described_fastlane_action(
         po_file_path: output_path,
         release_version: '1.0',
         source_files: {
@@ -57,7 +57,7 @@ RSpec.shared_examples 'update_metadata_source_action' do |options|
       whats_new_path = File.join(dir, 'whats_new.txt')
       File.write(whats_new_path, "- something new\n- something else new")
 
-      run_described_action(
+      run_described_fastlane_action(
         po_file_path: output_path,
         release_version: '1.23',
         source_files: {
@@ -95,7 +95,7 @@ RSpec.shared_examples 'update_metadata_source_action' do |options|
       file_2_path = File.join(dir, '2.txt')
       File.write(file_2_path, 'value 2')
 
-      run_described_action(
+      run_described_fastlane_action(
         po_file_path: output_path,
         release_version: '1.0',
         source_files: {
@@ -137,7 +137,7 @@ RSpec.shared_examples 'update_metadata_source_action' do |options|
       release_notes_path = File.join(dir, 'release_notes.txt')
       File.write(release_notes_path, "- release notes\n- more release notes")
 
-      run_described_action(
+      run_described_fastlane_action(
         po_file_path: output_path,
         release_version: '1.23',
         source_files: {
--- a/spec/shared_examples_for_update_metadata_source_action.rb
+++ b/spec/shared_examples_for_update_metadata_source_action.rb
@@ -19,8 +19,9 @@ RSpec.shared_examples 'update_metadata_source_action' do |options|
       file_2_path = File.join(dir, '2.txt')
       File.write(file_2_path, 'value 2')
 
-      described_class.run(
+      run_described_action(
         po_file_path: output_path,
+        release_version: '1.0',
         source_files: {
           key1: file_1_path,
           key2: file_2_path
@@ -56,7 +57,7 @@ RSpec.shared_examples 'update_metadata_source_action' do |options|
       whats_new_path = File.join(dir, 'whats_new.txt')
       File.write(whats_new_path, "- something new\n- something else new")
 
-      described_class.run(
+      run_described_action(
         po_file_path: output_path,
         release_version: '1.23',
         source_files: {
@@ -94,8 +95,9 @@ RSpec.shared_examples 'update_metadata_source_action' do |options|
       file_2_path = File.join(dir, '2.txt')
       File.write(file_2_path, 'value 2')
 
-      described_class.run(
+      run_described_action(
         po_file_path: output_path,
+        release_version: '1.0',
         source_files: {
           key1: file_1_path,
           key2: file_2_path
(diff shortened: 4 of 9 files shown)


spec/ios_merge_strings_files_spec.rb
```diff
@@ -1,41 +1,39 @@
-describe Fastlane do
-  describe Fastlane::FastFile do
-    let(:test_data_dir) { File.join(File.dirname(__FILE__), 'test-data', 'translations', 'ios_l10n_helper') }
+describe Fastlane::Actions::IosMergeStringsFilesAction do
```
mokagio, line 1: With the new `run_described_action` helper, we get to use the actual type we're testing here which I think makes the test a tiny bit clearer.

spec/shared_examples_for_update_metadata_source_action.rb
```diff
@@ -94,8 +95,9 @@
       file_2_path = File.join(dir, '2.txt')
       File.write(file_2_path, 'value 2')
 
-      described_class.run(
+      run_described_action(
         po_file_path: output_path,
+        release_version: '1.0',
```
mokagio: @AliSoftware case in point for using this approach over `run`. It highlighted the fact that `release_version` is a required `ConfigItem`, even though not passing it to the `run` call doesn't result in an error 👍

spec/spec_helper.rb
```diff
@@ -67,6 +67,22 @@ def expect_shell_command(*command, exitstatus: 0, output: '')
   expect(Open3).to receive(:popen2e).with(*command).and_yield(mock_input, mock_output, mock_thread)
 end
 
+# If the `described_class` of a spec is a `Fastlane::Action` subclass, it runs
+# it with the given parameters.
+#
+def run_described_action(parameters, lane_name = 'test')
+  expect(described_class).to be < Fastlane::Action, "Only call `#{__callee__}` from a spec describing a `Fastlane::Action` subclass."
```
mokagio: I went for a test failure over a `raise` so that we'd get this error as part of the test run result. What do you think?

spec/spec_helper.rb
```diff
@@ -75,3 +91,19 @@ def in_tmp_dir
...
+  hash.map do |key, value|
+    # rubocop:disable Style/CaseLikeIf
+    if value.is_a?(Hash)
+      "#{key}: {\n#{stringify_for_fastlane(value)}\n}"
+    elsif value.is_a?(String)
+      "#{key}: \"#{value}\""
+    else
+      "#{key}: #{value}"
+    end
+    # rubocop:enable Style/CaseLikeIf
```
mokagio: I tried to use `case-when`, but I couldn't find a neat way to put `value` in the `case` call, because I need to call `is_a?` on it. The only thing I could do was:

```ruby
case
when value.is_a?(Hash)
  ...
when ...
  ...
else
  ...
end
```

Which raised a different lint error saying I should have used `if-elsif` instead 🙃 ♻️ 

Keen to know if there's a different way to write this.

spec/spec_helper.rb
AliSoftware: I'm fine either way, though I'd have used `raise` because if called from the wrong place in the test code itself, that's a programmer error from whoever write the spec/test, as opposed to a correct test which goes red and fail an expectation because the SUT / implementation it's testing changed behavior.

To be a failed expectation suggests that the SUT was expected to behave in a certain way, and might have been for a while (making the test green for a while) but suddenly we change something in the implementation code and the test goes red because our expectations of how the SUT is supposed to behave were invalidated. So it's a behavior error and a notification that the behavior changed (after some refactoring of the implementation maybe).

By contrast, calling this method from a test which does not have an `Action` subclass as its SUT is a programmer error in the way you wrote the test, not an expectation of the SUT itself risking changing, so to me a `raise` would make more sense here. But to be honest, that's also not that big a deal 🤷

spec/shared_examples_for_update_metadata_source_action.rb
AliSoftware: Ah! Interesting!

As a side note (and I think we have this tracked somewhere, just not sure to remember where) one of the improvements we might want to implement on the replacement for those actions is to auto-detect the release_version an stop having to pass it to the actions. Given the action should already have everything it needs to guess/determine the current release version (either by reading it from the xcconfig file or by the release/* branch name), it would be nice to remove the need for it during your future refactoring 🙂

spec/spec_helper.rb
AliSoftware:
```ruby
    case value
    when Hash
      puts 'Hash!'
    when String
      puts 'String!'
    else
      puts "What's that?"
    end
```

spec/spec_helper.rb
```diff
@@ -67,6 +67,22 @@ def expect_shell_command(*command, exitstatus: 0, output: '')
   expect(Open3).to receive(:popen2e).with(*command).and_yield(mock_input, mock_output, mock_thread)
 end
 
+# If the `described_class` of a spec is a `Fastlane::Action` subclass, it runs
+# it with the given parameters.
+#
+def run_described_action(parameters, lane_name = 'test')
```
AliSoftware:
```suggestion
def run_described_action(parameters, lane_name: 'test')
```
I'd suggest using a named parameters here especially when the previous arguments `parameters` is of variable length (in that case typically an implicit Hash, making it read like if it was a variadic number of named parameters)

Though even with a named parameter I'm curious how the call site would look like if you wanted to provided a different lane name. I think the only option at call site would be to make the Hash of parameters explicitly wrapped in `{ … }` in that case, to avoid having `… lane_nane: 'foo'` be considered part of the `parameters` Hash otherwise.

All that being said, I wonder what's the point of allowing to customize the lane name here, given it's only used as a dummy value to run it once parsed… so I think you can even totally remove it after all.

```suggestion
def run_described_action(parameters)
```

AliSoftware: Also for clarity I'd probably name this helper method `run_described_fastlane_action` (especially given the term "action" can be interpreted and read differently depending on the context the reader is in when seeing this), wdyt?

spec/spec_helper.rb
```diff
@@ -75,3 +91,19 @@ def in_tmp_dir
     end
   end
 end
+
+private
+
+def stringify_for_fastlane(hash)
+  hash.map do |key, value|
+    # rubocop:disable Style/CaseLikeIf
+    if value.is_a?(Hash)
```
AliSoftware: This fails to cover the case of `Array` as well. But it happens that some of our actions expect arrays of strings or arrays of hashes as parameters.

For example `an_localize_libs_action` has its `libs_strings_path` parameter which is expected to be an Array of Hashes…

spec/spec_helper.rb
```diff
@@ -75,3 +91,19 @@ def in_tmp_dir
...
+      "#{key}: {\n#{stringify_for_fastlane(value)}\n}"
+    elsif value.is_a?(String)
+      "#{key}: \"#{value}\""
```
AliSoftware: This will cause trouble if:
(1) `value` contains some interpolation or backslash-escapes, as they would then be interpreted
(2) `value` contains literal quotes in it (e.g. if `value = 'translators: "%1$@" is the app name'` — which could totally be a value you want to pass to your future replacement of the action generating the `.po` file from scratch if you wanted to pass the comments to write when generating each block…)

I'm wondering if just using `.inspect` instead of implementing your own `stringify_for_fastlane` wouldn't be enough?
Sure it would not format it as nicely (e.g. it would generate `{:paths=>[\"foo\", \"bar\\\"x\\\".txt\"], :destination=>\"out.txt\"}` instead of a nice `{ paths: ["foo", "bar\"x\".txt"], destionation: "out.txt" }` with nice newlines and all… but I don't think we care, as even if as humans we tend to use the latter syntax (implicit Hash and `key: value` instead of `:key => value` syntax, to make it look more like if it were named parameters) both syntaxes are valid and I think the former should still be recognized by `Fastfile.new.parse(…)` anyway, which is the important part…

AliSoftware (dismissed):
I LOVE the idea of that helper ❤️ 

Unfortunately, the `stringify_for_fastlane` method has a potential for having a lot of issues and edge cases, so we need to work deeper on a solution for doing what this method is doing… And this might end up being quite tricky to be sure to cover all cases…

Worth testing if Ruby's built-in `.inspect` wouldn't be enough after all to format that like we need, and cover all cases?

In any case, don't forget to test that with action parameters that are a bit more convoluted in their structure (but that we unfortunately still have cases for right now), like parameters containing arrays of strings or even arrays of hashes (cf. `an_localize_libs_action.rb` amongst others)

PS: I also didn't check if Fastlane themselves didn't have a helper method on their own that would do something similar 🤔

spec/spec_helper.rb
AliSoftware: Oh, and also even if we don't _yet_ have any case like that currently, iirc we talked about the idea of starting to sometimes use small DTOs as parameters for some actions that currently require a convoluted and less readable Hash structure for some of their parameters (`an_localize_libs_action` comes to mind, but I believe the one we were talking about back then was for cases like the Hash of the locales map (see [this draft](https://github.com/wordpress-mobile/release-toolkit/pull/296/files#diff-4d1e645f1d455e17026d35880050ced2a9f0201d00ef958024c9a1108bf25a85R30)) for which we might consider in the future that some of our actions could accept `[Locale]` instances instead of `[Hash]` for readability in our Fastfiles… though for that one, I don't think there would be a magic solution to stringify those cases for us 😒  but just thought I'd mention that case to provide some water to the mill of brainstorming solutions for that. Tho in the end we might consider just not supporting those special cases… 🤷

mokagio:
Thanks @AliSoftware. I went for that basic version of the method mostly because it's the first option that worked for me. I figured if we had issues down the line, we'd deal with them then.

> In any case, don't forget to test that with action parameters that are a bit more convoluted in their structure (but that we unfortunately still have cases for right now), like parameters containing arrays of strings or even arrays of hashes (cf. `an_localize_libs_action.rb` amongst others)

I didn't think of array of hashes, but the arrays of string case was already covered by adopting the new method in `ios_merge_strings_files_spec.rb` :smile:

> Worth testing if Ruby's built-in `.inspect` wouldn't be enough after all to format that like we need, and cover all cases?

I just did that, thank you for the suggestion. All the existing tests still pass, which is a great place to start 😄

I'll keep working on this as my focus for the day, given it's soft-blocking you, as mentioned in https://github.com/wordpress-mobile/release-toolkit/pull/331

spec/android_merge_translators_strings_spec.rb
```diff
@@ -35,7 +35,7 @@
 def amts_run_test(script)
   test_script = @amtsTestUtils.get_test_from_file(script)
   @amtsTestUtils.create_test_data(test_script)
-  Fastlane::Actions::AndroidMergeTranslatorsStringsAction.run(strings_folder: @amtsTestUtils.test_folder_path)
+  run_described_fastlane_action(strings_folder: @amtsTestUtils.test_folder_path)
```
mokagio, line 38: Going through `Fastlane` to run the action can add considerable noise in the test output because the action's log now appear in it:

image

I'm sure there's a way to disable this, but I haven't found it yet 😞

spec/spec_helper.rb
mokagio: See [`635f9ca` (#330)](https://github.com/wordpress-mobile/release-toolkit/pull/330/commits/635f9cad73653fb4db59a7962a25981616473283)

mokagio:
@AliSoftware , I should have addressed your suggestions with my latest round of changes.

I haven't looked into whether Fastlane has code for this and I also haven't done any extra testing for the `#inspect` approach. I think that, if we run in some weird edge case, we can either look into dealing with it then or fallback to calling `#run` on the `< Action` itself. 

What I did, though, was making sure the implementation worked for all existing usages. That's obviously a subset of the possible cases, but I'm okay with it because it's the one we care about right now, and we're not trying to share that helper method as a catch all solution to the wider community :simple_smile:

Let me know what you think when you have the time 👍 🙇‍♂️

spec/android_merge_translators_strings_spec.rb
AliSoftware: This looks like the output of `UI.deprecated` calls, so maybe just add `allow(FastlaneCore::UI).to receive(:deprecated)` around the helper would be enough to catch and mute them?

spec/an_localize_libs_action_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      XML
+      expect(File.read(app_strings_path)).to eq(android_xml_with_content(expected))
+    end
+  end
+end
+
+def android_xml_with_content(content)
+  # I couldn't find a way to interpolate a multiline string preserving its
+  # indentation in the heredoc below, so I hacked the following transformation
+  # of the input that adds the desired indentation to all lines.
```
AliSoftware, line 45: 👍 

FYI iinm you can add post-processing methods to the heredoc prelude, e.g. `expected = <<~XML.gsub(/^/,'  ')` and similar if you need such thing. Haven't used that in ages but I believe that's the syntax.

That being said, that wouldn't really help here because the only thing you want to indent is the `content`, but not the surrounding `<resources …>` tags in the rest of that heredoc… so I think the way you did it is fine and probably the cleanest we can get given the context 🙃

spec/an_localize_libs_action_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      )
+
+      # Notice the extra indentation in the library strings. The action doesn't
+      # modify the app's strings content indentation, but it applies its own
+      # standard to the values read from the given library strings
+      expected = <<~XML
+        <string name="a_string">test from app</string>
+          <string name="a_lib1_string">test from lib 1</string>
+          <string name="a_lib2_string">test from lib 2</string>
+      XML
```
AliSoftware, line 36: Wait, does that mean that in practice, the merged XML is expected to contain different indentations, i.e. a certain indentation for strings coming from the `app_strings_path` but a different indentation for strings coming from libraries?

Never noticed this discrepency in practice 🙀  (and that surprises me given iirc, we merge the XMLs using Nokogiri, which re-formats the XML when it saves it on disk, so I'd expect it to normalize the indentation? 🤷🤔 

Besides, I don't see such inconsistent indendation in https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/fastlane/resources/values/strings.xml so I'm not sure what to make of this test and why we'd have a different indentation in that test but not in real life in the client project?

AliSoftware:
> Going through Fastlane to run the action can add considerable noise in the test output because the action's log now appear in it:

@mokagio Took the liberty to push 8e96443 to fix this myself directly, by stubbing `is_deprecated?` to return `false` when testing actions.

AliSoftware:
> I think that, if we run in some weird edge case, we can either look into dealing with it then or fallback to calling #run on the < Action itself.

Totally agree, it's already a great starting point and covers all our use cases anyway, so I'm all good with that approach 👍

AliSoftware (commented):
Looks mostly good, thanks for providing this, this will nicely clean stuff up 👍  👏 

Note that I've taken the liberty to push a change to the `run_described_fastlane_action` to mute [all those deprecation logs that you were seeing when running the tests](https://github.com/wordpress-mobile/release-toolkit/pull/330#discussion_r796770356). (I also took the occasion to switch the guard statement from an `expect` to a `raise`, as suggested [here](https://github.com/wordpress-mobile/release-toolkit/pull/330#discussion_r794631417), and use a build-in Fastlane helper for the check); I hope you don't mind.

---

Was also surprised by the indentation glitch you commented about in the XML stuff… quite curious about this.

Anyway, I'm only leaving this review as a comment and not a formal approval to leave you a chance to [add the missing entry in the `CHANGELOG.md`](https://github.com/wordpress-mobile/release-toolkit/pull/330#issuecomment-1024340052) without auto-merge kicking in beforehand 😛 ; but once you've added the entry, this should be all good to merge 👍 :shipit:

spec/an_localize_libs_action_spec.rb
mokagio: I think that's simply due to the indentation in the test using 2 spaces, as it's the default style for this codebase, and the action indenting XML using 4 spaces.

I found that interesting and left it there. If it makes for a confusing test, maybe we could update the examples to all use 4 spaces and leave a note about it being the "default"?

I did a quick search and apparently there is no recommended indentation level for XML. The spec mentions [white space handling](https://www.w3.org/TR/2008/REC-xml-20081126/#sec-white-space):

> In editing XML documents, it is often convenient to use "white space" (spaces, tabs, and blank lines) to set apart the markup for greater readability. Such white space is typically not intended for inclusion in the delivered version of the document. On the other hand, "significant" white space that should be preserved in the delivered version is common, for example in poetry and source code.

I'm guessing because most indentation is only meant for readability, then it's up to each author (or formatter author) to choose what they find appropriate.

AliSoftware: Ah so this would in practice be an artefact of the tests and how the content used for those XML tests are inlined and assembled in the specs themselves (as opposed to if you were using fixture files) then? While in real use case in client apps there's no such indentation discrepancy between the tags from the app and the ones from the libs?

If that's so then yeah better rewrite the tests to make this less confusing and more importantly make it match reality then 😅

AliSoftware: Ah wait yeah actually I just realized why that is. This is because the first line of your `content` already gets 2 extra spaces from the heredoc itself! Which is why you have to remove 2 spaces from the first line of your content here to compensate.

So this is just misleading and there is in fact no indentation difference in the XML itself.

You could make it less confusing by adding the 2 spaces on the first line here (to match the rest of the lines), and removing the 2 extra leading spaces in your helper method that includes it in the heresoc block.

mokagio:
> Note that I've taken the liberty to push a change to the `run_described_fastlane_action` to mute all those deprecation logs that you were seeing when running the tests. 

Thank you for that. Interestingly, you ended up using `allow(Fastlane::Actions).to receive(:is_deprecated?).and_return(false)` which is pretty cool.

Did you by any chance try the other approaches like `allow(Fastlane::UI).to receive(:deprecated)`? I did try a bunch of those and couldn't get rid of the messages.

> (I also took the occasion to switch the guard statement from an expect to a raise, as suggested here, and use a build-in Fastlane helper for the check); I hope you don't mind.

That's okay 👍 Your rationale is solid, it was my expectation for what `raise` would have done was incorrect. I was expecting something like in iOS where the whole suite crashes, but instead RSpec reports it as just another failure:

image

Also, using `is_class_action?` is pretty neat.

> Anyway, I'm only leaving this review as a comment and not a formal approval to leave you a chance to add the missing entry in the CHANGELOG.md without auto-merge kicking in beforehand 😛

Thank you for thinking of it, too. I didn't think this was worth an entry, but you obviously do and it's easy to do.

spec/an_localize_libs_action_spec.rb
AliSoftware: Mmmh actually re-checked your heredoc and the 2 leasing spaces I thought were there aren't 🤨 so scratch what I just said. Still, pretty strange and worth looking into…

mokagio:
I've dismissed the change request as per the following [review](https://github.com/wordpress-mobile/release-toolkit/pull/330#pullrequestreview-869462488):

> Anyway, I'm only leaving this review as a comment and not a formal approval to leave you a chance to add the missing entry in the `CHANGELOG.md` without auto-merge kicking in beforehand 😛 ; but once you've added the entry, this should be all good to merge 👍 :shipit:

Having added the entry in [`8448e02`](https://github.com/wordpress-mobile/release-toolkit/pull/330/commits/8448e02b4b474858d0ffb78067631115619e7235), I'm going to admin merge as soon as CI is green.

spec/an_localize_libs_action_spec.rb
mokagio: I took a bit of digging, but I discovered the difference in indentation to be due to the code adding lines from the XML adding 4 spaces. See:
 https://github.com/wordpress-mobile/release-toolkit/blob/da405f2c5ca90d696857ba2ad01da2753daa60dc/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L67

I opened https://github.com/wordpress-mobile/release-toolkit/pull/332 to hopefully make the test clearer.

AliSoftware: Yeah I remember that our code hardcodes 4 spaces like that, what I found unclear is not why we'd have 4 spaces indentation (that I expected) but instead why the first line With have a different indentation than the rest of the lines.

Your other PR is clearer, thanks for opening it, and it actually does match exactly what I expected the test (and expected XML) would look like from the start so that change in #332 definitively makes more sense to me.

That being said I'm still puzzled as to how the code in this present PR even passed before in its current state (without #332) then. Because the change you made in #332 consists of removing 2 spaces from all the lines of what `content` looked like **except the first line**, while in the code where you later inject that `content` into the XML heredoc you added 2 spaces (switching from 2 to 4 leading spaces on each line) to **all** the lines. So the 2 extra spaces in `indented_content` balance the 2 spaces removed on `content` just like I expect… except for that first line of `content` which gets 2 extra space from the `map` but didn't get 2 spaces removed from `content` to balance.

If I'm not missing anything from ☝️ (but I'm sure I do otherwise that doesn't make sense) that would mean that the string you use for the expectation changed between this PR and #332 (gaining 2 leading spaces only on the first line, as the rest of the lines balance out) yet your specs were green both before and after that change on the first line of the expectation string?!

I think I need some rest, I must be missing something obvious…

AliSoftware:
> Did you by any chance try the other approaches like `allow(Fastlane::UI).to receive(:deprecated)`? I did try a bunch of those and couldn't get rid of the messages.

I tried at first but when I saw that it wasn't working, I [looked at fastlane's source code to understand when and how those messages were printed](https://github.com/fastlane/fastlane/blob/e874a47c6e2e0e61590a03d3b71e75e5a505d1ce/fastlane/lib/fastlane/runner.rb#L256-L261), which is what led me to this solution instead 😉 

> That's okay 👍 Your rationale is solid, it was my expectation for what `raise` would have done was incorrect. I was expecting something like in iOS where the whole suite crashes, but instead RSpec reports it as just another failure:

🙂 👌 

> Thank you for thinking of it, too. I didn't think this was worth an entry, but you obviously do and it's easy to do.

Eh actually I thought of it mostly because I saw the Peril comment about it and didn't think about it twice… but now that you mention it and I think about it, yeah it would have been totally ok to omit it (I just didn't even take 2s to ponder about that when I commented, and instead just reacted to Peril mindlessly there 😅 ). Eh, doesn't hurt to have one anyway 👍

AliSoftware:
> I've dismissed the change request as per the following review

Oh, sorry about that, I thought that by adding a new PR review as "comment" (and not "approval" nor "request changes") it would automatically replace my previous "request changes" review; I should have double-checked and dismissed it myself indeed!

PR #331, opened 2022-01-31
[L10n Tooling: i3] Re-implement download-from-glotpress logic from .py and .swift scripts

AliSoftware:
Part of the L10n Modernization project paaHJt-2Ib-p2, especially item [i3].

## What

This introduces a new `ios_download_strings_files_from_glotpress` action, which is mostly the re-implementation of:
 - The `Scripts/update-translations.rb` ruby script currently hosted in WPiOS/WCiOS repos
 - The `Scripts/fix-translation` swift script currently hosted in WPiOS/WCiOS repos

PS: Obviously, once this lands and the client repos have migrated to use the future new version of the toolkit including that action and start using it, we will completely delete the scripts it replaces from the client repos' `Scripts/` folders.

## Handling invalid files and empty translations

I've decided to only _warn_ (using a `UI.error`) about the potential case of having empty translations in downloaded exports — as opposed to trying to fix it ourselves like the previous `Scripts/fix-translation` script was trying to do. There are multiple reasons for this choice:
 1. The issue that led us to have some empty translations (`"key" = "";`) in exports in the past seems to have been resolved a long time ago and not be present anymore, at least to my experience and smoke checks in our Mag16 locales
 2. [The logic to check if there were any empty translations remaining after the fix](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/Scripts/update-translations.rb#L160) seemed to assume the `.strings` files (downloaded from GlotPress then fixed) were in UTF16 (given the bytes used on the `grep`), while in practice all the downloaded files are all in UTF-8 nowadays (`find WordPress/Resources -name 'Localizable.strings' -exec file {} \;` in WPiOS only lists the one from `en.lproj`, aka the original, to be UTF16, and all others downloaded from GlotPress to be UTF8). Which shows how the check was outdated and not even checking correctly what we expected it to fix (as the `grep` would not find those UTF16 bytes in the UTF-8 files anyway)
 3. The fix back in the way it was implemented by `Scripts/fix-translations` consisted in [replacing the empty `"key" = "";` with `"key" = "key";`](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/Scripts/fix-translation#L16); but given we want to be open to the possibility of (and start actually adopting) using keys that are different to the English copy (see Project Thread explanations), this would not have been the correct fix anymore, as that would not have been the correct fallback anymore for those cases.
 4. A potential fix would have instead be to replace lines matching `/= "";$/` with empty lines, or comment those lines out, so that the fallback would instead be to use the `value:` parameter from `NSLocalizedString(_ key:, value:, comment:)` in the codebase (which would be the English copy, as opposed to the key). But manipulating the textual `.strings` file in that way starts raising risks and issues when trying to deal with text encodings (and GlotPress serves those files as `application/octet-stream`, without explicitly exposing the actual encoding used, even if by experience it looks like UTF8 but we shouldn't make such an assumption if we want our code to be resilient)

Given that the bug seems to have been fixed since a long time I figured better avoid trying to fix something that does not exist — and risk messing things up rather than improving them — and instead just detect in case it reappears one day. If we end up seeing it reappear, it will always be time to try to implement a workaround in the action's code, but… YAGNI.

## What's next

While this implements most of what's needed to cover item [i3] of the plan (paaHJt-2Ib-p2), we will still need to add another action on top of this one (in a separate, future PR) to re-implement the logic from `Scripts/extract-framework-translations.swift` in a separate fastlane action in order to finally complete [i3] (and [i6]). Most of the logic to do that last step is already present in the `ios_l10n_helper` so that future PR should be straightforward though.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
@@ -30,16 +30,16 @@ module Fastlane
         return unless File.exist?(destination) # If the file failed to download, don't try to validate an non-existing file. We'd already have a separate error for the download failure anyway.
 
         translations = Fastlane::Helper::Ios::L10nHelper.read_strings_file_as_hash(path: destination)
-        empty_keys = translations.select { |_, value| value.nil? || value.empty? }.keys
+        empty_keys = translations.select { |_, value| value.nil? || value.empty? }.keys.sort
         unless empty_keys.empty?
           UI.error(
-            "Found empty translations in `#{destination}` for the following keys: #{empty_keys}.\n" \
+            "Found empty translations in `#{destination}` for the following keys: #{empty_keys.inspect}.\n" \
               + "This is likely a GlotPress bug, and will lead to copies replaced by empty text in the UI.\n" \
               + 'Please report this to the GlotPress team, and fix the file locally before continuing.'
           )
         end
       rescue StandardError => e
-        UI.error("Error while validating the file exported from GlotPress (`#{destination}`) - #{e.message}")
+        UI.error("Error while validating the file exported from GlotPress (`#{destination}`) - #{e.message.chomp}")
       end
 
       #####################################################
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
@@ -1,0 +1,105 @@
+module Fastlane
+  module Actions
+    class IosDownloadStringsFilesFromGlotPressAction < Action
+      def self.run(params)
+        # TODO: Once we introduce the `Locale` POD via #296, check if the param is an array of locales and if so convert it to Hash{glotpress=>lproj}
+        locales = params[:locales]
+
+        locales.each do |glotpress_locale, lproj_name|
+          # Download the export in the proper `.lproj` directory
+          destination = File.join(params[:download_dir], "#{lproj_name}.lproj", "#{params[:table_basename]}.strings")
+          Fastlane::Helper::Ios::L10nHelper.download_glotpress_export_file(
+            project_url: params[:project_url],
+            locale: glotpress_locale,
+            filters: params[:filters],
+            destination: destination
+          )
+          # Do a quick check of the downloaded `.strings` file to ensure it looks valid
+          validate_strings_file(destination) unless params[:skip_file_validation]
+        end
+      end
+
+      # Validate that a `.strings` file downloaded from GlotPress seems valid and does not contain empty translations
+      def self.validate_strings_file(destination)
+        translations = Fastlane::Helper::Ios::L10nHelper.read_strings_file_as_hash(destination)
+        empty_keys = translations.select { |_, value| value.nil? || value.empty? }.keys
+        unless empty_keys.empty?
+          UI.error(
+            "Found empty translations in `#{destination}` for the following keys: #{empty_keys}.\n" \
+              + "This is likely a GlotPress bug, and will lead to copies replaced by empty text in the UI.\n" \
+              + 'Please report this to the GlotPress team, and fix the file locally this continuing.'
+          )
+        end
+      rescue StandardError => e
+        UI.error("Error while validating the file exported from GlotPress (`#{destination}`) - #{e.message}")
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        'Downloads the `.strings` files from GlotPress for the various locales'
+      end
+
+      def self.details
+        <<~DETAILS
+          Downloads the `.strings` files from GlotPress for the various locales,
+          validates them, and saves them in the relevant `*.lproj` directories for each locale
+        DETAILS
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :project_url,
+                                       env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_PROJECT_URL',
+                                       description: 'URL to the GlotPress project',
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :locales,
+                                       env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_LOCALES',
+                                       description: 'The map of locales to download, each entry of the Hash corresponding to a { glotpress-locale-code => lproj-folder-basename } pair',
+                                       type: Hash), # TODO: also support an Array of `Locale` POD/struct type when we introduce it later (see #296)
+          FastlaneCore::ConfigItem.new(key: :download_dir,
+                                       env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_DOWNLOAD_DIR',
+                                       description: 'The parent directory containing all the `*.lproj` subdirectories in which the downloaded files will be saved',
+                                       type: Hash),
+          FastlaneCore::ConfigItem.new(key: :table_basename,
+                                       env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_TABLE_BASENAME',
+                                       description: 'The basename to save the `.strings` files under',
+                                       type: String,
+                                       optional: true,
+                                       default_value: 'Localizable'),
+          FastlaneCore::ConfigItem.new(key: :filters,
+                                       env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_FILTERS',
+                                       description: 'The GlotPress filters to use when requesting the translations export',
+                                       type: Hash,
+                                       optional: true,
+                                       default_value: { status: 'current' }),
+          FastlaneCore::ConfigItem.new(key: :skip_file_validation,
+                                       env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_SKIP_FILE_VALIDATION',
+                                       description: 'If true, skips the validation of `.strings` files after download',
+                                       type: Fastlane::Boolean,
+                                       optional: true,
+                                       default_value: false),
+        ]
+      end
+
+      def self.return_type
+        # Describes what type of data is expected to be returned
+        # see RETURN_TYPES in https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/action.rb
+      end
+
+      def self.return_value
+        # Textual description of what the return value is
+      end
+
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include?(platform)
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
@@ -27,7 +27,7 @@ module Fastlane
           UI.error(
             "Found empty translations in `#{destination}` for the following keys: #{empty_keys}.\n" \
               + "This is likely a GlotPress bug, and will lead to copies replaced by empty text in the UI.\n" \
-              + 'Please report this to the GlotPress team, and fix the file locally this continuing.'
+              + 'Please report this to the GlotPress team, and fix the file locally before continuing.'
           )
         end
       rescue StandardError => e
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
@@ -24,7 +24,9 @@ module Fastlane
 
       # Validate that a `.strings` file downloaded from GlotPress seems valid and does not contain empty translations
       def self.validate_strings_file(destination)
-        translations = Fastlane::Helper::Ios::L10nHelper.read_strings_file_as_hash(destination)
+        return unless File.exist?(destination) # If the file failed to download, don't try to validate an non-existing file. We'd already have a separate error for the download failure anyway.
+
+        translations = Fastlane::Helper::Ios::L10nHelper.read_strings_file_as_hash(path: destination)
         empty_keys = translations.select { |_, value| value.nil? || value.empty? }.keys
         unless empty_keys.empty?
           UI.error(
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
@@ -4,10 +4,13 @@ module Fastlane
       def self.run(params)
         # TODO: Once we introduce the `Locale` POD via #296, check if the param is an array of locales and if so convert it to Hash{glotpress=>lproj}
         locales = params[:locales]
+        download_dir = params[:download_dir]
+
+        UI.user_error!("The parent directory `#{download_dir}` (which contains all the `*.lproj` subdirectories) must already exist") unless Dir.exist?(download_dir)
 
         locales.each do |glotpress_locale, lproj_name|
           # Download the export in the proper `.lproj` directory
-          lproj_dir = File.join(params[:download_dir], "#{lproj_name}.lproj")
+          lproj_dir = File.join(download_dir, "#{lproj_name}.lproj")
           destination = File.join(lproj_dir, "#{params[:table_basename]}.strings")
           FileUtils.mkdir(lproj_dir) unless Dir.exist?(lproj_dir)
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
@@ -65,7 +65,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :download_dir,
                                        env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_DOWNLOAD_DIR',
                                        description: 'The parent directory containing all the `*.lproj` subdirectories in which the downloaded files will be saved',
-                                       type: Hash),
+                                       type: String),
           FastlaneCore::ConfigItem.new(key: :table_basename,
                                        env_name: 'FL_IOS_DOWNLOAD_STRINGS_FILES_FROM_GLOTPRESS_TABLE_BASENAME',
                                        description: 'The basename to save the `.strings` files under',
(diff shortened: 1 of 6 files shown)


spec/ios_download_strings_files_from_glotpress_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it 'uses the proper filters when exporting the files from GlotPress'
+    it 'uses a custom table name for the `.strings` files if provided'
+  end
+
+  describe 'error handling' do
+    it 'shows an error if an invalid locale is provided (404)'
+    it 'shows an error if the file cannot be written in the destination'
+    it 'reports if a downloaded file is invalid by default'
+    it 'does not report invalid downloaded files if skip_file_validation:true'
+  end
```
AliSoftware: PR is still Draft, I plan to implement these soon… though having the helper from #330 merged first to then use it to implement those would help 🙂

spec/ios_l10n_helper_spec.rb
```diff
@@ -160,4 +160,82 @@ def file_encoding(path)
...
+        # Act
+        described_class.download_glotpress_export_file(project_url: gp_fake_url, locale: 'fr', filters: nil, destination: dest)
+        # Assert
+        expect(stub).to have_been_made.once
+        expect(dest.string).to eq('content')
+      end
+
+      it 'passes the expected params when a list of filters is provided' do
+        # Arrange
+        stub = stub_request(:get, "#{gp_fake_url}/fr/default/export-translations?format=strings&filters%5Bstatus%5D=current&filters%5Bterm%5D=foobar").to_return(body: 'content')
```
mokagio: Just a thought: this stub depends on the order of the parameters as hardcoded in the URL. 

TIL [webmock has an API to specify query parameters using an hash](https://github.com/bblimke/webmock/tree/33d8810c2828fc17010e15cc3f21ad2c726a966f#matching-query-params-using-hash).

```ruby
stub_request(:get, "www.example.com").with(query: {"a" => ["b", "c"]})
```

I haven't looked into it deeper, but I assume that approach would not depend on the order of the parameters.

There are also `hash_including` and `hash_excluding` methods that can be used to build the query hash, but I don't think we need them here as we're interested in the query containing all and only those values.

spec/ios_l10n_helper_spec.rb
```diff
@@ -160,4 +160,82 @@ def file_encoding(path)
...
+      end
+    end
+
+    describe 'destination parameter' do
+      # it 'accepts an IO (File)' # Not really needed, as all previous tests already cover this by using a `StringIO`
+
+      it 'accepts a String (path)' do
+        Dir.mktmpdir('a8c-release-toolkit-tests-') do |tmp_dir|
+          # Arrange
+          # Note: in practice it seems that GlotPress's `.strings` exports are using UTF8 (but served as `application/octet-stream`)
```
mokagio: Nitpick:

```suggestion
          # Note: in practice it seems that GlotPress's `.strings` exports are using UTF-8 (but served as `application/octet-stream`)
```

I can't find a page that says "You shall always add a dash", but that's how the different encoding are named in official docs, such as [the techinical FAQ page on the unicode.org site](https://www.unicode.org/faq/utf_bom.html).

spec/ios_l10n_helper_spec.rb
```diff
@@ -160,4 +160,82 @@ def file_encoding(path)
...
+          #       but it does not hurt to ensure the download to a file can work with UTF16 ()and copy the binary stream verbatim)
```
mokagio: Other naming nitpick:
```suggestion
          #       but it does not hurt to ensure the download to a file can work with UTF-16 ()and copy the binary stream verbatim)
```

mokagio: Typo (?) fix:

```suggestion
          #       but it does not hurt to ensure the download to a file can work with UTF16 (and copy the binary stream verbatim)
```

spec/ios_l10n_helper_spec.rb
```diff
@@ -160,4 +160,82 @@ def file_encoding(path)
...
+        # Arrange
+        stub = stub_request(:get, "#{gp_fake_url}/invalid/default/export-translations?format=strings").to_return(status: [404, 'Not Found'])
+        error_messages = []
+        allow(FastlaneCore::UI).to receive(:error) { |message| error_messages.append(message) }
+        dest = StringIO.new
+        # Act
+        described_class.download_glotpress_export_file(project_url: gp_fake_url, locale: 'invalid', filters: nil, destination: dest)
+        # Assert
+        expect(stub).to have_been_made.once
+        expect(error_messages).to eq(['Error downloading locale `invalid` — 404 Not Found'])
```
mokagio, line 224: Why did you decide to append the errors to an array when there's only one returned, rather than having a `String` value? Just in case we get no errors (in which case empty array is clearer than `nil`) or, more interesting, multiple ones?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -116,6 +117,30 @@ def self.generate_strings_file_from_hash(translations:, output_path:)
...
+        # Also post-process the downloaded content to comment out any potential line with an empty (`""`) translation,
+        # to ensure those would fall back to the English/Base locale at runtime instead. These case should theoretically
+        # never happen as GlotPress should not include empty translations in its exports, but we've seen issues in the
+        # past, so better safe than sorry.
+        #
+        # @param [String] project_url The URL to the GlotPress project to export from, e.g. `"https://translate.wordpress.org/projects/apps/ios/dev"`
+        # @param [String] locale The GlotPress locale code to download strings for.
+        # @param [Hash{Symbol=>String}] filters The hash of filters to apply when exporting from GlotPress.
+        #                               Typical examples include `{ status: 'current' }` (the default) or `{ status: 'review' }`.
+        # @param [String, IO] destination The path or `IO`-like instance, where to write the downloaded file on disk.
```
mokagio, line 127: Just curious: Is it common in Ruby to expect either a `String` for the path or an `IO` (sub)class? 

It works well for the tests and the code doesn't need any extra handling for this duplicity 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_download_strings_files_from_glotpress.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      # Validate that a `.strings` file downloaded from GlotPress seems valid and does not contain empty translations
+      def self.validate_strings_file(destination)
+        translations = Fastlane::Helper::Ios::L10nHelper.read_strings_file_as_hash(destination)
+        empty_keys = translations.select { |_, value| value.nil? || value.empty? }.keys
+        unless empty_keys.empty?
+          UI.error(
+            "Found empty translations in `#{destination}` for the following keys: #{empty_keys}.\n" \
+              + "This is likely a GlotPress bug, and will lead to copies replaced by empty text in the UI.\n" \
+              + 'Please report this to the GlotPress team, and fix the file locally this continuing.'
```
mokagio:
```suggestion
              + 'Please report this to the GlotPress team, and fix the file locally before continuing.'
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -116,6 +117,30 @@ def self.generate_strings_file_from_hash(translations:, output_path:)
...
+        #
+        def self.download_glotpress_export_file(project_url:, locale:, filters:, destination:)
+          query_params = (filters || {}).transform_keys { |k| "filters[#{k}]" }.merge(format: 'strings')
```
mokagio, line 130: Would this work to keep the chain that evaluates `query_params` a "linear"

```suggestion
        def self.download_glotpress_export_file(project_url:, locale:, destination:, filters: {})
          query_params = filters.transform_keys { |k| "filters[#{k}]" }.merge(format: 'strings')
```

Of course, that would mean omitting `filters: nil` when one doesn't want filters, which I'm guessing is intentional in the design because we expect callers to always pass a filter (as shown by the default value of the param in the action above being `{ status: 'current' }`. 

So I think the answer to this question is "yes it would make the implementation neater, but it would result in a worse ergonomic, and we should keep it as it is."

mokagio:
Took the liberty of adding this to the [Localization Toolchain Modernization & Fixes](https://github.com/orgs/wordpress-mobile/projects/105) project.

mokagio (commented):
Looking good so far 👍  Also easier to follow than the original scripts. Looking forward to have everything centralized here.

spec/ios_l10n_helper_spec.rb
AliSoftware: Nice, TIL, yeah I think that makes sense to switch to that `with(query:)` matcher indeed, nice find!

spec/ios_l10n_helper_spec.rb
AliSoftware: Yeah my guess is that the official name is `UTF-8`, and the places where we don't use a dash are places where this is used as a symbol name in code or similar contexts where a dash is not a valid character?

spec/ios_l10n_helper_spec.rb
AliSoftware: Yep, exactly: this expectation not only check that I have the one `UI.error`, but also ensures that we didn't get any other, unexpected ones, but only the one I expect.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: I'm not sure if it's common (as in "idiomatic"), but I've seen it in a couple of places.

It allows you to provide an IO if you already have one open instead of having just a path, not only for tests but also for some other contexts at call site where you might already have an IO opened and around to pass (e.g. if you want to concatenate various texts, or add a comment header at start of the file, you can `File.open`, then `f.write(header)`, then pass the `f` IO to the method to add the rest, then `f.close`. Granted, this is not super common nor really a use case for us here, but it's nice to have the flexibility, for tests or otherwise.

I would probably not add such a flexibility if it required a lot of additional code in the implementation just to support both. But given that here it doesn't cost anything (as `IO.copy_stream` already supports both), I figured it was worth it (especially given that it made the tests way nicer to write 😛 )

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
AliSoftware: Yeah tbh I went back and forth on that one, and at some point I even had `filters: { status: 'current' }` as a default.

But then it didn't read as nice, I thought, and ultimately I decided in 4952fba58ec92373f1f4bf947942fd2e74fe822f that it made more sense to have the default value on the action anyway (see commit's message) rather than having to decide here if we wanted `nil` or `{ status: 'current' }` as a default, and only to potentially have a different default (and no single source of truth for a default value) than the one provided as `default_value` on the `ConfigItem` of the action itself.

Besides, having the `filters:` parameter closer to the `project_url` and `locale` (given all 3 parameters are influencing the request made to GlotPress) and the `destination:` parameter separate at the end made the call site read nicer as well, rather than having `filters:` at the end (which `rubocop` enforces when we provide a default value for it)

spec/ios_l10n_helper_spec.rb
```diff
@@ -160,4 +160,82 @@ def file_encoding(path)
...
+    end
+
+    describe 'destination parameter' do
+      # it 'accepts an IO (File)' # Not really needed, as all previous tests already cover this by using a `StringIO`
+
+      it 'accepts a String (path)' do
+        Dir.mktmpdir('a8c-release-toolkit-tests-') do |tmp_dir|
+          # Arrange
+          # Note: in practice it seems that GlotPress's `.strings` exports are using UTF-8 (but served as `application/octet-stream`)
+          #       but it does not hurt to ensure the download to a file can work with UTF-16 ()and copy the binary stream verbatim)
```
AliSoftware: Previous GitHub suggestion for typo by Gio conflicted with another one he did on the same line, so GitHub ended up only applying one of the two 🤦 . Re-fixing it here.

```suggestion
          #       but it does not hurt to ensure the download to a file can work with UTF-16 (and copy the binary stream verbatim)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
mokagio: 👍 👍 👍

spec/ios_l10n_helper_spec.rb
AliSoftware: This actually just saved my bacon as I fixed something else in the implementation and made a typo, as this expectation then caught a second error message that I didn't expect to be there (and was because of said typo). So that helped me fix the implementation change I made  in order to make that test go green again and return to only one error message  💪

AliSoftware:
@mokagio This should now be ready for review 🎉

spec/ios_download_strings_files_from_glotpress_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+        stub_zh = gp_stub(locale: 'zh-cn', query: expected_gp_params).to_return(body: '"key" = "zh-copy";')
+
+        # Act
+        run_described_fastlane_action({
+          project_url: gp_fake_url,
+          locales: locales_subset,
+          download_dir: tmp_dir,
+          table_basename: tablename,
+          filters: filters
+        }.compact)
```
AliSoftware, line 30: This `.compact` ensures that if we pass a `nil` value for `tablename` or `filters`, then the corresponding keys / parameters passed when calling the action (i.e. `table_basename: …` and `filters: …` won't be included in the call site; allowing them to use the default value of their `ConfigItem` instead, just as intended (as opposed to providing the parameters but passing them a `nil` value which would overwrite the default)

spec/ios_download_strings_files_from_glotpress_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+      act = lambda do
+        run_described_fastlane_action(
+          project_url: gp_fake_url,
+          locales: { 'fr-FR': 'fr' },
+          download_dir: download_dir
+        )
+      end
+
+      # Assert
+      expect { act.call }.to raise_error(FastlaneCore::Interface::FastlaneError, "The parent directory `#{download_dir}` (which contains all the `*.lproj` subdirectories) must already exist")
```
AliSoftware, line 103: This is the exception raised by `UI.user_error!`.

PS: Note that we can't really use `allow(FastlaneCore::UI).to receive(:user_error!) { |m| … }` here because if we catch and stub those, that won't interrupt the action (i.e. the call [here on L9](https://github.com/wordpress-mobile/release-toolkit/pull/331/files#diff-db88a4ce40ab03bd414faa6cafc8fde0008f5fdf43b7db981943fd9ab40fd73fR9) would not raise if stubbed so the rest of the `run` method would continue executing).

_(And we could do `allow(FastlaneCore::UI).to receive(:user_error!) do |e| raise e }` to be sure that it still interrupts the code even when stubbed… but then what would be the point of stubbing the method while the real implementation already raises…)_

AliSoftware: Btw, wdyt of using a `lambda` like this here (which is basically the Ruby equivalent of an anonymous closure in Swift) so that it allowed me to still keep the `# Act` and `# Assert` separate?

Is it readable and worth it? Or too obscure and you'd instead prefer to avoid the `lambda` and directly use `expect { run_described_fastlane_action(…) }.to …`… even if that means mixing the "Act" and "Assert" steps together and breaking the usual A/A/A structure of the tests just to be able to catch the exception raised by the "act" part?

mokagio: > Btw, wdyt of using a `lambda` like this here...

I think it's a neat trick that doesn't affect the code readability much. Having said that, as much as I like AAA I think it would be okay to break the pattern here to keep the code concise.

Up to your preference.

spec/ios_download_strings_files_from_glotpress_spec.rb
```diff
@@ -55,16 +54,127 @@ def test_gp_download(filters:, tablename:, expected_gp_params:)
...
+          project_url: gp_fake_url,
+          locales: { 'unknown-locale': 'Base' },
+          download_dir: tmp_dir
+        )
+
+        # Assert
+        expect(stub).to have_been_made.once
+        file = File.join(tmp_dir, 'Base.lproj', 'Localizable.strings')
+        expect(File).not_to exist(file)
+        expect(error_messages).to eq(['Error downloading locale `unknown-locale` — 404 Not Found'])
```
mokagio: Nitpick: Since `file` is only used one, inlining it wouldn't make the code less DRY and also result in a nice `expect` only assert section.

```suggestion
        expect(stub).to have_been_made.once
        expect(File).not_to exist(File.join(tmp_dir, 'Base.lproj', 'Localizable.strings'))
        expect(error_messages).to eq(['Error downloading locale `unknown-locale` — 404 Not Found'])
```

Not a big deal at all, happy for it to stay as it is. Just throwing it here in case you like it too...

spec/ios_download_strings_files_from_glotpress_spec.rb
mokagio: Nice. What do you think of adding the comment in the code? I can see myself scratching my head for a minute when looking at this code in the future.

mokagio (approved):
I appreciated the prod to test code ratio. Even accounting for the boilerplate required by file system access and Fastlane action and `UI`, it's still very thorough. Thank you. It makes me trust the code a lot more and will make modifying it in the future easier.

GTG, only left minor non-necessary suggestions.

spec/ios_download_strings_files_from_glotpress_spec.rb
AliSoftware: Good point, will add one, thanks for the suggestion  👍

spec/ios_download_strings_files_from_glotpress_spec.rb
AliSoftware: Addressed in fd678842e1ef8a6fe3ffbd33bd403f1354b420e5

spec/ios_download_strings_files_from_glotpress_spec.rb
AliSoftware: Added via ba0fdc6d9ee7671b3888f4c838c323644e11caca 🙂

spec/ios_download_strings_files_from_glotpress_spec.rb
AliSoftware: 👍  I'll keep it then 🙂

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -116,6 +117,30 @@ def self.generate_strings_file_from_hash(translations:, output_path:)
...
           File.write(output_path, builder.to_xml)
         end
+
+        # Downloads the export from GlotPress for a given locale and given filters.
+        #
+        # Also post-process the downloaded content to comment out any potential line with an empty (`""`) translation,
+        # to ensure those would fall back to the English/Base locale at runtime instead. These case should theoretically
+        # never happen as GlotPress should not include empty translations in its exports, but we've seen issues in the
+        # past, so better safe than sorry.
+        #
```
AliSoftware: This whole doc was written before I started doing the implementation, but the plan changed and I decided to not post-process the downloaded file after all, but instead only validate it (and also leave that validation for the action rather than the helper).

```suggestion
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -116,6 +117,30 @@ def self.generate_strings_file_from_hash(translations:, output_path:)
...
+        # @param [String] project_url The URL to the GlotPress project to export from, e.g. `"https://translate.wordpress.org/projects/apps/ios/dev"`
+        # @param [String] locale The GlotPress locale code to download strings for.
+        # @param [Hash{Symbol=>String}] filters The hash of filters to apply when exporting from GlotPress.
+        #                               Typical examples include `{ status: 'current' }` (the default) or `{ status: 'review' }`.
```
AliSoftware: This is no longer the default value as far as the helper method is concerned (that default value has since been moved on the action's `ConfigItem` instead)
```suggestion
        #                               Typical examples include `{ status: 'current' }` or `{ status: 'review' }`.
```

PR #332, opened 2022-02-01
Clarify expectation on XML indentation in `AnLocalizeLibsAction` spec

mokagio:
This hopefully makes the test clearer. See discussion at https://github.com/wordpress-mobile/release-toolkit/pull/330#discussion_r796838599

### DIFF
--- a/spec/an_localize_libs_action_spec.rb
+++ b/spec/an_localize_libs_action_spec.rb
@@ -26,13 +26,10 @@ describe Fastlane::Actions::AnLocalizeLibsAction do
         ]
       )
 
-      # Notice the extra indentation in the library strings. The action doesn't
-      # modify the app's strings content indentation, but it applies its own
-      # standard to the values read from the given library strings
       expected = <<~XML
         <string name="a_string">test from app</string>
-          <string name="a_lib1_string">test from lib 1</string>
-          <string name="a_lib2_string">test from lib 2</string>
+        <string name="a_lib1_string">test from lib 1</string>
+        <string name="a_lib2_string">test from lib 2</string>
       XML
       expect(File.read(app_strings_path)).to eq(android_xml_with_content(expected))
     end
@@ -40,10 +37,17 @@ describe Fastlane::Actions::AnLocalizeLibsAction do
 end
 
 def android_xml_with_content(content)
+  # Under the hood, the code hardcodes the generated XML indentation as 4
+  # spaces. Let's "force" it to be the same in here for consistency.
+  #
+  # See
+  # https://github.com/wordpress-mobile/release-toolkit/blob/da405f2c5ca90d696857ba2ad01da2753daa60dc/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L67
+  default_indentation = ' ' * 4
+
   # I couldn't find a way to interpolate a multiline string preserving its
   # indentation in the heredoc below, so I hacked the following transformation
   # of the input that adds the desired indentation to all lines.
-  indented_content = content.chomp.lines.map { |l| "  #{l}" }.join
+  indented_content = content.chomp.lines.map { |l| "#{default_indentation}#{l}" }.join
 
   return <<~XML
     <?xml version="1.0" encoding="UTF-8"?>
--- a/spec/an_localize_libs_action_spec.rb
+++ b/spec/an_localize_libs_action_spec.rb
@@ -43,7 +43,7 @@ def android_xml_with_content(content)
   #
   # The desired indentation is 4 spaces to stay aligned with the production
   # code applies when merging the XMLs.
-  indented_content = content.chomp.lines.map { |l| "    #{default_indentation}#{l}" }.join
+  indented_content = content.chomp.lines.map { |l| "    #{l}" }.join
 
   return <<~XML
     <?xml version="1.0" encoding="UTF-8"?>
--- a/spec/an_localize_libs_action_spec.rb
+++ b/spec/an_localize_libs_action_spec.rb
@@ -37,17 +37,13 @@ describe Fastlane::Actions::AnLocalizeLibsAction do
 end
 
 def android_xml_with_content(content)
-  # Under the hood, the code hardcodes the generated XML indentation as 4
-  # spaces. Let's "force" it to be the same in here for consistency.
-  #
-  # See
-  # https://github.com/wordpress-mobile/release-toolkit/blob/da405f2c5ca90d696857ba2ad01da2753daa60dc/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L67
-  default_indentation = ' ' * 4
-
   # I couldn't find a way to interpolate a multiline string preserving its
   # indentation in the heredoc below, so I hacked the following transformation
   # of the input that adds the desired indentation to all lines.
-  indented_content = content.chomp.lines.map { |l| "#{default_indentation}#{l}" }.join
+  #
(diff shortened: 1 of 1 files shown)


spec/an_localize_libs_action_spec.rb
```diff
@@ -26,24 +26,28 @@
...
 end
 
 def android_xml_with_content(content)
+  # Under the hood, the code hardcodes the generated XML indentation as 4
+  # spaces. Let's "force" it to be the same in here for consistency.
+  #
+  # See
+  # https://github.com/wordpress-mobile/release-toolkit/blob/da405f2c5ca90d696857ba2ad01da2753daa60dc/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L67
+  default_indentation = ' ' * 4
+
```
AliSoftware: FWIW and tbh, I don't think it's really worth having that lengthy comment and a dedicated constant for all that here 🙃 . You can just use 4 spaces literally in the map and be done with it, I think it's enough.

(In other words, remove all those lines 40-46 and just use `{ |l| "    #{l}" }` for the `map` block.)

The real puzzle is why you needed a different indentation for the first line (compared to all the other lines) in the `expected = <<~XML` heredoc (L30-32) before, and not why the indentation were 4 spaces… so it's ok to just use the 4 spaces without comment here imho 🙂

mokagio: > why you needed a different indentation for the first line

**I think it's some Nokogiri weirdness / us not using it properly**. It looks like the XML content of the main strings file is not modified, i.e. it's indentation stays the same. That's despite us rewriting the file while also passing `indent: 4`:

https://github.com/wordpress-mobile/release-toolkit/blob/b533fe2376454931c3196dddcfb672b3c1e329c8/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L130-L132

We can verify this by feeding an XML with no indentation as the main one to the action and verify that, once the action is done, the nodes from it still have no indentation.

```diff
@@ -10,7 +10,14 @@ describe Fastlane::Actions::AnLocalizeLibsAction do
   it 'merges the strings from the given array into the given main strings file' do
     in_tmp_dir do |tmp_dir|
       app_strings_path = File.join(tmp_dir, 'app.xml')
-      File.write(app_strings_path, android_xml_with_content('<string name="a_string">test from app</string>'))
+      xml_with_no_indent = <<~XML
+        <?xml version="1.0" encoding="UTF-8"?>
+        <resources xmlns:tools="http://schemas.android.com/tools">
+        <!-- no indentation here, to show how the action doesn't modify the main XML -->
+        <string name="a_string">test from app</string>
+        </resources>
+      XML
+      File.write(app_strings_path, xml_with_no_indent)
 
       lib1_strings_path = File.join(tmp_dir, 'lib1.xml')
       File.write(lib1_strings_path, android_xml_with_content('<string name="a_lib1_string">test from lib 1</string>'))
@@ -27,11 +34,15 @@ describe Fastlane::Actions::AnLocalizeLibsAction do
       )
 
       expected = <<~XML
+        <?xml version="1.0" encoding="UTF-8"?>
+        <resources xmlns:tools="http://schemas.android.com/tools">
+        <!-- no indentation here, to show how the action doesn't modify the main XML -->
         <string name="a_string">test from app</string>
-        <string name="a_lib1_string">test from lib 1</string>
-        <string name="a_lib2_string">test from lib 2</string>
+            <string name="a_lib1_string">test from lib 1</string>
+            <string name="a_lib2_string">test from lib 2</string>
+        </resources>
       XML
-      expect(File.read(app_strings_path)).to eq(android_xml_with_content(expected))
+      expect(File.read(app_strings_path)).to eq(expected)
     end
   end
 end
```

Conversely, regardless of the indentation we use in the input files for the libraries, they are appended using 4 spaces as per the line linked in my original code comment.

```diff
@@ -13,10 +13,22 @@ describe Fastlane::Actions::AnLocalizeLibsAction do
       File.write(app_strings_path, android_xml_with_content('<string name="a_string">test from app</string>'))
 
       lib1_strings_path = File.join(tmp_dir, 'lib1.xml')
-      File.write(lib1_strings_path, android_xml_with_content('<string name="a_lib1_string">test from lib 1</string>'))
+      xml_with_no_indent = <<~XML
+        <?xml version="1.0" encoding="UTF-8"?>
+        <resources xmlns:tools="http://schemas.android.com/tools">
+        <string name="a_lib1_string">test from lib 1</string>
+        </resources>
+      XML
+      File.write(lib1_strings_path, xml_with_no_indent)
 
       lib2_strings_path = File.join(tmp_dir, 'lib2.xml')
-      File.write(lib2_strings_path, android_xml_with_content('<string name="a_lib2_string">test from lib 2</string>'))
+      xml_with_eight_indent = <<~XML
+        <?xml version="1.0" encoding="UTF-8"?>
+        <resources xmlns:tools="http://schemas.android.com/tools">
+                <string name="a_lib2_string">test from lib 2</string>
+        </resources>
+      XML
+      File.write(lib2_strings_path, xml_with_eight_indent)
 
       run_described_fastlane_action(
         app_strings_path: app_strings_path,
```

I added some `puts` here and there during the manipulation of the Nokogiri object `main_strings` and the output always had the original level of indentation (be that 2 spaces, no spaces, 4 spaces, etc.).

I'm satisfied with this answer, even though I don't see why calling `to_xml(indent: 4)` wouldn't apply that indentation to the output. I don't know much about Nokogiri, but I'm sure I'll get to know more about it as we keep working on this [project](https://github.com/orgs/wordpress-mobile/projects/105).

AliSoftware: Oh shoot I remember now having seen that in the past, where Nokogiri considering those spaces (that we use as intent) as being `XML::Text` notes (i.e. just like content text inside the surrounding XML tag, like when you write `<tag>  hello  </tag>`) rather than interpreting those leading spaces as indentation… which is why I think the `to_xml(indent: 4)` does not work like we expect it to. Which for me is a bug on Nokogiri's side (and probably be why we had to append those 4 spaces per line manually and explicitly in the link you linked in your original comment after all…) 🤦

AliSoftware: > I'm satisfied with this answer

Yes, works for me as well. :shipit: 

PS: I'll personally still probably remove the lengthy comment and use 4 literal spaces instead of `default_indentation`, as it feels a bit over-verbose… but up to you if you prefer to keep it. Rest is ready to ship 👍

mokagio: I went for a middle-ground solution in 7ad1d9d: removed the `default_indentation` definition and the link to the production code but still left a (shorter) comment to say that we use 4 spaces here to be aligned with the production code.

mokagio:
@wpmobilebot 

> Please add an entry in the CHANGELOG.md file to describe the changes made by this PR

I'm ignoring this because it's a minor test code enhancement.

spec/an_localize_libs_action_spec.rb
AliSoftware: > I went for a middle-ground solution in [7ad1d9d](https://github.com/wordpress-mobile/release-toolkit/commit/7ad1d9d85dc267ee871743246bed6767ad67c3a4): removed the `default_indentation` definition and the link to the production code but still left a (shorter) comment to say that we use 4 spaces here to be aligned with the production code.

Perfect, nice balance 👌 👏

AliSoftware:
> I'm ignoring this because it's a minor test code enhancement.

👍  agreed

AliSoftware (approved):

PR #333, opened 2022-02-02
Add Buildkite Trigger Build Action

jkmassel:
Adds an action to trigger builds via the Buildkite API.

Uses https://github.com/Shopify/buildkit under the hood.

**To Test**
1. Obtain a Buildkite Personal API Token with the `write_builds` scope.
2. Run `bundle exec fastlane run buildkite_trigger_build`, and enter the details for an existing branch. If you'd like to test this PR with https://github.com/wordpress-mobile/WordPress-iOS/pull/17862, you can use these values:
- `branch` : `add/buildkite-release-builds`
- `pipeline file`: `release-builds.yml`
- `access_token`: [Your Access Token]
- `pipeline name`: `wordpress-ios`


Gemfile.lock
```diff
@@ -276,7 +280,8 @@ GEM
     naturally (2.2.1)
     netrc (0.11.0)
     no_proxy_fix (0.1.2)
-    nokogiri (1.12.5-x86_64-darwin)
+    nokogiri (1.12.5)
+      mini_portile2 (~> 2.6.1)
```
mokagio, line 284: Interesting that, when the dependency changed to be the non-platform specific, `mini_portile2` came with it, too 🤔

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -28,6 +28,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'diffy', '~> 3.3'
   spec.add_dependency 'nokogiri', '~> 1.11' # Needed for AndroidLocalizeHelper
   spec.add_dependency 'octokit', '~> 4.18'
+  spec.add_dependency 'buildkit', '~> 1.4.5'
+
```
mokagio: Have you considered relaxing to:
 
```suggestion
  spec.add_dependency 'buildkit', '~> 1.4'

```

mokagio: Also, is the empty line intentional to differentiate the block of dependencies above from the one below?

In case it isn't:

```suggestion
  spec.add_dependency 'buildkit', '~> 1.4.5'
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
+module Fastlane
+  module Actions
+    class BuildkiteTriggerBuildAction < Action
+      def self.run(params)
+        require 'buildkit'
+
+        UI.message "Triggering build on branch #{params[:branch]} at #{params[:commit]}"
```
mokagio: Nitpick:

```suggestion
        UI.message "Triggering build on branch #{params[:branch]}, commit #{params[:commit]}, using pipeline from #{params[:pipeline_file]}"
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        pipeline_name = {
+          PIPELINE: params[:pipeline_file]
```
mokagio, line 10: I assume Buildkit checks whether the pipeline file exists, right?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+          params[:buildkite_organization],
+          params[:buildkite_pipeline],
+          {
+            branch: params[:branch],
+            commit: params[:commit],
+            env: params[:environment].merge(pipeline_name)
+          }
+        )
+
+        response.state == 'scheduled' ? UI.message('Done!') : UI.user_error!("Failed to start job\nError: [#{response}]")
```
mokagio: Nitpick: How would you feel about using [`crash!`](https://docs.fastlane.tools/advanced/actions/#interacting-with-the-user) instead? It's borderline, but the error in the response might be caused by something other than a user error (e.g. network timeout)

```suggestion
        response.state == 'scheduled' ? UI.message('Done!') : UI.crash!("Failed to start job\nError: [#{response}]")
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      def self.description
+        'Triggers a job on Buildkite'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :buildkite_token,
+            env_name: 'BUILDKITE_ACCESS_TOKEN',
```
mokagio: Nitpick:

```suggestion
            key: :buildkite_token,
            env_name: 'BUILDKITE_TOKEN',
```

Just to keep the two names consistent

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+            key: :buildkite_organization,
+            env_name: 'BUILDKITE_ORGANIZTION',
+            description: 'The Buildkite organization that contains your pipeline',
+            type: String,
+            default_value: 'wordpress-mobile'
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :buildkite_pipeline,
+            env_name: 'BUILDKITE_PIPELINE',
+            description: 'The Buildkite pipeline you\'d like to build',
```
mokagio: I wasn't expecting that escape to work, but it does 😄 

```
➜ irb
irb(main):001:0> s = 'The Buildkite pipeline you\'d like to build'
=> "The Buildkite pipeline you'd like to build"
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+            env_name: 'BUILDKITE_ACCESS_TOKEN',
+            description: 'Buildkite Personal Access Token',
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :buildkite_organization,
+            env_name: 'BUILDKITE_ORGANIZTION',
+            description: 'The Buildkite organization that contains your pipeline',
+            type: String,
+            default_value: 'wordpress-mobile'
```
mokagio: I'd like to suggest removing the default value. 

I'm saying this because, in the long run, the wordpress-mobile org won't hold the majority of usages for this action. At that point, it might be confusing to have such a default. If we don't set it now, we also save ourselves the trouble of removing it in the future (which would be a breaking change, no?) and keep all usages more verbose but consistent.

```suggestion
            type: String
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+            key: :environment,
+            description: 'Any additional environment variables to provide to the job',
+            type: Hash,
+            default_value: {}
+          ),
+        ]
+      end
+
+      def self.authors
+        ['jkmassel']
```
mokagio:
```suggestion
        ['Automattic']
```

I wonder if Fastlane has like a `.env` that we can put in `~` or the root of our Automattic dev folder where to set this 🤔

mokagio (commented):
Looking good so far, just dropping my initial notes before manual testing in case I get interrupted. 👍 

I got a bit sad that we (you) had to write all this boilerplate to do things the Fastlane way 😞 I wondered if there was an action already, but the [list of available actions](https://docs.fastlane.tools/actions/) doesn't include Buildkite (or `buildk`, even) at this time.

I guess I'm in a grumpy mood because I haven't had lunch yet but having a wrapper of an API being itself wrapped makes me sad.

toy story troikas

I mean, I understand why we'd want to do it this way and I also can't suggest a neater way that wouldn't include completely ripping apart our tooling. So, just sharing a feeling. I hope you find the GIF funny 😅

Gemfile.lock
AliSoftware: Yeah that has always been the case since we started to have this back-and-forth of platform-specific to non-platform-specific iirc… 🤷

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
AliSoftware: +1 on removing the default value here.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
AliSoftware: Alternatively, you could use:
```suggestion
            description: %(The Buildkite pipeline you'd like to build),
```
(but it's just a matter of preference, really, so 🤷 )

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+        'Triggers a job on Buildkite'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :buildkite_token,
+            env_name: 'BUILDKITE_ACCESS_TOKEN',
+            description: 'Buildkite Personal Access Token',
+            type: String
```
AliSoftware: TIL about [the `sensitive` option](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L57-L59), so might as well use it (and probably worth adding to our other existing actions at some point too, but that's for a separate PR)
```suggestion
            type: String,
            sensitive: true
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+            key: :branch,
+            description: 'The branch you want to build',
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :commit,
+            description: 'The commit hash you want to build',
+            type: String,
+            default_value: 'HEAD'
+          ),
```
AliSoftware, line 66: Should we [make those exclusive to one another](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L48-L49), so that only one is valid? (Otherwise, what happens if we provide both?!)

```suggestion
          FastlaneCore::ConfigItem.new(
            key: :branch,
            description: 'The branch you want to build',
            type: String,
            conflicting_options: [:commit]
          ),
          FastlaneCore::ConfigItem.new(
            key: :commit,
            description: 'The commit hash you want to build',
            type: String,
            default_value: 'HEAD'
            conflicting_options: [:branch]
          ),
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
```diff
@@ -0,0 +1,90 @@
...
+          FastlaneCore::ConfigItem.new(
+            key: :pipeline_file,
+            description: 'The name of the pipeline file in the project',
+            type: String
```
AliSoftware, line 70: Wouldn't it make sense to default this to `pipeline.yml`?
Or alternatively (and might make better sense), default it to `nil`, then handle the `nil` case on L9-L11 (i.e.
`pipeline_name = params[:pipeline_file] ? { PIPELINE: params[:pipeline_file] } : {}`)?

AliSoftware (commented):
Agree with most of @mokagio feedback, and adding my own notes and suggestions.

PS: I looked at https://docs.fastlane.tools/plugins/available-plugins/ and couldn't find an existing action with "buildkite" in its name either 🤷

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
mokagio: Update: It does not.

image

image

In hindsight, that makes total sense! Buildkit is a wrapper around the Buildkite APIs, for it to expect to be called from a file system location that includes the pipeline file it's asked to run would be reductive.

This actually makes me feel better about my earlier remarks about wrapper wrapping wrappers. Given the different of purpose, we can specialize our own wrapper to add those expectations and make it worthwhile.

I think it would be fair for our action to enforce that the pipeline file exists. Fastlane actions are most usually used within the context of a single project.

mokagio: Update: It does not.

image

image

In hindsight, that makes total sense! Buildkit is a wrapper around the Buildkite APIs, for it to expect to be called from a file system location that includes the pipeline file it's asked to run would be reductive.

This actually makes me feel better about my earlier remarks about wrapper wrapping wrappers. Given the different of purpose, we can specialize our own wrapper to add those expectations and make it worthwhile.

I think it would be fair for our action to enforce that the pipeline file exists. Fastlane actions are most usually used within the context of a single project.

mokagio: > I think it would be fair for our action to enforce that the pipeline file exists. Fastlane actions are most usually used within the context of a single project.

I guess there's two routes we can take:

1. We could work on this as a super-flexible Fastlane-Buildkite integration to possibly extract as its own plugin. If we do that, then the check for the pipeline file existence in the local folder might limit the flexibility.
2. We could double down on using it constrained to a single project. That wouldn't preclude us from bundling as a dedicated action, but might make it less appealing (?) because of its assumption. In that case, IMHO, checking the pipeline file existence would be appropriate.

Gemfile.lock
jkmassel: I tried reverting it, but `bundle` insisted on it 🤔 

I guess this'll just be a heisendependency until the bundler folks sort this out?

fastlane-plugin-wpmreleasetoolkit.gemspec
jkmassel: Good call on these, thanks!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: Ya this'll be nice to have the extra info!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: So...this is tricky, because this is an implementation detail that's specific to how we do things. To the best of my knowledge, there's no global way to say "hey Buildkite, run this specific pipeline", so instead, our default project configuration does `buildkite-agent pipeline upload ./.buildkite/${PIPELINE:-pipeline.yml}`, so the `PIPELINE` environment variable determines what's run, if it's present.

We could try encoding this into the action, but then we'd need to replicate Buildkite's search structure (it searches the project root, the `.buildkite` directory, and possibly other places) in order to validate this. 

While I like the idea of helping the developer with this, I wonder if we're over-replicating work Buildkite already does (after all, IMHO the error message Buildkite emits that you posted above is fairly descriptive of the problem, and makes it pretty easy to see where you went wrong).

That said, I don't feel super strongly about it either, so interested to hear your thoughts 😄

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: Solid point! :)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: Yep makes sense – since we'll all have to modify our local machines this is a good time to change things.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: TIL!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: Yep – was back and forth on this because [the current one works this way](https://github.com/wordpress-mobile/release-toolkit/blob/b533fe2376454931c3196dddcfb672b3c1e329c8/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/circleci_trigger_job_action.rb#L33-L37).

FWIW this is for the Buildkite account, not for the GitHub org, but still, this kinda pushes me over the edge to say "let's manually specify it".

Gemfile.lock
mokagio: There's no 🤷‍♂️ -reaction, so I'll just 🤷‍♂️ -comment here.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: Ah this is better

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: [Both are required by the Buildkite API](https://buildkite.com/docs/apis/rest-api/builds#create-a-build), so they can't be mutually exclusive.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: I don't think so, because it'd be very rare that we'd want to run `pipeline.yml` using a webhook – instead we'd want to use a "special" one to do API tasks (otherwise we'd end up with a ton of conditionals in our `pipeline.yml`, when IMHO separating release stuff out into its own keeps things clean.

Open to having my mind changed here though 🙂

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: One day I'll replace this correctly when duplicating a file 🤦

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
AliSoftware: Oh. How strange… I wonder how Buildkite reacts then if you provide a commit sha1 that is not on the branch you also provide but on a different branch instead… 🤔  (My guess would be: it doesn't care and use the commit to do the checkout, and the branch name is just there for branch filtering…?)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
mokagio: > So...this is tricky, because this is an implementation detail that's specific to how we do things. To the best of my knowledge, there's no global way to say "hey Buildkite, run this specific pipeline", so instead, our default project configuration does `buildkite-agent pipeline upload ./.buildkite/${PIPELINE:-pipeline.yml}`, so the `PIPELINE` environment variable determines what's run, if it's present.

How. Did. I. Miss. That?!

Okay cool, then that solves it for me. This is leaning way more towards a custom bit of tooling, for which we don't need to worry as much about flexibility but we can optimize to work well for us.

> We could try encoding this into the action, but then we'd need to replicate Buildkite's search structure (it searches the project root, the .buildkite directory, and possibly other places) in order to validate this.

Does it? Don't _we_ pass an exact path to `buildkite-agent pipeline upload`?

If my assumption is correct, we _could_ update our command in the Buildkite UI to be like:

```sh
buildkite-agent pipeline upload ${PIPELINE:-./.buildkite/pipeline.yml}
```

And then pass a proper path there.

But... I think, in particular because of the "internal tool" ramble above, that we can easily live with `pipeline_file` backing in certain assumptions specific to our coding standards and maybe compensate for it in the docs.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
AliSoftware: No that makes total sense, and I am all for the idea of having separate `yml` files for API tasks and release builds (rather than tons of conditional), so 👌

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
mokagio: > FWIW this is for the Buildkite account, not for the GitHub org, but still, this kinda pushes me over the edge to say "let's manually specify it".

Oh, right 🤦 

That changes things a bit, actually 🤔 I'm not so strongly for removing it. But yeah, an extra line of verbosity won't hurt.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
mokagio: Looks like Buildkite prioritizes the commit and... just trust you know the branch exists?

image

I started it with:

```
➜ bf run buildkite_trigger_build buildkite_organization:automattic \
  buildkite_pipeline:wordpress-ios \
  branch:will-this-be-ignored-question-mark \
  commit:b461d3c3eff4aba1cf91ca0482e7c05bfc9454e1 \
  pipeline_file:pipeline.yml
```

I should have used a commit different from the tip of the branch it belongs to, but the test turned out interesting anyway 🤔 

I wonder if that's a bug in the API?

AliSoftware: Well the docs Jeremy linked to says:
> [Branch]: Branch the commit belongs to. This allows you to take advantage of your pipeline and step-level branch filtering rules.

And given a commit can belong to multiple branches (example case being a train of PRs, with branch B built on top of branch A, so commit at tip of branch A would be in both A and B), `git branch -a --contains <sha1>` does not always return a single result, so I guess that makes sense for Buildkite to request us to tell which branch this is supposed to be triggered from in order to apply the correct branch filtering rules accordingly 🤷

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: Yeah, in this case _we_ have a specific way of doing it, but IIRC Pocket Casts is one example of a non-standard setup (non-standard == "not like our other projects"). If we were going to add support for _any_ Buildkite project we'd need to handle all of the file paths listed on https://buildkite.com/docs/agent/v3/cli-pipeline#uploading-pipelines and that seems like a lot of work.

I think we are in agreement! 😃

jkmassel:
Oh – also just commenting to note that I found the gif very funny 😃

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
jkmassel: This is probably the correct answer, because it seems weird otherwise!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
mokagio: > but IIRC Pocket Casts is one example of a non-standard setup (non-standard == "not like our other projects")

I updated Pocket Casts to follow the Automattic way™ not long ago 🤓 👌

mokagio (approved):
:shipit:

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb
AliSoftware: Ah! I was actually wondering how passing the `PIPELINE` as parameter here was working with the `pipeline-upload` step (which is hardcoded in the Buildkite project's web UI' settings, not in files in the git repo) and by what magic would Buildkite use this `PIPELINE` var.

I initially first thought that this was kind of a built-in an magic variable that Buildkite would know about and would in such case skip the `pipeline-upload` step defined in web settings and use the custom pipeline instead, which felt a bit too magical. The fact that it's all us explicitly using `${PIPELINE:-pipeline.yml}` makes way more sense, and I like that because it's way more explicit as well.

I completely missed that tbh. (I'm not sure if it you recently updated the pipeline setting on the web UI to change hardcoded `pipeline.yml` into `${PIPELINE:-pipeline.yml}` in order to support that PR, or if it was already there before?) A small nit I'd suggest then would be to name this env var something more explicit not that I understand that we do have full control over it, like `BUILDKITE_PIPELINE_FILE` rather than just `PIPELINE`?

PR #334, opened 2022-02-04
Release 3.0.0 into trunk

mokagio:
New version 3.0.0. 

Be sure to create a GitHub Release and tag once this PR gets merged.

---

I'd like to release a new version so that we can [use the `buildkite_trigger_build` action in WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/17893) and ditch CircleCI for releases, ideally starting in 19.2 🤞 ⏱️ 

I haven't enabled auto-merge so I can merge, tag, push, and release once approved.

PR #336, opened 2022-02-07
Remove custom check for CI in favor of using Fastlane's one

mokagio:
Addresses a small `TODO` comment as we prepare for a new release. See also https://github.com/wordpress-mobile/release-toolkit/pull/334#issuecomment-1029969292.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
@@ -10,10 +10,7 @@ module Fastlane
       end
 
       def source_contents
-        # TODO: This works only on CircleCI. I chose this variable because it's the one checked by Fastlane::is_ci.
-        # Fastlane::is_ci doesn't work here, so reimplementing the code has been necessary.
-        # (This should be updated if we change CI or if fastlane is updated.)
-        return File.read(secrets_repository_file_path) unless self.encrypt || ENV.key?('CIRCLECI')
+        return File.read(secrets_repository_file_path) unless self.encrypt || FastlaneCore::Helper.is_ci?
         return nil unless File.file?(encrypted_file_path)
 
         encrypted = File.read(encrypted_file_path)
--- a/spec/file_reference_spec.rb
+++ b/spec/file_reference_spec.rb
@@ -69,19 +69,17 @@ describe Fastlane::Configuration::FileReference do
 
     describe '#source_contents' do
       it 'gets the contents from the secrets repo' do
-        set_circle_env(false) do
-          allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
-          expect(subject.source_contents).to eq('source contents')
-        end
+        allow(FastlaneCore::Helper).to receive(:is_ci?).and_return(false)
+        allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
+        expect(subject.source_contents).to eq('source contents')
       end
     end
 
     describe '#source_contents on ci' do
       it 'gets the contents from the secrets repo' do
-        set_circle_env(true) do
-          allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
-          expect(subject.source_contents).to eq(nil)
-        end
+        allow(FastlaneCore::Helper).to receive(:is_ci?).and_return(true)
+        allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
+        expect(subject.source_contents).to eq(nil)
       end
     end
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -25,24 +25,6 @@ RSpec.configure do |config|
   config.filter_run_when_matching :focus
 end
 
-def set_circle_env(define_ci)
-  is_ci = ENV.key?('CIRCLECI')
-  orig_circle_ci = ENV['CIRCLECI']
-  if define_ci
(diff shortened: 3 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/models/file_reference.rb
```diff
@@ -10,10 +10,7 @@ def initialize(params = {})
       end
 
       def source_contents
-        # TODO: This works only on CircleCI. I chose this variable because it's the one checked by Fastlane::is_ci.
-        # Fastlane::is_ci doesn't work here, so reimplementing the code has been necessary.
-        # (This should be updated if we change CI or if fastlane is updated.)
-        return File.read(secrets_repository_file_path) unless self.encrypt || ENV.key?('CIRCLECI')
+        return File.read(secrets_repository_file_path) unless self.encrypt || FastlaneCore::Helper.is_ci?
```
AliSoftware, line 13: Any idea why the comment initially said that "`Fastlane::is_ci?` doesn't work here "?

mokagio, line 13: See `is_ci?` source [here](https://github.com/wordpress-mobile/release-toolkit/pull/334#issuecomment-1029969292).

To test this change, I run the tests as usual first, the re-run them for this file with `CI=1`:

```
CI=1 bundle exec rspec spec/file_reference_spec.rb
```

I got the following failure:

```
  1) Fastlane::Configuration::FileReference without encryption #source_contents gets the contents from the secrets repo
     Failure/Error: expect(subject.source_contents).to eq('source contents')

       expected: "source contents"
            got: nil

       (compared using ==)
```

That's exactly the failure I would expect to get: `is_ci?` returned `true` and we didn't read from the stubbed local secrets but tried to read the encrypted source, which didn't exist.

mokagio: I'm pretty sure the reason is that `is_ci` is [a Fastlane action](https://github.com/fastlane/fastlane/blob/e874a47c6e2e0e61590a03d3b71e75e5a505d1ce/fastlane/spec/actions_specs/is_ci_spec.rb) and can't be called directly.

If we put it in here, we get a no method error.

AliSoftware: Ah, right. As opposed to calling the helper directly as you did here 👍👌

AliSoftware: That also means that the test is failing on CI though 😓😛

mokagio: 🤦‍♂️

mokagio: 🤦‍♂️  is not enough of a reaction...

AliSoftware: image

spec/file_reference_spec.rb
```diff
@@ -69,19 +69,17 @@
...
 
     describe '#source_contents on ci' do
       it 'gets the contents from the secrets repo' do
-        set_circle_env(true) do
-          allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
-          expect(subject.source_contents).to eq(nil)
-        end
+        allow(FastlaneCore::Helper).to receive(:is_ci?).and_return(true)
+        allow(File).to receive(:read).with(subject.secrets_repository_file_path).and_return('source contents')
+        expect(subject.source_contents).to eq(nil)
```
mokagio, line 82: Having to work on this made me wonder whether it is appropriate for `FileReference` to know whether its code runs on CI or not. 

I was tempted to leave a `FIXME` mentioning this and suggesting to move the knowledge up the abstraction ladder, but I didn't because:

1. `FileReference` is a fat model, not a mere data representation objects. For example, it knows how to diff and apply changes to the file system
2. This objects is used for the secrets management and, in the long run, we want to use the [Rust implementation](https://github.com/automattic/configure) so I wouldn't recommend working on stuff that is not a necessary bug fix

mokagio:
@wpmobilebot 

> Please add an entry in the `CHANGELOG.md` file to describe the changes made by this PR

Ignoring because this is just a refactor of an internal method.

spec/file_reference_spec.rb
AliSoftware: Yeah I've never liked the way that this was all hosted in FileReference, this would definitively benefit some restructuration and refactoring… but that's something to keep for later™

AliSoftware:
> Ignoring because this is just a refactor of an internal method.

To be fair one of the goal of the CHANGELOG is also to be able to, if something stops working with secrets and configure_apply soon after the release, be aware that this might be because of that change here. So that users who notice such a change can look into the CHANGELOG and say "oh, something related to this changed in 3.0.0 so that might be it" and be able to jump in the PR which did the change… 🙃  (I know, we can also use `git blame` and `git bisect`, but `CHANGELOG` is faster to identify when key stuff changed).

I don't mind skipping it for this one (that's not really a big deal anyway tbh), but good to keep in mind for the future :)

mokagio:
@AliSoftware should I consider your last comment an implicit approval? If so, feel free to hit merge when you read this 😄 

Otherwise, let me know what I missed. Thanks.

AliSoftware (approved):
Oh yeah, sorry. Approved 👍

PR #337, opened 2022-02-07
Remove deprecated, unused actions

mokagio:
A little cleanup in preparation for the 3.0.0 release, see https://github.com/wordpress-mobile/release-toolkit/pull/334#issuecomment-1031184199.

All the removed actions were unused in our projects (WordPress, WooCommerce, Simplenote, and PocketCasts. I don't have access to Day One and AFAIK Tumblr doesn't yet use the release toolkit), or used only in Simplenote Android which is currently paused.

There's another deprecated action still in the codebase: [`ios_localized_project`](https://github.com/wordpress-mobile/release-toolkit/blob/c198441d2050c52714d40c8791d0020bd1c4f752/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_localize_project.rb#L31). It's used in WooCommerce (e.g. [here](https://github.com/woocommerce/woocommerce-ios/blob/81ed7b18078c58d3337f53a91d0f09c802bf3c3f/fastlane/Fastfile#L122)) and Simplenote (which is less pressing because that's on pause).

I've decided not to remove it to make the 3.0.0 adoption easier. I think Buildkite trigger support and `trunk` as default are useful things that we should rollout ASAP, and wouldn't want the friction —hoverer small— of addressing the `ios_localized_project` deprecation to block that.


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_update_metadata.rb
```diff
@@ -1,52 +0,0 @@
...
-  module Actions
-    module SharedValues
-      ANDROID_UPDATE_METADATA_CUSTOM_VALUE = :ANDROID_UPDATE_METADATA_CUSTOM_VALUE
-    end
-
-    class AndroidUpdateMetadataAction < Action
-      def self.run(params)
-        require_relative '../../helper/android/android_git_helper'
-
-        Fastlane::Helper::Android::GitHelper.update_metadata(ENV['validate_translations'])
```
AliSoftware, line 11: The corresponding method in the `android_git_helper.rb` should also be removed in parallel to removing the action.

Note that this helper method is also marked as `@todo` so the fact that you missed it makes me wonder if you used case-sensitive search when looking for TODOs/FIXMEs. We should definitively try to be consistent in using `@TODO`/`@FIXME` in all uppercase everywhere (I _think_ that's the ruby convention?) but 🤷‍♂️ 

---

PS: As a bonus it would be nice to remove the `./tools/update-translations.sh` script (which is the script the `update_matadata` helper ends up calling) from client repos in parallel to that PR

AliSoftware (dismissed):
Don't forget to remove the helper method associated with the action as well 🙃

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_update_metadata.rb
mokagio: Good catch, thanks. Addressed in [`fae13ff` (#337)](https://github.com/wordpress-mobile/release-toolkit/pull/337/commits/fae13ff5e44a2eb173da48bcafe7a9a8c2255b1e)

> As a bonus it would be nice to remove the `./tools/update-translations.sh` script 

Bonus points to @JavonDavis for doing that a while back:

- https://github.com/woocommerce/woocommerce-android/commit/7fec286c18#diff-0baa79b5bfd3b41d3e2b326f974f71a14a7c91b4a0b01bad6c55ef882912f64f
- https://github.com/wordpress-mobile/WordPress-android/commit/28af2e3823#diff-0baa79b5bfd3b41d3e2b326f974f71a14a7c91b4a0b01bad6c55ef882912f64f

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -28,23 +28,6 @@ def self.commit_version_bump(include_deliverfile: true, include_metadata: true)
...
-        # Calls the `Scripts/localize.py` script in the project root folder and push the `*.strings` files
-        #
-        # That script updates the `.strings` files with translations from GlotPress.
-        #
-        # @env PROJECT_ROOT_FOLDER The path to the git root of the project
-        # @env PROJECT_NAME The name of the directory containing the project code (especially containing the `build.gradle` file)
-        #
-        # @deprecated This method is only used by the `ios_localize_project` action, which is itself deprecated
-        #             in favor of the new `ios_generate_strings_file_from_code` action
-        # @todo [Next Major] Remove this method once we fully remove `ios_localize_project`
```
AliSoftware: Isn't `ios_localize_project` still used though (not for long, but until we finish our work on the Localization Tooling Modernization and finally migrate all client apps to the new actions)?

> There's another deprecated action still in the codebase: [ios_localized_project](https://github.com/wordpress-mobile/release-toolkit/blob/c198441d2050c52714d40c8791d0020bd1c4f752/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_localize_project.rb#L31). It's used in WooCommerce (e.g. [here](https://github.com/woocommerce/woocommerce-ios/blob/81ed7b18078c58d3337f53a91d0f09c802bf3c3f/fastlane/Fastfile#L122)) and Simplenote (which is less pressing because that's on pause).

mokagio: Ouch! I'm so glad we have code review 😓 

I _did_ run the tests after removing this, but I guess it was a naive of just being happy with that given how old this code is.

AliSoftware (approved):
All good now 🎉 :shipit:

PR #338, opened 2022-02-10
Introduce ios_extract_keys_from_strings_files action

AliSoftware:
This is part of parent project paaHJt-2Ib-p2, especially tackling the last part of [i3] and [i6].

## What

This PR introduces the new `ios_extract_keys_from_strings_files` fastlane action.

## Why

This action will be used right after we download the translations back from GlotPress, to extract the relevant translations subset from the various downloaded `*.lproj/Localizable.strings` files into the `InfoPlist.strings` files (and `Sites.strings` files for the Siri intent definition target).

---

<details><summary>Obsolete notes from previous version of this PR</summary>

### Handle multiple outputs (`target_tablenames`) in parallel in one run

I've considered allowing that single action to handle multiple output files via a single pass, which would allow us to call this action only once for both `InfoPlist.strings` and `Sites.strings`. This first implementation can only tackle one output table at a time (which suggest we'd call that action twice from our WPiOS `Fastfile`), but I'm willing to give it a try tomorrow to make it possible to pass multiple output filenames at once, which would simplify the call site and also improve performance (by parsing the input files only once)

### Test cases

I'm open to ideas for more test cases, if you think of some edge cases, please share ideas!

### Alternative action names?

I'm also open to alternative names for this action.
 - I hesitated naming it `ios_extract_translations_from_strings_files` but that would suggest an action that would only extract the translations without the keys or something.
 - I went with "extract keys" because the idea is to only extract a specific subset of specific keys (determined dynamically by looking at the keys present in the original locale), but not sure that's super clear.
 - Also, `extract` might suggests that the keys would be removed from original file (`Localizable.strings`)? – while it's not the case as we don't want risking touching anything in an OpenStep/ASCII file…

Naming is hard!

</details>


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,99 @@
...
+                                       type: String,
+                                       default_value: 'Localizable'),
+          FastlaneCore::ConfigItem.new(key: :target_tablename,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_TARGET_TABLENAME',
+                                       description: 'The basename of the `.strings` file (without the extension) to extract the translations subset to',
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :base_locale,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_BASE_LOCALE',
+                                       description: 'The basename of the `.lproj` directory which acts as reference/originals (e.g. `en-US` or `Base`). Used to determine which entries to extract, by looking at the keys present in `<base_locale>.lproj/<<target_tablename>.strings`,
+                                       type: String),
```
AliSoftware: I considered providing a `default_value` here but figured that `en`, `en-US` and `Base` would all be valid default values, so decided to not provide any and enforce the call site to be explicit.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,99 @@
...
+          FastlaneCore::ConfigItem.new(key: :source_tablename,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_SOURCE_TABLENAME',
+                                       description: 'The basename of the `.strings` file (without the extension) to extract the keys and translations from',
+                                       type: String,
+                                       default_value: 'Localizable'),
+          FastlaneCore::ConfigItem.new(key: :target_tablename,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_TARGET_TABLENAME',
+                                       description: 'The basename of the `.strings` file (without the extension) to extract the translations subset to',
+                                       type: String),
```
AliSoftware: Improvement idea that I'll try tomorrow: make this be able to also accept an Array of Strings to extract keys in multiple output files (`InfoPlist.strings` + `Sites.strings`) in parallel, without having to invoke the action twice (which would also make it parse the input files twice, increasing BigO complexity overall)

spec/ios_extract_keys_from_strings_files_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'spec_helper'
+require 'tmpdir'
+
+describe Fastlane::Actions::IosExtractKeysFromStringsFilesAction do
+  let(:test_data_dir) { File.join(File.dirname(__FILE__), 'test-data', 'translations', 'ios_extract_keys_from_strings_files') }
+
+  it 'extracts the right keys from Localizable.strings to InfoPlist.strings for all locales but the base one' do
```
AliSoftware: I only have that single test for now… if you have ideas for other test cases to add, I'd gladly welcome them 🙂

spec/test-data/translations/ios_extract_keys_from_strings_files/source/en.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,2 @@
+NSLocationUsageDescription = "WordPress would like to add your location to posts on sites where you have enabled geotagging.";
+NSCameraUsageDescription = "To take photos or videos to use in your posts.";
```
AliSoftware: Notice how the keys are not quoted here. This is totally valid to not quote the keys in the OpenStep `.strings` file format (very similar to quotes being optional in YAML if there's no ambiguity or special characters in the string to represent), and in fact quite common to not quote them for `InfoPlist.strings` (because the string keys have to match keys in the `Info.plist` file/dictionary anyway, so they never contain special characters or spaces or whatnot anyway).

This is a nice bonus to have them unquoted here, so that we double-validate that our helper methods reading those files (esp. `read_strings_file_as_hash`) do support those cases as well (while I'm not sure custom RegExes or libs like `apfel` or `apstrings` would support that, for example…)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,99 @@
...
+            target_strings_file = File.join(lproj_dir, target_strings_filename)
+            UI.message("Extracting #{keys_to_extract.count} keys into #{target_strings_file}...")
+
+            translations = Fastlane::Helper::Ios::L10nHelper.read_strings_file_as_hash(path: source_strings_file)
+            translations.slice!(*keys_to_extract) # only keep the keys/translations we want to extract
+            Fastlane::Helper::Ios::L10nHelper.generate_strings_file_from_hash(translations: translations, output_path: target_strings_file)
+          rescue StandardError => e
+            UI.error("Error while extracting keys from #{source_strings_file} into #{target_strings_file}: #{e.message}")
+          end
+        end
```
AliSoftware: I've implemented the full logic directly in the action for now, because even if it might look a tad long, most of those are about declaring intermediate variables or rescue blocks which is mostly what makes this 28 lines long, so I'm not sure there would be much sense to try and extract that in helper methods (where would we split the implementation into smaller pieces? Would that make it smaller anyway?).

Maybe I will end up moving part of that into various methods in the helper after I tackle the "support multiple output table names" improvement tomorrow and that ends up making the code more complicated, we'll see. But overall, it might also be beneficial to keep the loop logic directly in the action for readability rather than having to jump back and forth between action and helper…? 🤷

spec/ios_extract_keys_from_strings_files_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    Dir.mktmpdir('a8c-release-toolkit-tests-') do |tmp_dir|
```
mokagio: FYI, we also have the following helper that saves some typing:

https://github.com/wordpress-mobile/release-toolkit/blob/06d9f926f2983d7c0b00b414746c83eff5b5b5f4/spec/spec_helper.rb#L70-L77

I'm fine using either one or the other. Calling `Dir.mktmpdir` removes one abstraction layer, but I find the helper makes the code less noisy. 🤷 ⚖️

spec/ios_extract_keys_from_strings_files_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      # Arrange
+      lproj_source_dir = File.join(tmp_dir, 'Resources')
+      FileUtils.cp_r(File.join(test_data_dir, 'source/'), lproj_source_dir) # slash at end of `source/` makes sure we copy the dir content, not the dir itself
```
mokagio: What do you think of this?

```suggestion
      FileUtils.cp_r(File.join(test_data_dir, 'source', '.'), lproj_source_dir)
```

spec/test-data/translations/ios_extract_keys_from_strings_files/source/en.lproj/InfoPlist.strings
mokagio: Nice!

What do you think of adding this explanation in the file itself?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,99 @@
...
+                                       description: 'The basename of the `.strings` file (without the extension) to extract the keys and translations from',
+                                       type: String,
+                                       default_value: 'Localizable'),
+          FastlaneCore::ConfigItem.new(key: :target_tablename,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_TARGET_TABLENAME',
+                                       description: 'The basename of the `.strings` file (without the extension) to extract the translations subset to',
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :base_locale,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_BASE_LOCALE',
+                                       description: 'The basename of the `.lproj` directory which acts as reference/originals (e.g. `en-US` or `Base`). Used to determine which entries to extract, by looking at the keys present in `<base_locale>.lproj/<<target_tablename>.strings`',
```
mokagio:
```suggestion
                                       description: 'The basename of the `.lproj` directory which acts as reference/originals (e.g. `en-US` or `Base`). Used to determine which entries to extract, by looking at the keys present in `<base_locale>.lproj/<target_tablename>.strings`',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
mokagio: That's fair. 👍

spec/ios_extract_keys_from_strings_files_spec.rb
mokagio: The only suggestion I have is for testing the unhappy paths. Namely, verify that, under the right conditions, the expected errors are raised:

- `UI.user_error!("Failed to read the keys to extract from originals file #{target_table_originals}: #{e.message}")`
- `UI.error("Error while extracting keys from #{source_strings_file} into #{target_strings_file}: #{e.message}")`

mokagio (commented):
Looking good 👌 

> Naming is hard!

Yes 😄 

You point out that "extract" might confuse the reader into thinking the keys are removed from the source file. How about the verb "port"?

`ios_port_keys_from_strings_files`? 

Or just `ios_port_keys` and rely on the required arguments to tell the rest of the story. Although maybe that's not such a good idea because one can scramble the order of an action  arguments (given they're in an `Hash`). Also, keys just like that might be too generic; it's not clear they're localization keys.

> Handle multiple outputs (`target_tablenames`) in parallel in one run

I like the idea of parallelizing. In fact, I've been meaning to do a bit of research and open a request to parallelize the GlotPress `.strings` downloads. I wonder how Fastlane deals with stuff running in parallel 🤔 

You mentioned you'll try that for this code soon. I'd say not to sweat it too much, it can always come as a followup in a later release.

spec/test-data/translations/ios_extract_keys_from_strings_files/source/en.lproj/InfoPlist.strings
AliSoftware: Good point 👍 

Not sure if it's best to add the comment in the fixture file vs in the spec, but will definitively comment somewhere 👍

spec/ios_extract_keys_from_strings_files_spec.rb
AliSoftware: Ohhh I always forgot about this one, good point 👍 

I might do a separate PR to update all the `Dir.mktmpdir` calls in other specs as well to update them all to `in_tmp_dir` for consistency 🤔

AliSoftware:
> You point out that "extract" might confuse the reader into thinking the keys are removed from the source file. How about the verb "port"?
> 
> `ios_port_keys_from_strings_files`?
> 
> Or just `ios_port_keys` and rely on the required arguments to tell the rest of the story. Although maybe that's not such a good idea because one can scramble the order of an action arguments (given they're in an `Hash`). Also, keys just like that might be too generic; it's not clear they're localization keys.

Also, maybe it's because I'm not a native speaker, but the verb "port" doesn't really speak to me and feels it might mean multiple things (and be either a verb or a noun…).

And also, `ios_port_keys_…` might be a bit too Harry-Potter-y 😂 
portkey

---

What about simply `copy`?

Though, then, `copy_keys_from_strings_file` fails to suggest that it will only extract a _subset_ of the keys, suggesting it will copy all of them…? 🤔

mokagio:
I can't give any more thoughts on this because I'm too busy laughing about Harry Potter portkeys 🤣

mokagio:
> Though, then, `copy_keys_from_strings_file` fails to suggest that it will only extract a subset of the keys, suggesting it will copy all of them…? 🤔

`copy_keys_subset_from_strings_file`? Bit verbose but clearer? Then a user can read the docs to understand how the subset is evaluated?

AliSoftware:
> `copy_keys_subset_from_strings_file`? Bit verbose but clearer? Then a user can read the docs to understand how the subset is evaluated?

Yeah… clearer but verbose and quite a mouthful. Thinking about it I am starting to circle back in liking `extract` (and then a user can read the docs to understand that it doesn't remove them from the source… 😛  – and who knows, maybe in the future that action could improve to actually remove them if we finally find a decent and safe way to do so confidently… 😄 ).

The thing I like about "extract" is mostly because it reflects _why_ we're doing this and calling this action rather than _how_. The keys we're interested in for `InfoPlist.strings` are in another `.strings` file and we want to _extract_ them from `Localizable.strings` in order to put them in `InfoPlist.strings`. Also "extract" kind of already contains the notion of "subset" to me (as in "extract the juice from this lemon" 🍋 😂 )

In a way, the fact that they are not deleted from the source is kind of an implementation detail (actually, it exactly is, because the reason we don't delete them is we don't want to risk messing up the `.strings` file in OpenStep format, otherwise we'd probably have removed them…)

Anyway, I'll sleep on that "naming" part and let the weekend pass and see how I feel about it when I jump back on this next week 😃

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
AliSoftware: Implemented in e9c57626545884901a97ac881af9da60391adedc — alongside the possibility for the target files (e.g. `InfoPlist.strings` and `Sites.strings` to be in completely different folders (e.g. `WordPress/Resources/` vs `WordPress/WordPressIntents/`) — which I didn't anticipate / thought about at first.

spec/ios_extract_keys_from_strings_files_spec.rb
AliSoftware: Added a lot of tests now. Some are still unimplemented yet, but at least I have the ideas now 🙂

spec/test-data/translations/ios_extract_keys_from_strings_files/source/en.lproj/InfoPlist.strings
AliSoftware: Addressed in 7b6b36445f0a9eeaca82540530525275ea9c17e8

spec/ios_extract_keys_from_strings_files_spec.rb
AliSoftware: Addressed at least for this new action in 47d1e39816e849ed59fa089ae3b5ac0b22cf5576

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
AliSoftware: Discarding this suggestion as the implementation has changed since, grouping the target tablename and the base locale together in a single param expecting the path to the original (e.g. `en.lproj/InfoPlist.strings`), which allowed me to make a nicer API to support having multiple target files in parallel

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,131 @@
...
+                                       type: String,
+                                       verify_block: proc do |value|
+                                         UI.user_error!("`source_parent_dir` should be a path to an existing directory, but found #{value}.") unless File.directory?(value)
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :source_tablename,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_SOURCE_TABLENAME',
+                                       description: 'The basename of the `.strings` file (without the extension) to extract the keys and translations from',
+                                       type: String,
+                                       default_value: 'Localizable'),
+          FastlaneCore::ConfigItem.new(key: :target_original_files,
```
AliSoftware, line 83: Ok after the week-end I think I'm ok with the action name (`extract_keys_from_strings_files`) after all… but now my next naming bikeshed is about this one.

The idea is to have a list of `<path>/<base_locale>/<target_table>.strings` paths, which, for each target path, will convey the information of { the relative path of the parent dir containing the `.lproj` for the _target_ files (which might not be the same as the source one) + the base locale + the target table name } all in a single string for each item of the array (as opposed to having something like an Array of Hashes).

_For example when we'll want to extract the keys from `Localizable.strings` (downloaded from GlotPress) into both `InfoPlist.strings` and `Sites.strings`, because those are for different Xcode targets and in different folders (`WordPress/Resources` vs `WordPress/WordPressIntents`, respectively), we need, for each target table, to provide both the path to the parent dir containing the `.lproj`, and the basename of the table._

---

❓  This approach is a nicer and more flexible API than the previous one I drafted imho… but I'm unsure if the name of the parameter (`target_original_files:`) is the base name we could come up then. I like the `target_*` prefix as it makes for a nice counterpart to `source_*` for the other params, but not sure if the name is clear enough. Maybe `target_files_originals` would be better, maybe something else… 🤷

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -108,7 +108,7 @@ def self.generate_strings_file_from_hash(translations:, output_path:)
             xml.comment('Warning: Auto-generated file, do not edit.')
             xml.plist(version: '1.0') do
               xml.dict do
-                translations.each do |k, v|
+                translations.sort.each do |k, v| # NOTE: use `sort` just in order to be deterministic over various runs
```
AliSoftware, line 111: ℹ️  Being deterministic is important both to ensure the Unit Tests will always pass regardless of the order the `Hash` is iterated in… but also in normal run and usage of the action, to ensure that the action is idempotent and won't risk creating useless changes in `git diff` between every runs, making the PRs harder to review.

spec/ios_extract_keys_from_strings_files_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  describe 'input parameters validation' do
+    it 'errors if the source dir does not exist'
+    it 'errors if one of the target original files does not exist'
+  end
+
+  describe 'error handling during processing' do
+    it 'errors it if fails to read the keys to extract'
+    it 'errors it if fails to write one of the target files'
+  end
```
AliSoftware: ~🔜  I'll implement those tomorrow.~ [EDIT] Done.

AliSoftware:
@mokagio PR is now ready for review.

I added a lot more error handling and tests, support for parallel extraction, and support for target files that are in a different folder (e.g. `WordPress/WordPressIntent/`) than the source (e.g. `WordPress/Resources`). Should hopefully now cover all the cases we'll need in client apps (at least WPiOS).

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,119 @@
...
+                                       description: 'The basename of the `.strings` file (without the extension) to extract the keys and translations from for each locale',
+                                       type: String,
+                                       default_value: 'Localizable'),
+          FastlaneCore::ConfigItem.new(key: :target_original_files,
+                                       env_name: 'FL_IOS_EXTRACT_KEYS_FROM_STRINGS_FILES_TARGET_ORIGINAL_FILES',
+                                       description: 'The path to the `<base-locale>.lproj/<target-tablename>.strings` file(s) for which we want to extract the keys to. ' \
+                                        + 'Those files, containing the originals (typical `en` or `Base` locale), will be used to determine which keys to extract from the `source_tablename`, ' \
+                                        + 'and the target file path(s) in which to extract the translations will be determined from those path, writing them in the file with the same basename ' \
+                                        +' in each of the other `*.lproj` sibling folders of those target paths',
```
AliSoftware: This feels both a bit wordy and not super easy to grasp, but also I've had a hard time to find a clear yet detailed description that would explain this parameter more easily.
It's also late so maybe my brain is having a hard time finding the right words to provide a nice and clear explanation…

TL;DR: any suggestion on how to improve the description/documentation of this parameter (while still make it clear how it works) is welcome.

spec/ios_extract_keys_from_strings_files_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+require 'tmpdir'
+
+describe Fastlane::Actions::IosExtractKeysFromStringsFilesAction do
+  let(:test_data_dir) { File.join(File.dirname(__FILE__), 'test-data', 'translations', 'ios_extract_keys_from_strings_files') }
+
+  describe 'extract the right keys from `Localizable.strings` to all locales' do
+    def assert_output_files_match(expectations_map)
+      expectations_map.each do |output_file, expected_file|
+        expect(File).to exist(output_file), "expect `#{output_file}` to exist"
+        expect(File.read(output_file)).to eq(File.read(File.join(test_data_dir, expected_file))), "expect content of `#{output_file}` to match `#{expected_file}`"
```
mokagio: Nitpick. When those strings are shown, the error already occurred.

```suggestion
        expect(File).to exist(output_file), "expected `#{output_file}` to exist"
        expect(File.read(output_file)).to eq(File.read(File.join(test_data_dir, expected_file))), "expected content of `#{output_file}` to match `#{expected_file}`"
```

RSpec uses the past tense, too. E.g. [here](https://github.com/rspec/rspec-core/blob/ea8554afd1a2b63677c6593059fa8f2476181deb/spec/support/matchers.rb#L13) and [here](https://github.com/rspec/rspec-core/blob/ea8554afd1a2b63677c6593059fa8f2476181deb/spec/support/matchers.rb#L30).

spec/ios_extract_keys_from_strings_files_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+            source_parent_dir: tmp_dir,
+            target_original_files: misleading_target_file
+          )
+        end.to raise_error(FastlaneCore::Interface::FastlaneError, "Expected `#{misleading_target_file}` (found in `target_original_files`) to be a path ending in a `*.lproj/*.strings`.")
+      end
+    end
+  end
+
+  describe 'error handling during processing' do
+    it 'errors if fails to read a source file' do
```
mokagio: Nitpick. What do you think of 

```suggestion
    it 'errors when failing to read a source file' do
```

or

```suggestion
    it 'errors when it failed to read a source file' do
```

spec/ios_extract_keys_from_strings_files_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+        FileUtils.cp_r(File.join(test_data_dir, 'Resources', '.'), lproj_source_dir)
+        broken_file_path = File.join(lproj_source_dir, 'fr.lproj', 'Localizable.strings')
+        File.write(broken_file_path, 'Invalid strings file content')
+
+        expect do
+          run_described_fastlane_action(
+            source_parent_dir: lproj_source_dir,
+            target_original_files: File.join(lproj_source_dir, 'en.lproj', 'InfoPlist.strings')
+          )
+        end.to raise_error(FastlaneCore::Interface::FastlaneError, /^Error while reading the translations from source file `#{broken_file_path}`: #{broken_file_path}: Property List error/)
```
mokagio, line 194: TIL about `raise_error` with type + regex message. RSpec is so versatile!

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,119 @@
...
+            target_strings_file = File.join(File.dirname(File.dirname(target_original_file)), lproj_dir_name, File.basename(target_original_file))
+            next if target_strings_file == target_original_file # do not generate/overwrite the original locale itself
+
+            keys_to_extract = keys_to_extract_per_target_file[target_original_file]
+            UI.message("Extracting #{keys_to_extract.count} keys into #{target_strings_file}...")
+
+            extracted_translations = translations.slice(*keys_to_extract)
+            FileUtils.mkdir_p(File.dirname(target_strings_file)) # Ensure path up to parent dir exists, create it if not.
+            Fastlane::Helper::Ios::L10nHelper.generate_strings_file_from_hash(translations: extracted_translations, output_path: target_strings_file)
+          rescue StandardError => e
```
mokagio, line 24: As far as I understand, [a rescue clause without an explicit `Exception` class will rescue all `StandardError`s (and only those)](https://www.rubydoc.info/stdlib/core/StandardError).

Is there a particular reason to be verbose about this here? That is, as opposed of:

```ruby
rescue => e
```

mokagio: > Is there a particular reason to be verbose about this here?

Because Rubocop says so!

```
Style/RescueStandardError: Avoid rescuing without specifying an error class.
```

😅

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      def self.return_type
+        # Describes what type of data is expected to be returned
+        # see RETURN_TYPES in https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/action.rb
+        nil
+      end
+
+      def self.return_value
+        # Freeform textual description of the return value
+      end
```
mokagio, line 107: Just a thought unrelated to the work here. Should we keep these overrides that do the [same as the base class](https://github.com/fastlane/fastlane/blob/df6e234a7dc7f5d725b40bb59286edfe8d1dc745/fastlane/lib/fastlane/action.rb#L70-L85)?

Removing them would make the code shorted, but hamper discoverability if we wanted to add a return value in the future. In that case, however, I think we can trust folks to look at how other actions do that and/or at the Fastlane documentation.

FWIW, I removed them locally and run the tests to smoke test. All green. That's not surprising because of how inheritance works, but I wanted to see if maybe Fastlane would complain about it.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
mokagio: Here's a suggestion, but not sure how much it helps...

```suggestion
                                       description: 'The path to the `<base-locale>.lproj/<target-tablename>.strings` file(s) for which we want to extract the keys to. ' \
                                        + 'Each file should contain the originals strings (typically `en` or `Base` locale) and will be used to determine which keys to extract from the `source_tablename`. ' \
                                        + 'The path(s) in which to extract the translations will be determined from each input by looking for the files with the same basename ' \
                                        + 'in each of the `*.lproj` sibling folders.',
```

The fact that this can't be explained clearly should give us pause. If we can't explain it clearly, how clear will it be to work with, both as a consumer or as a maintainer in the future? 🤔 

On the other hand, the whole setup is quite advanced: localizing a project and its libraries via a single input for translators convenience that we then deconstruct into multiple `.strings` files. So a bit of complexity in the descriptions is granted.

I thought of suggesting to split the `base-locale` value from the `target-tablename` one, but I don't think the `verify_block` below would then be able to read `base-locale` to check for the file existence. 😞

mokagio (approved):
Ready to merge with or without my suggestions.

spec/ios_extract_keys_from_strings_files_spec.rb
AliSoftware: Oh the old wording I had was `it 'errors if it fails …'` but I later realized that it had a typo which made it read `it 'errors if if fails …'`… and I mis-fixed the typo by getting rid of one of the double `if` words entirely, instead of fixing it to an `it` as I should have. That explains why it did read strange there 😓 🤦 

Will pick your suggestion instead 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
AliSoftware: Yep, exactly that 😄

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
AliSoftware: Yeah I always wonder about that as well when I create a new action.

But given most of us do a copy/paste of an old action to get a starting point when creating a new one, and also because every time I add those if I copied this from an old action that didn't have them:
 - I often then forget to re-introduce them if they end up being useful
 - Or even forget to just _consider_ and pause and ponder if my new action should have a return value or not in the first place (while having them around from the copied action makes you wonder about it which is a good thing imho)
 - And when I decide they are missing but should be added to my new action, I have to search for both the name of those, but also their role (e.g. "wait, is `return_type` supposed to be a freeform text? a Ruby type like `Array`? A symbol like `:array`?" and "is `return_value` supposed to be some symbol, or the actual value of a variable? Or just freeform text describing it?"), and it's easier to have them around and empty but documented rather than having to dig back into fastlane's own source code to find the parent class and retro-engineer their role and values

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
AliSoftware: > I thought of suggesting to split the base-locale value from the target-tablename one, but I don't think the verify_block below would then be able to read base-locale to check for the file existence. 😞

Also, splitting `target_tablename` and `base_locale` is how it was before the 2nd pass of my work of that PR, but I changed it to that because of the need for this action to support multiple `{target_tablename + target_parent_dir}` pairs (see [this comment above](https://github.com/wordpress-mobile/release-toolkit/pull/338#discussion_r806176480) for details).

---

Because we need to support the case of having both `WordPress/Resources/*/InfoPlist.strings` and `WordPress/WordPressIntents/*/Sites.strings` as target files to generate — but can't really _actually_ expect to use that `*` literal in those paths when passing the values for that parameter at call site — this is why it's difficult to split those into more individual parameters.

One option would have been to split that into `base_locale` and `target_tables`… and make `target_tables` be an Array of Hashes, like:
```ruby
  target_tables: [
    { dir: `WordPress/Resources`, table: `InfoPlist` },
    { dir: `WordPress/WordPressIntents`, table: `Sites` },
  ]
```
or like:
```ruby
  target_tables: [
    { `WordPress/Resources`: `InfoPlist` },
    { `WordPress/WordPressIntents`: `Sites` },
  ]
```

We couldn't have it be directly a `Hash` (of `{directory => tablename}` values) because nothing tells us that we won't have a case in the future of having to extract to more tables, with 2 target tables ending up being in the same parent dir… but Hashes keys must me unique.

We could then have an `Hash<String, Array<String>>` to do `{ directory => Array<tablenames> }` to support multiple target tables in same parent dir… but at that point that seems even more convoluted than any other option

---

Overall this above has been my reasoning when trying to find the final API, and the train of thoughts which led me to use a simple `Array` like here, with the drawback that it's a tad hard to describe in docs… but I think once you understand the expected values it's quite simple to use.

PS: Note that as an additional bonus, the current solution I ended up using here should theoretically also allow us to have target files that use a different base locale (e.g. `WordPress/Resources/en.lproj/InfoPlist.strings` for the main app but `WordPress/WordPressIntents/en-US.lproj/Sites.strings` or `WordPress/WordPressIntents/Base.lproj/Sites.strings` for the Siri Intents target, or something like that) should we encounter such an odd setup in one of our client apps and targets. That scenario seems unlikely as it's more likely that we try to have the base locales be consistent (including given GlotPress project and thus original locale would be the same)… but we never know what odd beast we might encounter… (also thinking about projects like DayOne and/or PocketCasts which might soon adopt our tooling but might not yet have a straight setup with all edges sharpened properly just yet)

AliSoftware: Ended up borrowing most of your wording suggestion here, but tweaked the second sentence a bit more, which led to 96a9a1186ba669cd80adda61fbab2a0e7606a9d7 and this description:
```ruby
description: 'The path(s) to the `<base-locale>.lproj/<target-tablename>.strings` file(s) for which we want to extract the keys to. ' \
  + 'Each of those files should containing the original strings (typically `en` or `Base` locale) and will be used to determine which keys to extract from the `source_tablename`. ' \
  + 'For each of those, the path(s) in which the translations will be extracted will be the files with the same basename in each of the other `*.lproj` sibling folders',
```
I'll merge and go with this but don't hesitate to comment if you have further feedback on that take and I can address them in a follow-up PR.

mokagio: All good 👍

PR #339, opened 2022-02-18
Add Upload to S3 Action

jkmassel:
Adds an "upload to S3" action inside the release toolkit that's useful for binary storage (particularly for installable builds).

**To Test:**
Note that https://github.com/woocommerce/woocommerce-android/pull/5862 passes and properly displays a comment with a download link – the upload functionality for this is powered by this new action.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ _None_
 ### New Features
 
 * Introduce new `ios_extract_keys_from_strings_files` action. [#338]
-* Add Upload to S3 Action
+* Add Upload to S3 Action. [#339]
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ _None_
 ### New Features
 
 * Introduce new `ios_extract_keys_from_strings_files` action. [#338]
+* Add Upload to S3 Action
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -15,9 +15,6 @@ module Fastlane
         bucket = params[:bucket]
         key = params[:key]
 
-        UI.user_error!('You must provide a valid bucket name') if bucket.empty?
-        UI.user_error!('You must provide a valid key') if key.is_a?(String) && key.empty?
-
         key = file_name if key.nil?
 
         if params[:auto_prefix] == true
@@ -78,13 +75,15 @@ module Fastlane
             key: :bucket,
             description: 'The bucket that will store the file',
             optional: false,
-            type: String
+            type: String,
+            verify_block: proc { |bucket| UI.user_error!('You must provide a valid bucket name') if bucket.empty? }
           ),
           FastlaneCore::ConfigItem.new(
             key: :key,
             description: 'The path to the file within the bucket',
             optional: true,
-            type: String
+            type: String,
+            verify_block: proc { |key| UI.user_error!('You must provide a valid key') if key.is_a?(String) && key.empty? }
           ),
           FastlaneCore::ConfigItem.new(
             key: :file,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -13,16 +13,14 @@ module Fastlane
         file_name = File.basename(file_path)
 
         bucket = params[:bucket]
-        key = params[:key]
-
-        key = file_name if key.nil?
+        key = params[:key] || file_name
 
         if params[:auto_prefix] == true
           file_name_hash = Digest::SHA1.hexdigest(file_name)
           key = [file_name_hash, key].join('/')
         end
 
-        UI.user_error!("File already exists at #{key}") if file_is_already_uploaded?(bucket, key)
+        UI.user_error!("File already exists in S3 bucket #{bucket} at #{key}") if file_is_already_uploaded?(bucket, key)
 
         UI.message("Uploading #{file_path} to: #{key}")
 
@@ -80,10 +78,10 @@ module Fastlane
           ),
           FastlaneCore::ConfigItem.new(
             key: :key,
-            description: 'The path to the file within the bucket',
+            description: 'The path to the file within the bucket. If `nil`, will default to the `file`'s basename',
             optional: true,
             type: String,
-            verify_block: proc { |key| UI.user_error!('You must provide a valid key') if key.is_a?(String) && key.empty? }
+            verify_block: proc { |key| UI.user_error!('The provided key must not be empty. Use nil instead if you want to default to the file basename') if key.is_a?(String) && key.empty? }
           ),
           FastlaneCore::ConfigItem.new(
             key: :file,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -70,7 +70,6 @@ module Fastlane
       end
 
       def self.details
-        # Optional:
         'Uploads a file to S3, and makes a pre-signed URL available in the lane context'
       end
 
@@ -93,10 +92,11 @@ module Fastlane
             description: 'The path to the local file on disk',
             optional: false,
             type: String
+            verify_block: proc { |f| UI.user_error!("Path `#{f}` does not exist.") unless File.file?(f) }
           ),
           FastlaneCore::ConfigItem.new(
             key: :auto_prefix,
-            description: 'Generate a derived prefix based on the filename that makes it harder to guess the URL to the uploaded object',
+            description: 'Generate a derived prefix based on the filename that makes it harder to guess the URL of the uploaded object',
             optional: true,
             default_value: true,
             type: Boolean
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -11,7 +11,6 @@ module Fastlane
       def self.run(params)
         file_path = params[:file]
         file_name = File.basename(file_path)
-        UI.user_error!("Unable to read file at #{file_path}") unless File.file?(file_path)
 
         bucket = params[:bucket]
         key = params[:key]
@@ -91,7 +90,7 @@ module Fastlane
             key: :file,
             description: 'The path to the local file on disk',
             optional: false,
-            type: String
+            type: String,
             verify_block: proc { |f| UI.user_error!("Path `#{f}` does not exist.") unless File.file?(f) }
           ),
           FastlaneCore::ConfigItem.new(
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -4,7 +4,7 @@ require 'digest/sha1'
 module Fastlane
   module Actions
     module SharedValues
-      UPLOADED_FILE_PATH = :UPLOADED_FILE_PATH
+      S3_UPLOADED_FILE_PATH = :S3_UPLOADED_FILE_PATH
     end
 
     class UploadToS3Action < Action
@@ -37,7 +37,7 @@ module Fastlane
 
         UI.success('Upload Complete')
 
-        Actions.lane_context[SharedValues::UPLOADED_FILE_PATH] = key
+        Actions.lane_context[SharedValues::S3_UPLOADED_FILE_PATH] = key
 
         return key
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -78,10 +78,13 @@ module Fastlane
           ),
           FastlaneCore::ConfigItem.new(
             key: :key,
-            description: 'The path to the file within the bucket. If `nil`, will default to the `file`'s basename',
+            description: 'The path to the file within the bucket. If `nil`, will default to the `file\'s basename',
             optional: true,
             type: String,
-            verify_block: proc { |key| UI.user_error!('The provided key must not be empty. Use nil instead if you want to default to the file basename') if key.is_a?(String) && key.empty? }
+            verify_block: proc { |key|
+              next if key.is_a?(String) && !key.empty?
+              UI.user_error!('The provided key must not be empty. Use nil instead if you want to default to the file basename')
+            }
           ),
           FastlaneCore::ConfigItem.new(
             key: :file,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -16,8 +16,8 @@ module Fastlane
         bucket = params[:bucket]
         key = params[:key]
 
-        UI.user_error!("You must provide a valid bucket name") if bucket.empty?
-        UI.user_error!("You must provide a valid key") if key.is_a?(String) && key.empty?
+        UI.user_error!('You must provide a valid bucket name') if bucket.empty?
+        UI.user_error!('You must provide a valid key') if key.is_a?(String) && key.empty?
 
         key = file_name if key.nil?
 
@@ -35,7 +35,7 @@ module Fastlane
             body: file,
             bucket: bucket,
             key: key
-        )
+          )
         rescue Aws::S3::Errors::ServiceError => e
           UI.crash!("Unable to upload file to S3: #{e.message}")
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -16,8 +16,13 @@ module Fastlane
         bucket = params[:bucket]
         key = params[:key]
 
+        UI.user_error!("You must provide a valid bucket name") if bucket.empty?
+        UI.user_error!("You must provide a valid key") if key.is_a?(String) && key.empty?
+
+        key = file_name if key.nil?
+
         if params[:auto_prefix] == true
-          file_name_hash = Digest::SHA1.base64digest(file_name)
+          file_name_hash = Digest::SHA1.hexdigest(file_name)
           key = [file_name_hash, key].join('/')
         end
 
@@ -26,11 +31,11 @@ module Fastlane
         UI.message("Uploading #{file_path} to: #{key}")
 
         File.open(file_path, 'rb') do |file|
-          Aws::S3::Client.new().put_object({
-                                             body: file,
-                                             bucket: bucket,
-                                             key: key
-                                           })
+          Aws::S3::Client.new().put_object(
+            body: file,
+            bucket: bucket,
+            key: key
+        )
         rescue Aws::S3::Errors::ServiceError => e
           UI.crash!("Unable to upload file to S3: #{e.message}")
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -52,7 +52,7 @@ module Fastlane
           bucket: bucket,
           key: key
         )
-        return (response[:content_length]).positive?
+        return response[:content_length].positive?
       rescue Aws::S3::Errors::NotFound
         return false
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -1,0 +1,107 @@
+require 'fastlane/action'
+require 'digest/sha1'
+
+module Fastlane
+  module Actions
+    module SharedValues
+      UPLOADED_FILE_PATH = :UPLOADED_FILE_PATH
+    end
+
+    class UploadToS3Action < Action
+      def self.run(params)
+        file_path = params[:file]
+        file_name = File.basename(file_path)
+        UI.user_error!("Unable to read file at #{file_path}") unless File.file?(file_path)
+
+        bucket = params[:bucket]
+        key = params[:key]
+
+        if params[:auto_prefix] == true
+          file_name_hash = Digest::SHA1.base64digest(file_name)
+          key = [file_name_hash, key].join('/')
+        end
+
+        UI.user_error!("File already exists at #{key}") if file_is_already_uploaded?(bucket, key)
+
+        UI.message("Uploading #{file_path} to: #{key}")
+
+        File.open(file_path, 'rb') do |file|
+          Aws::S3::Client.new().put_object({
+                                             body: file,
+                                             bucket: bucket,
+                                             key: key
+                                           })
+        rescue Aws::S3::Errors::ServiceError => e
+          UI.crash!("Unable to upload file to S3: #{e.message}")
+        end
+
+        UI.success('Upload Complete')
+
+        Actions.lane_context[SharedValues::UPLOADED_FILE_PATH] = key
+
+        return key
+      end
+
+      def self.file_is_already_uploaded?(bucket, key)
+        response = Aws::S3::Client.new().head_object(
+          bucket: bucket,
+          key: key
+        )
+        return (response[:content_length]).positive?
+      rescue Aws::S3::Errors::NotFound
+        return false
+      end
+
+      def self.description
+        'Uploads a given file to S3'
+      end
+
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.return_value
+        'Returns the object\'s derived S3 key'
+      end
+
+      def self.details
+        # Optional:
+        'Uploads a file to S3, and makes a pre-signed URL available in the lane context'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :bucket,
+            description: 'The bucket that will store the file',
+            optional: false,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :key,
+            description: 'The path to the file within the bucket',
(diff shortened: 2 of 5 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+                                             bucket: bucket,
+                                             key: key
+                                           })
+        rescue Aws::S3::Errors::ServiceError => e
+          UI.crash!("Unable to upload file to S3: #{e.message}")
+        end
+
+        UI.success('Upload Complete')
+
+        Actions.lane_context[SharedValues::UPLOADED_FILE_PATH] = key
```
AliSoftware, line 40: Given `SharedValues` is a wide and quite global namespace I'd  suggest use a more specific constant name here

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
+require 'fastlane/action'
+require 'digest/sha1'
+
+module Fastlane
+  module Actions
+    module SharedValues
+      UPLOADED_FILE_PATH = :UPLOADED_FILE_PATH
```
AliSoftware: Given `SharedValues` is a wide and quite global namespace I'd  suggest use a more specific constant name here.

```suggestion
      S3_UPLOADED_FILE_PATH = :S3_UPLOADED_FILE_PATH
```

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+* Add Upload to S3 Action
```
```suggestion
* Add Upload to S3 Action. [#339]
```

AliSoftware (commented):
Why not use https://github.com/fastlane-community/fastlane-plugin-s3 instead of reimplementing our own?!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: Addressed in 1dcc36bb4c5cb9d682dbcd6b762b3d29f632e1be to also update its usage.

jkmassel:
> Why not use https://github.com/fastlane-community/fastlane-plugin-s3 instead of reimplementing our own?!

Good question – I'd be fine to use it if we could (though I'm curious whether it'd make more sense to have this plugin depend on that one, or just use both in the top-level app – I don't feel strongly either way).

I came across [this issue ](https://github.com/fastlane-community/fastlane-plugin-s3/issues/39) which I believe is a blocker, but I'll try it out in a branch and see if it's actually been resolved already (as the thread indicates it might have been).

jkmassel:
@AliSoftware – I took another look at using the built-in S3 plugin, and there are a few other issues (from what I can see):
- By default it'll read the result of `gradle` or `gym` and upload the generated file – this isn't necessarily what we want, and I don't see a way to opt out of it.
- Relatedly, when we upload an `apk` this plugin will generate (what I assume is) a manifest for the binary and an HTML page to download it. We can disable this, but it's the default, and I'm not sure I want to constantly have to go "against the grain" for simple functionality like this.
- We can provide a `files` list for it to upload, but not a list of keys to use for those files, which given the open nature of the artifacts bucket makes the URLs guessable (which is not good).

jkmassel:
Rebased to fix merge conflicts.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+        file_path = params[:file]
+        file_name = File.basename(file_path)
+        UI.user_error!("Unable to read file at #{file_path}") unless File.file?(file_path)
+
+        bucket = params[:bucket]
+        key = params[:key]
+
+        if params[:auto_prefix] == true
+          file_name_hash = Digest::SHA1.base64digest(file_name)
+          key = [file_name_hash, key].join('/')
```
mokagio, line 20: Just curious. Is there an advantage of doing this over `"#{file_name_hash}/#{key}"?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+        UI.user_error!("File already exists at #{key}") if file_is_already_uploaded?(bucket, key)
+
+        UI.message("Uploading #{file_path} to: #{key}")
+
+        File.open(file_path, 'rb') do |file|
+          Aws::S3::Client.new().put_object({
+                                             body: file,
+                                             bucket: bucket,
+                                             key: key
+                                           })
```
mokagio: Indentation nitpick.

```suggestion
          Aws::S3::Client.new().put_object(
            {
              body: file,
              bucket: bucket,
              key: key
            }
          )
```

mokagio: Or, more compact:

```suggestion
          Aws::S3::Client.new().put_object({ body: file, bucket: bucket, key: key })
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+        rescue Aws::S3::Errors::ServiceError => e
```
mokagio, line 33: Would it be worth rescuing other errors, too?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+
+        return key
+      end
+
+      def self.file_is_already_uploaded?(bucket, key)
+        response = Aws::S3::Client.new().head_object(
+          bucket: bucket,
+          key: key
+        )
+        return (response[:content_length]).positive?
```
mokagio: Is the wrapping in `( )` necessary?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.return_value
+        'Returns the object\'s derived S3 key'
+      end
+
+      def self.details
+        # Optional:
```
mokagio:
```suggestion
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :file,
+            description: 'The path to the local file on disk',
+            optional: false,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :auto_prefix,
+            description: 'Generate a derived prefix based on the filename that makes it harder to guess the URL to the uploaded object',
```
mokagio: I'm not sure if this suggestion is correct, but just in case:

```suggestion
            description: 'Generate a derived prefix based on the filename that makes it harder to guess the URL of the uploaded object',
```

mokagio (approved):
Sad that there's no tests for this action, but there's not much to do when the bulk of the logic is around a third party dependency.

Well, actually, it looks like [the AWS SDK has stubbing methods for testing](https://aws.amazon.com/blogs/developer/advanced-client-stubbing-in-the-aws-sdk-for-ruby-version-3/). It might be worth checking them out, but definitely not worth blocking this PR for me, also given this is required the [Woo Android installable builds](https://github.com/woocommerce/woocommerce-android/pull/5862).

I implicitly verified this code's happy path in https://github.com/woocommerce/woocommerce-android/pull/5862. Approving.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: Ahah I had the exact same question.

Actually during my first review my skimming gave me the wrong impression that the `params[:key]` provided by the call site would be unused if `auto_prefix` was true, because this line would overwrite its value with `key = …`. Not sure if using `"#{file_name_hash}/#{key}"` would have prevented my skimming to have the same wrong assumption (lack of coffee was probably the real culprit) but I was still caught off-guard by this 😆 

---

I'm not fully an S3 expert, so I have to ask: what's the result of having a `/` in that prefixed `key`? Is this still considered to be a single key name (with literal character`/` in it) in S3 API, or will that result into an actual intermediate directory in the path to the file? I believe it's the latter, right? If so, then maybe even using `File.join` would make even more sense here then?

```suggestion
          key = File.join(file_name_hash, key)
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: I even think, given how Ruby works and interprets named parameters at call site vs Hashes at def site, that you should be able to do:
```suggestion
          Aws::S3::Client.new().put_object(body: file, bucket: bucket, key: key)
```
At least that's how most of those typical Ruby client implementations (e.g. Octokit for GitHub API client) expect you to call those methods anyway 🙂

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: ☝️

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: ☝️

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+          FastlaneCore::ConfigItem.new(
+            key: :bucket,
+            description: 'The bucket that will store the file',
+            optional: false,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :key,
+            description: 'The path to the file within the bucket',
+            optional: true,
```
AliSoftware, line 82: Optional? 🤔  What happens if we don't provide a key parameter to that one?

AliSoftware (commented):
Left a couple of style nitpicks, otherwise seems to be working.
I still think unit tests would be quite valuable to that action, despite it interacting with a 3rd party service.

We can `stub_requests` or make use of `double` / `instance_double` and similar to easily stub the S3 side of things, and focus the tests on the behavior at parsing and interpreting the input `ConfigItems` params and checking that it calls `put_object` with the expected parameters as a result.

For example, writing a test that calls `run_described_fastlane_action` with a `nil` key, and checks that it would either auto-generate a key for us, or crash with a `user_error!` if that is not permitted (I actually have no idea which of the 2 behaviors would be the correct one for that case, hence using that as a good example for a spec 😛 )

That being said, if that is a blocker, we could also decide to merge this PR without the tests in order to unblock the migration of remaining app projects to Buildkite, and then implement the tests in a subsequent PR, but I still think we should implement those tests at some point.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :file,
+            description: 'The path to the local file on disk',
+            optional: false,
+            type: String
+          ),
```
AliSoftware:
```suggestion
            type: String
            verify_block: proc { |f| UI.user_error!("Path `#{f}` does not exist.") unless File.file?(f) }
          ),
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+  module Actions
+    module SharedValues
+      S3_UPLOADED_FILE_PATH = :S3_UPLOADED_FILE_PATH
+    end
+
+    class UploadToS3Action < Action
+      def self.run(params)
+        file_path = params[:file]
+        file_name = File.basename(file_path)
+        UI.user_error!("Unable to read file at #{file_path}") unless File.file?(file_path)
```
AliSoftware: You can move this to the `verify_block` of the `ConfigItem` (see [this suggestion](https://github.com/wordpress-mobile/release-toolkit/pull/339#discussion_r811862892)) to let Fastlane report the error earlier in the process while parsing args (without having to create a Runner for nothing)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,107 @@
...
+
+        bucket = params[:bucket]
+        key = params[:key]
+
+        if params[:auto_prefix] == true
+          file_name_hash = Digest::SHA1.base64digest(file_name)
```
AliSoftware: Note: the Base64 alphabet happens to contain the `/` character. Will that risk causing an issue or result in misleading or strange cases?

For example if the input file is named `input_file_1` then the base64 digest of its SHA1 is `k5w5OY2yQF55HiBXeP9w+F3/Yg4=` with a `/` in the middle. Worse, I imagine there could be input file names that lead to a base64 SHA1 to _end with_ a `/`, which would lead to a key containing two consecutive `/` once we assemble that prefix with the filename, and I can imagine a key with `foo//bar` to lead to errors with S3…

Discovered that use case by implementing Unit Tests for this in #340 btw… so they _are_ really useful (in case that was still needed to be proven) 😛 

I suggest we should use `hexdigest` instead of `base64digest` to avoid such situations:
```suggestion
          file_name_hash = Digest::SHA1.hexdigest(file_name)
```

AliSoftware (changes requested):
Actually found a bug while giving a try at implementing Unit Tests in #340 😉

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: I initially used `File.join`, then realized that the result would be invalid on Windows, where the path separator is `\`. 

I'm not sure we're likely to have this running on windows boxes _soon_, but I figured this was safe since it's S3-specific 🙂

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: I thought about that, but my thinking was that most common issue will be "S3 is having issues or there's internet weather or AWS is down _again_", and if something _else_ is wrong, I'd rather rather see a stack trace in order to figure out exactly where the problem happened (vs a simple exception `message` that might not be helpful). 🤷‍♂️

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: Probably not! 😅 

Removed in ab376e740849c0c13d10544a709549b77218a532

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: Good q – addressed in the testing PR, but now it'll emit an error.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: Not sure either – I'll take it though! 😂

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: Done in e8748eeb55beb5fd542561b9dc42a5f32e2a7b9e, then extended the pattern to other validation in 684fad43220ea8766b6f7e0f14e1bebacaa4dcf8

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
jkmassel: Yep – it's funny, I reviewed another PR and suggested dropping the hash syntax, but somehow missed it here. I've simplified it to just the regular named parameter syntax (though I retained the multi-line call in case of future changes).

jkmassel:
> what's the result of having a / in that prefixed key?

If that happens, it'll just be nested in "directories" within S3. Doesn't hurt anything, but there's a possibility of a crash if the string _ends with_ `/`, so the `hexdigest` is for sure the way to go 🙂

jkmassel:
@AliSoftware – I'll wait to ensure your concerns are addressed before merging if that's ok?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: Good point 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    class UploadToS3Action < Action
+      def self.run(params)
+        file_path = params[:file]
+        file_name = File.basename(file_path)
+
+        bucket = params[:bucket]
+        key = params[:key]
+
+        key = file_name if key.nil?
```
AliSoftware:
```suggestion
        key = params[:key] || file_name
```
(This use of `||` is the Ruby equivalent of `??` in Swift)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,110 @@
...
+
+        if params[:auto_prefix] == true
+          file_name_hash = Digest::SHA1.hexdigest(file_name)
+          key = [file_name_hash, key].join('/')
+        end
+
+        UI.user_error!("File already exists at #{key}") if file_is_already_uploaded?(bucket, key)
```
AliSoftware:
```suggestion
        UI.user_error!("File already exists in S3 bucket #{bucket} at #{key}") if file_is_already_uploaded?(bucket, key)
```
Mostly to emphasis that the "file already exist" is on the S3 side, as opposed to being about a file in the local filesystem

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,110 @@
...
+          FastlaneCore::ConfigItem.new(
+            key: :bucket,
+            description: 'The bucket that will store the file',
+            optional: false,
+            type: String,
+            verify_block: proc { |bucket| UI.user_error!('You must provide a valid bucket name') if bucket.empty? }
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :key,
+            description: 'The path to the file within the bucket',
```
AliSoftware:
```suggestion
            description: 'The path to the file within the bucket. If `nil`, will default to the `file`'s basename',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -0,0 +1,110 @@
...
+            optional: true,
+            type: String,
+            verify_block: proc { |key| UI.user_error!('You must provide a valid key') if key.is_a?(String) && key.empty? }
```
AliSoftware: Was confused at first that you had both this test which tests for a key… and also [added support for `nil` key](https://github.com/wordpress-mobile/release-toolkit/pull/339/files#diff-1f39f81d942cbca33ca74186835e32e40390e15d0e5481026c8f9c48291af468R18), then realized that this was only erroring if the key was the empty string, but still allows `nil` key (to default to the file basename).

Probably worth making the distinction in the error message, wdyt?
```suggestion
            verify_block: proc { |key| UI.user_error!('The provided key must not be empty. Use nil instead if you want to default to the file basename') if key.is_a?(String) && key.empty? }
```

spec/spec_helper.rb
```diff
@@ -32,6 +32,12 @@ def allow_fastlane_action_sh
   allow(FastlaneCore::Helper).to receive(:sh_enabled?).and_return(true)
 end
 
+# Allow us to do `.with` matching against a `File` instance to a particular path in RSpec expectations
+# Because `File.open(path)` returns a different instance of `File` for the same path on each call)
```
AliSoftware: Just saw my typo there…
```suggestion
# (Because `File.open(path)` returns a different instance of `File` for the same path on each call)
```

spec/spec_helper.rb
```diff
@@ -32,6 +32,12 @@ def allow_fastlane_action_sh
...
+RSpec::Matchers.define :file_instance_of do |path|
```
AliSoftware, line 37: nit: maybe I should have named this `file_instance_to(path)` rather than `file_instance_of(path)` 🤔 🤷

spec/spec_helper.rb
```diff
@@ -75,3 +81,26 @@ def in_tmp_dir
...
 end
+
+# Executes the given block with a temporary file with the given `file_name`
+def with_tmp_file_path_for_file_named(file_name)
+  in_tmp_dir do |tmp_dir|
+    file_path = File.join(tmp_dir, file_name)
+
+    File.write(file_path, '')
+    yield file_path
+    File.delete(file_path)
```
AliSoftware: You want to make sure the intermediate file is deleted _even if_ the code crashes in the middle of the execution:

```suggestion
  ensure
    File.delete(file_path)
```

PS: I'm not _fully sure_ that we can use `ensure` if we don't even have a `rescue` in-between… but I think that's valid?

PPS: I actually thought about writing such a helper at first, but wasn't sure it was worth it 🤷 . I also thought about using `Tempfile.create` which is dedicated to a very similar purpose… except that `Tempfile` does not let you enforce a specific filename (and instead create the temp file in `/tmp` and find a unique basename, instead of making the uniqueness by using an intermediate temp dir), which wouldn't have worked for our use case given the SHA1 is dependent on the basename used in the unit test…

spec/spec_helper.rb
```diff
@@ -75,3 +81,26 @@ def in_tmp_dir
...
+def with_tmp_file_path
+  in_tmp_dir do |tmp_dir|
+    file_name = ('a'..'z').to_a.sample(8).join # 8-character random file name
+    file_path = File.join(tmp_dir, file_name)
+
+    File.write(file_path, '')
+    yield file_path
+    File.delete(file_path)
+  end
+end
```
AliSoftware: I think it would be both simpler (and DRY) and more readable to merge those two helpers in a single one, with a file name generated using `sample(8)` like you did if the file provided as parameter is `nil` (which could be the default). That would look like this:

```ruby
# Executes the given block with a temporary file with the given `file_name`
def with_tmp_file(named: nil, content: '')
  in_tmp_dir do |tmp_dir|
    file_name = named || ('a'..'z').to_a.sample(8).join # 8-character random file name if nil
    file_path = File.join(tmp_dir, file_name)
    File.write(file_path, content)
    yield file_path
  ensure
    File.delete(file_path)
  end
end
```

And then you'd simply replace all your calls to `with_tmp_file_path` with this `with_tmp_file` (without parameters), and all your calls to `with_tmp_file_path_for_file_named(x)` with `with_tmp_file(named: x)`.

spec/spec_helper.rb
```diff
@@ -75,3 +81,26 @@ def in_tmp_dir
     end
   end
 end
+
+# Executes the given block with a temporary file with the given `file_name`
+def with_tmp_file_path_for_file_named(file_name)
+  in_tmp_dir do |tmp_dir|
+    file_path = File.join(tmp_dir, file_name)
+
+    File.write(file_path, '')
```
AliSoftware: I'd suggest to provide an optional content parameter to the method to allow custom content to be written as well instead of an empty file.

i.e. making the signature of the method look more like: `def with_tmp_file(basename:, content: '')` and then use the `content` parameter on this line.

spec/upload_to_s3_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+      .and_return(Aws::S3::Types::HeadObjectOutput.new(content_length: content_length))
+  end
+
+  def stub_s3_file_exists(key)
+    stub_s3_head_request(key, 1)
+  end
+
+  def stub_s3_file_does_not_exist(key)
+    stub_s3_head_request(key, 0)
+  end
```
AliSoftware: You might as well refactor this as a single helper like this:
```suggestion
  def stub_s3_file_exists(key, exists)
    content_length = exists ? 1 : 0
    allow(client).to receive(:head_object)
      .with(bucket: test_bucket, key: key)
      .and_return(Aws::S3::Types::HeadObjectOutput.new(content_length: content_length))
  end
```

spec/upload_to_s3_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  describe 'uploading a file' do
```
AliSoftware: This ends up being the only `describe` at that level, which makes it kinda pointless, and also containing quite a lot of tests in a single group.

I'd suggest you to split it instead, for example:
 - `describe 'use the proper key'` (L28-L117)
 - `describe 'errors on invalid input'` (the rest)

spec/upload_to_s3_spec.rb
```diff
@@ -0,0 +1,168 @@
+require 'tempfile'
```
AliSoftware: This is a leftover from when I tried to use `Tempfile.create` to generate the temporary files — until I realized that the name of those files would not be deterministic leading to non-deterministic SHA1 for the prefixed keys…

```suggestion
```

AliSoftware (commented):
Overall LGTM by skimming the new code 👍  but left a couple of suggestions to improve the wording, error messages, and code readability.

I'll let you address those nits during your day, and review the PR again first thing on my tomorrow — at which point I expect we should finally be all good to go 🤞

spec/upload_to_s3_spec.rb
jkmassel: Fixed in 281c1818a2e555933885dcddb96607eff53ab072 – I tried to keep to the "reads as a sentence" format

PR #340, opened 2022-02-22
Introduce specs for upload_to_s3 action

AliSoftware:
This is built on top of https://github.com/wordpress-mobile/release-toolkit/pull/339, to add a couple of unit tests on the new action.

Feel free to push new commits to this PR to implement the remaining tests.

This PR can be merged either in `add/s3-binary-upload` first, _before_ we merge #339 tun `trunk`, or it can be merged in `trunk` _after_ #339 lands if we don't want to block #339.

## To Test

 - Run `be rspec spec/upload_to_s3_spec.rb` to focus on running just those new tests
 - Alternatively, just let CI run `rspec` tests

### DIFF
--- a/spec/upload_to_s3_spec.rb
+++ b/spec/upload_to_s3_spec.rb
@@ -23,7 +23,7 @@ describe Fastlane::Actions::UploadToS3Action do
   end
 
   describe 'happy path' do
-    it 'generates a prefix for the key when using auto_prefix:true' do
+    it 'generates a prefix for the key by default' do
       in_tmp_dir do |tmp_dir|
         file_path = File.join(tmp_dir, 'input_file_1')
         File.write(file_path, 'Dummy content')
@@ -43,7 +43,47 @@ describe Fastlane::Actions::UploadToS3Action do
       end
     end
 
-    it 'uses the provided key verbatim when using auto_prefix:false'
+    it 'generates a prefix for the key when using auto_prefix:true' do
+      in_tmp_dir do |tmp_dir|
+        file_path = File.join(tmp_dir, 'input_file_1')
+        File.write(file_path, 'Dummy content')
+        expected_key = 'k5w5OY2yQF55HiBXeP9w+F3/Yg4=/subdir/a8c-key1'
+
+        stub_s3_head_request(expected_key, 0) # File does not exist in S3
+        expect(client).to receive(:put_object).with(body: file_instance_of(file_path), bucket: test_bucket, key: expected_key)
+
+        return_value = run_described_fastlane_action(
+          bucket: test_bucket,
+          key: 'subdir/a8c-key1',
+          file: file_path,
+          auto_prefix: true
+        )
+
+        expect(return_value).to eq(expected_key)
+        expect(Fastlane::Actions.lane_context[Fastlane::Actions::SharedValues::S3_UPLOADED_FILE_PATH]).to eq(expected_key)
+      end
+    end
+
+    it 'uses the provided key verbatim when using auto_prefix:false' do
+      in_tmp_dir do |tmp_dir|
+        file_path = File.join(tmp_dir, 'input_file_1')
+        File.write(file_path, 'Dummy content')
+        expected_key = 'subdir/a8c-key1'
+
+        stub_s3_head_request(expected_key, 0) # File does not exist in S3
+        expect(client).to receive(:put_object).with(body: file_instance_of(file_path), bucket: test_bucket, key: expected_key)
+
+        return_value = run_described_fastlane_action(
+          bucket: test_bucket,
+          key: 'subdir/a8c-key1',
+          file: file_path,
+          auto_prefix: false
+        )
+
+        expect(return_value).to eq(expected_key)
+        expect(Fastlane::Actions.lane_context[Fastlane::Actions::SharedValues::S3_UPLOADED_FILE_PATH]).to eq(expected_key)
+      end
+    end
   end
 
   describe 'error reporting' do
--- a/spec/upload_to_s3_spec.rb
+++ b/spec/upload_to_s3_spec.rb
@@ -45,16 +45,16 @@ describe Fastlane::Actions::UploadToS3Action do
 
     it 'generates a prefix for the key when using auto_prefix:true' do
       in_tmp_dir do |tmp_dir|
-        file_path = File.join(tmp_dir, 'input_file_1')
+        file_path = File.join(tmp_dir, 'input_file_2')
         File.write(file_path, 'Dummy content')
-        expected_key = 'k5w5OY2yQF55HiBXeP9w+F3/Yg4=/subdir/a8c-key1'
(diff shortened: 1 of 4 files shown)


spec/upload_to_s3_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  RSpec::Matchers.define :file_instance_of do |path|
+    match { |actual| actual.is_a?(File) && actual.path == path }
+  end
+
+  describe 'happy path' do
+    it 'generates a prefix for the key when using auto_prefix:true' do
+      in_tmp_dir do |tmp_dir|
+        file_path = File.join(tmp_dir, 'input_file_1')
+        File.write(file_path, 'Dummy content')
+        expected_key = 'k5w5OY2yQF55HiBXeP9w+F3/Yg4=/subdir/a8c-key1'
```
AliSoftware: The `SHA1.base64digest('input_file_1')` seems to lead to a base64 string that contains `/`, which I believe could cause issues (especially if that `/` ends up at the very start or very end of the SHA1 base64), so I think we'll need to fix the implementation (e.g. using `hexdigest` instead) to avoid such cases

spec/upload_to_s3_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+    end
+  end
+
+  describe 'error reporting' do
+    it 'errors if bucket is empty or nil'
+    it 'errors if key is nil' # Or should it auto-generate a key based on the filename or something instead in that case?
+    it 'errors if local file to upload does not exist'
+    it 'reports an error if the file already exists on S3'
+  end
```
AliSoftware: @jkmassel I've only implemented the Unit Tests for the happy path, to give you a starting point, but I'll leave it to you to implement the remaining test cases, especially these suggested ones 🙂

AliSoftware:
> 1 Warning
> ⚠️	Please add an entry in the CHANGELOG.md file to describe the changes made by this PR
> Generated by 🚫 [Danger](https://danger.systems/)

Depending if we merge this one in `add/s3-binary-upload` branch before merging #339 or way later, we might or might not need a dedicated entry for the addition of unit tests in this PR; but I think the one in #339 will be enough to act for both PRs anyway.

spec/upload_to_s3_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        expect(return_value).to eq(expected_key)
+        expect(Fastlane::Actions.lane_context[Fastlane::Actions::SharedValues::S3_UPLOADED_FILE_PATH]).to eq(expected_key)
+      end
+    end
+
+    it 'generates a prefix for the key when using auto_prefix:true' do
+      in_tmp_dir do |tmp_dir|
+        file_path = File.join(tmp_dir, 'input_file_2')
+        File.write(file_path, 'Dummy content')
+        expected_key = 'i94aegQwDfJ7UvQ4PcmX5fu/8YA=/subdir/a8c-key2'
```
AliSoftware: [Ditto](https://github.com/wordpress-mobile/release-toolkit/pull/340/files#r811927276) with the base4 of SHA1 of `input_file_2` apparently…

spec/upload_to_s3_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    allow(client).to receive(:head_object)
+      .with(bucket: test_bucket, key: key)
+      .and_return(Aws::S3::Types::HeadObjectOutput.new(content_length: content_length))
+  end
+
+  # Allow us to do `.with` matching against a `File` instance to a particular path in RSpec expectations
+  # Because `File.open(path)` returns a different instance of `File` for the same path on each call)
+  RSpec::Matchers.define :file_instance_of do |path|
+    match { |actual| actual.is_a?(File) && actual.path == path }
+  end
```
AliSoftware: I wonder if it would be worth it to move this in `spec_helper.rb` for other, future tests to use… 🤷

jkmassel: Moved in 8d543c0ad2d9e80c8989a0fdf28efd3207590f70, because I could see it being helpful for translation stuff too (really anything where we're pointing at a file).

spec/upload_to_s3_spec.rb
jkmassel: It'd be _super_ rare that it'd cause a problem, but I agree - better to avoid the risk altogether – I used `hexdigest` for cleaner output – nice catch 🙂

spec/upload_to_s3_spec.rb
jkmassel: Thanks for the head start on this!! I've fleshed out the tests and tweaked a few little things – including a little bit of DRY in ab376e740849c0c13d10544a709549b77218a532.

spec/upload_to_s3_spec.rb
AliSoftware: To be fair, for translation actions I did TDD, which led me to [create APIs that could equally take either a file path (`String`) or an `IO`/`File`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb#L127-L128) for versatility.

This allowed me to [use `StringIO` everywhere in my Unit Tests where I needed a pseudo-`File`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/ios_l10n_helper_spec.rb#L183), which avoided me having to create actual temp files in the filesystem altogether, and instead have everything IO-related be in-memory [during all my tests (except the only test checking that passing a `String` worked, obviously)](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/ios_l10n_helper_spec.rb#L193-L195).

---

We could have considered going the same route as well for your `s3_upload_file` action, but that would have required some logic on the implementation's side (to convert the `String` file path to an `IO` via something like `file = File.open(file) if file.is_a?(String)` … but then that begs the question of how to close it only if it was opened by us (i.e. was a `String`) but not close it ourselves if the caller was the one who did the `open`… so probably not worth the added complexity in the implementation just to support passing an `IO` — which in the case of this action, let's be honest, would only be used in unit tests anyway, but unlikely in real world call sites.

That was easier for my `ios_l10n_helper` case because the implementation already used `IO.copy_stream` which already supports either an `IO` or a `String` path out of the box, so I was able to support that duality at zero cost there. Unlike in your case, for which it is probably not be worth the effort, and which is why I ended up keeping your action's API always expecting a `String`, and using the custom matcher on `File` there, instead of trying to use `StringIO` in those tests.

AliSoftware: 💡  An alternative idea would have been to do use this in our unit tests:
```ruby
file_instance = StringIO.new
allow(File).to receive(:open).with(file_path).and_return(file_instance)
```
And then we would have been able to match using `.with(file_instance, …)` directly, without the need for the custom matcher above… 🤷  

Eh, it's too late now, and not worth changing anymore, but just wanted to drop this here anyway just for your information and our own Ruby culture (or for ideas for future specs)

PR #341, opened 2022-03-01
Update Dependencies

jkmassel:
Updates dependencies – one of which is a security warning.


CHANGELOG.md
```diff
@@ -18,7 +18,13 @@ _None_
 
 ### Internal Changes
 
-_None_
+* Updated the following internal dependencies: [#341]
+	* buildkit (1.4.5) -> (1.5.0)
```
mokagio: `buildkit` is part of the public dependencies list below

```suggestion
```

CHANGELOG.md
```diff
@@ -18,7 +18,13 @@ _None_
...
+	* nokogiri (1.12.5) -> (1.13.3)
+	* oj (3.13.9) -> (3.13.11)
+	* rake-compiler (1.1.1) -> (1.1.9)
+* Updated the following public dependency: [#341]
+	* buildkit (1.4.5) -> (1.5.0)
```
mokagio: I was surprised by the indentation here, which uses tabs. But, this is also the first occurrence of an indented list, so I guess that's okay as the standard for the file 🤷

mokagio (approved):
LGTM.

I like the principle of tracking this kind of dependency updates in the changelog, which to me is to make it easier to understand how things changed and trace back any future bug. 

But, I wonder if this will result in a bit of extra work and noise in the file? 

Also, can I double check I got the difference between internal and public dependencies right? All of the four dependencies mentioned are part of the `.gemspec` as `spec.add_dependency` 🤔 But the only one with the version requirement changed is `buildkit`, which is also the only one in the "public dependencies" list. So... is the difference that "internal" means consumers can resolve as they please and the version in the code is currently `x`, while "public" means that the value consumers can resolve to has been update, too?

CHANGELOG.md
AliSoftware: tbh I'm not sure this was worth it to list the details of the dependencies being updated directly in the CHANGELOG 🤷 . The CHANGELOG already references `[#341]` (aka this PR) so if anyone is interested in knowing the details of what dependencies were changed/updated, they can just head here in this PR and look at the `Gemfile.lock` in the diff 🙃

jkmassel: Yeah, fair – I just went over over-explaining – no need for it in the future 🤷‍♂️

mokagio:
🤔 CI is [failing](https://buildkite.com/automattic/release-toolkit/builds/407#e695aab8-9f0d-494b-8636-8caaaa77124c/337-696) because

```
Can't find the ImageMagick library or one of the dependent libraries.
```

Is it possible that #348 passing in CI was a false positive? By looking at the two successful builds from that PR, we can see that neither built the gems from scratch but fetched them from the cache.

- https://buildkite.com/automattic/release-toolkit/builds/401#fbe3ba5e-db26-4d87-9751-ae4f42f5d269/337-339
- https://buildkite.com/automattic/release-toolkit/builds/402#43649c24-0e5e-43c2-9c61-c0bc6ca046b4/337-339

I'm saying this because `bundle install` took ~10s and the output says "Using <gem>":

image

I guess there reason the cached was used is because there was no change in the `Gemfile.lock`.

Conversely, this PR updates the lockfile, resulting in CI building new gems but failing on RMagick because it can't find ImageMagick. 

Also notice the difference in the `bundle install` output:

image

---

I don't think we can remove the RMagick dependency yet, so I'd suggest going back to installing ImageMagick in CI, then file a request to get rid of it once and for all?

AliSoftware:
I've made https://github.com/wordpress-mobile/release-toolkit/pull/349 to revert the changes from #348.

If it ends up being approved and land in `trunk`, we should then merge the new `trunk` (post-merge of 349) back in this PR once again and validate that it makes it go back to green.

PR #342, opened 2022-03-01
Add strings validation helper capable of identifying duplicated keys

mokagio:
Adds a new, "simple" helper object to parse `.strings` files looking for duplicated keys.

To test: You can see there are a lot of different kinds of duplicated keys we test against, and because this object is currently unused (more on that in a second) if those test pass we should be good to merge. Feel free to add more tests for edge cases I couldn't see, though.

Note: The helper is currently unused. As mentioned [in the comments](https://github.com/wordpress-mobile/release-toolkit/pull/342#discussion_r829707870) the next step for me would be to create an action that uses this. I'm keen on doing that in a dedicated PR so we can isolate the conversation on the action API there and on the helper behavior here.

Also note: I didn't add a `CHANGELOG.md` entry, saving that for the action PR.


spec/strings_file_validation_helper_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+          next if key.nil?
+
+          # Check if that key was already encountered in the past
+          existing = found_keys[key]
+
+          duplicated_keys.append(key) unless existing.nil?
+          # TODO: Use proper Fastlane UI methods for this, or better rely on
+          # the consumer to read the return Hash (itself a TODO) and print an
+          # error. This is an helper, after all.
```
mokagio: @AliSoftware can you give me a 30% feedback on the direction I'm taking here?

Aside from the fact that I have the prod and test code in the same file (which is handy for me as I mess with this but obviously will change before calling this ready) what do you think of:

- Having this as a `Fastlane::Helper` – I don't see why it needs to belong to Fastlane, but all the helpers I looked at for inspiration do, so I've been consistent
- The idea of returning an Hash with key and an array of the line numbers where they occur.
- The idea of letting the caller look into the returned Hash and decided what to do with it, which right now should be something like emitting a warning via the `FastlaneCore::UI`

Thanks 🙇

spec/strings_file_validation_helper_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+          # Memorize the line at which this key appeared in the source `.strings`
+          # file
+          found_keys[key] = index
+        end
+
+        duplicated_keys
+      end
+
+      def self.extract_key(line:)
+        regexp = /\s*"([^"]+.*)"\s*=/
```
mokagio: We discussed the fact that, when it comes to `.strings`, parsing is tough and it might very well be one of those case where you try to solve a problem with a RegExp and you're now left with two problems.

But, I'm not sure how else to go about finding duplicated keys otherwise.

I'd say that since the purpose of this RegExp is quite limited, and we can "test the heck out of it", it's an okay approach in this case.

You might recognize the code around here. I lifted it from your original suggestion at paaHJt-2S5-p2#comment-5716. However, I modified the RegExp to avoid the strings substitution, which I think makes the code clearer.

spec/test-data/translations/file-with-duplicated-keys.strings
```diff
@@ -0,0 +1,66 @@
...
+"\U0025 key" = "\U0025 is the % symbol";
+/* ...and not consecutive */
+"\U0026 key" = "\U0025 is the & symbol";
+"unicode\U0020key" = "\U0020 is the space character";
+"\U0026 key" = "\U0025 is the & symbol";
+/* Special case: \"%@\" can be tricky to detect */
+"The Google account \"%@\" doesn't match any account on WordPress.com" = "localized translation here";
+/* A red herring that might be seen as duplicate to the previous one if our RegExp does not correctly account for escaped quotes */
+"The Google account \"%@\" is invalid" = "localized translation here";
+/* Duplicated \"%@\" */
```
mokagio:
```suggestion
/* Duplicated keys that include escaped quotes */
```

spec/test-data/translations/file-with-duplicated-keys.strings
```diff
@@ -0,0 +1,66 @@
...
+"key with \"%@\" character" = "localized translation here";
+"key with \"%@\" character" = "localized translation here";
+"key with multiple \"%@\" escapes \":)\" in it" = "localized translation here";
+"key with multiple \"%@\" escapes \":)\" in it" = "localized translation here";
+/* This is just a check for Gio's implementation. We might want to remove it? */
```
mokagio:
```suggestion
/* Duplicated keys with trailing spaces */
```

spec/strings_file_validation_helper_spec.rb
AliSoftware: The fact that our helpers are in `Fastlane::Helper` namespace is likely a bit of a mix between legacy, and I'm assuming what the template for whatever `fastlane plugin create` command (or whatever it's called) might create as a scaffold when creating a new plugin?

It has always a bit bothered me tbh because that means we might accidentally clash with present or future helpers with the same name that fastlane itself would decide to use. I'd agree in refactoring all those to be in a `module WPMRT` or `module A8C` namespace instead at some point, if that makes more sense to all of us. This is part of the very long list of little things I'd love to refactor in our release-toolkit one day to clean things up and make it more structured, but never got the time to…

---

As for the return value and error emitting:
 - I think it makes more sense for the error emitting to be handled by either the action, or the client caller in the Fastfile, rather than the Helper. It makes more sense for the Helper to either `raise` or return a list of errors (like you did), and let the caller process them, because a Helper is more in the "business model" category than the "UI/View" category anyway
 - That being said, I like when the action at least offers an easy way to do the calls to `UI.message`/ `UI.user_error!`/… itself — as opposed to having the `Fastfile` calling the action be responsible for it. Because error handling and processing is quite repetitive and often the same, and it's easy to forget to handle it at call site if it's optional.

So my approach for that is usually:
 - Make the Helper just return the list or errors (as you did)
 - Make the action default to do the error handle that it received back from the helper, i.e. call `UI.user_error!` or `UI.message` and the like as necessary
 - Optionally, add a `ConfigItem` to the action to disable the error raising handled by the action, or choose a mode (e.g. between `:error`, `:warn`, `:silent`) in the rare case where the caller in `Fastfile` would prefer to handle the error reporting themselves.

That way we would have a sensible default that would avoid us to let those errors go unnoticed/unreported accidentally (for anyone just discovering that action and not realizing they'd have to handle the errors themselves), but an opt-out option for when you want more control.

spec/strings_file_validation_helper_spec.rb
AliSoftware: Yeah, tough call there.

Notice that for example that means that the action won't work on `InfoPlist.strings` file which typically does not use `""` quotes around key names… not sure if that was one of the potential file we'd want to target with this action, but worth noting anyway.

It also won't work with keys that contain `\"` in their name — because of the `[^"]+` thing. The `.*` that follows might accidentally workaround this, actually rendering the preceding `[^"]+` kinda useless, but that also risks making the `.*` be greedy (e.g. `"Login to a \"%@\" account" = "Login to a \"%@\" account"; // Here %@ = the service name, like Google`)

While we'd hope in an ideal world that we'd try to avoid such keys, and despite the fact that our new tooling will now accept keys that are not the English copy (`AppLocalizedString("key", value: "English copy", comment: "blah")`) which would even lessen the risk of having such cases, we know that those cases currently exist in the current codebase and `Localizable.strings` files…

I don't have a magic solution for this though, tbh. Maybe using a Parser Combinator or at least a parsing simple state machine would make it easier than using a RegEx when it comes to handling backslash escapes and quote balancing? 🤔  Though maybe not, and maybe that'd also be killing a fly with a bazooka…

spec/test-data/translations/file-with-duplicated-keys.strings
```diff
@@ -0,0 +1,66 @@
...
+"\U0025 key" = "\U0025 is the % symbol";
+"\U0025 key" = "\U0025 is the % symbol";
+/* ...and not consecutive */
+"\U0026 key" = "\U0025 is the & symbol";
+"unicode\U0020key" = "\U0020 is the space character";
+"\U0026 key" = "\U0025 is the & symbol";
+/* Special case: \"%@\" can be tricky to detect */
+"The Google account \"%@\" doesn't match any account on WordPress.com" = "localized translation here";
+/* A red herring that might be seen as duplicate to the previous one if our RegExp does not correctly account for escaped quotes */
+"The Google account \"%@\" is invalid" = "localized translation here";
```
AliSoftware, line 58: This passes because the RegEx actually bases its detection of the end of the range containing the key… based on `=` (the `[^*]+` is kinda pointless given you have a `.*` right after it, so the real thing that makes it stop at end of key is the `=` in your RegEx after all)

Which means that while this red herring passes your test, the same line but with also some addition `=` somewhere in the line (be it in the key, the translation, or as a `/* comment */` at the end of the line) would still throw you off rails.

mokagio: In [`6478d58`](https://github.com/wordpress-mobile/release-toolkit/pull/342/commits/6478d58557f477bee78278b0c2e42498de5c4889) I added more tests inputs based on my understanding of your [suggestion](https://github.com/wordpress-mobile/release-toolkit/pull/342/#discussion_r816760766) and the observation above: 

> It also won't work with keys that contain \" in their name — because of the [^"]+ thing. The .* that follows might accidentally workaround this, actually rendering the preceding [^"]+ kinda useless, but that also risks making the .* be greedy (e.g. "Login to a \"%@\" account" = "Login to a \"%@\" account"; // Here %@ = the service name, like Google)

The tests passed on my machine. I need to jump offline so I don't know if they'll pass on CI, too, but I don't see why not (_last famous words_).

If I misunderstood your suggestion, could you suggest more inputs for `file-with-duplicated-keys.strings` so we can get the tests failing and go from there? Thanks.

spec/test-data/translations/file-with-duplicated-keys.strings
```diff
@@ -56,11 +56,19 @@ Another line.";
...
 "key with \"%@\" character" = "localized translation here";
+/* Duplicated keys that include escaped quotes, interleaved */
 "key with multiple \"%@\" escapes \":)\" in it" = "localized translation here";
+"Login to a \"%@\" account" = "Login to a \"%@\" account"; // Here %@ = the service name, like Google
 "key with multiple \"%@\" escapes \":)\" in it" = "localized translation here";
-/* This is just a check for Gio's implementation. We might want to remove it? */
+"Login to a \"%@\" account" = "Login to a \"%@\" account"; // Here %@ = the service name, like Google
+/* Duplicated keys with escaped quotes and equals */
+"key with \"%@\" and = character" = "localized translation here";
+"key with \"%@\" and = character" = "localized translation here";
```
AliSoftware, line 72: The idea (from what I talked in my comment) would have been to have a red herring here instead of an actual duplicate, to show that the RegEx would mistakenly see `"key with \"%@\" and =` as the key + equal sign left of the key, and not see the rest of the key (i.e. ` character`)

So I think the goal of the test here should instead be:

```suggestion
"key with \"%@\" and = character for case 1" = "localized translation here";
"key with \"%@\" and = character for case 2" = "localized translation here";
```
And then make the test ensure that those would NOT be considered the same, duplicate keys, but be properly seen as different ones.

AliSoftware: @mokagio Played around with https://regex101.com/ to find some more red herring cases, and added one of them in 8d4eb9e, which makes the test fail — proving that the current RegEx is not resilient enough 😒 

The more I think about it the more I think a mini-state-machine line parser — which would read one character at a time, notice that it's in an escaping context when it encounters a backslash and eats the following character without interpreting it in such case, and notice when we enter/exit a quoting context… — might be a safer option?

_(As a bonus, such a solution would also support unquoted keys — like we have in `InfoPlist.strings` files; though that action to find duplicates is more likely to be run on files that we get back from GlotPress rather than files manually maintained, and those GlotPress files probably always get their keys quoted during export… 🤷 )_

Some alternatives to going the mini-state-machine route would be:
 - Either to first `gsub('\\', '\u0001').'gsub('\"', '\u0001')` (using `\0001` for its high unlikeliness to be used in real keys) before running the RegEx, thus avoiding it to risk being tripped by escaped quotes, but also double-tripped by "escaped backslash right before proper unescaped quote" (eg `"key ends in backslash char \\"`. And then we can just us a RegEx that will match `^\s*"([^"]+)"`. Not sure how confident we could be of such a hacky solution though…
 - Or alternatively, work even harder in a RegEx that would be able to represent a rule of "any character other than a quote… except if that quote is preceeded by a backslash". But again we might get tripped by "double-backslash before quote" case… so that might make the RegEx quite complicated, unreadable, and hard to trust as well.

Hence the idea of the mini-state-machine starting to feel more and more appealing to me… wdyt?

AliSoftware: @mokagio Couldn't resist at taking a stab at how a state machine would look 😅  You can see my attempt [in the `duplicate-keys-state-machine` branch](https://github.com/wordpress-mobile/release-toolkit/compare/add-helper-to-lint-for-duplicated-keys-in-strings..duplicate-keys-state-machine) 🙃 

Make what you will of it, build on it or toss it in the trash and ignore it completely if you think that's not the right approach, up to you… but figured I'd share it anyway 😃 

Btw, that attempt made me discover that your example `.strings` fixture had a syntax error — which my state machine detected, while your code couldn't 😛 — where a `;` was missing at the end of line 14 😇

mokagio: @AliSoftware 

> The idea (from what I talked in my comment) would have been to have a red herring here instead of an actual duplicate

Duh 🤦‍♂️ I focused more on the "with `=`" part than on the actual inner workings of the RegEx and how to trick it.

>  Couldn't resist at taking a stab at how a state machine would look 😅 

Oh sweet. Thanks. Given it works better than the current implementation, I think we should use it.

It is more verbose than the substitution approach, but I feel like the substitution is "dirty". It's a neat trick, but the state machine approach seems more solid. You also mention working even harder on the RegEx, but I agree that we might end up shooting ourselves in the foot with that.

mokagio: @AliSoftware I cherry-picked your work, then made it the only implementation via [`f4b5db2`](https://github.com/wordpress-mobile/release-toolkit/pull/342/commits/f4b5db2ece4e90441844a36849e56d86aaebcf6b).

I'll keep working on this next week, my idea is to extract this into the helper as discussed [above](https://github.com/wordpress-mobile/release-toolkit/pull/342#discussion_r816748689). Then, I think I'd call this PR "done" and open a followup to use the new helper in a Fastlane action.

spec/strings_file_validation_helper_spec.rb
```diff
@@ -5,6 +5,98 @@ module Fastlane
...
+        state = State.new(context: :root, buffer: StringIO.new, in_escaped_ctx: false, found_key: nil)
+
+        File.readlines(file).each_with_index do |line, line_no|
+          line.chars.each_with_index do |c, col_no|
+            # Handle escaped characters at a global level
+            if (!state.in_escaped_ctx && c == '\\') || state.in_escaped_ctx
+              state.buffer.write(c) if state.context == :in_quoted_key
+              state.in_escaped_ctx = !state.in_escaped_ctx
+              next
+            end
```
AliSoftware: I just realized that doing this at global level independently of the current context means the state machine accepts escaped chars in `:root` context (i.e. outside of any comment or quoted key or value), while in practice this is not valid.

We should:
 - either keep it at global level but add more conditions to the `if`, especially `&& [:in_quoted_key, :in_quoted_value].include?(state.context)`
 - or find a way to include this escape handling as part of the `TRANSITIONS` table — but going that route could prove a bit tricky and make it quite unreadable, because for example when in `:in_quoted_key` context, we'd have to act differently when matching `'"'` depending on if `in_escaped_ctx` is `true` (then just read the char in buffer) or `false` (do what the current `lambda` does). Otoh that would allow us to avoid having that escape processing being separate to the rest of the rules…

You might want to experiment with different solutions (handle escapes separately/globally vs at each state transition)  to find the one that is most readable.

AliSoftware (dismissed):

spec/strings_file_validation_helper_spec.rb
AliSoftware: Dammit, "approved" the PR instead of just "commenting" apparently… but can't find a way to discard my accidentaly approval (probably because the PR is still Draft?). Hoping that re-requesting my own review will do it?

EDIT: Confirmed that I had to undrafting the PR to be able to dismiss my accidental approval (before drafting the PR again). 🤷

mokagio: > EDIT: Confirmed that I had to undrafting the PR to be able to dismiss my accidental approval (before drafting the PR again). 🤷

TIL

mokagio:
@AliSoftware "I'll keep working on it next week" turned out to be more like "I'll be able to pick this back up in two weeks+"... 

Anyways, I think I'm "done" with this part. Let me know what you think.

spec/test-data/translations/strings-with-escape-character-in-root-context.strings
```diff
@@ -0,0 +1,6 @@
+/* This file has characters root context (not in a comment, or key, or value), which is not a valid `.strings` syntax */
+"key" = "value";
+"other key" = "other value";
+\\
```
mokagio, line 8: @AliSoftware is this what you were thinking of when talking about escaped characters in the root context?

Is there any other example we should add a it's own `.strings` file to ensure the helper catches it?

Thanks

AliSoftware: Yes that's what I was thinking about when I talked about escaped characters in the root context.
(And if I'm not mistaken, such a `\\` in such a position in the `.strings` file is not supposed to be valid)

AliSoftware:
> @AliSoftware "I'll keep working on it next week" turned out to be more like "I'll be able to pick this back up in two weeks+"...

Same feel here, as I thought I'd have time to review this PR today but didn't… and now I'm tempted to say I'll review it tomorrow or later this week, but hoping that I am not jinxing it 😄

spec/strings_file_validation_helper_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      expect(
+        described_class.find_duplicated_keys(
+          file: File.join(test_data_dir, 'ios_l10n_helper', 'expected-merged.strings')
+        )
+      ).to be_empty
+      expect(
+        described_class.find_duplicated_keys(
+          file: File.join(test_data_dir, 'ios_extract_keys_from_strings_files', 'Resources', 'en.lproj', 'Localizable.strings')
+        )
+      ).to be_empty
```
mokagio: Something I forgot in latest round of comments: Looking at the test data to write this example made me remember the helper currently only supports text-formatted `.strings` file, not XML or binary-plist.

I think you, @AliSoftware, mentioned it already, actually.

I don't recall whether we came to a conclusion, but I wonder whether we should expand to support multiple formats, or start with the assumption this will only be used on the "standard" text-formatted `Localizable.strings` and not worry about other formats for now. 

I like the second option because it would allow us to start using this sooner rather than later.

AliSoftware: Yes I'd go with second option too, especially because we already are late on our schedule/deadline for the whole Localization Tooling project, and this action was just a bonus/security anyway (given the core issue of duplicate keys seems to have been fixed since anyway) — while we really need to start working on other things like `.po` file generation rewrite and stuff (but unexpected stuff kept getting in the way) 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
```diff
@@ -0,0 +1,129 @@
+module Fastlane
+  module Helper
+    class StringsFileValidationHelper
```
AliSoftware: Should probably be in an iOS-related namespace and be named with the `ios_` prefix, given that this is all specific to iOS `.strings` files and not with Android's `strings.xml` files?

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+
+        File.readlines(file).each_with_index do |line, line_no|
+          line.chars.each_with_index do |c, col_no|
+            # Handle escaped characters at a global level. This is more
+            # straightforward than having a `TRANSITIONS` table that account
+            # for it.
+            if state.in_escaped_ctx || c == '\\'
+              # Just because we check for escaped characters at the global
+              # level, it doesn't mean we allow them in every context.
+              raise "Found escaped character outside of allowed contexts (current context: #{state.context})" unless %i[in_quoted_key in_quoted_value in_block_comment].include?(state.context)
```
AliSoftware: I think `:in_line_comment` should also be one of the valid context where escaped characters should be allowed?

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+      def self.get_keys_from_strings(file:)
+        keys_with_lines = {}
+
+        state = State.new(context: :root, buffer: StringIO.new, in_escaped_ctx: false, found_key: nil)
+
+        File.readlines(file).each_with_index do |line, line_no|
+          line.chars.each_with_index do |c, col_no|
+            # Handle escaped characters at a global level. This is more
+            # straightforward than having a `TRANSITIONS` table that account
+            # for it.
```
AliSoftware: Not the first time I see you do hard-line-wrap instead of letting your IDE do the soft-wrapping depending on your IDE's preferences and line length settings 🤔 

Any strong reason for it? I mean, I know that it's gonna be akin to the old "tabs vs spaces" eternal debate asking this here 😅  but what surprises me most is that you always hard-wrap at 80 characters, while [even our current `rubocop` settings sets a way higher limit of 300](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/.rubocop.yml#L60-L61) there.

I'm not completely against hard-wrapping as opposed to soft-wrapping… as long as it's consistent with the line length we have set ourselves for all the rest of the codebase (i.e. use the same limit for the comments as the one forced by rubocop for the code itself) 🙃 

Also (and this is mostly the main reason why I've noticed your 80-cols manual hard-wrapping in the first place), the fact that doing so leads to odd splits with comment lines sometimes only containing one or 2 words makes it a bit of an odd ball imho, as it increases vertical space those doc comments take just to account for 2 lone words 🤷 

Again, not a big deal, and very opinionated take from me here obviously, but since that's not the first time I see such lone words isolated on their own comment line _just_ to make that hard-wrap at 80-cols, I figured I'd still mention it 😃 . Personally I prefer more lose wrapping rules and using CRLFs for semantic breaks, e.g.:

```suggestion
            # Handle escaped characters at a global level.
            # This is more straightforward than having a `TRANSITIONS` table that account for it.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+            if state.in_escaped_ctx || c == '\\'
+              # Just because we check for escaped characters at the global
+              # level, it doesn't mean we allow them in every context.
```
AliSoftware: Likewise here: without forcing yourself to do the hard-wrapping, that comment line end up being as long as the `raise "…" unless …` line below it anyway… so it feels a bit strange to hard-wrap the comment… while we don't hard-wrap the code itself 🙃 

```suggestion
              # Just because we check for escaped characters at the global level, it doesn't mean we allow them in every context.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
AliSoftware: fix applying to the 2 previous suggestions in above comments at once:
```suggestion
              # Just because we check for escaped characters at the global level, it doesn't mean we allow them in every context.
              raise "Found escaped character outside of allowed contexts (current context: #{state.context})" unless %i[in_quoted_key in_quoted_value in_line_comment in_block_comment].include?(state.context)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+
+            state.context = next_context.is_a?(Proc) ? next_context.call(state, c) : next_context
+            next unless state.found_key
+
+            # If we just exited the :in_quoted_key context and thus have found
+            # a new key, process it
+            key = state.found_key.dup
+            state.found_key = nil
+
+            keys_with_lines[key] = keys_with_lines[key].nil? ? [line_no + 1] : keys_with_lines[key].push(line_no + 1)
```
AliSoftware: Now that I think about it, I think there's a way in Ruby to provide [a default value when none is associated with a key](https://ruby-doc.org/core-3.1.1/Hash.html#method-c-new), something like:

```ruby
 # Use this on line 70 instead of the keys_with_lines = {}
keys_with_lines = Hash.new([])
…
# then here, that should simplify this code to just this, given that `keys_with:
keys_with_lines[key] += [line_no + 1]
```

_PS: Note that if we go that route, we would need to use `+= [x]` instead of `.push(x)` because `hash[key]` only **returns** the default value if the `hash` does not have one for `key`, but would not **set** that default value back in the `hash` (so `hash[key].push(x)` would just have `hash[key]` return a `[]` empty array instance, push a value `x` into it… then discard the result, without ever storing it back into `hash`, while `+=` does store it back._

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+        get_keys_from_strings(file: file).reduce([]) do |accumulator, (key, lines)|
+          if lines.length >= 2
+            # Duplicate key, keep
+            accumulator.append({ key: key, lines: lines })
+          else
+            # Valid key, discard
+            accumulator
+          end
+        end
+      end
```
AliSoftware: I think you should be able to avoid the `reduce` complexity completely here and just call `delete_if` or `keep_if` on the hash instead:
```suggestion
      def self.find_duplicated_keys(file:)
        get_keys_from_strings(file: file).keep_if { |_, lines| lines.length > 1 }
      end
```

AliSoftware (commented):
I left a couple of comments and suggestions to fix a missing context where escaped is allowed, the namespacing of the helper to iOS, and simplify code significantly.

_I also just started a tab-vs-space-like war with you 😅 which is not really specific to this PR, so feel free to discuss the topic in P2 or Slack if you want to discuss it more in details though_

Also, I just noticed that this only adds the helper, but not yet an action calling it. I'm totally fine merging the helper PR first and adding the action in a subsequent PR, but wonder if it wouldn't be easier to include the action in the PR to give you the occasion to validate that the API will match exactly what you need when you start using it at call site in the action vs if you realize while starting using it that you might have missed a useful case or something.

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
mokagio: > Any strong reason for it? I mean, I know that it's gonna be akin to the old "tabs vs spaces" eternal debate asking this here 😅 but what surprises me most is that you always hard-wrap at 80 characters

No _strong_ reason. My rationale for doing it is to avoid having to scroll horizontally when looking at diffs:

Screen Shot 2022-04-07 at 12 38 37 pm

I'm sure there's ways to introduce soft-wraps in most IDEs and diff tool. [GitHub _used_ to do it](https://github.blog/2013-12-03-soft-wrapping-on-prose-diffs/). Still, by enforcing short line in the source, I don't have to worry about it.

It obviously comes with tradeoffs. On top of what you already highlighted, it's also annoying to have to reselect and reformat comments every time they're changed because of that. But the result is net positive, IMHO.

Maybe this is a good chance to discuss the approach to take in this repo and document the choice?

mokagio: > [GitHub used to do it](https://github.blog/2013-12-03-soft-wrapping-on-prose-diffs/).

Interestingly, the suggestions render with soft wrap, but the diff in the PR `/files` do not 🤔

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
mokagio: Brilliant! I'll try that.

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
mokagio: Wow, cool methods 😄 

To use `keep_if`, we'd need to update the return type from `get_keys_from_strings` to be already in the `{ key:, lines: }` format. That's okay for me. 

Originally, I called `.select` then `.map`, then replaced with `reduce` to iterate only once. To be fair, though, because the `.map` would happen after the `.keep_if` or `.select`, it would only touch the limited list of results, which would make the second iteration negligible in most case. Also... this is all premature optimization 🙃 

I was thinking about this after signing off and I'm in fact a bit unsure about the two methods returning data in a different format. But, while `[{ key:, line: }]` seems good for `find_duplicated_keys(file:)` (which I now realized I should rename to `find_repeated_keys(file:)`, `Hash<key:, line:>` seemed better for the other.

🤔 Maybe I could get the best of both worlds by returning this from `get_keys_from_strings`:

```ruby
{
		foo: { key: 'foo', lines: [1] },
		bar: { key: 'bar', lines: [2] }
}
```

In `find_repeated_keys` I could then call `.values .keep_if`. What do you think?

AliSoftware: Tbh I'm not sure I see the point of using the `Array<Hash<key:,lines:>>` intermediate format (ie `[{key: 'foo', lines: [2,6,12]}, {key: 'bar', lines: [7,42,1337]}]`) while this is essentially an unordered mapping listing an array of lines associated with each key anyway … aka a `Hash`

So personally I'd go with `Hash<String, Array<Int>>` (ie `{'foo': [2,6,12], 'bar': [7,42,1337]}`) for both cases.

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
AliSoftware: > My rationale for doing it is to avoid having to scroll horizontally when looking at diffs:

I'm not sure if this is because they're currently A/B testing their new beta UI and I have it but your don't, or if it's a user setting in GitHub prefs that would make it differ between what you see in GitHub and what I see…

…but personally when I look at the diff in `/files` I have **plenty of space** left horizontally (largely enough room to fit waayyy more than 80 characters! 😛 ) before I'd need to scroll horizontally when viewing the diff on GitHub 🤔 

image

mokagio: Well if you put it like that...

Screen Shot 2022-04-11 at 1 26 18 pm

🙃

I formed this habit before MacBooks had Retina displays or I could afford and extra wide external monitor, and it's based on really [old conventions](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting) from back in the day. Maybe it's time to revisit the length at which I make my editor hard-wrap...

What would be the "best" value for you @AliSoftware? I'll open a dedicated PR for that.

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
mokagio: Fair point... I thought of the keys to make the data clearer to access, but the use case and data footprint is so small that the simpler `{ foo: [2,6,12] }` doesn't cause any confusion. 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
mokagio: Ah! I kept it generic because it seemed clear to me "strings file" referred to `.strings`. I hadn't considered the `strings.xml` naming convention in Android 🤦‍♂️  Thanks.

spec/test-data/translations/strings-with-escape-character-in-root-context.strings
mokagio: > And if I'm not mistaken, such a `\\` in such a position in the `.strings` file is not supposed to be valid

That seems to be the case:

image

However, if I move the `\\` down in the file, I don't get the error 🤔 

image

Not sure whether that's an Xcode issue or not 🤔

mokagio:
@AliSoftware, I should have implemented all your suggestions, minus the line length tweak.

- [Helper namespace in `Ios`](https://github.com/wordpress-mobile/release-toolkit/pull/342/commits/3cda14f17abd46fc8f8c7505e24d4841ae45f6cd)
- [Allow escaped characters in inline comments](https://github.com/wordpress-mobile/release-toolkit/pull/342/commits/65f6507e5180525cc69f1ce39465b0707c3f6d2e)
- [Use a simpler `{ 'key' => [1,2,3] }` format](https://github.com/wordpress-mobile/release-toolkit/pull/342/commits/e35206ba77c917338c0e3fd6bd1e27641df803ea)

Let me know what you think. Thanks!

spec/test-data/translations/strings-with-escape-character-in-root-context.strings
AliSoftware: I've seen `plutil` consisting an invalid Plist (or .strings) file to be valid but truncated before (and I'm assuming Xcode might rely on the same logic as plutil for this?). Especially I remember a while ago having a strings file missing a semicolon at the end of one line and `plutil -lint` telling me that the file was linting OK… But `plutil -p` just stopping parsing after that line and totally ignoring any content after that missing semicolon like if no content existed after the syntax error.

I wonder what you get, and especially if there might be missing the "final key" in the output, when running `plutil -p` on that second example that Xcode fail to warn you about … could it be the same bug? If so, our parser might be smarter at detecting syntax errors than Xcode's / `plutil` 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/strings_file_validation_helper.rb
AliSoftware: > I formed this habit before MacBooks had Retina displays or I could afford and extra wide external monitor

Oh yes, and I do remember that time where I limited everything to 80 cols too and I had smaller monitors too 😛 But computer screen sizes have evolved since, and most importantly this length limit was for back in the days when Terminals were also 80 cols. Even if I were to run `bat` to list code in my Terminal, my terminal's preferences is set to use 240x60 by default, so… 🙃 

> based on really [old conventions](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting) from back in the day

Btw for _commit messages_ I'm ok with trying to limit the length of the first line; but I'd not hard-wrap the following lines (aka the "summary"). So my rule here would be maybe something like 50/∞ instead of the 50/72 mentioned in your SO link 😛 

> What would be the "best" value for you @AliSoftware? I'll open a dedicated PR for that.

I have no strict limit to give you, as in practice:
 - I don't really set a hard limit at which I should hard-wrap
 - And instead I generally don't hard wrap (but I go to the next line to break long prose in paragraphs…)
 - Or when I do, I just use my gut feel to cut at a point in the very long lines (>200, 300?) where the hard-wrap would (1) not feel too odd in this part of the sentence, and (2) does not create imbalance with the rest of the line being wrapped. For example if a line is 310 characters long, I'll not hard-wrap it at 300 characters and leave the remaining 10 characters in the 2nd line, but instead maybe cut around 160/150? Or 120, 120, 70?

So basically I don't really have a hard rule myself… except that the rule that I don't have a hard rule 😛  and a hard line length limit that I'd follow blindly.
That being said, this repo is [currently configured to have a line length of 300 already for ruby code](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/.rubocop.yml#L60-L61), so if you really want a number I'd go with that one for no particular reason other than that's what our config currently says (and what I believe rubocop suggested as a value back when we set it up based on its analysis of our current code base and max line lengths used throughout our existing code) 🙃

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_strings_file_validation_helper.rb
```diff
@@ -0,0 +1,124 @@
...
+        # Inspects the given `.strings` file for duplicated keys, returning them
+        # if any.
+        #
+        # @param [String] file The path to the file to inspect.
+        # @return [Array<Hash>] Array of all the duplicated keys. Each entry is a
+        #         a `Hash` with keys: `key` for the key value and `lines` with an
+        #         array of line numbers at which the key occurs.
+        def self.find_duplicated_keys(file:)
+          get_keys_from_strings(file: file).keep_if { |_, lines| lines.count > 1 }
+        end
```
AliSoftware: Given how simple this one has become, I wonder if it's still necessary to have 2 separate methods now.

Could make for some nice tidying to remove the `get_keys_from_strings(file:)` method, and move its whole implementation directly into this `find_duplicated_keys` instead of this one-liner of one calling the other, in order to merge the 2 methods into one, wdyt?

spec/test-data/translations/strings-with-escape-character-in-root-context.strings
AliSoftware: Bingo, just tested it and that's exactly such a case of `plutil` falling into a trap (and Xcode syntax checker falling into the same one I guess), and ending the parsing on encountering that the syntax error instead of erroring:

```
✓ $ cat spec/test-data/translations/strings-with-escape-character-in-root-context.strings      
/* This file has characters root context (not in a comment, or key, or value), which is not a valid `.strings` syntax */
"key" = "value";
"other key" = "other value";

/* Escaped characters should be allowed in block comments \\ */
"yet another key" = "yet another value"; // Escapes should be allowed in inline comments, too \\

\\
/* The parsing shouldn't reach this, line. It should tell us the escape up there is not valid. */
"final key" = "final value";
                                                                                                                                                                                                                                                
✓ $ plutil -lint spec/test-data/translations/strings-with-escape-character-in-root-context.strings
spec/test-data/translations/strings-with-escape-character-in-root-context.strings: OK
                                                                                                                                                                                                                                                
✓ $ plutil -p spec/test-data/translations/strings-with-escape-character-in-root-context.strings   
{
  "key" => "value"
  "other key" => "other value"
  "yet another key" => "yet another value"
}
```
Notice how lint says the file is OK (despite the `\\` being invalid), and then `plutil -p` printing only the keys before the `\\`, but failing to see the `"final key"` and omitting it completely.

So… known bug of `plutil` and `Xcode` (that our parser manages to avoid!)

spec/test-data/translations/strings-with-escape-character-in-root-context.strings
```diff
@@ -0,0 +1,10 @@
+/* This file has characters root context (not in a comment, or key, or value), which is not a valid `.strings` syntax */
+"key" = "value";
+"other key" = "other value";
+
+/* Escaped characters should be allowed in block comments \\ */
+"yet another key" = "yet another value"; // Escapes should be allowed in inline comments, too \\
```
AliSoftware, line 6: I was about to comment in the PR that we should add a case of a `\\` in an inline comment as well… but then just saw you thought of that too with this case here 👍 👌

AliSoftware (commented):
The latest changes look good 👍 

Not sure if you want to address the comments' line length in this PR or the next? So I'll leave that open for you to decide if you want to finish it here or later.
I've also left a nitpick/suggestion to cleanup helper methods now that you've simplified the code by both removing the `reduce` and making the return value format be uniform as a `Hash`. Take it or leave it 🙃

AliSoftware:
Added this PR to [our L10n GitHub Project](https://github.com/orgs/wordpress-mobile/projects/105/views/8) (I always forget about adding PRs to this one too… and I should take some time to update its cards a bit to reflect the progress)

spec/test-data/translations/strings-with-escape-character-in-root-context.strings
mokagio: > that our parser manages to avoid!

🕺 💃

spec/test-data/translations/strings-with-escape-character-in-root-context.strings
```diff
@@ -0,0 +1,10 @@
...
+
+\\
+/* The parsing shouldn't reach this, line. It should tell us the escape up there is not valid. */
```
AliSoftware:
```suggestion
/* The parsing shouldn't reach this line. It should tell us the escape up there is not valid. */
```
Typo I just spotted btw

mokagio:
> Not sure if you want to address the comments' line length in this PR or the next?

[I'll open a dedicated PR for that](https://github.com/wordpress-mobile/release-toolkit/pull/342#discussion_r846921049), so I can take the time to address all (hopefully) the short lines / hard wraps.

I address the rest. I haven't enabled auto-merge because I want to squash the WIP commits at the start of the branch (🙄 🤦‍♂️ me). Apologies in advance if there'll be noise post-approval.

AliSoftware (approved):
:shipit:

AliSoftware:
@mokagio I've encountered an issue in [another PR](https://github.com/wordpress-mobile/release-toolkit/pull/354#issuecomment-1097088421) where [CI started to fail](https://buildkite.com/automattic/release-toolkit/builds/472#23320b23-b856-4e15-ae7f-694e43ffc719) due to the specs introduced by this PR:
```


1) FastlaneCore::Helper::Ios::StringsFileValidationHelper when there are no duplicated keys returns an empty array
--
  | Failure/Error: raise "Invalid character `#{c}` found on line #{line_no + 1}, col #{col_no + 1}" if next_context.nil?
  |  
  | RuntimeError:
  | Invalid character `I` found on line 21, col 1

```
I took a very quick look and it appears that this line 21 of the file used on this specific example is [this one](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/test-data/translations/ios_l10n_helper/expected-merged.strings#L21), in other words it's on a key merged from `InfoPlist.strings` that is not quoted.

Thinking about it, I'm not very surprised that the test fails after all, because our state machine does not support unquoted keys (as you can see from the `TRANSITIONS` constant. That was kind of intentional of me to leave that unsupported in my initial implementation of that state machine, to keep it simple, and also because I'd expect us to run your helper only on `.strings` files that come back from GlotPress, which would always have their keys quoted anyway.

What I'm more surprised about, is how the CI for this PR passed and allowed us to merge that PR in the first place though.

I'm not sure what's the best way forward:
 - Make [the spec use a test file other that this `expected-merged.strings` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/ios_strings_file_validation_helper_spec.rb#L45) you're piggy-backing on, since the root cause is that this file contains unquoted keys
 - Improve our state machine to also support unquoted keys (is it worth it considering the use case for that helper though?)

spec/ios_strings_file_validation_helper_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  context 'when there are no duplicated keys' do
+    it 'returns an empty array' do
+      # Piggy back on some of the `.strings` from other tests to ensure this
+      # behaves correctly
+      expect(
+        described_class.find_duplicated_keys(
+          file: File.join(test_data_dir, 'ios_l10n_helper', 'expected-merged.strings')
```
AliSoftware, line 45: This one contains entries with unquoted keys, which our state-machine does not support… so that spec now fails on all follow-up PRs 😢 
No idea why it didn't fail on CI before we merged that PR, that's quite suspicious 🤔

AliSoftware: Closing the loop: I've fixed the test failure this introduced by creating https://github.com/wordpress-mobile/release-toolkit/pull/357

PR #343, opened 2022-03-01
Update `Gemfile.lock` (via `bundle install`) after 3.1.0 release

mokagio:
I guess we should add the check from the version in the `Gemfile.lock` matching that from `version.rb` as part of the CI release process? Or maybe update the `Rakefile` task to run it for us before committing the version bump? Probably both, just in case 😅


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    fastlane-plugin-wpmreleasetoolkit (3.0.0)
+    fastlane-plugin-wpmreleasetoolkit (3.1.0)
```
mokagio, line 4: I noticed this because Peril told me [here](https://github.com/wordpress-mobile/release-toolkit/pull/342#issuecomment-1054957608)
Screen Shot 2022-03-01 at 3 39 48 pm

AliSoftware (approved):
Definitively worth automating the run of `bundle install` during the `rake new_release` automation 💯

AliSoftware:
> Definitively worth automating the run of `bundle install` during the `rake new_release` automation 💯

Mmmmh wait, went to create a PR to add this, until I realized this was already handled by the Rakefile [here](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/Rakefile#L89)… so either there's a glitch in the logic, or whoever made the `3.1.0` release (cc @jkmassel ?) didn't use the script — which is why this step was missed?… 🤔

mokagio:
Okay, good to know.

I skimmed the file before suggesting, but clearly didn't get to the bottom 🙃

We'll, then this is something to keep an eye out next time around. 🤞

PR #345, opened 2022-03-09
[L10n actions] Allow iOS merge and extract strings actions to add/remove prefixes to keys

AliSoftware:
## Why?

A project can have multiple `InfoPlist.strings` files, or can have keys that the same in multiple `.strings` files but we still want to distinguish them when merging the `.strings` files into one for GlotPress import, and when extracting then back after GlotPress export.

A typical use case is in WPiOS (in fact I discovered/realized about it while working on https://github.com/wordpress-mobile/WordPress-iOS/pull/18061), where there are 3 `InfoPlist.strings` files in the whole repo: one for the `WordPress` target, one for the `Jetpack` target, and one for the `WordpressDraftActionExtension` target (to provide the name of the action when it appears in the iOS share sheet).

## How?

To support such cases, I implemented a mechanism to add an optional prefix during `ios_merge_strings_files` to all the keys of a file being merged (same prefix for all the keys of a given file, but we can and should use a different prefix for different files, to allow us to distinguish where the original keys come from).

Then during `ios_extract_keys_from_strings_files`, a similar mechanism allows us to remove prefixes from those keys during the extraction back into the separate files.

<details><summary><h2>Note on prefixes support not changing the extraction strategy</h2></summary>

Note that while, by adding that feature, we could have decided to then rely on the key prefixes to decide which keys to extract in which files (i.e. "all keys prefixed with `draft-extension.` would go in the `WordPressDraftActionExtension/*.lproj/InfoPlist.strings` files), instead of using the current strategy I've relied on so far which is looking at the keys in the original file (`WordPressDraftActionExtension/en.lproj/InfoPlist.strings`) to know which keys should be extracted from the translated files.

I decided to keep the current strategy (rely on the keys found in the original `en.lproj`) rather than rely on prefixes, because:
 - Prefixes are still an optional feature that one can use or not. It's totally valid for the caller to decide in their `Fastfile` to use an empty prefix (or `nil` prefix) for some of the files being merged, or to use the same prefix for more than one of the merge files, etc.
 - So the use of prefixes is not mandatory when calling the action in your client apps, and their goal is more to add more flexibility and as a way to avoid duplicate keys during merge than to enforce a redistribution strategy

For example, in the WCiOS repo, which only has one app target (as opposed to WP + JP in WPiOS repo) and no app extension target, there will only be the `Localizable.strings` and the `InfoPlist.strings` files of the main app target to merge, nothing else, and it's quite unlikely that keys from `InfoPlist.strings` (like `NSCameraUsageDescription`) would clash with in-app string keys. So there is no obligation for WCiOS to use prefixes (unlike WPiOS which will need them), yet we still want the strategy to know which key to extract to still work in that context.

</details>


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
```diff
@@ -40,8 +41,8 @@ def self.available_options
           FastlaneCore::ConfigItem.new(
             key: :paths,
             env_name: 'FL_IOS_MERGE_STRINGS_FILES_PATHS',
-            description: 'The paths of all the `.strings` files to merge together',
-            type: Array,
+            description: 'A hash of the paths of all the `.strings` files to merge together, with the prefix to be used for their keys as associated value',
+            type: Hash,
```
AliSoftware: Note that this is source-breaking and will require a major version bump (`4.0.0`).

I've considered trying to support _both_ `Array` and `Hash` — which would mean that the old API would still work, and call site would be simpler to use in case you don't need any prefix (like WCiOS). But for the time being I've judged that, given that new action is not used in any of the client repos yet, better keep a simple and consistent API rather that try to support both `Array` and `Hash` for flexibility and for avoiding breaking change.

I could have my mind changed though; I think supporting that would just require us to use `skip_type_validation: true` on that `ConfigItem`, do the type check ourselves to accept both `Hash` and `Array`, and then do some `paths = paths.map { |f| [f, nil] }.to_h if paths.is_a?(Array)` in the action's body… and finally add the relevant Unit Tests in the action's spec.

But supporting both would also **complicate the already-quite-convoluted documentation** (`description:` field) for that `ConfigItem`, but even more so for [the similar one on `ios_extract_keys_from_strings_files`](https://github.com/wordpress-mobile/release-toolkit/pull/345/files#r824132357), so that's also one more reason to maybe not try too hard to support both. 

wdyt?

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 
 ### Breaking Changes
 
-_None_
+* Update the API of `ios_merge_strings_files` and `ios_extract_keys_from_strings_files` to support using prefixes for string keys when merging/splitting the files.
+  The actions now expect a `Hash` (instead of an `Array`) for the list of files to provide an associated prefix (or `nil` or `''` when none) for each file to merge/split. [#345]
```
AliSoftware, line 10: See [this comment](https://github.com/wordpress-mobile/release-toolkit/pull/345/files#r822930611) and also [that one](https://github.com/wordpress-mobile/release-toolkit/pull/345/files#r824132357) for the rationale about this being a breaking change (and pros/cons of making it backwards-compatible or not)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -82,15 +84,16 @@ def self.available_options
...
-                                       type: Array,
+                                       description: 'The path(s) to the `<base-locale>.lproj/<target-tablename>.strings` file(s) for which we want to extract the keys to, and the prefix to remove from their keys. ' \
+                                        + 'Each key in the Hash should point to a file containing the original strings (typically `en` or `Base` locale), and will be used to determine which keys to extract from the `source_tablename`. ' \
+                                        + 'For each key, the associated value is an optional prefix to remove from the keys (which can be useful if you used a prefix during `ios_merge_strings_files` to avoid duplicates). Can be nil or empty if no prefix was used during merge for that file.' \
+                                        + 'Note: For each entry, the path(s) in which the translations will be extracted to will be the files with the same basename as the key in each of the other `*.lproj` sibling folders. ',
```
AliSoftware, line 107: It was already a bit tricky to explain that parameter while maintaining both brevity and clarity when it was expecting an `Array`. I feel like that documentation for the `Hash` makes sense and is understandable while still providing all the needed info though.

Note that if we were to decide to go the extra mile to make that action support **both** `Hash` and `Array` (see [this other comment](https://github.com/wordpress-mobile/release-toolkit/pull/345/files#r822930611)), that would make it a challenge of another level to add that case to the explanation, and would probably lead to something a bit too complex to document properly imho.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
```diff
@@ -4,7 +4,8 @@ class IosMergeStringsFilesAction < Action
       def self.run(params)
         UI.message "Merging strings files: #{params[:paths].inspect}"
 
-        duplicates = Fastlane::Helper::Ios::L10nHelper.merge_strings(paths: params[:paths], output_path: params[:destination])
+        destination = params[:destination] || params[:paths].keys.first # Use first path if destination not provided, to do an in-place merge.
```
AliSoftware: The "Use first path if destination not provided, to do an in-place merge" feature was documented ([here](https://github.com/wordpress-mobile/release-toolkit/pull/345/files#diff-ac9f9a8b31559602b0cd66aac55b4cebb35bf91cb9ce95b6d65182bb9b381177R51) on the `ConfigItem`) but in practice not implemented.
This missing case was previous missed by the unit tests, because I always provided an explicit value to that param in my tests — even when testing the case for that destination being the same as one of the input paths. This has been fixed by adding [this new test case](https://github.com/wordpress-mobile/release-toolkit/pull/345/files#diff-0b56c04c18a71437c58dce3f37b480545b0b090611e0ba2582319e7760398f46R40-R60).

spec/test-data/translations/ios_l10n_helper/InfoPlist-utf8.strings
```diff
@@ -1,3 +1,4 @@
-"InfoKey1" = "Info String 1";
-"InfoKey2" = "Info String 2️⃣";
-"InfoKey3" = "Info String 3";
+InfoKey1 = "Info String 1";
+InfoKey2 = "Info String 2️⃣.
+Please accept.";
```
AliSoftware, line 3: Notice the new trap to ensure we also support multiline translations which use literal newlines rather than `\n` escaped ones 😈

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -48,24 +48,33 @@ def self.strings_file_type(path:)
...
               # Read line-by-line to reduce memory footprint during content copy; Be sure to guess file encoding using the Byte-Order-Mark.
-              File.readlines(input_file, mode: 'rb:BOM|UTF-8').each { |line| tmp_file.write(line) }
+              File.readlines(input_file, mode: 'rb:BOM|UTF-8').each do |line|
+                if prefix
+                  # We need to ensure the line and RegExp are using the same encoding, so we transcode everything to UTF-8.
+                  line.encode!(Encoding::UTF_8)
+                  # The `/u` modifier on the RegExps is to make them UTF-8
+                  line.gsub!(/^(\s*")/u, "\\1#{prefix}") # Lines starting with a quote are considered to be start of a key; add prefix right after the quote
+                  line.gsub!(/^(\s*)([A-Z0-9_]+)(\s*=\s*")/ui, "\\1\"#{prefix}\\2\"\\3") # Lines starting with an identifier followed by a '=' are considered to be an unquoted key (typical in InfoPlist.strings files for example)
+                end
```
AliSoftware: I'm not super happy with having to use RegExps here to modify the input `.strings` file, especially given one of my recurring motto while working on this modernization of our Translation Tooling was that I wanted to avoid at all cost having to parse or write `.strings` files that are in OpenStep (textual) format like here, because that format is poorly documented (no official strict spec/grammar published afaik) and can be full of edge cases. This is the whole reason why I've been using `plutil` to parse the key/value pairs of those files instead of trying to parse the file myself, and why the `.strings` files we generate with those actions use the XML output format and not the undocumented textual format.

Alas, I'm not sure there's a better solution in order to insert those prefixes in the keys (and without losing the original `/* comments */` of those strings files so that GlotPress have them too on import, hence why we need to keep the textual OpenStep format for those files we plan to make GlotPress import…).

I'm hoping the couple of [red herrings I've added to my Unit Test fixtures](https://github.com/wordpress-mobile/release-toolkit/pull/345/files#diff-8fa1a86ac87d73ab1777cffbe9502a139b2385523e2caace86d38009bf3e42d0R15) to try to trip this (and show that we didn't fall into those traps) would be enough to cover us, despite still not liking having to resort to that…

spec/test-data/translations/ios_l10n_helper/expected-merged-prefixed.strings
```diff
@@ -0,0 +1,34 @@
...
+
+/* comment on key 1 */
+"main.key1" = "string 1";
+
+/* comment on key 2 */
+"main.key2" = "string 2️⃣";
+
+/* multiline comment on key 3.
+   %d = number of lines
+   and = a red herring for when merge_strings tries to add prefixes to lines it identifies as pairs using unquoted keys */
```
AliSoftware, line 15: Notice the traps to ensure that the code adding prefixes to keys does not accidentally add prefixes to those lines (and does not unexpectedly transform them to `"main.%d" = number of lines` or `"main.and" = a red herring …`)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
```diff
@@ -4,7 +4,8 @@ class IosMergeStringsFilesAction < Action
...
+        duplicates = Fastlane::Helper::Ios::L10nHelper.merge_strings(paths: params[:paths], output_path: destination)
         duplicates.each do |dup_key|
           UI.important "Duplicate key found while merging the `.strings` files: `#{dup_key}`"
         end
```
AliSoftware, line 15: 💡  Could be worth adding an additional `UI.important` help message at the end to suggest using that new prefix feature I just introduced in those cases:
```ruby
UI.important 'Tip: You might want to consider providing different prefixes in the `Hash` you used for the `paths:` parameter to avoid those key conflicts.' unless duplicates.empty?
```
_(Not writing this as a GitHub suggestion because such a change in the `UI.important` messages will likely have to be mirrored in the specs testing the action in parallel to here)_

spec/test-data/translations/ios_extract_keys_from_strings_files/WordPress-Resources/en.lproj/Localizable.strings
```diff
@@ -0,0 +1,28 @@
...
+"jetpack.infoplist.NSLocationUsageDescription" = "Jetpack would like to add your location to posts on sites where you have enabled geotagging.";
+"jetpack.infoplist.NSCameraUsageDescription" = "To take photos or videos to use in your posts.";
+
+/* MARK: - Sites.strings */
+
+"BOl9KQ" = "There are ${count} options matching ‘${site}’.";
+"ILcGmf" = "Site";
+"cyajMn" = "Site";
+"gpCwrM" = "Select Site";
+"s4dJhx" = "Just to confirm, you wanted ‘${site}’?";
```
AliSoftware: 💡  Maybe it could be worth to extend the Unit Tests to also use a prefix for those keys (like `ios-widget.`) and validate that they are also extracted in the `Sites.strings` files unprefixed (just like for the case of the 2 `InfoPlist.strings` files above)?

 - On one end, if the current test already proves it works for the case of the 2 `InfoPlist.strings` files, no reason it wouldn't work for `Sites.strings`, and we could thus decide to even remove those keys in that fixture.
 - On the other end, since I kept those keys in that fixture (mostly because I did a simple copy/paste of the similar-but-unprefixed fixture from `Resources/` to create that file, and did not think about removing those afterwards), why not take the occasion to use them and make our test cover more examples… 🤔

AliSoftware: Ok was quite quick to adjust the test to include them, so I already did it in 50d0a597331399611e7806af705d69e187e2d997 after all.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
AliSoftware: Implemented in f9f1dafee6922d017a6e9af26938dc65d10da5f7

mokagio (commented):
> same prefix for all the keys of a given file, but we can and should use a different prefix for different files, to allow us to distinguish where the original keys come from

Have you considered automating the prefix generation, perhaps based on the file path or using a hash of its path?

The approach might make the GlotPress output noisy, but not that much more confusing than keys with "human readable" prefixes that mean something only to developers.

I'm asking about that to avoid the consumers having to decide the prefix value. The idea came to mind because reading the usages in the tests seemed confusing, both

```ruby
target_original_files: {
            File.join(wp_resources_dir, 'en.lproj', 'InfoPlist.strings') => 'wordpress.infoplist.',
```

and the prefix-less alternative

```ruby
target_original_files: { File.join(wp_resources_dir, 'en.lproj', 'InfoPlist.strings') => nil }
```

AliSoftware:
> Have you considered automating the prefix generation, perhaps based on the file path or using a hash of its path?

Mmmh that's an interesting thought.

But ultimately I still want to avoid adding prefixes when not needed, especially because I want to avoid using the RegExp / `gsub` implementation in the 99% of cases/strings. (see also commit message in [55d8ac0](https://github.com/wordpress-mobile/release-toolkit/pull/345/commits/55d8ac06ec6d8d1edd5b7a0db21a52ebf2b89547))

Especially, for the keys in the main `Localizable.strings` (the ones extracted by `genstrings`) I don't want a prefix to be added to them (while precomputing a prefix automatically based on file path or similar) in real project situations, because:
 - Currently those keys are the English copy for most (if not all) the app strings. Adding a prefix to all of them would feel a bit weird in GlotPress
 - That `Localizable.strings` file is the one, at least in real projects, that contains most of the strings: probably like 99% of all the strings that will end up in the merged file will come from those.
 - Adding a prefix to _all_ those 99% of keys would be useless and quite verbose (and would lead to a lot more of `gsub` calls… with the potential performance implications) for not much gain, but more importantly would make us rely a lot more on those RegExp-based substitutions for adding prefixes, compared to only use those `gsub` calls for the 1% of strings from the `InfoPlist.strings` and `Sites.strings` (probably totaling not more than ≈10 strings needing those prefixes out of the… 1k? strings total)

So TL;DR: I want to reduce the use of prefixes to the **minimum** number of keys (`InfoPlist.strings` and `Sites.strings`) because I don't trust using RegExp on OpenStep `.strings` files that could be filled with edge cases.

---

As for generating the prefix automatically (eg we could consider use `true/false` for the `Hash` values instead of the prefix string, to decide which files should add a prefix or not to their keys), I'm not against the idea… but:
- I like the fact that by providing the prefix ourselves, we could make it human readable in GlotPress; and that also helps us humans if we need to debug stuff and make sense of which file a prefixed key is associated with
- By contrast, if we autogenerate the prefix, we can't only base it on the file's basename (as all `InfoPlist.strings` files — like WP, JP and the WordPressDraftActionExtension targets in the real case of the WPiOS repo — would have the same basename) while the very existence of those 3 `InfoPlist.strings` similar files is the very reason why we will need to use different prefixes for each
 - And basing the prefix on the file path might be dangerous and not resilient in case the call site use paths like `File.join(PROJECT_ROOT, 'path', 'to', 'file.strings')` which would end up being absolute paths and thus would lead to different prefixes depending on the machine where the action would be run from (and thus the Release Manager running the command)

spec/ios_l10n_helper_spec.rb
```diff
@@ -61,14 +61,30 @@ def file_encoding(path)
...
-        expect(File.read(paths.first)).to eq(File.read(fixture('expected-merged.strings')))
+        paths = paths.map { |f| [File.join(tmp_dir, File.basename(f)), nil] }.to_h
+        described_class.merge_strings(paths: paths, output_path: paths.keys.first)
+        expect(File.read(paths.keys.first)).to eq(File.read(fixture('expected-merged.strings')))
+      end
+    end
+
+    it 'adds prefixes to string keys when merging files' do
+      paths = {
+        fixture('Localizable-utf16.strings') => 'main.',
```
AliSoftware, line 79: Note that I used a `main.` prefix for the main `Localizable.strings` **even though in practice in real world situations, we will very likely _not_ use a prefix at all for the main `Localizable.strings` file** (which contains 99% of our strings), and only use prefixes for the way smaller `.strings` file we merge into it (`InfoPlist.strings`, `Sites.strings`).

See also commit comment in 5c509a1f62e980836e9063ba5f71e0481771ec67, and full rationale in https://github.com/wordpress-mobile/release-toolkit/pull/345#issuecomment-1064783710

The reason I used a `main.` prefix in that unit test here is mostly so I can test as much use cases as possible, especially ensure we avoid the various traps I left in the various `.strings` files when encountering edge cases of the OpenStep format (unquoted strings, values on multiple lines, `=` in comments, escaped characters…), so I wanted the unit test to apply the 2 `gsub` RegExps on as many cases as possible to make the test stronger.

That example is not supposed to be a reflection on how we will typically chose when to use prefixes or not in the real-world projects though, even though I realize now that this is probably what confused @mokagio into thinking so and led him to originally leave [this comment](https://github.com/wordpress-mobile/release-toolkit/pull/345#pullrequestreview-906759082)).

AliSoftware: Rationale above added as a code comment in the spec file itself via 6c08deb

spec/ios_l10n_helper_spec.rb
```diff
@@ -69,6 +69,13 @@ def file_encoding(path)
 
     it 'adds prefixes to string keys when merging files' do
       paths = {
+        # NOTE: In real-world applications we typically won't use a prefix for the main `Localizable.strings` file which contains most of the strings,
+        # but instead only use a prefix on **secondary** files, with low number of strings — because we want to avoid polluting most of the strings in
+        # GlotPress with a prefix if they don't need one, and also want to reduce the number of applications of the RexExp adding the prefixes to a minimum.
+        #
+        # But for the sake of this Unit Test, since the goal of this test is to validate the prefix insertion on as many test cases as possible,
+        # using a prefix on all the files — including the main one — allows us to widen the coverage of the test. This is not supposed to reflect
+        # how this action is going to be used in real-world projects (where the main strings file would typically not use a prefix in most scenarios).
```
mokagio, line 78: This is very useful, thank you.

I hope future readers will look at tests to understand how things work, and this clarification on testing a behavioral edge vs real world usage will be helpful.

mokagio (approved):
Thank you for the clarifications 👍 

I'm saddened by the extra complexity, but there's no way out of it, short of reverting to apps only in English or with no libraries or extensions.

What makes it fell a bit better is that this is mostly a set-and-forget configuration. Once the `Fastfile` automation is setup, no one needs to work with it anymore, only run it. You already provided good docs and examples in the unit tests that will help future users navigate the code and its usage 👍 

Let's ship this.

AliSoftware:
@mokagio actually I just thought of one possibility which could allow us to make this whole "no prefix on main file in typical real world usage " thing and the corresponding API call site maybe less awkward:

Instead on having `ios_merge_strings_files` have an API which allows the case of the first input file and the output file to not be the same, while in practice in real world projects it will always be the case, maybe we could change the semantics of that action so that it always _append_ to the existing destination file.

What I mean by that is that, with such a change, at call site:
 - The list of files to merge would no longer contain the main `Localizable.strings` file
 - Instead we'd have to pass that path to the main `Localizable.strings` file to the `destination` parameter of the action (which already exists, but so far is not used at call site with the current semantics because we make it default to the first input file… which would not be the case with the new semantics)

In addition to making the input list parameter less "mixed" than it currently is (with currently the first and main path in the list being the odd one out with no prefix), that would also make it less convoluted to share the list between the merge and extract actions call sites.

Because currently call site [looks like this](https://github.com/wordpress-mobile/WordPress-iOS/pull/18061/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R455-R458) and has to make use of the `{ main_file => nil, **MANUALLY_MAINTAINED_STRINGS_FILES }` hash splatting operator trick (which is not a Ruby feature known to many) but with the new semantics we'd just have:
```ruby
  paths: MANUALLY_MAINTAINED_STRINGS_FILES,
  destination: main_file
  ```

wdyt? Worth that small change before we commit to the final API via a major release?

Such a change would be quite straightforward to implement in the current code (basically we could just reassemble the destination and the rest of the list in `self.run` right before calling the helper function and nothing else would change in the merge logic) but would make a less convoluted call site — which currently only exists for a case that we will never need anyway (aka merging the other strings file in a new file rather than appending to the existing main one) so…

**Comment by mokagio** [comment]
@AliSoftware brilliant! Removing the odd one out will make the existence of the prefixes more straightforward.

I only have one suggestion. What do you think of `paths_to_merge` instead of `paths`? Repeating the "merge" word from the action name is a bit redundant, but I feel like it goes well with `destination`. No big deal, though.

**Comment by AliSoftware** [comment]
> I only have one suggestion. What do you think of paths_to_merge instead of paths? Repeating the "merge" word from the action name is a bit redundant, but I feel like it goes well with destination. No big deal, though.

I don't have any objection against `paths_to_merge` 👍  Will get to that change during my day.

**Comment by AliSoftware** [comment]
@mokagio The API change discussed above is now implemented in c1887ee8d8ae055fce224afa1a8405ba3ea3d321 (+ 4586e39f79ac99837a7da3264964c43ccc5dc30c).

I also made `ios_extract_keys_from_strings_files` return the list of written files (via 9e8f697) as I figured it could be useful at call site in order to know which files to commit (instead of blindly commit all `**/*.strings` files found in the repo)

Finally, I've updated the call site accordingly [in WPiOS's PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/18061) via https://github.com/wordpress-mobile/WordPress-iOS/pull/18061/commits/151657b761691fc149a768014468b96c01bd6738.

Feel free to re-review those last minute changes during your day and then Merge the PR yourself if you agree with the new changes 🙂

File: lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb [hunk 12]
```diff
@@ -2,10 +2,14 @@ module Fastlane
...
-        UI.message "Merging strings files: #{params[:paths].inspect}"
+        destination = params[:destination]
+        # Include the destination as the first of the files (without key prefixes) to be included in the merged result
+        all_paths_list = { destination => nil }.merge(params[:paths_to_merge])
 
-        destination = params[:destination] || params[:paths].keys.first # Use first path if destination not provided, to do an in-place merge.
-        duplicates = Fastlane::Helper::Ios::L10nHelper.merge_strings(paths: params[:paths], output_path: destination)
+        UI.message "Merging strings files #{all_paths_list.inspect}"
+
+        File.write(destination, '') unless File.exist?(destination) # Create empty destination file if it does not exist yet
```
mokagio, line 11: Micro readability nitpick. How do you feel about moving this logic into its own, well named, method? 

Something like:

```diff
- File.write(destination, '') unless File.exist?(destination) # Create empty destination file if it does not exist yet
+ File.create_empty_if_not_existing(destination)
```

mokagio (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_merge_strings_files.rb
AliSoftware: I'm not sure it's worth it tbh, especially given it's only used in one place here (and makes for a long and potentially debatable name for an extension method 😛

AliSoftware:
> Micro-DRY suggestion: Extract `File.exist?(path) && File.size(path).zero?` into its own method or `File` _extension_.

Why not. Also, that method extension could also be able to do more than just check if size is zero, e.g. check if file only contains whitespaces and newline… which reminds me of [that similar helper I actually wrote in a different file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_linter_helper.rb#L195-L203).

So maybe we should just move that already-existing `only_empty_lines?` helper method into an extension on `File` and then use it here to replace the uses of `exist?` + `size.zero?` after all 🤔 
Or rename it `File.empty?(allow_only_whitespaces: false)`? 🤔

AliSoftware:
🤦 TIL [`File.empty?` already exists](https://ruby-doc.org/core-3.1.0/File.html#method-c-empty-3F)…

PR #348, opened 2022-03-17
Speed up Build & Test on CI

AliSoftware:
Make the "Build and Test" step stop installing `pkg-config` and `imagemagick`, since `drawText` has been removed from the repo and published as independent tool since 3.0.0 (via #312)

This should avoid all the time spent installing those 2 tools and thus speed up the CI on that "Build and Test" step.

---

PS: We still need to use a Mac agent for that step because it still requires `homebrew` to be able to `brew install git-lfs` (which is still required for some of our tests, at least so far).
We might consider getting rid of `git-lfs` and all the code and tests related to it in the future as I think it might not really be used anymore? But that's for another PR

### DIFF
--- a/.buildkite/gem-push.sh
+++ b/.buildkite/gem-push.sh
@@ -1,12 +1,14 @@
 #!/bin/bash -eu
 
+GEM_NAME="fastlane-plugin-wpmreleasetoolkit"
+
 echo "--- :hammer: Build Gemspec"
-gem build fastlane-plugin-wpmreleasetoolkit.gemspec
+gem build "$GEM_NAME.gemspec" -o "$GEM_NAME.gem"
 
 echo "--- :sleuth_or_spy: Validate Gem Install"
-gem install --user-install fastlane-plugin-wpmreleasetoolkit-*.gem
+gem install --user-install "$GEM_NAME.gem"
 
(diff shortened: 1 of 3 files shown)


AliSoftware:
Note: `imagemagick` is still used by `promo_screenshot_helper`, so technically one would think that it would still be needed for the tests on that helper to run. But incidentally, the [`promo_screenshots_helper_spec` is commented all the way](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/promo_screenshots_helper_spec.rb) so those tests don't run at all those days.

If we were to decide to re-enable those tests on `promo_screenshot_helper` at some point in the future, I think it would make more sense to re-write those tests in a way that would completely mock or test-double all `ImageMagick` call anyway, so that the tests would not require the binary to be installed. Which is why I think it still makes sense to remove its installation from Buildkite step anyway.

AliSoftware:
- [Latest CI build (release tag), before that change](https://buildkite.com/automattic/release-toolkit/builds/399): "Build and Test" took 3'49 to run
 - [CI build of this PR, after that change](https://buildkite.com/automattic/release-toolkit/builds/402): "Build and Test" took only 1'47 to run 🎉

jkmassel (approved):

PR #351, opened 2022-04-07
Add option to identify the source library when merging strings.xml files

AliSoftware:
This improves the existing `an_localize_libs` action to adds support for the `source_id` option.

This option aims to helps identify which library each `<string>` merged into the main `strings.xml` came from.

## Why

 - This is prep work towards the plan drafted in paaHJt-3fg-p2
   - The future goal here would be to later update the merge logic to delete all the old entries that came from a previous merge of a library, before merging the new strings again. That way, keys deleted from the lib's `strings.xml` in a later version would properly be removed from the app's `strings.xml`, instead of being kept in the apps' `strings.xml` forever while not being used nor useful anymore
   - I decided to keep that deletion feature for later though, because implementing that feature would likely require a larger change in the current implementation and logic of the `an_localize_libs`'s helper code, and also might require a deeper analysis of potential consequences before going that route (e.g. what if a string is deleted from a lib later… but that key was still used by the app itself too?). Still, adding an attribute to identify the libs the strings were coming from still felt a useful first step anyway
 - But also, it will just help us (both the developers and the platform team) understand where each string from the `strings.xml` is coming from
   - This will be useful to developer to understand, when a string they see in the `WordPress/src/main/res/strings.xml`, where the source of truth for that string actually comes from (the app, or a lib?) and thus where to change that value if they need to (no point in changing the copy in the app's `strings.xml` file if the key will be re-overwritten by the next lib merge!)
   - It will also allow them to better reason about when to use `content_override` if they do want the key from the app's `strings.xml` to always take over and not be replace by the same key from a lib in the future (useful for example from a key like `app_name` not to be overwritten by a merge of the about library if it were to define one as well…)

## How

The action now accepts a new key in the `Hash` passed to its `libs_strings_path` parameter.
That new key is named `source_id` and allows the caller to provide an (optional) "identifier" string for the library. If provided, that identifier will be added as the value of a custom `a8c-src-lib` attribute of the `<string …>` XML nodes added or updated during the merging of the strings of said library.

You can see the result in action in [this draft WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/16280).

This PR also takes the occasion to:
 - Fix a crash if the caller provided a `nil` value for the `exclusions` key in the Hash
 - Use `UI.message` and `UI.verbose` instead of `puts` when the helper printed information about the strings being merged or updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ _None_
 
 * Add the option for `an_localize_libs` to provide a `source_id` for each library being merged.  
   If provided, that identifier will be added as an `a8c-src-lib` XML attribute to the `<string>` nodes being updated with strings from said library.
-	This can be useful to help identify where each string come from in the resulting, merged `strings.xml`. [#351]
+  This can be useful to help identify where each string come from in the resulting, merged `strings.xml`. [#351]
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ _None_
 
 ### New Features
 
-_None_
+* Add the option for `an_localize_libs` to provide a `source_id` for each library being merged.  
+  If provided, that identifier will be added as an `a8c-src-lib` XML attribute to the `<string>` nodes being updated with strings from said library.
+	This can be useful to help identify where each string come from in the resulting, merged `strings.xml`. [#351]
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -24,7 +24,7 @@ module Fastlane
 
         # Checks if string_name is in the excluesion list
         def self.skip_string_by_exclusion_list(library, string_name)
-          return false unless library.key?(:exclusions)
+          return false if library[:exclusions].nil?
 
           skip = library[:exclusions].include?(string_name)
           if skip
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ _None_
 
 ### New Features
 
-_None_
+* Add the option for `an_localize_libs` to provide a `source_id` for each library being merged.  
+  If provided, that identifier will be added as an `a8c-src-lib` XML attribute to the `<string>` nodes being updated with strings from said library.
+	This can be useful to help identify where each string come from in the resulting, merged `strings.xml`. [#351]
```
mokagio: I like the hard wrapping here, but I wonder if the tab indentation is intentional? 

In case it isn't
```suggestion
* Add the option for `an_localize_libs` to provide a `source_id` for each library being merged.  
  If provided, that identifier will be added as an `a8c-src-lib` XML attribute to the `<string>` nodes being updated with strings from said library.
  This can be useful to help identify where each string come from in the resulting, merged `strings.xml`. [#351]
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -9,11 +9,13 @@ module Fastlane
...
     module Android
       module LocalizeHelper
+        LIB_SOURCE_XML_ATTR = 'a8c-src-lib'.freeze
+
         # Checks if string_line has the content_override flag set
         def self.skip_string_by_tag(string_line)
           skip = string_line.attr('content_override') == 'true' unless string_line.attr('content_override').nil?
           if skip
-            puts " - Skipping #{string_line.attr('name')} string"
+            UI.message " - Skipping #{string_line.attr('name')} string"
```
mokagio, line 18: Minor fix, but always appreciated 👨‍🍳 👌

mokagio (approved):
Tested this via https://github.com/wordpress-mobile/WordPress-Android/pull/16280. Worked as described. Nothing to comment 👍

CHANGELOG.md
AliSoftware: It wasn't, great eye, thanks!

PR #352, opened 2022-04-11
Refactor `.po` generation – Part 1 of `n`

mokagio:
I didn't get to write an RFC today as I was hoping, but I wanted to at least track a tiny bit of self-contained work I've done while wrapping my head around the current structure.

- Adds a spec for the `ios_update_metadata_source` action
- Removes an unnecessary `require` (where by unnecessary I mean the tests pass anyway without it)

### DIFF
--- a/spec/ios_update_metadata_source_spec.rb
+++ b/spec/ios_update_metadata_source_spec.rb
@@ -2,12 +2,12 @@ require 'spec_helper'
 require 'shared_examples_for_update_metadata_source_action'
 
 describe Fastlane::Actions::IosUpdateMetadataSourceAction do
-  before {
+  before do
     # This works around the `ensure_git_status_clean` call within the action.
     # It would be quite cumbersome to iterate on the code if we had to commit
     # every change _before_ running the tests...
     allow(Fastlane::Actions::EnsureGitStatusCleanAction).to receive(:run)
-  }
+  end
 
   include_examples 'update_metadata_source_action', whats_new_fails: false
 end
--- a/spec/ios_update_metadata_source_spec.rb
+++ b/spec/ios_update_metadata_source_spec.rb
@@ -4,8 +4,13 @@ require 'shared_examples_for_update_metadata_source_action'
 describe Fastlane::Actions::IosUpdateMetadataSourceAction do
   before do
     # This works around the `ensure_git_status_clean` call within the action.
-    # It would be quite cumbersome to iterate on the code if we had to commit
-    # every change _before_ running the tests...
+    #
+    # We can't easily remove the need to stub the call here without removing the check in the action.
+    # In the tests, we move into a temp folder, but then create files in that directory.
+    # So, `ensure_git_status_clean` will fail anyway, unless we add more cruft to the tests.
+    #
+    # See also conversation in
+    # https://github.com/wordpress-mobile/release-toolkit/pull/352
     allow(Fastlane::Actions::EnsureGitStatusCleanAction).to receive(:run)
   end
 
--- a/spec/ios_update_metadata_source_spec.rb
+++ b/spec/ios_update_metadata_source_spec.rb
@@ -1,0 +1,13 @@
+require 'spec_helper'
+require 'shared_examples_for_update_metadata_source_action'
+
+describe Fastlane::Actions::IosUpdateMetadataSourceAction do
+  before {
+    # This works around the `ensure_git_status_clean` call within the action.
+    # It would be quite cumbersome to iterate on the code if we had to commit
+    # every change _before_ running the tests...
+    allow(Fastlane::Actions::EnsureGitStatusCleanAction).to receive(:run)
+  }
+
+  include_examples 'update_metadata_source_action', whats_new_fails: false
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_update_metadata_source.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_update_metadata_source.rb
@@ -1,4 +1,3 @@
-require 'fastlane/action'
 require_relative '../../helper/metadata_update_helper'
 
 module Fastlane
--- a/spec/shared_examples_for_update_metadata_source_action.rb
+++ b/spec/shared_examples_for_update_metadata_source_action.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 RSpec.shared_examples 'update_metadata_source_action' do |options|
   it 'updates any block in a given .po file with the values from the given sources' do
-    Dir.mktmpdir do |dir|
+    in_tmp_dir do |dir|
       output_path = File.join(dir, 'output.po')
       dummy_text = <<~PO
         msgctxt "key1"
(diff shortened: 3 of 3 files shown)


spec/ios_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+require 'shared_examples_for_update_metadata_source_action'
+
+describe Fastlane::Actions::IosUpdateMetadataSourceAction do
+  before do
+    # This works around the `ensure_git_status_clean` call within the action.
+    # It would be quite cumbersome to iterate on the code if we had to commit
+    # every change _before_ running the tests...
```
mokagio: @AliSoftware, I wrote this comment before our discussion on soft vs hard wrapping. Once that's concluded, I'll update here, like with semantic line-breaks given the second line would still be "short".

spec/ios_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+    allow(Fastlane::Actions::EnsureGitStatusCleanAction).to receive(:run)
+  end
+
+  include_examples 'update_metadata_source_action', whats_new_fails: false
```
mokagio, line 17: Gotta love shared examples, even though having to use them is a smell IMHO (one which we're aware of in this case).

spec/ios_update_metadata_source_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+require 'shared_examples_for_update_metadata_source_action'
+
+describe Fastlane::Actions::IosUpdateMetadataSourceAction do
+  before do
+    # This works around the `ensure_git_status_clean` call within the action.
+    # It would be quite cumbersome to iterate on the code if we had to commit
+    # every change _before_ running the tests...
+    allow(Fastlane::Actions::EnsureGitStatusCleanAction).to receive(:run)
```
AliSoftware, line 14: Drawback/Risk of this (which, if I understand correctly, is only there to ease our pain while we work on the implementation of that feature and don't want to commit before testing) is that:
 - This suggests that the tests itself runs in a current-working-directory that is the release-toolkit itself (and not e.g. some temporary directory we created and `cd`'d into to run the test in isolation
 - Which also means that, if the unit test accidentally modifies files in the current-working-directory it currently operates in, if that pwd is the release-toolkit with some uncommited files ("because we're in the middle of working on the implementation") that means running the test could overwrite the modifications of the release toolkit (e.g. the fixtures) you were working on…

So I might be wrong but I think having to mock that is a code smell and a sign that there's something wrong with the tests (especially, suggests that it works inside the toolkit repo), and that mocking would thus be the wrong fix?

spec/ios_update_metadata_source_spec.rb
AliSoftware: 😅

AliSoftware (commented):
I know that this is very WIP, but just to to share my take on this in case you start progressing further on this without me having the occasion to react on your future RFC beforehand:

 - I like that you're adding more unit tests to the existing methods
 - But I think ultimately for this refactor, it would make more sense to deprecate those existing `gp_update_metadata_source` and `ios_update_metadata_source` actions completely and replace it with brand new ones.
   - The naming for the existing action has always been ambiguous and while naming is hard, the fact that "update" can mean both directions does not help.
   - The current logic for parsing the existing `.po` using `block_parsers` etc will not match the idea we previously discussed on just re-generating the `.po` file from scratch based on the `.txt` files, so the new implementation will be quite different to the current logic anyway, which is another argument pushing in the direction of a brand new action

I am pretty sure you already share those ideas and direction as I think we discussed and agreed on those points before, but just wanted to be sure you wouldn't start working in the wrong direction. In any case, adding unit tests to the existing actions (and having fixtures ready that would check that the behavior will be consistent between old and new during refactoring is always 👌 👍 🙂 )

### Brain dump of ideas

Given that, to re-generate the `.po` files from scratch given the content of the `.txt` files, we'll still have to know which translator comments to attach to each block/string, the new action might still have to explicitly list which `.txt` files we want to include in the `.po` file (so that we can provide the comment to associate with it)… as opposed to just implicitly deduce this from the list of `.txt` files found in the appropriate `metadata/*` directory.

We could consider having that comment hard-coded in the action itself (after all, those comments should all be the same for the similar keys we have in our various Android and iOS repos, `title.txt` will always be the app title and always limited to the same number of characters for all repos, etc). On the other hand, keep in mind that we also have files not handled by fastlane in those `.po`, especially copies for promo screenshots, for which the comments would likely be specific to each case, app and repo.

Then we can make use of some gem that supports generating `gettext` files (`.po`) to assemble everything together (and let the gem do all the heavy lifting like character escaping etc)

mokagio:
Since I opened this PR, we decided to park the `.po` refactor/restructure for the time being. I think this PR is still relevant and a tiny win that we can safely merge, though. 

@AliSoftware, if that's okay with you, could you approve it? Thanks.

spec/ios_update_metadata_source_spec.rb
mokagio: Gosh... I started looking into this but never wrote about it and now I'm not sure where it was 🤔 

I run into a circular issue. Our tests already use `tmp` directories, but because they also create files in there, the status of that Git repository appears dirty. The solution would be to read the inputs and output the generated `.po` in a location that doesn't have a Git repo, while also running the tests in a different folder that has a git repo, but clean. (_Something along those lines, I'm writing this from memory without searching for my notes or commits_)

Not sure it's worth for the sake of this one test.

If this smell is too smelly to merge, I think we can live with the codebase not having this test. After all, if we'll ever need to modify the existing code, I'd very much hope we'd take the time to implement the rewrites we discussed, which would make this test redundant.

AliSoftware:
> > Unfortunately, this doesn't seem to do the trick. If the stub call is removed, the test fails.
> 
> After spending more time with this, that's not surprising. `in_tmp_dir` moves away from the repo and into a temp directory, but then test creates files in that directory, so `git_ensure_status_clean` will fail anyways 🤦‍♂️

Ah got it, so it's useful after all, but not for the reasons we initially thought (and that the comment points to).

Could you thus update [the code comment here](https://github.com/wordpress-mobile/release-toolkit/pull/352/files#diff-91e5a51ed2b505af7ec98685272296764e433b10bf6762aea02b4432a3b242a9R6-R8) accordingly to update the explanation as to why the stub is necessary before we merge this? That way when we'll get back to the code later (God knows when) we would not have lost the context on that WIP work even long after this PR and its comments has landed 🙂

spec/ios_update_metadata_source_spec.rb
AliSoftware: > The solution would be to read the inputs and output the generated .po in a location that doesn't have a Git repo, while also running the tests in a different folder that has a git repo, but clean. (Something along those lines, I'm writing this from memory without searching for my notes or commits)

I feel like this would be even dirtier. Like, it would be trying to trick the system by running the action in a different folder than the thing it's working on… which tbh if that dirty trick worked, could actually be considered a bug, because we'd probably instead want to check that the repo root _corresponding to the input folder_ would not be dirty — as opposed to the pwd — since the whole point I think was to ensure we did not forget to commit some of the input files the generated `.po` was based on… So I'd vote against such a hack.

I think we should:
1. Either keep the mocking and update the comment to explain why it's necessary (i.e. because in the special context of tests we create temp files just for the sake of the test, and in a temp dir that is not tracked in git anyway)
2. Or make the temp test dir be a git repo (after all we do something similar in our `git_helper_spec.rb`… though that spec is specifically about git so that make sense there but less sense here…)
3. Or remove the test from the codebase for now if we feel it's too smelly to merge.

Personally I'm ok with (1).

mokagio:
@AliSoftware 

> Could you thus update [the code comment here](https://github.com/wordpress-mobile/release-toolkit/pull/352/files#diff-91e5a51ed2b505af7ec98685272296764e433b10bf6762aea02b4432a3b242a9R6-R8) accordingly to update the explanation as to why the stub is necessary before we merge this?

Done. Thank you for the suggestion.

AliSoftware (approved):

PR #353, opened 2022-04-11
Fix `ios_l10n_linter` to allow missing translations in `.strings` files not being flagged as violations

AliSoftware:
## Why?

The previous implementation used a `diff` of textual outputs from SwiftGen, which worked so far because in the host app projects, locales which were missing translations for some keys were previously replaced by `”key” = “key”` entries (this was back when we assumes that our keys were always going to be the English copies)

But since the improvements in our Localization Tooling, this is not the case anymore, and any key that does not have a translation in a given locale is now just omitted and not present in that locale’s `Localizable.strings` — just as it has always been intended.

This made the logic of the linter break as it relied on this assumption to make the `diff` it ran on the textual outputs be useful. But now that the `Localizable.strings` files for translations might actually miss a lot of the keys present in the reference locale, we needed to be smarter

## How?

This changes the approach from doing a plain textual `diff` of the outputs, to parsing the outputs of SwiftGen into a `Hash<Key, ExpectedParametersList>`, so we can then filter and ignore the keys that are missing in each locale, and only raise violations for keys that are present in both the localization and the reference locale (typically English), in other words, only for keys for which there are actual translations to compare the placeholders for.

## To Test

 - Run `bundle exec rspec` to run the unit tests — this is actually already run by CI, so as long as CI is green we should be good
 - You can look at https://github.com/wordpress-mobile/WordPress-iOS/pull/18348 for a practical example where this fix is put to use

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Fix `ios_lint_localizations` action so that it no longer mistakely report missing keys not yet translated in the other locales' `.strings` as violations. [#353]
 
 ### Internal Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ _None_
 
 ### Bug Fixes
 
-* Fix `ios_lint_localizations` action so that it no longer mistakely report missing keys not yet translated in the other locales' `.strings` as violations. [#353]
+* Fix `ios_lint_localizations` action so that it no longer mistakely reports missing keys not yet translated in the other locales' `.strings` as violations. [#353]
 
 ### Internal Changes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_linter_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_linter_helper.rb
@@ -58,8 +58,7 @@ module Fastlane
         #
         # @param [String] input_dir The path (ideally absolute) to the directory containing the `.lproj` folders to parse
         # @param [String] base_lang The code name (i.e the basename of one of the `.lproj` folders) of the locale to use as the baseline
-        # @return [Hash<String, String>] A hash whose keys are the language codes (basename of `.lproj` folders) for which violations were found,
-        #         and the values are the output of the `diff` showing these violations.
+        # @return [Hash<String, Array<String>>] A hash of violations, keyed by language code, whose values are the list of violation messages for that language
         #
         def run(input_dir:, base_lang: DEFAULT_BASE_LANG, only_langs: nil)
           check_swiftgen_installed || install_swiftgen!
@@ -87,7 +86,7 @@ module Fastlane
           <<~TEMPLATE
             {% macro recursiveBlock table item %}
               {% for string in item.strings %}
-            "{{string.key}}" => [{{string.types|join:","}}]
+            {{string.key}} ==> [{{string.types|join:","}}]
               {% endfor %}
               {% for child in item.children %}
               {% call recursiveBlock table child %}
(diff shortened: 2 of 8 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_linter_helper.rb
```diff
@@ -139,28 +138,27 @@ def generate_swiftgen_config!(input_dir, output_dir, only_langs: nil)
...
+        def placeholder_types_for_keys(dir, lang)
           file = File.join(dir, output_filename(lang))
           return nil unless File.exist?(file)
 
-          sorted_lines = File.readlines(file).sort
-          File.write(file, sorted_lines.join)
-          return file
+          File.readlines(file).map do |line|
+            line.match(/^(.*) ==> (\[[A-Za-z,]*\])$/)&.captures
+          end.compact.to_h
```
AliSoftware, line 153: One of the main changes: 
 - Before we were just returning the file generated by SwiftGen (which lists the keys and their associated expected types from placeholders found in the translation) directly, and then run a plain `diff` on the textual result
 - Now the new strategy is to parse that file generated by SwiftGen — which always has one-key-per-line, even for keys containing translations that are on multiple lines, see tests — to build a map (`Hash`) of keys -> list-of-types

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_linter_helper.rb
```diff
@@ -172,34 +170,22 @@ def find_diffs(input_dir:, base_lang:, only_langs: nil)
...
-        # Returns true if the file only contains empty lines, i.e. lines that only contains whitespace (space, tab, CR, LF)
-        def only_empty_lines?(file)
-          File.open(file) do |f|
-            while (line = f.gets)
-              return false unless line.strip.empty?
-            end
+              violations = params_for_lang.map do |key, param_types|
+                next "`#{key}` was unexpected, as it is not present in the base locale." if params_for_base_lang[key].nil?
+                next "`#{key}` expected placeholders for #{params_for_base_lang[key]} but found #{param_types} instead." if params_for_base_lang[key] != param_types
+              end.compact
```
AliSoftware, line 184: …Then here instead of doing the `diff -U0` as we did before, we can now iterate on the `Hash` we constructed during `placeholder_types_for_keys`, to:
 - Only iterate on the keys that are present in the translation, ignoring any key that haven't been translated in the inspected locale yet
 - Detect keys that are in the translations but not in the source locale (those would be fishy and not expected), and keys that are present in both but have not the same list of parameters derived from the placeholders in their respective translations

The big difference is that this will no longer generate a huge list of missing keys when run on WPiOS, which has now adopted the new tooling, for which one of the major changes is to not include untranslated strings (as `"key" = "key"`) in the translated `.strings` files downloaded from GlotPress — see paaHJt-3bh-p2

spec/ios_lint_localizations_spec.rb
```diff
@@ -54,6 +54,34 @@
...
+      result = run_described_fastlane_action(
+        install_path: install_dir,
+        input_dir: @test_data_dir,
+        base_lang: 'en'
+      )
+
+      # Assert
+      expect(result).to eq(yml['result'])
+    end
+
```
mokagio, line 84: Nitpick: I liked this at the bottom of the file so that a reader would scroll to the tests first, and only dive into this implementation detail if they needed to. Would it make sense to move it at the end of the `context` block?

_Fine by me to disregards, in particular since there's no other comment from me at this time._

mokagio (approved):
Tested this as part of https://github.com/wordpress-mobile/WordPress-iOS/pull/18348#pullrequestreview-938844003. 👍 

Not having the plain text diffs looks much better.

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Fix `ios_lint_localizations` action so that it no longer mistakely report missing keys not yet translated in the other locales' `.strings` as violations. [#353]
```
```suggestion
* Fix `ios_lint_localizations` action so that it no longer mistakely reports missing keys not yet translated in the other locales' `.strings` as violations. [#353]
```

spec/ios_lint_localizations_spec.rb
AliSoftware: Fair point, but given that all other specs which declare helper `def` functions in a similar way already also have them at the top (e.g. [here](https://github.com/wordpress-mobile/release-toolkit/blob/4fc2e0c52f730268fb81dd01c39a9029b303a947/spec/ios_extract_keys_from_strings_files_spec.rb#L8-L13) or [here](https://github.com/wordpress-mobile/release-toolkit/blob/08602aa689138061d278bbc9f3658c7332054f3c/spec/ios_download_strings_files_from_glotpress_spec.rb#L14-L43))… I figure it would make sense to keep consistency here 🤷

mokagio: > Fair point, but given that all other specs which declare helper `def` functions in a similar way already also have them at the top (e.g. [here](https://github.com/wordpress-mobile/release-toolkit/blob/4fc2e0c52f730268fb81dd01c39a9029b303a947/spec/ios_extract_keys_from_strings_files_spec.rb#L8-L13) or [here](https://github.com/wordpress-mobile/release-toolkit/blob/08602aa689138061d278bbc9f3658c7332054f3c/spec/ios_download_strings_files_from_glotpress_spec.rb#L14-L43))… I figure it would make sense to keep consistency here 🤷

Good point. I wasn't aware of that 👍

PR #354, opened 2022-04-12
Allow auto-addition of `tools:ignore="UnusedResources"` when merging Android strings

AliSoftware:
This PR adds the ability for the `an_localize_libs` action (which merges `strings.xml` files from libraries into the main one in order to send their strings to translation) to automatically add `tools:ignore="UnusedResources"` to the merged strings when appropriate. — as discussed in Slack and Zoom with @ParaskP7 

## Merging strings from binary libraries

When merging `strings.xml` files from **binary dependencies** (typically precompiled and downloaded from S3), since the linter is unable to detect when a string it sees merged in the app's `strings.xml` is used in the binary library we pulled it from, we need to add `tools:ignore="UnusedResources"` to that merged string, to ensure lint won't complain about it.

> Note: If we want to detect strings that are _actually_ unused even in the binary libraries themselves, we would rely on running lint in the repos of the binary libraries themselves, while this PR is only about adding `tools:ignore="UnusedResources"` on the strings after they are merged in the app's `strings.xml`, not from a point of view of the libraries.

## Merging strings from local modules

When merging `strings.xml` files from local modules (like `image-editor` or `WordPress-Editor`), [because we currently have the `checkDependencies true` option enabled for lint in the `WordPress/build.gradle`](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/WordPress/build.gradle#L265), that means that running lint on the WordPress project _will_ detect when strings merged in the `strings.xml` of the app that are in fact used by the module, and will then not consider them unused.

So for that case, the linter complaining that a string merged from one of our local module is unused would be a genuine warning that would indeed warrant manual investigation(†)
Thus, for those local modules, and at least as long as we have `checkDependencies true` set, we should NOT make our tooling add `tools:ignore="UnusedResources"` automatically for all the strings merged from those modules.

_(†) A warning in such a case would not always necessarily mean that the string is indeed unused and could be deleted blindly from that module without priori investigation, because that module could itself declare that string… to override the value of one of its transitive dependency… so we should still investigate such cases manually, and decide to either delete the genuinely unused string from the module, or add `tools:ignore="UnusedResources"` to that string manually if we consider it's expected._

## Solution

This PR allows the caller of `an_localize_libs` to specify, for each lib, if it should add the `tools:ignore="UnusedResources"` attribute automatically when merging the libs' strings or not — via the new `add_ignore_attr: true` option.

The implementation made `an_localize_libs` smart enough to also support cases where a `tools:ignore` attribute would already exist on the string being merged, either in the app's `strings.xml` or in the lib's one, and merge the `UnusedResources` value with the existing value of the attribute.

## To Test

 - Read the unit tests (specs) carefully to ensure that the `expected = […]` values I used in each of the unit test would indeed correspond to what we really expect from such a merge and configuration of the action
 - Ensure that `bundle exec rspec` passes (this one is already run by CI, so checking that CI is green should be enough)

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -116,7 +116,7 @@ module Fastlane
         #                         library's `strings.xml` will be skipped and won't be merged into the main one.
         #                       - `:source_id`: An optional `String` which will be added as the `a8c-src-lib` XML attribute
         #                         to strings coming from this library, to help identify their source in the merged file.
-        #                       - `:add_ignore_attr`: If set to true, will add `tools:ignore="UnusedResources"` to merged strings.
+        #                       - `:add_ignore_attr`: If set to `true`, will add `tools:ignore="UnusedResources"` to merged strings.
         #
         # @return [Boolean] True if at least one string from the library has been added to (or has updated) the main strings file.
         #
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -33,11 +33,20 @@ module Fastlane
           end
         end
 
+        # Adds the appropriate XML attributes to an XML `<string>` node according to library configuration
+        def self.add_xml_attributes!(string_node, library)
+          if library[:add_ignore_attr] == true
+            existing_ignores = (string_node['tools:ignore'] || '').split(',')
+            existing_ignores.append('UnusedResources') unless existing_ignores.include?('UnusedResources')
+            string_node['tools:ignore'] = existing_ignores.join(',')
+          end
+          string_node[LIB_SOURCE_XML_ATTR] = library[:source_id] unless library[:source_id].nil?
+        end
+
         # Merge string_line into main_string
         def self.merge_string(main_strings, library, string_line)
           string_name = string_line.attr('name')
           string_content = string_line.content
-          lib_src_id = library[:source_id]
 
           # Skip strings in the exclusions list
           return :skipped if skip_string_by_exclusion_list(library, string_name)
@@ -61,14 +70,14 @@ module Fastlane
               else
                 # It has the tools:ignore flag, so update the content without touching the other attributes
                 this_string.content = string_content
-                this_string[LIB_SOURCE_XML_ATTR] = lib_src_id unless lib_src_id.nil?
+                add_xml_attributes!(this_string, library)
                 return result
               end
             end
           end
 
           # String not found, or removed because needing update and not in the exclusion list: add to the main file
-          string_line[LIB_SOURCE_XML_ATTR] = lib_src_id unless lib_src_id.nil?
+          add_xml_attributes!(string_line, library)
           main_strings.xpath('//string').last().add_next_sibling("\n#{' ' * 4}#{string_line.to_xml().strip}")
           return result
         end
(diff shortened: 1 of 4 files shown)


AliSoftware:
@mokagio CI seems to fail on this error, which is not relateed to the changes from this PR, but rather on the helper to find duplicated strings that we recently merged:
```
1) FastlaneCore::Helper::Ios::StringsFileValidationHelper when there are no duplicated keys returns an empty array
--
  | Failure/Error: raise "Invalid character `#{c}` found on line #{line_no + 1}, col #{col_no + 1}" if next_context.nil?
  |  
  | RuntimeError:
  | Invalid character `I` found on line 21, col 1
```
No idea why it would fail in this PR (and on my Mac)… but didn't in your PR adding that helper and those specs in the first place… See also my comment on https://github.com/wordpress-mobile/release-toolkit/pull/342#issuecomment-1097101886

Can I let you take a closer look during the week? (I'll be AFK until next Tuesday)

AliSoftware:
> @mokagio CI seems to fail on this error, which is not related to the changes from this PR, but rather on the helper to find duplicated strings that we recently merged:
> 
> ```
> 1) FastlaneCore::Helper::Ios::StringsFileValidationHelper when there are no duplicated keys returns an empty array
> --
>   | Failure/Error: raise "Invalid character `#{c}` found on line #{line_no + 1}, col #{col_no + 1}" if next_context.nil?
>   |  
>   | RuntimeError:
>   | Invalid character `I` found on line 21, col 1
> ```
> 
> No idea why it would fail in this PR (and on my Mac)… but didn't in your PR adding that helper and those specs in the first place… See also my comment on [#342 (comment)](https://github.com/wordpress-mobile/release-toolkit/pull/342#issuecomment-1097101886)
> 
> Can I let you take a closer look during the week? (I'll be AFK until next Tuesday)

Ended up tackling this on my own today via https://github.com/wordpress-mobile/release-toolkit/pull/357

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -103,8 +112,14 @@ def self.verify_string(main_strings, library, string_line)
         # @param [Hash] library Hash describing the library to merge. The Hash should contain the following keys:
         #                       - `:library`: The human readable name of the library, used to display in console messages
         #                       - `:strings_path`: The path to the strings.xml file of the library to merge into the main one
-        #                       - `:exclusions`: An array of strings keys to exclude during merge. Any of those keys from the library's `strings.xml` will be skipped and won't be merged into the main one.
+        #                       - `:exclusions`: An array of strings keys to exclude during merge. Any of those keys from the
+        #                         library's `strings.xml` will be skipped and won't be merged into the main one.
+        #                       - `:source_id`: An optional `String` which will be added as the `a8c-src-lib` XML attribute
+        #                         to strings coming from this library, to help identify their source in the merged file.
+        #                       - `:add_ignore_attr`: If set to true, will add `tools:ignore="UnusedResources"` to merged strings.
```
mokagio: Nitpick:

```suggestion
        #                       - `:add_ignore_attr`: If set to `true`, will add `tools:ignore="UnusedResources"` to merged strings.
```

spec/an_localize_libs_action_spec.rb
```diff
@@ -1,54 +1,226 @@
 require 'spec_helper'
 
 describe Fastlane::Actions::AnLocalizeLibsAction do
-  # This test is more of a way of ensuring `run_described_fastlane_action` handles array
-  # of hashes properly than a comprehensive test for the
-  # `an_localize_libs_action` action.
-  #
-  # Please consider expanding this test if you'll find yourself working on its
-  # action.
```
mokagio: I'm glad to see you took on my suggestion to expand the test 🙇‍♂️

mokagio (approved):
LGTM 👍 

Once the unit tests fix from #357 will land in this PR, CI will be green and this will be merge-able.

@AliSoftware since [I didn't get to address the test failure](https://github.com/wordpress-mobile/release-toolkit/pull/354#issuecomment-1102946394) and the current merge conflicts are due to my work with #356, I'd be happy to return the favor and solve them here. Let me know.

AliSoftware:
> @AliSoftware since [I didn't get to address the test failure](https://github.com/wordpress-mobile/release-toolkit/pull/354#issuecomment-1102946394) and the current merge conflicts are due to my work with #356, I'd be happy to return the favor and solve them here. Let me know.

~I'll gladly take you up on that offer~ 🙂

AliSoftware:
> > @AliSoftware since [I didn't get to address the test failure](https://github.com/wordpress-mobile/release-toolkit/pull/354#issuecomment-1102946394) and the current merge conflicts are due to my work with #356, I'd be happy to return the favor and solve them here. Let me know.
> 
> ~I'll gladly take you up on that offer~ 🙂

Change of plans, I ended up doing the conflict resolution myself as it was otherwise blocking me to move forward with working on https://github.com/wordpress-mobile/release-toolkit/pull/358 (which had to be built on top of this one to avoid conflicts between the 2 PRs making different improvements but on the same action and specs)

spec/an_localize_libs_action_spec.rb
```diff
@@ -1,48 +1,223 @@
...
+        run_described_fastlane_action(
+          app_strings_path: app_strings_path,
+          libs_strings_path: [
+            { library: 'lib', strings_path: lib_strings_path, exclusions: [] },
+          ]
+        )
+
+        expected = [
+          '<string name="override-true" content_override="true">from app override-true</string>',
+          '', '', # FIXME: Current implementation adds empty lines; we should get rid of those at some point
```
AliSoftware, line 78: This will actually already be addressed by https://github.com/wordpress-mobile/release-toolkit/pull/358 (which builds on top of this PR) 🎉

PR #356, opened 2022-04-19
Soft-wrap some previously hard-wrapped comments

mokagio:
See discussion at https://github.com/wordpress-mobile/release-toolkit/pull/342#discussion_r843929362.

I use a script with `awk awk 'length>70 && length<81'`. Some of the results were comments starting with `##` or `###`. I ignored those under the assumption they'll get addressed at some point in a dedicated PR that converts them to `#` (or organically disappear)

--- 

The CI failure is unrelated to this PR.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -8,14 +8,11 @@ module Fastlane
       # Fallback default branch of the client repository.
       DEFAULT_GIT_BRANCH = 'trunk'.freeze
 
-      # Checks if the given path, or current directory if no path is given, is
-      # inside a Git repository
+      # Checks if the given path, or current directory if no path is given, is inside a Git repository
       #
-      # @param [String] path An optional path where to check if a Git repo
-      #        exists.
+      # @param [String] path An optional path where to check if a Git repo exists.
       #
-      # @return [Bool] True if the current directory is the root of a git repo
-      #         (i.e. a local working copy) or a subdirectory of one.
+      # @return [Bool] True if the current directory is the root of a git repo (i.e. a local working copy) or a subdirectory of one.
       #
       def self.is_git_repo?(path: Dir.pwd)
         # If the path doesn't exist, find its first ancestor.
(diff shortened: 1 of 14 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -106,7 +99,7 @@ def self.create_branch(branch_name, from: nil, push: true)
       #
       # @param [String] message The commit message to use
       # @param [String|Array<String>] files A file or array of files to git-add before creating the commit.
-      #        use `nil` or `[]` if you already added the files in a separate step and don't wan't this method to add any new file before commit.
+      #        Use `nil` or `[]` if you already added the files in a separate step and don't wan't this method to add any new file before commit.
```
mokagio, line 102: Out of scope, but so tiny it was worth adding it while touching the fil.

AliSoftware:
> The CI failure is unrelated to this PR.

I've created https://github.com/wordpress-mobile/release-toolkit/pull/357 to fix said CI failure. Feel free to rebase this PR on top of `trunk` once that fix lands to benefit from it here and make this PR turn green.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_strings_file_validation_helper.rb
```diff
@@ -73,12 +72,10 @@ def self.find_duplicated_keys(file:)
 
           File.readlines(file).each_with_index do |line, line_no|
             line.chars.each_with_index do |c, col_no|
-              # Handle escaped characters at a global level. This is more
-              # straightforward than having a `TRANSITIONS` table that account
-              # for it.
+              # Handle escaped characters at a global level.
+              # This is more straightforward than having a `TRANSITIONS` table that account for it.
```
AliSoftware: Completely unrelated to the "change hard-wrap to soft-wrap" thing but just realized that this comment (of mine) was not super clear (i.e. this might suggest that "we don't have a `TRANSITIONS` table, and would have to add one to account for that" as opposed to the intended meaning of "we would have to account for it in the _existing_ `TRANSITIONS` table); so could be the occasion to improve it to:
```suggestion
              # This is more straightforward than having to account for it in the `TRANSITIONS` table.
```

AliSoftware (approved):
Thanks for taking care of this! :shipit:

mokagio:
Thank you @AliSoftware for researching and fixing the test failure 🙇‍♂️ 

I just merged #357, I'm going to `git rebase` now, then auto-merge should do the rest.

mokagio:
@wpmobilebot I don't think this PR grants a `CHANGELOG.md` entry. Not even an `[internal]` one.

PR #358, opened 2022-04-20
Avoid moving XML nodes during `an_localize_libs` action 

AliSoftware:
## What?

This PR should address:

 - Platform Request ref: paaHJt-2pH-p2
 - Item `[a2]` of our "Localization Toolchain Modernization Fixes" project (ref: paaHJt-2Ib-p2)

## How?

This PR changes the implementation logic of `an_localize_libs` (88b5b89) from the "remove existing XML node, then add the lib one again at the end" approach that it used so far, to the new "replace the XML node in place with the lib one" for the same case.

Note: I took the occasion to also quickly rename some local variables used in the helper code (1fbd208) while I was trying to reason about the current implementation, which will hopefully make it easier to understand/decipher in the future.

## Why?

 - This should stop our XML nodes from moving around in the final XML post-code-freeze
    - Which will not only make our life easier when reviewing PR diffs…
    - …but also ensure that any XML `<string>` node that were placed in a specific location in the XML by the devs, for example to be grouped with other related strings, or to be below a specific `<!-- explanation comment -->`, to not lose those.
 - As a bonus, this will also stop introducing empty lines (as you can see from the `FIXME`s disappearing from the unit tests thanks to those changes — 53e4bf8) in the final XML

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ _None_
 ### Bug Fixes
 
 * Fix `ios_lint_localizations` action so that it no longer mistakely reports missing keys not yet translated in the other locales' `.strings` as violations. [#353]
+* Fix `an_localize_libs` so that it does not move XML nodes around when merging lib strings (and replace them in-place instead). [#358]
 
 ### Internal Changes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -11,19 +11,19 @@ module Fastlane
       module LocalizeHelper
         LIB_SOURCE_XML_ATTR = 'a8c-src-lib'.freeze
 
-        # Checks if string_line has the content_override flag set
-        def self.skip_string_by_tag(string_line)
-          skip = string_line.attr('content_override') == 'true' unless string_line.attr('content_override').nil?
+        # Checks if `string_node` has the content_override flag set
+        def self.skip_string_by_tag?(string_node)
+          skip = string_node.attr('content_override') == 'true' unless string_node.attr('content_override').nil?
           if skip
-            UI.message " - Skipping #{string_line.attr('name')} string"
+            UI.message " - Skipping #{string_node.attr('name')} string"
             return true
           end
 
           return false
         end
 
-        # Checks if string_name is in the excluesion list
-        def self.skip_string_by_exclusion_list(library, string_name)
+        # Checks if `string_name` is in the exclusion list
+        def self.skip_string_by_exclusion_list?(library, string_name)
           return false if library[:exclusions].nil?
 
           skip = library[:exclusions].include?(string_name)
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -43,66 +43,65 @@ def self.add_xml_attributes!(string_node, library)
...
               # The string needs an update
-              result = :updated
-              if this_string.attr('tools:ignore').nil?
-                # It can be updated, so remove the current one and move ahead
-                this_string.remove
-                break
+              if main_string_node.attr('tools:ignore').nil?
+                # No `tools:ignore` attribute; completely replace existing main string node with lib's one
+                add_xml_attributes!(lib_string_node, library)
+                main_string_node.replace lib_string_node
```
AliSoftware, line 68: This is the key change in the implementation that will avoid having XML nodes being moved around (the rest of the diff being mostly local variables renaming to make the code easier to reason about)

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -11,19 +11,19 @@ module Android
       module LocalizeHelper
         LIB_SOURCE_XML_ATTR = 'a8c-src-lib'.freeze
 
-        # Checks if string_line has the content_override flag set
-        def self.skip_string_by_tag(string_line)
-          skip = string_line.attr('content_override') == 'true' unless string_line.attr('content_override').nil?
+        # Checks if `string_node` has the content_override flag set
```
mokagio, line 14: Nitpick, given one is surrounded in backticks, we might want to do the same for the other:

```suggestion
        # Checks if `string_node` has the `content_override` flag set
```

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ _None_
 ### Bug Fixes
 
 * Fix `ios_lint_localizations` action so that it no longer mistakely reports missing keys not yet translated in the other locales' `.strings` as violations. [#353]
+* Fix `an_localize_libs` so that it does not move XML nodes around when merging lib strings (and replace them in-place instead). [#358]
```
mokagio, line 21: I'm not sure that was a bug. Maybe an oversight of the first implementation? Something that became obvious only after we started using it "at scale"?

Anyways, that's just me being pedantic. 🙃

mokagio (approved):
+1 for the renames. We're manipulating XMLs and they are composed of nodes, I like how the code reflects that 👍 

I didn't test it IRL, but I was fresh from reviewing a recent PR in which I remarked how nice it was that you added many new tests for the action. The tests:

- Noticed the changes in behavior 
- Are green in CI

:shipit:

AliSoftware:
Unit Tests FTW, baby! 🙌

CHANGELOG.md
AliSoftware: Fair; but even if the "Bug Fixes" wording might not be the best for that specific case, that is still the most appropriate SemVer section matching this kind of change, as this is not a "Breaking Change", nor really a "New Feature", but this is not just "Internal" either as it has impacts on the result.

And overall I think it still makes sense to consider this was a bug, even if not affecting the behavior of the app using the resulting `strings.xml`, but as it was affecting developers working on the file (introducing blank lines, but most importantly dissociating XML nodes from their nearby associated XML comment) and could lead to misleading understanding of the XML file by developers later due to those unexpected changes. So not a "bug" as in "the software using this wasn't  working", but still one as in "not producing the output that developers expected" 🙃

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
AliSoftware: Missed that before merging, but still included the suggested change in the #359 via fb72adc before doing the release 😅

PR #361, opened 2022-05-04
Update GlotPress `export-translations` requests to avoid rate limiting

mokagio:
~~Left to do: Make the "Error downloading locale..." print the URL used for the download, to make it easier to triage.~~ See #362.

---

Starting yesterday, both I and @AliSoftware have been experiencing rate-liming when trying to download translations for the iOS app via this tooling:

```
[05:56:38]: -------------------------------------------------------
[05:56:38]: --- Step: ios_download_strings_files_from_glotpress ---
[05:56:38]: -------------------------------------------------------
[05:56:38]: Downloading translations for 'ar' from GlotPress (ar) [{:status=>"current"}]...
[05:56:39]: Error downloading locale `ar` — 429 Too Many Requests
[05:56:39]: Downloading translations for 'bg' from GlotPress (bg) [{:status=>"current"}]...
[05:56:40]: Error downloading locale `bg` — 429 Too Many Requests
...
```

I [reached out in the WordPress.org Slack](https://wordpress.slack.com/archives/C02RP50LK/p1651607938132789) and the issue was identified as genuine rate-limiting applied to our requests (I still don't know why it started only recently, though).

@dd32 [suggested in the WordPress.org Slack](https://wordpress.slack.com/archives/C02RP50LK/p1651640726217119?thread_ts=1651607938.132789&cid=C02RP50LK) to:

> - Set the user agent
> - Fix the URI you’re requesting, you need to request …./export-translations/?… otherwise you’re doubling the requests needed (It doesn’t include a trailing slash at present)
> - Check if your SSL dependancies are up-to-date, older SSL libraries might trigger some new checks

This PR address the first two points. Regarding SSL, we don't have any specific dependency, so I'm guessing it's all up to the `URI.open` implementation from the Ruby runtime on which the tooling runs. We try to keep those fairly up to date. Our major apps are all on Ruby 2.7.4 (not the latest, but fairly recent still)

- [WooCommerce Android](https://github.com/woocommerce/woocommerce-android/blob/d1d41b3de1debc47e8f5b7ab1b017e78f43b38cf/.ruby-version)
- [WooCommerce iOS](https://github.com/woocommerce/woocommerce-ios/blob/df682188bca96766594fc2c77090bdc20acef251/.ruby-version)
- [WordPress Android](https://github.com/wordpress-mobile/WordPress-Android/blob/64101c4cb0ed2e4806a6e7f39dbbc9bb551c9460/.ruby-version)
- [WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/d51d3f7d864e69c9c2df2e2616a600d6e020bff4/.ruby-version)

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/user_agent.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/user_agent.rb
@@ -1,0 +1,5 @@
+module Fastlane
+  module Wpmreleasetoolkit
+    USER_AGENT = 'Automattic App Release Automator; https://github.com/wordpress-mobile/release-toolkit/'.freeze
+  end
+end
--- a/spec/android_localize_helper_spec.rb
+++ b/spec/android_localize_helper_spec.rb
@@ -61,7 +61,7 @@ describe Fastlane::Helper::Android::LocalizeHelper do
         Dir[File.join(stubs_dir, '*.xml')].each do |path|
           # Each file in stubs_dir is a `{locale_code}.xml` whose content is what we want to use as stub for glotpress requests to `locale_code`
           locale_code = File.basename(path, '.xml')
-          url = "#{gp_fake_url.chomp('/')}/#{locale_code}/default/export-translations?filters%5Bstatus%5D=current&format=android"
+          url = "#{gp_fake_url.chomp('/')}/#{locale_code}/default/export-translations/?filters%5Bstatus%5D=current&format=android"
           stub_request(:get, url).to_return(status: 200, body: File.read(path))
         end
 
@@ -215,7 +215,7 @@ describe Fastlane::Helper::Android::LocalizeHelper do
 
         # Arrange: Prepare request stubs
         custom_gp_urls = TEST_LOCALES_MAP.map do |locale|
-          "#{gp_fake_url.chomp('/')}/#{locale[:glotpress]}/default/export-translations?filters%5Bstatus%5D=custom-status&filters%5Bwarnings%5D=yes&format=android"
+          "#{gp_fake_url.chomp('/')}/#{locale[:glotpress]}/default/export-translations/?filters%5Bstatus%5D=custom-status&filters%5Bwarnings%5D=yes&format=android"
         end
         custom_gp_urls.each do |url|
           stub_request(:get, url)
@@ -244,7 +244,7 @@ describe Fastlane::Helper::Android::LocalizeHelper do
         statuses = %w[current waiting fuzzy]
         statuses.each do |status|
           stub_path = File.join(fixtures_dir, 'filters', "#{status}.xml")
-          stub_request(:get, "#{gp_fake_url.chomp('/')}/fakegploc/default/export-translations?filters%5Bstatus%5D=#{status}&format=android")
+          stub_request(:get, "#{gp_fake_url.chomp('/')}/fakegploc/default/export-translations/?filters%5Bstatus%5D=#{status}&format=android")
             .to_return(status: 200, body: File.read(stub_path))
         end
 
--- a/spec/android_localize_helper_spec.rb
+++ b/spec/android_localize_helper_spec.rb
@@ -264,5 +264,36 @@ describe Fastlane::Helper::Android::LocalizeHelper do
         expect(File.read(generated_file_path)).to eq(expected_merged_content)
       end
     end
+
+    it 'sets a predefined User Agent so GlotPress will not rate-limit us' do
+      # Arrange
+      #
+      # Note that in this test we don't care about what is downloaded or how it's processed,
+      # only whether the request is made with the expected User Agent.
+      # Therefore there is only minimum fixtures setup, unlike in the examples before
+      FileUtils.mkdir_p(File.dirname(generated_file(nil)))
+      FileUtils.cp(expected_file(nil), generated_file(nil))
+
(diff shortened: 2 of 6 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/user_agent.rb
```diff
@@ -0,0 +1,5 @@
+module Fastlane
+  module Wpmreleasetoolkit
+    USER_AGENT = 'Automattic App Release Automator; https://github.com/wordpress-mobile/release-toolkit/'.freeze
+  end
+end
```
mokagio, line 5: I'm not sure whether this is the best way to define a constant scoped to the whole toolkit codebase. I copied this approach from `version.rb`.

spec/ios_l10n_helper_spec.rb
```diff
@@ -187,10 +187,31 @@ def file_encoding(path)
...
+               # - https://github.com/bblimke/webmock/issues/276#issuecomment-28625436
+               headers: { 'User-Agent' => 'Automattic App Release Automator; https://github.com/wordpress-mobile/release-toolkit/' }
+             ).to_return(body: 'content')
+      dest = StringIO.new
+      # Act
+      described_class.download_glotpress_export_file(project_url: gp_fake_url, locale: 'fr', filters: nil, destination: dest)
+      # Assert
+      expect(stub).to have_been_made.once
+      expect(dest.string).to eq('content')
+    end
```
mokagio, line 209: Nitpick: The android test has this example at the bottom, while here is at the top. _I think we can live with that..._

mokagio (commented):
While working and reviewing the code, the fact that the logic to build the URL is duplicated in several places clearly stood out to me.  I decided not to refactor it to keep this fix focused and fast to review, given this is currently a blocking issue for at least the WordPress iOS release automation.

mokagio:
@AliSoftware this is still a draft, but I'd love to hear your thoughts on it. Thanks!

I think the approach is on the right track. However... I tested it locally from WordPress iOS (using `path: 'PATH TO MY RELEASE TOOLKIT')` in the `Pluginfile` and I still got 429s 🤔 But that might be due to my IP being in a block list now, or GlotPress not knowing it should allow our User Agent.

spec/android_localize_helper_spec.rb
```diff
@@ -264,5 +264,36 @@ def generated_file(code)
...
+      # only whether the request is made with the expected User Agent.
+      # Therefore there is only minimum fixtures setup, unlike in the examples before
+      FileUtils.mkdir_p(File.dirname(generated_file(nil)))
+      FileUtils.cp(expected_file(nil), generated_file(nil))
+
+      stub = stub_request(:get, "#{gp_fake_url}fr/default/export-translations/")
+             .with(
+               query: { format: 'android', 'filters[status]': 'current' },
+               # Note that the syntax below merely checks that the given headers are present in the request,
+               # it does not restrict the request to have only those headers.
```
AliSoftware, line 281: 👍

AliSoftware (approved):

AliSoftware:
@mokagio Decided to go ahead and merge it because I need to do a hotfix (that has already been delayed a bit in last 2 days so I don't want to delay it more) which means that I'll also need to do a new beta (for Android `versionCode` technical reasons) before I can submit the hotfix to PlayStore alongside with it… so I'm hoping to point the `release-toolkit` to `trunk` to include that fix while I try to do the beta. Hopefully I'll avoid the 429 with that fix of yours, and if not that would have been a good excuse to test it anyway.

That being said I'm indeed not sure that's enough for the server to not rate-limit us. Sure it will divide the requests by 2 by avoiding the intermediate redirect, but from my initial tests setting the User-Agent was still not reducing the rate limiting. So we'll still have to do some follow-up with folks who handle the server side of this.

Some options I can imagine:

 - Quickfix / Temporary fix to land something asap: ask them to whitelist our special User-Agent for the time being, until they have time to implement a better and stronger solution (see below), to at least unblock us for the next betas to come (otherwise that will really become a big problem real quick…).
 - Ask them to lift the rate-limiting if we use an authenticated request using a token that they would whitelist on their side. That way they'd still rate-limit malicious actors against attacks but still allow trusted actors like us to do the downloads
 - Ask them to implement the bulk-download logic so that we can download a zip of all the locales at once.
    - That would require work from both sides (changing the implementation logic of our toolkit a bit more than just adding an auth token to the request, for example) so will take a tad more time to make something land, but if they agree to lift the rate limiting at least for our User-Agent as a temporary patch in the meantime, that could be acceptable.
    - That being said, while bulk-download would divide our number of requests by 30+ (the number of locales we download individually nowadays) we could still hit the rate-limit in the case where e.g. our linter notice some issues in the downloaded translations and we have to fix them in GlotPress and redownload them soon after. Way less likely to hit rate-limit in those cases as it won't be as many requests and as close to one another than currently, but still not ideal to have the mere possibility that betas and final builds would be blocked by possibility of that odd occurrence.

My main concern with all that is timing. We really have to find a solution to stop having those 429 soon, because we will definitively need them for the next betas we do, be it the one for localization testing next Wednesday… or more importantly, the final build at the end of the sprint.

mokagio:
> So we'll still have to do some follow-up with folks who handle the server side of this.

For reference: https://make.wordpress.org/systems/2022/05/05/mobile-release-tools-unable-to-fetch-from-glotpress/

PR #362, opened 2022-05-04
Print URLs that failed to download from GlotPress

mokagio:
I wanted to make this part of https://github.com/wordpress-mobile/release-toolkit/pull/361, but I'm not too happy with the implementation, so here's a dedicated PR for it and we can iterate.

I'm not happy with the implementation because the tests look odd and rigid with all those hardcoded URLs.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
@@ -151,7 +151,7 @@ module Fastlane
           begin
             IO.copy_stream(uri.open(options), destination)
           rescue StandardError => e
-            UI.error "Error downloading locale `#{locale}` — #{e.message}"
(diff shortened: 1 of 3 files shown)


AliSoftware (approved):
To be honest I don't find it _that_ odd that the URLs show up in the unit tests.

Sure, it makes the text in the expectation quite long. But on the other end, it also allows us to see what we'd expect such a failure message to look like — including how the URL will be logged in the error message — and allow us to have the test validate that this URL will be logged in full (escapes and query params and all) when such an error occurs… So overall I think that's exactly what we'd expect from a unit test 🙂

mokagio:
> To be honest I don't find it that odd that the URLs show up in the unit tests.

Yeah fair enough... I guess whenever I see URL construction logic, I always wish there was a dedicated object that does that, and for the consumers' tests to deal with a simpler URL. URL construction and how to deal with the requesting and the response are different concerns.

But... legacy, legacy. We'll get there one day. Or just live with it because it works fine enough. It's already great that we have tests, so many more tests than when I or you started.

---

Merging. Glad I run it past you first 👍

PR #355, opened 2022-04-18
Add Firebase Test Lab Support

jkmassel:
Adds Firebase Test Lab support to allow for running instrumented/connected tests.

Can be tested alongside https://github.com/woocommerce/woocommerce-android/pull/6422

PR #364, opened 2022-05-10
Implement App size metrics actions

AliSoftware:
This work is part of the App Metrics project [ref: paaHJt-37V-p2] and implements the fastlane actions necessary to send iOS and Android app size metrics [ref: paaHJt-3od-p2#comment-6098] to the [App Metrics server](https://github.com/Automattic/apps-metrics).

## What

Implements the actions to send iOS and Android App Size metrics to our future metrics API/system

## How

 - Implements an `AppSizeMetricsHelper` class — and its unit tests — to help us build a the payload for the metrics we want to send
 - Implements the new `ios_send_app_size_metrics` action which can take an `.ipa`, and optionally an `app-thinning.plist` (auto-detected if present next to the `.ipa`), and generates/sends the app size metrics for it for all device variants
 - Implements the new `android_send_app_size_metrics` action which can take an `.aab` and generates/sends the app size metrics for it and all the `.apk` splits built from it using `bundletool`

Each of those two new actions can take an `api_url` which points to:
 - Either an `http://` or `https://` endpoint (typically once the Metrics server is up and running, it will be the URL to its `/api/grouped-metrics` endpoint described [here](https://github.com/Automattic/apps-metrics/wiki/Queue-Group-of-Metrics)
 - Or a `file://` URL (in which case the `api_token` is not required) in order to write the payload to disk — which can be useful to either debug the payload, or store it as a CI artefact, etc.

## To Test

 - Ensure that CI runs the unit tests and that they all pass
 - **iOS**
    - Add a test lane on one of the client app's `Fastfile` which would build the `.ipa` using `thinning: '<thin-for-all-variants>'` and `method: 'enterprise'` as part of `gym`'s `export_options`…
    - …Then calls this action, passing it a `file://` URL to write the payload to disk (since we don't yet have a Metrics system hosted on a stable URL)
    - 👉  This has been done and tested via [this PR in WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/18541) for example
 - **Android**
    - Add a test lane on one of the client app's `Fastfile` which would build the `.aab` bundle…
    - …Then calls this action, passing it a `file://` URL to write the payload to disk (since we don't yet have a Metrics system hosted on a stable URL)
    - 👉  This has been done and tested via [this PR in WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/16580) for example

## What's Next

Since it was created, that PR was since followed by https://github.com/wordpress-mobile/release-toolkit/pull/365 which built on top of it to add support for Universal APKs.

If you already started to review this PR — especially the parts about the Android action — I suggest to review and land this PR first and review the changes to the Android action afterwards. If you haven't started the review of this PR yet, especially haven't started to look at the Android action and its specs, feel free to merge 365 in this PR first (to merge the code changes applied to the Android action) to review the Android action and specs code all in one go.


lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
```diff
@@ -0,0 +1,95 @@
...
+      #
+      def send_metrics(to:, api_token:, use_gzip: true)
+        uri = URI(to)
+        json_payload = use_gzip ? Zlib.gzip(to_h.to_json) : to_h.to_json
+
+        # Allow using a `file:` URI for debugging
+        if uri.is_a?(URI::File)
+          UI.message("Writing metrics payload to file #{uri.path} (instead of sending it to a server)")
+          File.write(uri.path, json_payload)
+          return 201 # To make it easy at call site to check for pseudo-status code 200 even in non-HTTP cases
```
AliSoftware:
```suggestion
          return 201 # To make it easy at call site to check for pseudo-status code 201 even in non-HTTP cases
```

AliSoftware:
Update: I've created another PR #365 built on top of this one to add support for "Universal APKs" on the `android_send_app_size_metrics` action.

_I've made it a separate PR on top of this one rather than pushing direclty on this one, because I consider that this PR is ready to review and merged as-is, is in a good state and can work standalone; so I didn't want the work on improvements from #365 to block this one from moving forward, and also because I wasn't sure if some people already started to review this PR (as it has been on ready for review for a couple of days already) and didn't want them to have to re-review everything from start if that was the case._

spec/app_size_metrics_helper_spec.rb
```diff
@@ -0,0 +1,153 @@
+require_relative './spec_helper'
+
+describe Fastlane::WPMRT::AppSizeMetricsHelper do
```
mokagio: I think `WPMRT` is an alternative to `Wpmreleasetoolkit`, but it's not an alias, right? I'm guessing you chose it because it's shorter to type? _FWIW IMHO it also_ looks _better_.

Given this library is a collection of tools, of which only a portion interacts together, I think it's okay to use a different namespace. But I wonder if you have any plan to migrate the rest of the code to this syntax.

Of course, this question is unrelated to the PR 😄 Any work on those line would be out of scope here.

spec/app_size_metrics_helper_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+          'Content-Type': 'application/json'
+        }
+        last_received_body = nil
+        stub = stub_request(:post, api_url).with(headers: expected_headers) do |req|
+          last_received_body = req.body
+        end.to_return(status: 201)
+
+        code = metrics_helper.send_metrics(to: api_url, api_token: token, use_gzip: false)
+
+        expect(code).to eq(201)
```
mokagio, line 125: When I first read this, I questioned whether it was valuable to assert `code` is `201` when that's the value we set in `stub_request`.

But I came to the conclusion that it is indeed valuable: To ensure `send_metrics` returns the HTTP response status  code it received 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
```diff
@@ -0,0 +1,95 @@
...
+  module WPMRT
+    # A helper class to build an App Size Metrics payload and send it to a server (or write it to disk)
+    #
+    # The payload generated (and sent) by this helper conforms to the API for grouped metrics described in
+    # https://github.com/Automattic/apps-metrics
+    #
+    class AppSizeMetricsHelper
+      # @param [Hash] group_meta Metadata common to all the metrics. Can be any arbitrary set of key/value pairs.
+      #
+      def initialize(group_meta = {})
```
mokagio: Nitpick: Would you consider being more verbose use "metadata" over "meta"?

I think "meta" is clear enough, but "metadata" is clearer 😄

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      # Adds a single metric to the group of metrics
+      #
+      # @param [String] name The metric name
+      # @param [Integer] value The metric value
+      # @param [Hash] meta The arbitrary dictionary of metadata to associate to that metric entry
+      #
+      def add_metric(name:, value:, meta: nil)
+        metric = { name: name, value: value }
+        meta = (meta || {}).compact # Remove nil values if any
```
mokagio: Thinking out loud: How would this code feel if there was a `remove_nil_values_if_any(meta)` call instead?

```ruby
def remove_nil_values_if_any(hash)
  (hash || {}).compact
end
```

The rationale for this suggestion comes from one of my favorite bits from ["Clean Code" by Uncle Bob](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882):

Screen Shot 2019-10-31 at 9 23 28 am

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
```diff
@@ -0,0 +1,95 @@
...
+      # @param [String] api_token The API bearer token to use to register the metric.
+      # @return [Integer] the HTTP response code
+      #
+      def send_metrics(to:, api_token:, use_gzip: true)
+        uri = URI(to)
+        json_payload = use_gzip ? Zlib.gzip(to_h.to_json) : to_h.to_json
+
+        # Allow using a `file:` URI for debugging
+        if uri.is_a?(URI::File)
+          UI.message("Writing metrics payload to file #{uri.path} (instead of sending it to a server)")
```
mokagio: Nitpick

```suggestion
          UI.message("Writing metrics payload to file #{uri.path} (instead of sending it to a remote API endpoint)")
```

spec/app_size_metrics_helper_spec.rb
mokagio: Followup question after reading the production code this test verifies (I try to read the tests first and the production code after).

How do you feel about the fact that `201` is also the value returned by the file URI code path? Could it result in a false positive here? Or, given we rely on webmock which does a lot of extra checks for us, we can rest assured that if the file URI code path is hit instead of the API one we'll know about it? I'm leaning towards the latter 👍

mokagio (commented):
I've been trying to review this PR for a while, but never had enough time set aside for it.

To avoid this becoming stale, here's a **partial review** covering `app_size_metrics_helper` only.

I like what I'm seeing so far 😄 👏 Can't wait to see this in action.

spec/ios_send_app_size_metrics_spec.rb
```diff
@@ -0,0 +1,102 @@
+require_relative './spec_helper'
+
+describe Fastlane::Actions::IosSendAppSizeMetricsAction do
+  let(:test_data_dir) { File.join(File.dirname(__FILE__), 'test-data', 'app_size_metrics') }
+  let(:fake_ipa_size) { 1337 } # The value used in the `app-thinning.plist` and `ios-metrics-payload.json` fixtures
```
mokagio, line 5: Is this an easter egg reference to the [Leet language](https://en.wikipedia.org/wiki/Leet)?

spec/ios_send_app_size_metrics_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+          { name: 'Build Type', value: 'beta' },
+          { name: 'Source', value: 'unit-test' },
+        ],
+        metrics: [
+          { name: 'File Size', value: 123_456 },
+        ]
+      }
+
+      test_app_size_action(
+        fake_ipa_size: 123_456,
```
mokagio, line 50: I didn't know Ruby allowed this for numbers like Swift, but in hindsight it's not surprising. Neat!

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+module Fastlane
+  module Actions
+    class IosSendAppSizeMetricsAction < Action
+      def self.run(params)
+        # Check input parameters
+        api_url = URI(params[:api_url])
+        api_token = params[:api_token]
+        if (api_token.nil? || api_token.empty?) && !api_url.is_a?(URI::File)
+          UI.user_error!('An API token is required when using an `api_url` with a scheme other than `file://`')
+        end
```
mokagio, line 13: Have you considered making the API token an `optional: false` parameter?

From a consumer point of view, we don't expect users to call this action with a `file://` API URL, but always and only with an `https://` URL plus a token. We could ignore the token in the `file://` case here, which is what the `send_metrics` method seems to do, too.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+            key: :api_url,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_API_URL',
+            description: 'The endpoint API URL to publish metrics to. (Note: you can also point to a `file://` URL to write the payload to a file instead)',
+            type: String,
+            optional: false
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :api_token,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_API_TOKEN',
+            description: 'The bearer token to call the API. Required, unless `api_url` is a `file://` URL',
```
mokagio, line 86: If we go with my suggestion to make this parameter required, we could update the description to

```diff
-            description: 'The bearer token to call the API. Required, unless `api_url` is a `file://` URL',
+            description: 'The bearer token to call the API. Ignored if `api_url` is a `file://` URL', 
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+            key: :use_gzip_content_encoding,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_USE_GZIP_CONTENT_ENCODING',
+            description: 'Specify that we should use `Content-Encoding: gzip` and gzip the body when sending the request',
+            type: FastlaneCore::Boolean,
+            default_value: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :app_name,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_APP_NAME',
+            description: 'The name of the app for which we are publishing metrics, to help filter by app in the dashboard',
```
mokagio: Nitpick: There might use cases other than a dashboard where we want to filter or group by app:

```suggestion
            description: 'The name of the app for which we are publishing metrics, to help with filtering and grouping',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+            type: String,
+            optional: false
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :app_version,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_APP_VERSION',
+            description: 'The version of the app for which we are publishing metrics, to help filter by version in the dashboard',
```
mokagio: Nitpick: There might use cases other than a dashboard where we want to filter or group by version:

```suggestion
            description: 'The version of the app for which we are publishing metrics, to help with filtering and grouping',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+            type: String,
+            optional: false
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :build_type,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_BUILD_TYPE',
+            description: 'The build configuration for which we are publishing metrics, to help filter by build config in the dashboard. E.g. `Debug`, `Release`',
```
mokagio: Nitpick: There might use cases other than a dashboard where we want to filter or group by build type:

```suggestion
            description: 'The build configuration for which we are publishing metrics, to help with filtering and grouping. E.g. `Debug`, `Release`',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+            type: String,
+            optional: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :source,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_SOURCE',
+            description: 'The type of event at the origin of that build, to help filter data in the dashboard. E.g. `pr`, `beta`, `final-release`',
```
mokagio: Nitpick: There might use cases other than a dashboard where we want to filter or group by event name:

```suggestion
            description: 'The type of event at the origin of that build, to help with filtering and grouping. E.g. `pr`, `beta`, `final-release`',
```

mokagio: Is there a reason why the example values here are lower case while in the previous item they are capitalized?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+            type: String,
+            optional: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :ipa_path,
+            env_name: 'FL_IOS_SEND_APP_SIZE_METRICS_IPA_PATH',
+            description: 'The path to the .ipa to extract size information from',
```
mokagio: For consistency with other descriptions:

```suggestion
            description: 'The path to the `.ipa` to extract size information from',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
```diff
@@ -0,0 +1,165 @@
...
+      def self.return_value
+        'The HTTP return code from the call. Expect a 201 when new metrics were received successfully and entries created in the database'
+      end
+
+      def self.authors
+        ['automattic']
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include? platform
```
mokagio: Is macOS appropriate here, when the action expect `.ipa` files?

```suggestion
        platform == :ios
```

mokagio (commented):
Another partial review, this time for the `ios_send_app_size_metrics` action

spec/app_size_metrics_helper_spec.rb
AliSoftware: Ah, yes, to be honest this helper started as a `model/` instead (`models/app_size_metrics_payload.rb`) and I didn't like the models to be directly at the root of the `Fastlane` namespace. But on the other hand, I didn't want to have a namespace be as long as `WordPressMobileReleaseToolkit` or even `WPMReleaseTollkit`, especially since:
 - We tangentially discussed this with Jeremy in https://github.com/wordpress-mobile/release-toolkit/pull/355#discussion_r863960510
 - I also pondered about what sub-namespace we could use for models when I was working on https://github.com/wordpress-mobile/release-toolkit/pull/296 a while ago, to avoid putting models direclty at the root of `module Fastlane` (and risking having potential conflicts with a future version of Fastlane itself)

On the other hand, I just noticed that we already use the name `module Wpmreleasetookit` in other places, like [here](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit.rb#L4)… so I should probably have gone with that instead.

---

All that being said, since I wrote all that, I've since changed my mind and made this class a `AppSizeMetricsHelper` instead of a `AppSizeMetricsPayload` model (mostly because that class ended up having the `send_metrics(…)` method, which felt wrong on a model and better suited for a "helper")… and all our common Helpers seems to be under `module Fastlane::Helper`… so I should probably align with that here after all.

spec/app_size_metrics_helper_spec.rb
AliSoftware: > How do you feel about the fact that 201 is also the value returned by the file URI code path? Could it result in a false positive here?

Well, as the code comment says [on this line](https://github.com/wordpress-mobile/release-toolkit/pull/364/files#diff-fabf5c60f39ba983a6eb9ba64e193689e04d4cab2694665dac19c8802f59b4afR64), the fact that the file URI code path returns `201` to mimick a typical server response — and make the call size easier and more generic to write, not having to change the "success" test in the calling `Fastfile` when you need to switch between real server and debugging file URL — is fully intentional here.

So in fact that's fully intentional that this spec tests that the `code` returns `201`, even for the cases when the file URI path is being tested.

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
AliSoftware: Good point. I used `"meta"` because that was the name of the key in the JSON payload… but the whole point of having a helper and API is to abstract away from the implementation details of the payload schema… so better be explicit in the API and docs 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
AliSoftware: The thing is, your suggestion for `def remove_nil_values_if_any` does more than just removing nil values if there are any (this part is just what `.compact` does, by definition of `compact`). In practice your method also defaults the `meta` to an **empty `Hash`** if a `nil` value was provided for the `meta` parameter.

So to match what is described/auto-documented by your method's name, the correct implementation should be:
```ruby
def remove_nil_values_if_any(hash)
  hash.compact
end
```
And at that point, just `.compact` is already term of art and self-explanatory.

Alternatively, you could keep your implementation and try to find a more explanatory method name… but that would start to become quite verbose and cumbersome:
```ruby
def default_to_empty_hash_if_nil_and_remove_any_nil_values_otherwise(hash)
  (hash || {}).compact
end
```

---

An alternative to maybe make this line of code clearer and still using term-of-art methods would be to exchange the order of the two actions (but still not worth extracting in their own method):
```ruby
meta = meta&.compact || {}
```
Since `compact` is term-of-art, `meta&.compact` is self-explanatory (ICYDK, `&.` is optional-chaining in ruby, just like `?.` in Swift), and given that the `x || {}` is very idiomatic in Ruby to provide default values (that's the equivalent of `x ?? [:]` in Swift), the full expression would seem self-explanatory to me overall, and probably easier to parse than the parenthesised and flipped `(meta || {}).compact` (where the `()` is probably what makes it harder to grasp?).

spec/ios_send_app_size_metrics_spec.rb
AliSoftware: It is 😛  (I often freely use both `42` and `1337` for random values as references to H2G2 and Leet language in my tests 😝  )

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
AliSoftware: I'm not sure I'm following your argument here 😅 

> we don't expect users to call this action with a file:// API URL

Which is why it has to be `optional: true`, otherwise fastlane itself will `user_error!` all by itself if you provide a `file://` URL but no `api_token` (and why would you provide an API token when you're using `file://` URL, and what would that token be anyway… be it for debugging while we don't yet have a metrics server and thus any token to provide, or if you use this action to store the metrics as artefacts on your CI rather than sending them to a metrics server)

So, yes, the `token` in the `file://` case is ignored here (that's exactly what those lines 11–13 are for, to ignore it and allow it to be optional provided… **_only_** if using a file URL, and be required **_only_** when using a non-file URL). 
Given it has a different optionality depending on the type of `api_url` used, fastlane won't let us use `optional: false` in the `ConfigItem` otherwise it will complain that it's missing regardless of if it was valid to omit it (with `file://` urls) or not. So the only way in custom fastlane actions to provide conditional optionality is to handle it ourselves, as `ConfigItem` does not provide a way for that directly.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
AliSoftware: If we go with parameter being required, that means the call site will still **have** to provide a (dummy) non-`nil` value for the `api_token`… even if it gets completely unused… seems wrong and confusing to me, API-design-wise?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
AliSoftware: Not really 😅 

The actual reason is that the examples for the other keys are values already existing and provided by the project configuration (e.g. `Debug` and `Release` are build configuration names spelled capitalized in Xcode's build settings; `Vanilla`, `Jalapeno`, etc (for the `android_send_size_metrics.rb`'s `product_flavor`) are also spelled capitalized in the `./gradlew assembleWordpressVanillaRelease` commands, etc… while this `source` key is designed to be totally arbitrary, independant to the build configurations and settings of the project, and more acting like "hashtags" to attribute to a build metric.

So I guess I thought about the `#hashtag` similarity here and that's why I went with lowercase in the examples… 🤷

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
AliSoftware: Good point. I have no idea if there's a similar concept of app thinning with macOS `.app` files (would probably make sense, as in Universal Binaries vs Intel-only vs M1-only…?)… but since we haven't tested it and the code wasn't written with this use case in mind so probably might need some tweaks (and specific unit tests!) to make sure we'd support macOS… probably better to not declare the support for now anyway 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
AliSoftware: ☝️ applied in a3660c8f12a45c8a5d7c824cd18ac347ee1055cf

spec/app_size_metrics_helper_spec.rb
AliSoftware: Addressed in fc90b313e29a55d4fc621e705ce636f598593bbe

spec/ios_send_app_size_metrics_spec.rb
AliSoftware: Actually `rubocop` even complains if I _don't_ add the `_` thousand separators 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
AliSoftware: Addressed in e8acd9d328088d94a36692a7c3bb25d227adb5a5

AliSoftware:
@mokagio Just wanted to let you know that since your last review, in addition to addressing your feedback, I also addressed an issue we encountered on Android CI, where the `apkanalyzer` tool that the Android action uses… was installed in a different location on disk.

I addressed this in ad5935c, by adding a new `ConfigItem` to allow providing an explicit path to the `apkanalyzer` binary in case we ever needed to be explicit… and updating the logic which tries to deduce/find the binary in `PATH` and `ANDROID_SDK_HOME` to account for the variations of the tool path used in our CI.
So, in practice, we still don't need to provide the explicit path to `apkanalyzer` for our case (as with the new logic it's now able to deduce/find it on its own… even on CI), but in case that default installation path changes in future Android AMIs or Docker images, we now have a way to also be explicit for it.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
AliSoftware: Ok I'm re-reading your comment and I get what you meant now (I was just missing some coffee earlier… ☕ ).

But in fact, even if in 80% of cases — especially once we'll have our Metrics server up and running at a stable URL — we will use a `https://` URL for `api_url`… there are still legitimate cases were using `file://` is not the exception or the "secret" case but a genuine usage we still want to cover properly (instead of treating it as if it was an easter egg):

 - Debugging the payloads is an obvious use case for sure, but I can see how this could be considered a rare case and not "normal operation"
 - But we could also imagine to use this to handle cases when the Metrics server has to go into maintenance (e.g. if we need to shut it down for a sprint or two while figuring out something broken or fixing some security issue on it…) and we would want to instead store the metrics on disk in the artefacts during that time, to be able to "replay" those payloads later once the Metrics server goes back up
 - Or even on a regular basis, if we decide to run the action twice, once to upload the metrics to a server, and another time to store the payload in the CI's artifacts for "backup"
 - Or when we want to generate the metrics locally, e.g. to compare the sizes of the app between two branches or two tags, and we'd then want to run the lane locally on our Macs in one branch then the next and compare the results
 - …

So imho while it definitively won't be the majority of cases, those are still legitimate ones and as such should be considered more than just easter eggs and the optionality of `api_token` depending on it should be handled properly as a result 😛

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
```diff
@@ -0,0 +1,212 @@
...
+            UI.message("[App Size Metrics] Generated #{apks.length} APKs.")
+
+            apks.each do |apk|
+              UI.message("[App Size Metrics] Computing file and download size of #{File.basename(apk)}...")
+              split_name = File.basename(apk, '.apk')
+              file_size = Action.sh(apkanalyzer_bin, 'apk', 'file-size', apk, print_command: false, print_command_output: false).chomp.to_i
+              download_size = Action.sh(apkanalyzer_bin, 'apk', 'download-size', apk, print_command: false, print_command_output: false).chomp.to_i
+              metrics_helper.add_metric(name: 'APK File Size', value: file_size, metadata: { split: split_name })
+              metrics_helper.add_metric(name: 'Download Size', value: download_size, metadata: { split: split_name })
+            end
```
AliSoftware, line 43: ℹ️  Note for the reviewers, in case you were tempted to make suggestions about those in this PR:
 - This code has since been extracted in a separate method during https://github.com/wordpress-mobile/release-toolkit/pull/365, to make the code cleaner, less "massive" and more readable
 - The names of the metrics (`APK File Size` and `Download Size` here, but same for all others used all over in `*_send_app_size_metrics`) have since been DRYed into constants in https://github.com/wordpress-mobile/release-toolkit/pull/365 as well

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
mokagio: > So imho while it definitively won't be the majority of case

Okay, cool. If that's the expectation, then it makes sense to keep the API consistent for both, rather than optimize it only for a scenario.

👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_send_app_size_metrics.rb
mokagio: > seems wrong and confusing to me, API-design-wise?

Granted. I considered it acceptable under the assumption that using `file://` was not an "official" use case, which I now appreciate was an incorrect assumption, as per https://github.com/wordpress-mobile/release-toolkit/pull/364/files#r880888176

spec/app_size_metrics_helper_spec.rb
mokagio: > So in fact that's fully intentional that this spec tests that the code returns 201, even for the cases when the file URI path is being tested.

Totally agree that it's appropriate to ensure `code` is `201` when _the file URI path_ is being tested 👍 

What I'm on the fence about is the `https://` behavior. This code is for the `https://` URI, though, and that's really what my question was about. 

The `https://` path [returns `response.code.to_i`](https://github.com/wordpress-mobile/release-toolkit/pull/364/files#diff-fabf5c60f39ba983a6eb9ba64e193689e04d4cab2694665dac19c8802f59b4afR91).

The tests stubs the response to return 201, the the same value that the `file://` code path would return. That's what made me wonder about the chance of a false positive, since, just by looking at `code`, I don't think we can which path the code took.

But "given we rely on webmock which does a lot of extra checks for us, we can rest assured that if the file URI code path is hit instead of the API one we'll know about it" because between webmock and the `expect(stub).to have_been_made.once` the test would fail. So, all in all, I think this is okay 👍 😄

mokagio (commented):
Posting a few more comments in case I get interrupted while looking at the Android side of things.

lib/fastlane/plugin/wpmreleasetoolkit/helper/app_size_metrics_helper.rb
mokagio: 👍

spec/android_send_app_size_metrics_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      # Asserts
+      expect(code).to eq(201)
+      expect(File).to exist(output_file)
+      gzip_disabled = other_action_args[:use_gzip_content_encoding] == false
+      generated_payload = gzip_disabled ? File.read(output_file) : Zlib::GzipReader.open(output_file, &:read)
+      # Compare the payloads as pretty-formatted JSON, to make the diff in test failures more readable if one happen
+      expect(JSON.pretty_generate(JSON.parse(generated_payload))).to eq(JSON.pretty_generate(expected_payload)), 'Decompressed JSON payload was not as expected'
+      # Compare the payloads as raw uncompressed data as a final check
+      expect(generated_payload).to eq(expected_payload.to_json)
```
mokagio, line 49: These two "triangulating" assertions (inappropriate because they're only two assertions, so not a triangle, but that's a term sometimes used for this pattern) are neat 😎

spec/app_size_metrics_helper_spec.rb
AliSoftware: Ah I see.

But note that this `expect` is there not only to ensure that the status code was 201 (which we know because we stub it)… _but_ also (and mainly?) to check that the action [forwards that http status code (that the request / stub returned) as a return value of the action itself](https://github.com/wordpress-mobile/release-toolkit/pull/364/files#diff-fabf5c60f39ba983a6eb9ba64e193689e04d4cab2694665dac19c8802f59b4afR91) — as opposed to returning nothing, or the JSON payload, or whatnot as a return value.

mokagio (approved):
:shipit:

PR #365, opened 2022-05-19
App Size Metrics: add support for Android Universal APKs

AliSoftware:
This work is part of the App Metrics project [ref: paaHJt-37V-p2] and implements the fastlane actions necessary to send iOS and Android app size metrics [ref: paaHJt-3od-p2#comment-6098] to the [App Metrics server](https://github.com/Automattic/apps-metrics).

## What

This builds on top of https://github.com/wordpress-mobile/release-toolkit/pull/364 and adds support for including "Universal APKs" metrics for Android.

_(I suggest to get #364 merged first before this #365, to keep the diffs of each PR separate and easier to review in isolation)_

## Why

This will mostly be useful if/when we want to send app size metrics for Installable Builds, which are build using `./gradlew assemble<app>JalapenoDebug` and thus produce a single, Universal `.apk` — as opposed to an `.aab` bundle like we do for beta and release builds that we upload to the Play Store)

## How

 - New `universal_apk_path` parameter for the `android_send_app_size_metrics` action
 - The `aab_path` parameter is no longer mandatory, to allow the case where we only pass the universal APK but no `.aab`. It's still required to have at least _one of_ `universal_apk_path` or `aab_path` (if not both) though.
 - Changed a key used in the Android metrics payload to the more explicit `"Optimized File Size"` for R8-optimized APKs (i.e. `file-size` as reported by `apkanalyzer`, as opposed to the `.apk` file size on disk).
    - Hopefully that will help differentiate in particular the Universal APK's file size on disk vs the Universal APK's optimized file size ("optimized" as in "after R8 optimizations and according to `apkanalyzer`", as opposed to "optimized for a specific device architecture", that latter being instead indicated by the `split:` metadata)
 - Moved the keys used in each metrics payload into constants at the top of the action, making them easier to spot and change

## To Test

 - Check that tests pass on this PR (CI green)
 - Check https://github.com/wordpress-mobile/WordPress-Android/pull/16580, which uses this branch to run the action with the Universal APKs built for WP and JP Installable Builds
    - Verify that CI is green on that WPAndroid PR
    - Verify that [the `*-app-size-metrics.json` files in the "Artifacts" tab of the WPAndroid PR's CircleCI job](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-Android/31475/workflows/e7f74f3f-4dcc-4cad-b217-ccead5ea8d94/jobs/124137/artifacts) look as expected.


lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
```diff
@@ -54,26 +58,79 @@ def self.run(params)
...
+        # The path where the `apkanalyzer` binary was found, after searching it:
+        #  - in priority in `$ANDROID_SDK_ROOT` (or `$ANDROID_HOME` for legacy setups), under `cmdline-tools/latest/bin/` or `cmdline-tools/tools/bin`
+        #  - and falling back by trying to find it in `$PATH`
+        #
+        # @return [String] The path to `apkanalyzer`
+        # @raise [FastlaneCore::Interface::FastlaneError] if it wasn't found in any of the above tested paths.
+        #
+        def find_apkanalyzer_binary!
+          apkanalyzer_bin = find_apkanalyzer_binary
+          UI.user_error!('Unable to find `apkanalyzer` executable in `$PATH` nor `$ANDROID_SDK_ROOT`. Make sure you installed the Android SDK Command-line Tools') if apkanalyzer_bin.nil?
```
mokagio: Nitpick: I can't point my finger on it, but that "nor" sounds odd. I usually see "nor" in "neither ... nor" or sentences. I think, with "unable" being a negative connotation verb, it works, but I'd suggest: 

```suggestion
          UI.user_error!('Unable to find `apkanalyzer` executable in either `$PATH` or `$ANDROID_SDK_ROOT`. Make sure you installed the Android SDK Command-line Tools') if apkanalyzer_bin.nil?
```

image

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
```diff
@@ -3,13 +3,23 @@
 module Fastlane
   module Actions
     class AndroidSendAppSizeMetricsAction < Action
+      # Keys used by the metrics payload
+      AAB_FILE_SIZE_KEY = 'AAB File Size'.freeze                     # value from `File.size` of the `.aab`
+      UNIVERSAL_APK_FILE_SIZE_KEY = 'Universal APK File Size'.freeze # value from `File.size` of the Universal `.apk`
+      UNIVERSAL_APK_SPLIT_NAME = 'Universal'.freeze                  # pseudo-name of the split representing the Universal `.apk`
```
mokagio, line 9: Just curious: Why is at "pseudo-name"?

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
```diff
@@ -178,6 +236,16 @@ def self.available_options
             type: FastlaneCore::Boolean,
             default_value: true
           ),
+          FastlaneCore::ConfigItem.new(
+            key: :universal_apk_path,
+            env_name: 'FL_ANDROID_SEND_APP_SIZE_METRICS_UNIVERSAL_APK_PATH',
+            description: 'The path to the Universal .apk to extract size information from',
```
mokagio:
```suggestion
            description: 'The path to the Universal `.apk` to extract size information from',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
```diff
@@ -178,6 +236,16 @@ def self.available_options
...
+            type: String,
+            optional: true, # We can have `aab_path` only, or `universal_apk_path` only, or both (but not none)
+            verify_block: proc do |value|
+              UI.user_error!('You must provide an path to an existing `.apk` file') unless File.exist?(value)
```
mokagio:
```suggestion
              UI.user_error!('You must provide a path to an existing `.apk` file') unless File.exist?(value)
```

mokagio (approved):
image

👍 :shipit:

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
AliSoftware: Yeah I agree, tbh I found it a bit odd-looking as well when I wrote it myself 😛

Should it be "in neither $PATH nor $ANDROID_SDK_HOME" then, given it's a negative form?

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
AliSoftware: Because for all the other splits, the name (i.e. the value used for the `split:` metadata) is provided by `bundletool` (it's the basename of the `.apk` produced by it in the `<tmpdir>/splits/*.apk`); while for the universal APK it's just a name of our choosing and not some official name provided by the Android tooling, so it could be any name we want it to be.

In fact if we wanted to be pedantic, the Universal APK should probably not be considered a "split" per se — it's not the name of a CPU architecture after all. So it's a bit like if we said for iOS that the Universal `.ipa` was a "thinned version for the _device_ named 'Universal'": `pseudo-name` because not really the name of a split/CPU-arch but one we made up to represent/mean "multiple archs in a single FAT file".

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb
mokagio: I think "Unable ... in neither ... nor ..." would make the sentence a [double negative](https://examples.yourdictionary.com/examples-of-double-negatives.html) and while they're not wrong, they "should be avoided".

But... we're hair splitting, really.

PR #366, opened 2022-05-25
Screenshots: Fix blocking errors in promo screenshots helper

rachelmcr:
Related: pdnsEh-hU-p2

## Description

This PR resolves errors with the `drawText` shell command in the promo screenshots helper, which were blocking screenshot generation.

Many thanks to @AliSoftware for the help investigating and resolving this!

## Changes

* Calls `Action.sh` instead of `sh` to resolve an unknown method error for the `drawText` shell command.
* Removes the escaped quotes in the `drawText` shell command. These quotes are not needed since the parameters are being passed separately, and instead cause errors as they are taken as verbatim quote characters.

## Testing

You can test these changes in Woo iOS with the testing steps in https://github.com/woocommerce/woocommerce-ios/pull/6933.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ _None_
 ### Bug Fixes
 
 * Update GlotPress `export-translations` requests to avoid rate limiting. [#361] [#362]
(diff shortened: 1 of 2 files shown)


AliSoftware (approved):
🎉

PR #370, opened 2022-06-02
Remove legacy iOS localization actions and related helpers

mokagio:
~~**Blocked!**~~ I was comfortable opening this because, once https://github.com/woocommerce/woocommerce-ios/pull/7032 will be merged, both WordPress and WooCommerce iOS won't depend on these actions anymore. Simplenote iOS and macOS still do, but they are on hold and we have Semantic Versioning (merging this would count as a breaking change) on our side to prevent issues there (besides, if we'll pick up development on those platforms it would only make sense to spend the time to upgrade the tooling then).

~~But... I forgot about Pocket Casts iOS. The app is actively in development and uses these legacy APIs. I think the best play for me is to work on converting that one, too, while I have all the know-how "fresh" in my mind.~~

**Update** Pocket Casts iOS [no longer uses the legacy APIs](https://github.com/Automattic/pocket-casts-ios/pull/399). 

---

We now have newer, better actions to achieve the same results:

- `ios_generate_strings_file_from_code`
- `ios_extract_keys_from_strings_files`
- `ios_download_strings_files_from_glotpress`
- `ios_merge_strings_files`


lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -28,48 +28,6 @@ def self.commit_version_bump(include_deliverfile: true, include_metadata: true)
...
-          Action.sh("cd #{get_from_env!(key: 'PROJECT_ROOT_FOLDER')} && ./Scripts/update-translations.rb")
-
-          Fastlane::Helper::GitHelper.commit(message: 'Update translations', files: strings_files, push: false)
-
-          Action.sh('cd fastlane && ./download_metadata.swift')
-
-          Fastlane::Helper::GitHelper.commit(message: 'Update metadata translations', files: './fastlane/metadata/', push: true)
-        end
-
-        def self.strings_files
```
mokagio, line 66: This method is only used in the two methods deleted above.

mokagio:
Superseded by #447

PR #371, opened 2022-06-02
Use Automattic as the Gem and Fastlane action author

mokagio:
I was looking at an action and noticed Lore was the hardcoded author. Seemed no longer appropriate given he has left 😢 

I took a couple of minutes to update everything to use "Automattic," which is the standard name we've been using for a while.


AliSoftware (approved):
thanks for standardizing this 🎉  :shipit:

PR #373, opened 2022-06-14
Remove the fastlane-specific `require`

jkmassel:
It references an empty file, and breaks our ability to use Dependabot.

**To Test**
If CI passes, this should be good to go, I'm not sure how it would break anything.


AliSoftware:
We should remove the empty `fastlane/Pluginfile` entirely from the repo as well, for consistency.

mokagio:
🤦‍♂️ 

I wanted to address @AliSoftware's straightforward comment regarding removing `Pluginfile` and merge this, but in doing so, I noticed an additional piece of work having to do with RuboCop. I performed it in https://github.com/wordpress-mobile/release-toolkit/pull/373/commits/4a7282fc55e920b93e72ccc856f862c2295dfe74, but at that point my additional diff was too big for me to be comfortable in merging this PR. That is, because it was more than removing a file we already agreed was unused, I would have wanted someone to review my changes before merging.

So, I created a new branch with the intent of opening a PR on top of this one with my changes.

Then I pushed my branch on top of this one because... muscle memory and tab-completion.

image

_Because my new branch and this one have the same root, tab-completing the branch name resulted in the original branch, not mine._

By the time I realized my blunder and hit `Ctrl C`, the push had already gone through.

I though of reverting my changes on this branch and open the PR as originally intended, but I think we can settle for someone else looking over my changes here and approve + merge it.

AliSoftware (approved):
@mokagio I actually like that you addressed the rubocop thing as well in this PR after all, it makes sense to do those together and it's not that big of a diff 🙂

Very nice catch spotting that rubocop exception being obsoleted by the file removal, btw!

mokagio:
> Very nice catch spotting that rubocop exception being obsoleted by the file removal, btw!

Thanks! 🙌  I've shipped so many half cooked cleanups that I'm (slowly and inconsistently) developing the habit to run a `git grep` for whatever I remove.

PR #377, opened 2022-06-24
Bump ruby version to 2.7.4

AliSoftware:
I can't seem to install ruby `2.6.4` (which the `release-toolkit` is currently using as part of `.ruby-version`) on my M1 Mac, so I figured it was high time to update the `.ruby-version` and `spec.required_ruby_version` of the toolkit to `2.7.4` after all.

<details><summary>Output of trying to <tt>rbenv install</tt> ruby 2.6.4 on my new M1 Mac</summary>

```
$ rbenv install
Downloading openssl-1.1.1n.tar.gz...
-> https://dqw8nmjcqpjn7.cloudfront.net/40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a
Installing openssl-1.1.1n...
Installed openssl-1.1.1n to /Users/olivierhalligon/.rbenv/versions/2.6.4

Downloading ruby-2.6.4.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.6/ruby-2.6.4.tar.bz2
Installing ruby-2.6.4...
ruby-build: using readline from homebrew

BUILD FAILED (macOS 12.4 using ruby-build 20220610)

Inspect or clean up the working tree at /var/folders/fn/g0dpb1lx345g5lg0s65wlp800000gn/T/ruby-build.20220624132740.11615.QGRLQo
Results logged to /var/folders/fn/g0dpb1lx345g5lg0s65wlp800000gn/T/ruby-build.20220624132740.11615.log

Last 10 log lines:
18 warnings generated.
linking shared-object stringio.bundle
5 warnings generated.
4 warnings generated.
linking shared-object zlib.bundle
1 warning generated.
linking shared-object psych.bundle
422 warnings generated.
linking shared-object date_core.bundle
make: *** [build-ext] Error 2
```
</details>

We already use ruby `2.7.4` in our client repos which have the `release-toolkit` as a dependency, which should be proof enough that the release-toolkit doesn't have any issue running with this version of Ruby.


.buildkite/pipeline.yml
```diff
@@ -1,7 +1,7 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   env: &xcode_image
-    IMAGE_ID: xcode-12.5.1
+    IMAGE_ID: xcode-13.4.1
```
AliSoftware, line 4: I also had to bump the CI image used here because our `xcode-12.5.1` image [didn't have ruby 2.7.4 installed on it](https://github.com/Automattic/buildkite-ci/blob/trunk/src/macos-vms/manifests/xcode-12.5.1.json#L7-L9). It was about time we update the CI image used to a more recent one anyway.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-_None_
+* Update the version of Ruby required from `2.6.4` to `2.7.4`. [#377]
```
AliSoftware, line 9: I figured it should be considered a breaking change, since we changed the `spec.required_ruby_version = '>= 2.7'` in the `.gemspec` (which we had to, otherwise `rubocop` warns us that that version and the `.ruby-version` are not in sync), and client projects using the `release-toolkit` but which might still be using ruby `2.6.4` will have to update their ruby to `2.7+` to continue using it, thus making it a breaking change in the pure semver definition of it.

I don't expect we have many client projects still using ruby `2.6.4`, but (1) if we do, that would be the right occasion to finally update them to `2.7.4` anyway (2) that update shouldn't be a huge cost anyway — mostly updating the same files like `.ruby-version` and the `buildkite/pipeline.yml` in those client projects, I'd guess.

mokagio (approved):
I remember reading about, and possibly facing myself, issues with older versions of Ruby on Apple Silicon.

+1 for focusing on upgrading over fixing this. We want to move forward 👍

PR #381, opened 2022-07-05
Fix a typo: "traslation"

mokagio:
I run into this while working on the Pocket Casts iOS localization tooling. Easy fix.


AliSoftware (approved):
This one has bothered me for a long time but never took the time to fix it; thanks!

PR #384, opened 2022-07-20
Release 5.0.1 into `trunk`

mokagio:
New version 5.0.1. Be sure to create a GitHub Release and tag once this PR gets merged.

---

I need this to continue working on 92-gh-Automattic/pocket-casts-ios

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    fastlane-plugin-wpmreleasetoolkit (5.0.0)
+    fastlane-plugin-wpmreleasetoolkit (5.0.1)
       activesupport (~> 5)
       bigdecimal (~> 1.4)
       buildkit (~> 1.5)
@@ -272,7 +272,6 @@ GEM
     method_source (0.9.2)
     mini_magick (4.11.0)
     mini_mime (1.1.2)
-    mini_portile2 (2.8.0)
     minitest (5.14.4)
     molinillo (0.8.0)
     multi_json (1.15.0)
@@ -282,8 +281,7 @@ GEM
     naturally (2.2.1)
     netrc (0.11.0)
     no_proxy_fix (0.1.2)
-    nokogiri (1.13.7)
-      mini_portile2 (~> 2.8.0)
+    nokogiri (1.13.7-x86_64-darwin)
       racc (~> 1.4)
     octokit (4.21.0)
       faraday (>= 0.9)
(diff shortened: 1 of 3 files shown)


mokagio:
@jkmassel I just saw your PR #386 which "invalidates" this release because it adds a new features which would require a 5.1.0 version number, not 5.0.1.

I'm going to close this PR, merge yours, and make a new release PR

mokagio:
> I'm going to close this PR, merge yours, and make a new release PR

And just as I wrote that came the "Daddy I'm hungry" cries from the kids... Might be a while. 🍳

Gemfile.lock
```diff
@@ -282,8 +281,7 @@ GEM
     naturally (2.2.1)
     netrc (0.11.0)
     no_proxy_fix (0.1.2)
-    nokogiri (1.13.7)
-      mini_portile2 (~> 2.8.0)
+    nokogiri (1.13.7-x86_64-darwin)
```
crazytonyli, line 284: I don't own an M1 (yet, Apple finally started to process my order), I could be talking total nonsense here but does this change mean we'll have to run `arch -x86_64 bundle install` instead of natively without Rosetta?

mokagio: Good catch! I had miss this in the diff, how silly of me given the whole reason I went on my Intel Mac was because of the `nokogiri` failure reported in #385.

I played around with the `bundle` configs in #387 and I seem to be able to run the release task from there, but I do wonder whether the problem is in the task itself, given `bundle install` succeeds from both my machines 🤔

mokagio:
Closing in favor of an upcoming 5.1.0 release PR that will include #386.

Gemfile.lock
AliSoftware: This has been a long standing issue with `bundler` and `nokogiri` toggling back and forth between the platform-agnostic version of `nokogiri` and the `x86_64-darwin` one. A [change in the `.bundle/config` to add `BUNDLE_SPECIFIC_PLATFORM: "false"`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/.bundle/config#L3) — as well as ensuring that [`PLATFORMS` just listed `ruby` in the `Gemfile.lock`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/Gemfile.lock#L417-L418) — was _supposed_ to solve it for good (see paaHJt-2Gx-p2)… but apparently not always… 😒 :sigh:

So at that point I don't understand why it keeps resolving it to `x86_64-darwin` sometimes, despite `specific_platform` being disabled… 🤷

Issue #385, opened 2022-07-20
`bundle exec rake new_release` fails


mokagio:
I just opened the PR for 5.0.1, #384, and run into an issue on my M1 Max. When I run the `bundle exec rake new_release` I got the following error:

```
➜ br new_release
>>> Current version is: 5.0.0
>>> Pending CHANGELOG:
| ### Bug Fixes
|
| - Fix metadata length computation logic [[#383](https://github.com/wordpress-mobile/release-toolkit/pull/383)]
|

New version to release? [default: 5.0.1]

git checkout -b release/5.0.1
Switched to a new branch 'release/5.0.1'
>>> Update `VERSION` constant in `version.rb`...
bundle install --quiet
Bundler could not find compatible versions for gem "nokogiri":
  In Gemfile:
    fastlane-plugin-wpmreleasetoolkit was resolved to 5.0.1, which depends on
      nokogiri (~> 1.11)

Could not find gem 'nokogiri (~> 1.11)' with platform 'ruby', which is required by gem 'fastlane-plugin-wpmreleasetoolkit', in locally installed gems.

The source contains the following gems matching 'nokogiri (~> 1.11)':
  * nokogiri-1.13.6-arm64-darwin
  * nokogiri-1.13.7-arm64-darwin
rake aborted!
Command failed with status (6): [bundle install --quiet...]
/Users/gio/Developer/a8c/release-toolkit/Rakefile:89:in `update_version_constant'
/Users/gio/Developer/a8c/release-toolkit/Rakefile:57:in `block in <top (required)>'
/Users/gio/Developer/a8c/release-toolkit/vendor/bundle/ruby/2.7.0/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
/Users/gio/.rbenv/versions/2.7.4/bin/bundle:23:in `load'
/Users/gio/.rbenv/versions/2.7.4/bin/bundle:23:in `<main>'
Tasks: TOP => new_release
(See full trace by running task with --trace)
```

`bundle install` works fine on the same machine.

I had to switch to my Intel Mac for the automation to succeed.

mokagio:
Now that I think about it, I wonder if running `rake` via `bundle exec` had something to do with the issue.

mokagio:
> Now that I think about it, I wonder if running `rake` via `bundle exec` had something to do with the issue.

[...]

PR #387, opened 2022-07-20
Update Bundler config to fix/workaround #385

mokagio:
`BUNDLE_FORCE_RUBY_PLATFORM` `true` should force `nokogiri` being build via source. This seems wasteful given Bundler should be able to use the appropriate platform for it depending on the machine it's running from,
but that hasn't been the case for me recently, see https://github.com/wordpress-mobile/release-toolkit/issues/385.

Without this setting, I am unable to run `bundle exec rake new_release`, getting the `nokogiri` platform errors reported in #385.

Something is on my mind, though: Why does `rm -rf vendor/bundle && bundle install`work on my machine but the`bundle install` run from the `new_release` Rake task fails?


mokagio:
Note: The build is currently [failing](https://buildkite.com/automattic/release-toolkit/builds/655#01821d7e-afe7-4179-b2f8-73251c0f79a8) because of an infrastructure issue.

image

AliSoftware:
Maybe `BUNDLER_PLATFORM_SPECIFIC: "false"` was not enough and the seemingly-random choice from bundler between platform-agnostic vs platform-specific `nokogiri` is still unstable even with this? (See paaHJt-2Gx-p2 and https://github.com/wordpress-mobile/release-toolkit/pull/384#discussion_r935903177)

So even if you didn't seem to have the issue when you made the `5.1.0`, since I remember this (bundler choosing between ruby vs `x86_64-darwin`) was kinda flaky and random… maybe it's still worth adding that `BUNDLE_FORCE_RUBY_PLATFORM` too after all?

AliSoftware:
Rebased this on top of the `release/5.2.0` branch (and updated the target branch of this PR accordingly) since I just encountered a similar issue while doing this new 5.2.0 release.

I will undraft and merge this once CI is green, and try updating `release/5.2.0` to get rid of the platform-specific nokogiri once it has landed in the release branch.

AliSoftware (approved):

mokagio:
👍 

Sad that it had to come to this because it's yet another instance of Bundler being a bit unstable in the past couple of year, at least as far as my experience with it has been.

At the same time, I'm glad my PR was useful. Case in point for extracting knowledge out of one's brain whenever possible, specially in a distributed async-first organization 🤓

AliSoftware:
> Case in point for extracting knowledge out of one's brain whenever possible, specially in a distributed async-first organization 🤓

💯 

tbh I'm still unsure if that definitively fixed the issue at its core or if I was lucky on the second run and all that was coincidental and that workaround didn't have a real effect on how `bundle` decides which version of `nokogiri` to pick… we'll see on subsequent releases, I guess.

PR #390, opened 2022-08-02
Add `tools:ignore="InconsistentArrays"` to `available_languages.xml`

AliSoftware:
This should address the lint issue on repos using multiple flavors for multiple apps, and thus needing multiple `available-languages.xml`, like WP vs JP do.

See also p8Qyks-351-p2#comment-1298 and https://github.com/wordpress-mobile/WordPress-Android/commit/35103522f8c410f0d4aa81906e97bfd42377f4e9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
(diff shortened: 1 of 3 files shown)


ParaskP7 (approved):
LGTM! 💯

PR #391, opened 2022-08-05
Fixes milestones being compared as strings

oguzkocer:
We had a few failures in `create_new_milestone` in WCiOS and WCAndroid returning `already_exists` error. This was due to `get_last_milestone` returning `9.9` as the last milestone when in fact `10.0` is the last one. It turns out we have been doing a string comparison for the parts of milestone instead of an integer one and Ruby says `"10" > "9" = false`, so 🤷 

This PR updates the milestone comparison to an integer one and introduces a unit test that fails without this fix.

**To test:**
* Revert `8d7bd7c19a59433ea3611d0c2c11f9a5991a43f6` and run the new unit test and verify that it fails
* Run the unit test again and verify that it succeeds (this is also covered by CI)

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -70,7 +70,7 @@ module Fastlane
           else
             begin
               last_vcomps = last_stone[:title].split[0].split('.')
-              last_stone = mile if mile_vcomps[0] > last_vcomps[0] || mile_vcomps[1] > last_vcomps[1]
+              last_stone = mile if Integer(mile_vcomps[0]) > Integer(last_vcomps[0]) || Integer(mile_vcomps[1]) > Integer(last_vcomps[1])
             rescue StandardError
               puts 'Found invalid milestone'
             end
--- a/spec/github_helper_spec.rb
+++ b/spec/github_helper_spec.rb
@@ -92,6 +92,30 @@ describe Fastlane::Helper::GithubHelper do
     end
   end
 
+  describe 'get_last_milestone' do
+    let(:test_repo) { 'repo-test/project-test' }
+    let(:last_stone) { mock_milestone('10.0') }
+    let(:client) do
+      instance_double(
+        Octokit::Client,
+        list_milestones: ['9.8 ❄️', '9.9'].map { |title| mock_milestone(title) }.append(last_stone)
+      )
+    end
+
+    before do
+      allow(described_class).to receive(:github_client).and_return(client)
+    end
+
+    it 'returns correct milestone' do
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -70,7 +70,7 @@ def self.get_last_milestone(repository)
           else
             begin
               last_vcomps = last_stone[:title].split[0].split('.')
-              last_stone = mile if mile_vcomps[0] > last_vcomps[0] || mile_vcomps[1] > last_vcomps[1]
+              last_stone = mile if Integer(mile_vcomps[0]) > Integer(last_vcomps[0]) || Integer(mile_vcomps[1]) > Integer(last_vcomps[1])
```
AliSoftware, line 73: Note that `Integer(x)` will _crash_ if `x` cannot be interpreted as an Integer — as opposed to `x.to_i` which will return `0` in such cases.

I'm not sure what's best and what we want, but in the (unlikely, but still possible) case that we have milestones like `alpha.1` in GitHub that will end up having `mile_vcomps[0]` be `'alpha'` and have this crash, I wonder if `.to_i` wouldn't be better? (That being said, we have a `rescue StandardError` block for if `Integer(x)` fails, so… 🤷 )

spec/github_helper_spec.rb
```diff
@@ -92,6 +92,30 @@
...
+    end
+
+    before do
+      allow(described_class).to receive(:github_client).and_return(client)
+    end
+
+    it 'returns correct milestone' do
+      expect(client).to receive(:list_milestones)
+      expect(described_class.get_last_milestone(repository: test_repo)).to eq(last_stone)
+    end
```
AliSoftware, line 112: Thanks for adding unit tests for those!

I think it would be worth to add some more, to cover for cases like unexpected milestone names (e.g. `alpha.1`), or the `last_stone` being the frozen one with the emoji, and the likes. That can totally be for a separate PR, so that we don't block this and can get the fix for the upcoming code freeze, but still a nice addition to keep in mind.

AliSoftware (approved):
Approved to unblock, but disabled auto-merge to let you the opportunity to check my feedback and decide if you want to apply my suggestion or not (I'm still not sure what's best tbh)

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
oguzkocer: I'd prefer a crash over unpredictable behaviour every time. If we have a special case, we should handle that and a crash is a great reminder for it.

AliSoftware: Good point.

PR #396, opened 2022-08-11
Add optional `release_notes_file_path` arg to `update_release_notes`

### LINKED ISSUE #395: Remove hardcoded references to `RELEASE-NOTES.txt`
We discussed internally (pdnsEh-aj-p2) that it would be nice to use `CHANGELOG.md` instead of `RELEASE-NOTES.txt`, which is the de-facto standard for tracking changes in open source projects these days.

@leandroalonso recently [noticed](104-gh-Automattic/pocket-casts-ios#discussion_r932156026) that a considerable blocker to implementing this change is the fact that the toolkit has hardcoded references to `RELEASE-NOTES.txt` in it. Example:

https://github.com/wordpress-mobile/release-toolkit/blob/b2a0ecb7276c48f167a299e78027a0196597cb56/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_update_release_notes.rb?rgh-link-date=2022-07-28T12%3A27%3A54Z#L11

Before we can begin renaming the files, we need to remove any hardcoded reference (which will be a breaking change) and replace them with configurable parameters (maybe defaulting to `CHANGELOG.md` for convenience).

leandroalonso:
Fixes #395 

This PR changes the fastlane actions `ios_update_release_notes` and `android_update_release_notes` to accept an optional `release_notes_file_path`. This way our projects are not forced to use `RELEASE-NOTES.txt`.

If a file path is not given, it will use `RELEASE-NOTES.txt` as it was before, not affecting any of the current projects that use it.

### To test

1. Run the tests added for this PR
2. Code review!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ _None_
 
 - Add `tools:ignore="InconsistentArrays"` to `available_languages.xml` to avoid a linter warning on repos hosting multiple app flavors. [#390]
 - Add the ability to provide a custom message for builds triggered via `buildkite_trigger_build` action [#392]
+- Add optional `release_notes_file_path` to `ios_update_release_notes` and `android_update_release_notes` [#396]
 
 ### Bug Fixes
 
--- a/spec/android_update_release_notes_spec.rb
+++ b/spec/android_update_release_notes_spec.rb
@@ -2,9 +2,10 @@ require 'spec_helper'
 
 describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
   let(:release_notes_txt) { File.join(File.dirname(__FILE__), 'RELEASE-NOTES.txt') }
+  let(:changelog_md) { File.join(File.dirname(__FILE__), 'CHANGELOG.md') }
 
   after do
-    FileUtils.remove_entry release_notes_txt
+    FileUtils.rm([release_notes_txt, changelog_md], force: true)
   end
 
   describe '#android_update_release_notes' do
@@ -21,6 +22,21 @@ describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
 
       # Assert
       expect(File.read(release_notes_txt)).to eq("1.1\n-----\n\n\n")
+    end
+
+    it 'adds a new section on the given file' do
+      # Arrange
+      ENV['PROJECT_ROOT_FOLDER'] = File.dirname(__FILE__)
+      File.write(changelog_md, '')
+
+      # Act
+      run_described_fastlane_action(
+        new_version: '1.0',
+        release_notes_file_path: changelog_md
+      )
+
+      # Assert
+      expect(File.read(changelog_md)).to eq("1.1\n-----\n\n\n")
     end
   end
 end
--- a/spec/android_update_release_notes_spec.rb
+++ b/spec/android_update_release_notes_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
+  let(:release_notes_txt) { File.join(File.dirname(__FILE__), 'RELEASE-NOTES.txt') }
+
+  after do
+    FileUtils.remove_entry release_notes_txt
+  end
+
+  describe '#android_update_release_notes' do
+    it 'adds a new section on RELEASE-NOTES.txt' do
+      # Arrange
+
+      ENV['PROJECT_ROOT_FOLDER'] = File.dirname(__FILE__)
+      File.write(release_notes_txt, '')
+
+      # Act
+      run_described_fastlane_action(
+        new_version: '1.0'
+      )
+
+      # Assert
+      expect(File.read(release_notes_txt)).to eq("1.1\n-----\n\n\n")
+    end
+  end
+end
--- a/spec/android_update_release_notes_spec.rb
+++ b/spec/android_update_release_notes_spec.rb
@@ -1,6 +1,10 @@
 require 'spec_helper'
 
 describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
+  after do
+    ENV['PROJECT_ROOT_FOLDER'] = nil
+  end
+
   describe '#android_update_release_notes' do
     it 'adds a new section on RELEASE-NOTES.txt' do
       in_tmp_dir do |tmp_dir|
--- a/spec/android_update_release_notes_spec.rb
+++ b/spec/android_update_release_notes_spec.rb
@@ -5,13 +5,37 @@ describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
     ENV['PROJECT_ROOT_FOLDER'] = nil
   end
 
+  let(:new_section) do
+    <<~CONTENT
+      1.1
+      -----
+
+
+    CONTENT
+  end
+  let(:content) do
+    <<~CONTENT
+      1.0
+      -----
+      - Item 1 for v1.0
+      - Item 2 for v1.0
+
+      // Comment in the middle
+
+      0.9.0
+      -----
+      - Item 1 for v0.9.0
+      - Item 2 for v0.9.0
+    CONTENT
+  end
+
   describe '#android_update_release_notes' do
     it 'adds a new section on RELEASE-NOTES.txt' do
       in_tmp_dir do |tmp_dir|
         # Arrange
         ENV['PROJECT_ROOT_FOLDER'] = tmp_dir
         release_notes_txt = File.join(tmp_dir, 'RELEASE-NOTES.txt')
-        File.write(release_notes_txt, ANDROID_FAKE_CONTENT)
+        File.write(release_notes_txt, content)
 
         # Act
         run_described_fastlane_action(
@@ -19,7 +43,7 @@ describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
         )
 
         # Assert
-        expect(File.read(release_notes_txt)).to eq(ANDROID_NEW_SECTION + ANDROID_FAKE_CONTENT)
+        expect(File.read(release_notes_txt)).to eq(new_section + content)
       end
     end
 
@@ -28,7 +52,7 @@ describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
         # Arrange
         ENV['PROJECT_ROOT_FOLDER'] = tmp_dir
         changelog_md = File.join(tmp_dir, 'CHANGELOG.md')
-        File.write(changelog_md, ANDROID_FAKE_CONTENT)
+        File.write(changelog_md, content)
 
         # Act
         run_described_fastlane_action(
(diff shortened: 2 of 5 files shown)


spec/android_update_release_notes_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
+  let(:release_notes_txt) { File.join(File.dirname(__FILE__), 'RELEASE-NOTES.txt') }
+  let(:changelog_md) { File.join(File.dirname(__FILE__), 'CHANGELOG.md') }
+
+  after do
+    FileUtils.rm([release_notes_txt, changelog_md], force: true)
+  end
```
AliSoftware: Please don't make the tests write in the directory of the repo itself.

It's better to use `Dir.mktmpdir` to create a temporary directory (somewhere in `/tmp`, the method will take care of using the system temp dir and find a unique name for that) and write the temp file in it.

We also have `in_tmp_dir` and `with_tmp_file` helpers defined in `spec_helper.rb` that will surely help there.

spec/ios_update_release_notes_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::IosUpdateReleaseNotesAction do
+  let(:release_notes_txt) { File.join(File.dirname(__FILE__), 'RELEASE-NOTES.txt') }
+  let(:changelog_md) { File.join(File.dirname(__FILE__), 'CHANGELOG.md') }
+
+  after do
+    FileUtils.rm([release_notes_txt, changelog_md], force: true)
+  end
```
AliSoftware: Same, please don't make running the test result in creating files in the repo itself.

This is especially problematic if your tests happen to crash and the files would not be deleted, unlike with using `Dir.mktmpdir` or `in_tmp_dir` or `with_tmp_file`.

leandroalonso:
@AliSoftware @mokagio JFYI: there are a few tests failing here but they're also failing on `trunk`.

AliSoftware (changes requested):
This mostly look good to me, except the fact that your tests should not write where your test code is (don't sh*t where you eat ^^), especially to avoid polluting the code and git status if your test suite crashes for example.

Also, I'd suggest you to take this occasion to also update the other actions dealing with release notes, especially the ones that extract the section of the release notes file corresponding to the version you're about to code freeze(in order to extract the bullet points to be sent to editorialization and then translation).

spec/android_update_release_notes_spec.rb
leandroalonso: Done in 6baf51dbd13b8667d9503cb60c63ad399b19f47b and 9d4b18a2ff1b408150c79588a89dd315392a6956

spec/ios_update_release_notes_spec.rb
leandroalonso: Done in 6baf51dbd13b8667d9503cb60c63ad399b19f47b and 9d4b18a2ff1b408150c79588a89dd315392a6956

leandroalonso:
> Also, I'd suggest you to take this occasion to also update the other actions dealing with release notes, especially the ones that extract the section of the release notes file corresponding to the version you're about to code freeze(in order to extract the bullet points to be sent to editorialization and then translation).

I'm not sure I follow that. What do you mean by updating them? The actions I changed are the only ones that had a `RELEASE-NOTES.txt` hardcoded in them, so I kept the PR scope small in order to fix #395.

spec/android_update_release_notes_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
+  describe '#android_update_release_notes' do
+    it 'adds a new section on RELEASE-NOTES.txt' do
+      in_tmp_dir do |tmp_dir|
+        # Arrange
+        ENV['PROJECT_ROOT_FOLDER'] = tmp_dir
+        release_notes_txt = File.join(tmp_dir, 'RELEASE-NOTES.txt')
+        File.write(release_notes_txt, '')
```
AliSoftware: It would be more resilient to have the fixture actually contain some text before we apply the action, to test if the action actually prepends to the existing content, instead of appending, or replacing the whole existing content.

Even if this use case is already covered in depth by [the existing specs for the `Fastlane::Helper::ReleaseNotesHelper`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/release_notes_helper_spec.rb) — which test the helper itself and not the wrapping actions using it like you're doing here — it doesn't hurt to make a stronger test here too.

AliSoftware (approved):
Approving to unblock now that you fixed the tests, but still left a suggestion to improve said tests a bit before merging 🙂

AliSoftware:
> > Also, I'd suggest you to take this occasion to also update the other actions dealing with release notes, especially the ones that extract the section of the release notes file corresponding to the version you're about to code freeze(in order to extract the bullet points to be sent to editorialization and then translation).
> 
> I'm not sure I follow that. What do you mean by updating them? The actions I changed are the only ones that had a `RELEASE-NOTES.txt` hardcoded in them, so I kept the PR scope small in order to fix #395.

I was thinking about [this `extract_release_notes_for_version` action](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/extract_release_notes_for_version_action.rb), but it happens that this other action [already had a `ConfigItem` to specify the release notes file path](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/extract_release_notes_for_version_action.rb#L71-L75), so nothing to fix on that other one after all!

spec/android_update_release_notes_spec.rb
leandroalonso: Yeah, my main reasoning was that this is already covered in another test. But having an empty fixture feels a little bit weird, changed in 8f1dea21b17f8920c3eb131ff8fea4b1cb535721 and cd42ef4abcf374721c61a716422d2b242b4cd45b

leandroalonso:
@AliSoftware last question: what about the failing unit tests (also failing on `trunk`)? Can I just go ahead and merge this PR?

AliSoftware:
> @AliSoftware last question: what about the failing unit tests (also failing on `trunk`)? Can I just go ahead and merge this PR?

Good question. I'm not sure how that test failure got into `trunk` in the first place without failing and preventing the PR that introduced the change to be merged…

Ideally I'd prefer that we fix the tests that are failing in `trunk` in a dedicated PR first, land that test-fixing PR in `trunk`, then merge trunk back in your PR here to have it go green.

Looking at the current test failures, I don't think that would be too hard to fix (it seems to be just some `expect` that didn't account for the path to a file to be an absolute path to a tmp dir instead of a relative path… so updating the `expect` in the tests according should be enough).
@leandroalonso Would you mind taking a look? (I'm a bit focused on something else right now)

spec/android_update_release_notes_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        )
+
+        # Assert
+        expect(File.read(changelog_md)).to eq(ANDROID_NEW_SECTION + ANDROID_FAKE_CONTENT)
+      end
+    end
+  end
+end
+
+ANDROID_FAKE_CONTENT = <<~CONTENT.freeze
```
AliSoftware: Prefer using a `let(:fake_content) { … }` to provide constant-like values in specs, rather than top-level constants (which will end up being publicly exposed/visible to the all rest of the codebase… as you realized in your last commit)

AliSoftware (changes requested):

spec/android_update_release_notes_spec.rb
AliSoftware: PS: I know, `release_notes_helper_spec` does this as well… but that was written before we knew better and were better at Ruby and rspec 😛  (ideally we should change the ones in `release_notes_helper_spec.rb` as well)

leandroalonso: I said nothing 😆

Done in f0ced0647ed21ea0ca3d33440a8d8a7fda27bb41

leandroalonso:
@AliSoftware the tests failing were actually my fault (plus a test that always fails on my machine). 😝 It all should be green now. 🟢

spec/android_update_release_notes_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::AndroidUpdateReleaseNotesAction do
+  after do
+    ENV['PROJECT_ROOT_FOLDER'] = nil
+  end
```
AliSoftware: What do you think to move this into an `ensure` block in each of the `in_tmp_dir` calls, instead of using an `after` block?

Or even better, given the actions are supposed to handle cases where `PROJECT_ROOT_FOLDER` is not set, and that the `release_notes_txt` or `changelog_md` path you pass to the action [here](https://github.com/wordpress-mobile/release-toolkit/pull/396/files#diff-b213d2371b4a182b49d9e5ae7f2b8794b1076ac56e15f0204bd4b49a12493fd0R61) are an **absolute paths** (which already includes the whole `tmp_dir` path in it)… I think your tests **don't need to define a value for `ENV['PROJECT_ROOT_FOLDER']` anymore to pass in the first place**… So you might as well remove all the lines in your tests that sets that env var and that should still pass?

leandroalonso: You're right! Removed in 091de20cabd5debfcc20a0a8df48f16e044b925b

AliSoftware (approved):
:shipit:

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ _None_
 
 - Add `tools:ignore="InconsistentArrays"` to `available_languages.xml` to avoid a linter warning on repos hosting multiple app flavors. [#390]
 - Add the ability to provide a custom message for builds triggered via `buildkite_trigger_build` action [#392]
+- Add optional `release_notes_file_path` to `ios_update_release_notes` and `android_update_release_notes` [#396]
```
AliSoftware, line 29: @leandroalonso Oh shoot I just realized that I failed to spot that this release notes entry was in the wrong place, it should have been added under the `trunk` section (as 5.2.0 has already been published some time ago, while any new PR is yet to be).

As a result, when @oguzkocer  did a new release a couple of minutes ago in #398, our release script didn't see that this was a new entry under "New Features" when it tried to determine the next version to bump to according to SemVer.

That is not dramatic because the version that got released luckily was also to ship a "New Feature" (ie minor version bump according to semver) so we ended up with the correct version number `5.3.0` being released after all.

Still, for consistency, it would be nice to fix that, by moving this entry from `5.2.0` to `5.3.0`, and also update the description of the `5.3.0` GitHub Release accordingly; even if that would be fixed in `trunk` only after the fact and release, better fix the inconsistency for the future.

oguzkocer: Fixed in #399.

leandroalonso: Thank you so much @oguzkocer!

PR #397, opened 2022-08-11
Support custom milestone duration

oguzkocer:
For Woo apps we are experimenting with weekly releases where we code freeze on Fridays and release on Mondays. This PR updates the `create_new_milestone` action and its helper to support a custom `milestone_duration` and `number_of_days_from_code_freeze_to_release`.

If I worked on this feature from scratch, I'd not implement it this way, but I wanted to avoid a breaking change so the changes are mostly replacing the hardcoded values with optional parameter values.

Unfortunately this action and its helper didn't have any unit tests and I don't currently have the bandwidth to implement them.

**To test**

Follow the testing instructions from https://github.com/woocommerce/woocommerce-android/pull/7173 or https://github.com/woocommerce/woocommerce-ios/pull/7458.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
@@ -12,11 +12,12 @@ module Fastlane
         last_stone = Fastlane::Helper::GithubHelper.get_last_milestone(repository)
         UI.message("Last detected milestone: #{last_stone[:title]} due on #{last_stone[:due_on]}.")
         milestone_duedate = last_stone[:due_on]
-        newmilestone_duration = params[:milestone_duration]
-        newmilestone_duedate = (milestone_duedate.to_datetime.next_day(newmilestone_duration).to_time).utc
+        milestone_duration = params[:milestone_duration]
+        newmilestone_duedate = (milestone_duedate.to_datetime.next_day(milestone_duration).to_time).utc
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
```diff
@@ -48,6 +50,18 @@ def self.available_options
                                        optional: true,
                                        is_string: false,
                                        default_value: false),
+          FastlaneCore::ConfigItem.new(key: :milestone_duration,
+                                       env_name: 'GHHELPER_MILESTONE_DURATION',
```
jkmassel, line 54: IMHO we don't need an `env_name` for this, as it IMHO encourages magic global variables set in the `Fastfile`. 

Replacing that with a constant in the `Fastfile` that's passed as an argument to this method is IMHO far better, even if it kinda looks the same code-wise – it'll hopefully push us toward better (and more testable) patterns.

jkmassel (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
oguzkocer: If it was up to me, I'd burn all of the ENV vars with eternal fire as it's nothing but global variables on steroids. I was only following the code conventions we currently have in the repository, but might have misinterpreted it.

The PR is auto-merged and I don't think it's worth following up right now. Let me know if you disagree and I'll open a new PR.

PR #401, opened 2022-08-16
Support Bulk Download from GlotPress

AliSoftware:
> **Warning**: This is a Work in Progress branch. I've only commited this to the remote and opened this draft PR to upload and keep a backup of my experiment while I refine the implementation, but this is still early stage of the implementation

 🚧  Next Steps / TODO

 - [ ] Fix `rubocop` violations
 - [ ] Implement robust Unit Tests for all the new classes and stages
 - [ ] Implement the fastlane actions that will use those new implementations
 - [ ] Undraft this PR

---

## Why?

With us encountering the `429 Too Many Requests` quota limitation more and more often while downloading translations from GlotPress during our release duties, I've recently asked our Team Global and Orbit (ref: pxLjZ-72D-p2) to install [the `gp-import-export` plugin](https://github.com/Automattic/gp-import-export) on our GlotPress instances (ref: 307-gh-Automattic/Orbit), so that we can bulk-download the translations for all the locales at once (as a ZIP).

This will allow us to only make a single network request to export them instead of one request per locale (which ends up being a lot of requests especially for WordPress and the 40+ locales it supports)

## How?

I've reimplemented the logic for the download of GlotPress exports into a more modern infrastructure / code architecture in our code base. I intend to deprecate `glotpress_helper` and `gp_downloadmetadata_action` once this new implementation is validated as more stable, and refactoring `android_download_translations_action` / `ios_downlaod_strings_files_from_glotpress` to use the new helpers and models (or deprecating those actions and creating new ones to replace them).

The new implementation/architecture is structured like this:

### Models to manipulate locales easily

This will supersede my #296 draft PR (which I opened… one year ago!)

 - `models/locale`: is a `Struct` which represents a single locale, exposing the various locale code representations for that locale — since there are multiple ISO standards for locale codes needed/used in various places of the codebase (ios folder names, android folder names, glotpress locale codes…)
 - `models/locales`: represents a set of `Locale` objects. That model allows to build a set of locales based on their locale codes (to have an object representing all the locales supported by a project/app, typically), build convenience sets (e.g. Mag16), easily add/remove/filter locales from sets, and find a known locale from the set of all known locales

### A class dedicated to download exports from GlotPress

The `helper/glotpress_downloader` class knows how to exports from GlotPress in `android`, `ios` or `json` formats.

 - You start by initializing an instance with a GlotPress URL host and URL project path
 - You can then call methods to download exports from that project:
    - Either for a single, given locale, one by one
[...]

PR #402, opened 2022-08-18
Add an option to retry when GlotPress download fails on "429 Too Many Requests" errors

AliSoftware:
## Why?

We still keep having `429 - Too Many Requests` errors from time to time when running the actions which download the app translations from GlotPress.

For example, on average I encounter those on at least 1-3 locales every time I'm doing a new beta or final release, forcing me to interrupt the lane, discard the unstaged changes in git, restart the lane from scratch again, and pray it will pass on those subsequent tries, otherwise I'll have to do all that again. And since the more you retry, the closer you are to the quota limit… things tend to actually get worse on subsequent retries.

This has become quite painful while doing betas and releases, especially for WordPress (which has more than 40 locales to download, so the chances of hitting the `429` error on at least one of them are higher)

## How?

This simply asks the user if they want to retry a the download of a locale if one fails.

 - This not only allows us to avoid having to restart the whole lane and thus the download of all locales from scratch (reducing the risk of hitting the quota once again)
 - But also allows the script to pause until the user replies to the `UI.confirm` prompt before retrying, which allows to leave some time before the first and next retry of the request, giving us even more chances for the quota limit to cool down before the next try.

## Related work

In parallel to that, other work is being done to help getting rid of this `429` error and quota limit. Especially, we've asked for the `gp-import-export` plugin to be installed on both of our GlotPress instances so we could do bulk-downloads of all the locales at once with a single request, and we plan to implement the code to in the `release-toolkit` to be able to use it hopefully soon (we already [have a Draft here](https://github.com/wordpress-mobile/release-toolkit/pull/401), but it's still very early stage so far).

Supporting bulk-download in the `release-toolkit` will require a bit more involved work though, for which I don't have the bandwidth for just yet. So this workaround of still using one-at-a-time locales download but allowing retry would hopefully help mitigate the issue in the meantime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- Propose to retry when the download of GlotPress translations failed for a locale (especially useful for occurrences of `429 - Too Many Requests` quota limits) [#402]
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -291,6 +291,7 @@ module Fastlane
             uri.open(options) { |f| Nokogiri::XML(f.read.gsub("\t", '    '), nil, Encoding::UTF_8.to_s) }
           rescue StandardError => e
             UI.error "Error downloading #{locale} - #{e.message}"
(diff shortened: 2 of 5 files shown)


spencertransier (approved):
These changes look good to me! I read through it and it all makes sense 👍

mokagio (approved):
I'd normally make a snarky remark to suggest adding tests for the retry logic. But I don't think it's needed in this instance because:

- Requires non-trivial test setup to cover a niche feature
- We already have a new implementation of the whole workflow in the works, so any extra time spent here has low ROI

PR #403, opened 2022-08-18
FTL: Support providing a `--test-targets` filter

AliSoftware:
## Why?

After migrating the code able to trigger Firebase Test Lab from CircleCI and [our related CircleCI orb](https://github.com/wordpress-mobile/circleci-orbs/blob/trunk/src/android/orb.yml#L81-L206) to [move it to a fastlane action in release-toolkit](https://github.com/wordpress-mobile/release-toolkit/pull/355) during the Buildkite migration, we forgot to add an option to filter the test targets, like we had and [used to use in e.g. WordPress Android to exclude screenshots from the instrumented/connected tests](https://github.com/wordpress-mobile/WordPress-Android/blob/2b4b4f1737ce387c2e95929ef8ebcdbdd854e68f/.circleci/config.yml#L156)

## How

Add an optional `test_targets` parameter (aka `ConfigItem`) to the `android_firebase_test` action which is then passed to the invocation of `gcloud`

## To test

The unit tests (`spec/firebase_test_runner_spec.rb`) have been updated to cover for cases with or without a `test_targets` parameter being provided, and for testing the proper escaping of the filter if one is provided.
So if CI is green, it should be all good to go.

## What's next

To workaround the fact that we couldn't filter the test targets during instrumented tests, @pachlava had to temporarily `@Ignore` the Screenshot tests in WordPress Android (https://github.com/wordpress-mobile/WordPress-Android/pull/16970).

Once this lands, we should then:
 - [x] Make a new release of the `release-toolkit`
 - [x] Point WordPress-Android to that new version of the toolkit
 - [x] Update [the call site of `android_firebase_test` in `build_and_run_instrumented_tests`](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/fastlane/lanes/test.rb#L24-L32) to add `test_targets: 'notPackage org.wordpress.android.ui.screenshots'` to the parameters
 - [x] Revert the changes made by https://github.com/wordpress-mobile/WordPress-Android/pull/16970 to un-ignore screenshot tests (and thus be able to progress on the work on Android screenshots)

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/models/firebase_test_runner.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/models/firebase_test_runner.rb
@@ -20,20 +20,21 @@ module Fastlane
     # @param [FirebaseDevice] device The virtual device to run tests on.
     # @param [String] type The type of test to run.
     #
-    def self.run_tests(project_id:, apk_path:, test_apk_path:, device:, type: 'instrumentation')
+    def self.run_tests(project_id:, apk_path:, test_apk_path:, device:, test_targets: nil, type: 'instrumentation')
       raise "Unable to find apk: #{apk_path}" unless File.file?(apk_path)
       raise "Unable to find apk: #{test_apk_path}" unless File.file?(test_apk_path)
       raise "Invalid Type: #{type}" unless VALID_TEST_TYPES.include?(type)
 
-      command = Shellwords.join [
-        'gcloud', 'firebase', 'test', 'android', 'run',
-        '--project', project_id,
-        '--type', type,
-        '--app', apk_path,
-        '--test', test_apk_path,
-        '--device', device.to_s,
-        '--verbosity', 'info',
-      ]
+      params = {
+        project: project_id,
+        type: type,
+        app: apk_path,
+        test: test_apk_path,
+        'test-targets': test_targets,
+        device: device.to_s,
+        verbosity: 'info'
+      }.compact.flat_map { |k, v| ["--#{k}", v] }
+      command = Shellwords.join(['gcloud', 'firebase', 'test', 'android', 'run', *params])
 
       log_file_path = Fastlane::Actions.lane_context[:FIREBASE_TEST_LOG_FILE_PATH]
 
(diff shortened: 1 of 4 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/models/firebase_test_runner.rb
```diff
@@ -20,20 +20,21 @@ def self.preflight(verify_gcloud_binary: true, verify_logged_in: true)
...
+      params = {
+        project: project_id,
+        type: type,
+        app: apk_path,
+        test: test_apk_path,
+        'test-targets': test_targets,
+        device: device.to_s,
+        verbosity: 'info'
+      }.compact.flat_map { |k, v| ["--#{k}", v] }
+      command = Shellwords.join(['gcloud', 'firebase', 'test', 'android', 'run', *params])
```
spencertransier, line 37: @AliSoftware Can you explain the benefits of switching to this syntax from the previous syntax? I'm not questioning it, just curious 😃

spencertransier (approved):
Approved with just a curious question from me about one of the changes 🙂

lib/fastlane/plugin/wpmreleasetoolkit/models/firebase_test_runner.rb
AliSoftware: This syntax allowed me to inject the `compact` call on the `Hash` before transforming it into an array of key-value-pairs.

The effect of that `.compact` is to remove any entry from the `Hash` with a `nil` value, which was necessary to completely remove `--test-targets <value>` from the params if the user didn't provide any value for it. Otherwise we'd be passing `--test-targets ""` (or actually even worse, just `--test-targets` without a corresponding value next to it) to the `gcloud` command, instead of skipping that param entirely when it's not provided by the user.

spencertransier: Makes sense, thank you for explaining!

spec/firebase_test_runner_spec.rb
```diff
@@ -31,6 +31,13 @@
       run_tests
     end
 
+    it 'includes and properly escape the test filters if any are provided' do
```
mokagio: Nitpick: 
```suggestion
    it 'includes and properly escapes the test targets if any are provided' do
```

The "targets" suggestion is because that's how they are referred to in the code.

mokagio (approved):

pachlava:
Thank you for this PR @AliSoftware! ⭐

PR #406, opened 2022-09-06
Handle 429 and prevent 301 in `gp_downloadmetadata`

mokagio:
During the WordPress 20.6 finalization, we noticed that `gp_downloadmetadata` didn't offer to retry the GlotPress request when receiving a 429 (see details in [this comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/19267#discussion_r960523291) and [this commit](https://github.com/wordpress-mobile/WordPress-iOS/pull/19275/commits/7c3289fcc7fe0b493aa9a49696bb3ba85a85646a))

This PR adds a crude retry mechanism to the action.

image

While working on this, I noticed that the URL used was lacking the `/`, consistently resulting in a 301 from the API. I addressed that, too.

`gp_downloadmetadata` is set to be removed in favor of a better alternative as soon as we'll have time for it. I tried to write new code that was good, but kept the refactoring and improvements to a minimum.

### Testing

I tested this by pointing my  WordPress iOS to my local copy of the repo.

```diff
--- a/fastlane/Pluginfile
+++ b/fastlane/Pluginfile
@@ -8,7 +8,8 @@ end

 # This comment avoids typing to switch to a development version for testing.
 # gem 'fastlane-plugin-wpmreleasetoolkit', git: 'git@github.com:wordpress-mobile/release-toolkit', branch: 'tru>
-gem 'fastlane-plugin-wpmreleasetoolkit', '~> 5.1'
+# gem 'fastlane-plugin-wpmreleasetoolkit', '~> 5.1'
+gem 'fastlane-plugin-wpmreleasetoolkit', path: '../../release-toolkit'
```

I then run `bundle exec fastlane download_wordpress_localized_app_store_metadata` till I got a 429.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-_None_
+- Propose to retry when `gp_downloadmetadata` receives a `429 - Too Many Requests` error. [#406]
 
 ### New Features
 
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- Update the URL used by `gp_downloadmetadata` to prevent consistent `301` responses. [#406]
 
 ### Internal Changes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
@@ -20,13 +20,13 @@ module Fastlane
         params[:locales].each do |loc|
           if loc.is_a?(Array)
             UI.message "Downloading language: #{loc[1]}"
-            complete_url = "#{params[:project_url]}#{loc[0]}/default/export-translations?filters[status]=current&format=json"
+            complete_url = "#{params[:project_url]}#{loc[0]}/default/export-translations/?filters[status]=current&format=json"
             downloader.download(loc[1], complete_url, loc[1] == params[:source_locale])
           end
 
           if loc.is_a?(String)
             UI.message "Downloading language: #{loc}"
-            complete_url = "#{params[:project_url]}#{loc}/default/export-translations?filters[status]=current&format=json"
+            complete_url = "#{params[:project_url]}#{loc}/default/export-translations/?filters[status]=current&format=json"
             downloader.download(loc, complete_url, loc == params[:source_locale])
           end
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
@@ -19,13 +19,13 @@ module Fastlane
 
         params[:locales].each do |loc|
           if loc.is_a?(Array)
-            puts "Downloading language: #{loc[1]}"
+            UI.message "Downloading language: #{loc[1]}"
             complete_url = "#{params[:project_url]}#{loc[0]}/default/export-translations?filters[status]=current&format=json"
             downloader.download(loc[1], complete_url, loc[1] == params[:source_locale])
           end
 
           if loc.is_a?(String)
-            puts "Downloading language: #{loc}"
+            UI.message "Downloading language: #{loc}"
             complete_url = "#{params[:project_url]}#{loc}/default/export-translations?filters[status]=current&format=json"
             downloader.download(loc, complete_url, loc == params[:source_locale])
           end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
@@ -16,12 +16,7 @@ module Fastlane
       def download(target_locale, glotpress_url, is_source)
         uri = URI(glotpress_url)
         response = Net::HTTP.get_response(uri)
-        response = Net::HTTP.get_response(URI.parse(response.header['location'])) if response.code == '301'
-
-        @alternates.clear
-        loc_data = JSON.parse(response.body) rescue loc_data = nil
-        parse_data(target_locale, loc_data, is_source)
-        reparse_alternates(target_locale, loc_data, is_source) unless @alternates.length == 0
+        handle_glotpress_download(response: response, locale: target_locale, is_source: is_source)
       end
 
       # Parse JSON data and update the local files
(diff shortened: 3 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
```diff
@@ -19,14 +19,14 @@ def self.run(params)
 
         params[:locales].each do |loc|
           if loc.is_a?(Array)
-            puts "Downloading language: #{loc[1]}"
-            complete_url = "#{params[:project_url]}#{loc[0]}/default/export-translations?filters[status]=current&format=json"
+            UI.message "Downloading language: #{loc[1]}"
```
mokagio, line 22: Converted the `puts` to `UI.message` for more consistent logs.

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
```diff
@@ -100,6 +95,33 @@ def delete_existing_metadata(target_locale)
...
+        when '301'
+          # Follow the redirect
+          UI.message("Received 301 for `#{locale}`. Following redirect...")
+          download(locale, response.header['location'], is_source)
+        when '429'
+          # We got rate-limited, offer to try again
+          if UI.confirm("Retry downloading `#{locale}` after receiving 429 from the API?")
+            download(locale, response.uri, is_source)
+          else
+            UI.error("Abandoning `#{locale}` download as requested.")
```
mokagio, line 119: I decided to make this `error` so it would print red to mark the difference in behavior.

image
_(The `301` is because I hadn't updated the URLs yet)_

I can see how `error`/red might be misleading though, and look like a wholesale failure. If you think that's too confusing, I can update to a `message`:

```suggestion
            UI.message("Abandoning `#{locale}` download as requested.")
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
```diff
@@ -16,12 +16,7 @@ def initialize(target_folder, target_files)
       def download(target_locale, glotpress_url, is_source)
         uri = URI(glotpress_url)
         response = Net::HTTP.get_response(uri)
```
crazytonyli, line 18: Just a thought, not related to this PR: it might be good to use faraday which has features like following redirects and retry with an interval (which could be a replacement for handling 429 error?).

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
```diff
@@ -100,6 +95,33 @@ def delete_existing_metadata(target_locale)
...
+          download(locale, response.header['location'], is_source)
+        when '429'
+          # We got rate-limited, offer to try again
+          if UI.confirm("Retry downloading `#{locale}` after receiving 429 from the API?")
+            download(locale, response.uri, is_source)
+          else
+            UI.error("Abandoning `#{locale}` download as requested.")
+          end
+        else
+          UI.error("Received unexpected #{response.code} from request to URI #{response.uri}.")
```
crazytonyli: Should we stop the execution when encounter this error?

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
mokagio: Yeah... I don't know about [faraday](https://lostisland.github.io/faraday/usage/) or other Ruby networking clients, but given the number of network requests we do with this toolkit, having a feature-rich way to do requests would be handy.

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
mokagio: Good question 🤔 I'm unsure. Is an error on one locale worth blocking the entire process? From one point of view it might be, so that we can go on and fix it. From another, HTTP comes with occasional transport errors and it would be good to update _some_ metadata at least.

AliSoftware: Middle-ground: ask the user 😄 

```suggestion
          message = "Received unexpected #{response.code} from request to URI #{response.uri}."
          UI.abort_with_message!(message) unless UI.confirm("#{message} Continue anyway?")
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
AliSoftware: iinm Farady is also the network client that fastlane itself uses, so it should already be part of our ruby dependencies anyway.

This is one of those items part of a long list of little things I'd love to improve / refactor to clean up the release-toolkit in many areas and modernize it… but never got the bandwidth to 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
AliSoftware: Given the error message is pretty clear and it's in reaction of us explicitly replying `n` in the prompt on the previous line, I vote for keeping it red with `UI.error` 👍

AliSoftware (dismissed):
You're missing a `CHANGELOG.md` entry, so I'm blocking the merge to avoid accidentally landing this without it 😉 
But apart from that, the rest LGTM 👍  So should be good to ship once you add that.

crazytonyli (approved):

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
mokagio: Nice, so there's less chances it'll be missed, too.

mokagio: Just a note to say that I "tested" this before merging:

image

CHANGELOG.md
```diff
@@ -6,15 +6,15 @@
 
 ### Breaking Changes
 
-_None_
+- Propose to retry when `gp_downloadmetadata` receives a `429 - Too Many Requests` error. [#406]
```
AliSoftware, line 9: @mokagio I'm not sure this is really a breaking change — as in, that won't require any change to client repos adopting the new release-toolkit to update their call sites or config or whatnot, will it?

I think it's not worth a major version bump on the next release, while a New Feature and minor bump would make more sense for this change 🙃

mokagio: Yep, not at all a breaking change. That was an oversight which I realized only after merging this. It's fixed in 39808dab6bace7ed77e403c36494422303127647.

AliSoftware: Oh, perfect, thanks for following-up 👍

PR #407, opened 2022-09-06
Remove call to `pod clean` from `ios_build_preflight`

mokagio:
Following up on what was discussed [here](https://github.com/woocommerce/woocommerce-ios/pull/7649#issuecomment-1236756962).

This call could lead to long wait times, in particular in WordPress iOS, as every pod was downloaded from scratch.

See:

- https://github.com/woocommerce/woocommerce-ios/pull/7649
- https://github.com/wordpress-mobile/WordPress-iOS/pull/19033#issuecomment-1183230212

---

I'm a bit unsure whether we should keep this whole action in the first place. Quoting from my comment in the WordPress PR:

> I considered removing the call to `ios_build_preflight` altogether, but maybe there's value in having them in those rare circumstances when we run builds locally? I might open an issue to discuss this option further [...]

I wonder whether it's not appropriate for this library to make assumptions on the kind of pre-requisites its clients should have. A pre-flight step might be useful, but maybe it should be up to each client to define it.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_build_preflight.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_build_preflight.rb
@@ -33,7 +33,6 @@ module Fastlane
           raise
         end
 
-        Action.sh('rake dependencies:pod:clean')
         other_action.cocoapods()
       end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@
 
 ### Breaking Changes
 
-- Propose to retry when `gp_downloadmetadata` receives a `429 - Too Many Requests` error. [#406]
+_None_
 
 ### New Features
 
-_None_
+- Propose to retry when `gp_downloadmetadata` receives a `429 - Too Many Requests` error. [#406]
 
 ### Bug Fixes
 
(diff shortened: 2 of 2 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_build_preflight.rb
```diff
@@ -33,7 +33,6 @@ def self.run(params)
           raise
         end
 
-        Action.sh('rake dependencies:pod:clean')
```
mokagio, line 36: Also, I feel good about removing the implicit assumption that a client would have a `Rakefile` with this task defined in it.

Luckily, this was the only direct call to `rake` I could find. There is another one, above, but is in a comment.

https://github.com/wordpress-mobile/release-toolkit/blob/9f78ca93a1d4c494db90e52abbf090126355f9b1/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_build_preflight.rb#L32

AliSoftware (approved):
> I wonder whether it's not appropriate for this library to make assumptions on the kind of pre-requisites its clients should have. A pre-flight step might be useful, but maybe it should be up to each client to define it.

I think that given you kept the call to `other_action.cocoapods()` still in there, that should be a good enough guarantee that if we ran this from our local machine we'd still have Pods be checked to be up-to-date. The likeliness of `Pods/` being in a wrong state with stale versions even after we do a `pod install` (aka `other_action.cocoapods()`) is very low, so I don't think that's a big risk anyway.

PR #409, opened 2022-09-16
Add Android AVD / emulator actions

AliSoftware:
Adds a helper and multiple actions to help us install Android system-images, create AVDs dynamically, and boot and shutdown Android emulators.

These are used by our `screenshots` lane in https://github.com/wordpress-mobile/WordPress-Android/pull/17096 to automate capture of raw screenshots.

> EDIT: Reworked to improve many error cases (e.g. `sdkmanager` installed via `brew` but `emulator` installed by Android Studio, leading to multiple SDK roots; or giving hints for the `PANIC: Broken AVD` message; etc.) and to support dynamic emulator TCP ports (i.e. non-hardcoded `emulator-5554` serial) to support running multiple emulators in parallel 💪 

### Base Scenario

 - Switch to the `tooling/screenshots` branch in WPAndroid (from https://github.com/wordpress-mobile/WordPress-Android/pull/17096)
 - Run `bundle install` to make sure you point the `fastlane/Pluginfile` and `release-toolkit` in that WPAndroid branch to this toolkit's `screenshots/emulator-helper` branch
 - Run `bundle exec fastlane screenshots app:wordpress locale:zh-CN` from the WPAndroid repo
 - Ensure that the screenshot automation finishes and worked (it can take a couple of minutes†…), in particular:
   - Check that the logs mention calling the `android_create_avd` action, installing the system image and creating the AVD successfully, for each case (phone, tablet)
   - Check that the logs mention calling the `android_launch_emulator` action, booting the emulator then waiting for it to finish booting, for each case (phone, tablet)
   - Check that the lane ended up creating the raw screenshots in Chinese for both phone and tablet device sizes successfully.
   - Once the full lane completes and the screenshot UI tests have finished running, check that the raw screenshots images have been generated and pulled from the emulator's sdcard, and are present in the HTML summary file.

> **Note**: _† For some reason it appears that sometimes there's a long pause between when the unit test has started and the app has launched… and when the first click of the UI test scenario finally starts. I'm still not sure what could explain those delays (they are not consistent and are hard to debug), so for the time just don't be surprised that it could sometime be between 2 to 8 minutes between when the app's first screen shows in the emulator and when the UI test finally starts its clickity-clackity work._ 🙃 

### Variants / configurations

Please consider testing all these different setups, to ensure all my error handling of possible scenarios encountered on different user's machines or CI hosts will be covered

 - 💻  Test on both an Intel and an M1 machine
    - especially to validate that the installation of the appropriate Android System Image by `sdkmanager` installs the appropriate system image depending on the architecture of the machine you run those lanes from.
 - 🔧  Test with different ways of installing the Android Command Line Tools
    - Uninstall the command line tools like `sdkmanager` and `avdmanager`, then re-run the `screnshots` lane and confirm that the action errors, with a suggestion to install them
    - Install the Android command line tools [using `brew`](https://formulae.brew.sh/cask/android-commandlinetools) — while you keep `emulator` installed by Android Studio and reachable from terminal (either via your `$PATH` or by having `$ANDROID_SDK_ROOT` defined). Then re-run the `screenshots` lane and confirm that the action errors, suggesting that you might have two different Android SDK locations (brew and AS) and that `sdkmanager` and `emulator` might not be installed in the same one. Then `brew uninstall` the Android command line tools once done.
    - Finally, re-install the Command Line Tools via the Android Studio's "SDK Manager" window, and test the lane again. That time it should succeed.
 - 🏎️  Test to run multiple emulators in parallel
    - Start any AVD from Android Studio that is neither `Pixel_3_API_28` nor `Nexus_9_API_28`. The goal here is to have the TCP port `5554` — which is usually the default one / first picked by `adb` to connect to an emulator — already in use.
    - In one terminal window, run `bundle exec fastlane screenshots app:wordpress locale:ar device:phone`
    - While the first command is still running, run `bundle exec fastlane screenshots app:wordpress locale:ar device:tenInch` in another terminal window
    - Verify that each command found the correct serial (e.g. `emulator-5556` for the phone one and `emulator-5558` for the tablet one — since `emulator-5554` was likely the one already used by the dummy AVD you launched from AS in step one)

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_create_avd_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_create_avd_action.rb
@@ -1,0 +1,90 @@
+module Fastlane
+  module Actions
+    class AndroidCreateAvdAction < Action
+      def self.run(params)
+        device_model = params[:device_model]
+        api_level = params[:api_level]
+        avd_name = params[:avd_name]
+        sdcard = params[:sdcard]
+
+        helper = Fastlane::Helper::Android::EmulatorHelper.new
+
+        # Ensure we have the system image needed for creating the AVD with this API level
+        system_image = params[:system_image] || helper.install_system_image(api: api_level)
+
+        # Create the AVD for device, API and system image we need
+        helper.create_avd(
+          api: api_level,
+          device: device_model,
+          system_image: system_image,
+          name: avd_name,
+          sdcard: sdcard
+        )
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        'Creates a new Android Virtual Device (AVD) for a specific device model and API level'
+      end
+
+      def self.details
+        <<~DESC
+          Creates a new Android Virtual Device (AVD) for a specific device model and API level.
+          By default, it also installs the necessary system image (using `sdkmanager`) if needed before creating the AVD
+        DESC
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :device_model,
+                                       env_name: 'FL_ANDROID_CREATE_AVD_DEVICE_MODEL',
+                                       description: 'The device model code to use to create the AVD. Valid values can be found using `avdmanager list devices`',
+                                       type: String,
+                                       optional: false),
+          FastlaneCore::ConfigItem.new(key: :api_level,
+                                       env_name: 'FL_ANDROID_CREATE_AVD_API_LEVEL',
+                                       description: 'The API level to use to install the necessary system-image and create the AVD',
+                                       type: Integer,
+                                       optional: false),
+          FastlaneCore::ConfigItem.new(key: :avd_name,
+                                       env_name: 'FL_ANDROID_CREATE_AVD_AVD_NAME',
+                                       description: 'The name to give to the created AVD. If not provided, will be derived from device model and API level',
+                                       type: String,
+                                       optional: true,
+                                       default_value: nil),
+          FastlaneCore::ConfigItem.new(key: :sdcard,
+                                       env_name: 'FL_ANDROID_CREATE_AVD_SDCARD',
+                                       description: 'The size of the SD card to use for the AVD',
+                                       type: String,
+                                       optional: true,
+                                       default_value: '512M'),
+          FastlaneCore::ConfigItem.new(key: :system_image,
+                                       env_name: 'FL_ANDROID_CREATE_AVD_SYSTEM_IMAGE',
+                                       description: 'The system image to use (as used/listed by `sdkmanager`). Defaults to the appropriate system image given the API level requested and the current machine\'s architecture',
+                                       type: String,
+                                       optional: true,
+                                       default_value_dynamic: true,
+                                       default_value: nil),
+        ]
+      end
+
+      def self.output
+      end
+
+      def self.return_value
+        'Returns the name of the created AVD'
+      end
+
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.is_supported?(platform)
+        platform == :android
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_shutdown_emulator_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_shutdown_emulator_action.rb
@@ -1,0 +1,48 @@
+module Fastlane
+  module Actions
+    class AndroidShutdownEmulatorAction < Action
+      def self.run(params)
+        helper = Fastlane::Helper::Android::EmulatorHelper.new
+        helper.shut_down_emulators!(serials: params[:serial])
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        'Shuts down Android emulators'
+      end
+
+      def self.details
+        description
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :serial,
+                                       env_name: 'FL_ANDROID_SHUTDOWN_EMULATOR_SERIAL',
+                                       description: 'The serial(s) of the emulators to shut down. If not provided (nil), will shut them all down',
+                                       type: Array,
+                                       optional: true,
+                                       default_value: nil),
+        ]
+      end
+
+      def self.output
+      end
+
+      def self.return_value
+        # If you method provides a return value, you can describe here what it does
+      end
+
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.is_supported?(platform)
+        platform == :android
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_shutdown_emulator_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_shutdown_emulator_action.rb
@@ -3,7 +3,7 @@ module Fastlane
     class AndroidShutdownEmulatorAction < Action
       def self.run(params)
         helper = Fastlane::Helper::Android::EmulatorHelper.new
-        helper.shut_down_emulators!(serials: params[:serial])
+        helper.shut_down_emulators!(serials: params[:serials])
       end
 
       #####################################################
@@ -20,8 +20,8 @@ module Fastlane
 
       def self.available_options
         [
-          FastlaneCore::ConfigItem.new(key: :serial,
-                                       env_name: 'FL_ANDROID_SHUTDOWN_EMULATOR_SERIAL',
+          FastlaneCore::ConfigItem.new(key: :serials,
+                                       env_name: 'FL_ANDROID_SHUTDOWN_EMULATOR_SERIALS',
                                        description: 'The serial(s) of the emulators to shut down. If not provided (nil), will shut them all down',
                                        type: Array,
                                        optional: true,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
@@ -24,7 +24,7 @@ module Fastlane
           package = system_image_package(api: api)
 
           UI.message("Installing System Image for Android #{api} (#{package})")
-          Actions.sh(@tools.sdkmanager, "--sdk_root=#{@tools.android_sdk_root}", '--install', package)
+          Actions.sh(@tools.sdkmanager, '--install', package)
           UI.success("System Image #{package} successfully installed.")
           package
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
@@ -1,0 +1,162 @@
+module Fastlane
+  module Helper
+    module Android
+      # Helper methods to manipulate System Images, AVDs and Android Emulators
+      #
+      class EmulatorHelper
+        BOOT_WAIT = 2
+        BOOT_TIMEOUT = 60
+
+        SHUTDOWN_WAIT = 2
+        SHUTDOWN_TIMEOUT = 60
+
+        def initialize
+          @tools = Fastlane::Helper::Android::ToolsPathHelper.new
+        end
+
+        # Installs the system-image suitable for a given Android `api`, with `google_apis`, and for the current machine's architecture
+        #
+        # @param [Integer] api The Android API level to use
+        #
+        # @return [String] The `sdkmanager` package specifier that has been installed
+        #
+        def install_system_image(api:)
+          package = system_image_package(api: api)
+
+          UI.message("Installing System Image for Android #{api} (#{package})")
+          Actions.sh(@tools.sdkmanager, '--install', package)
+          UI.success("System Image #{package} successfully installed.")
+          package
+        end
+
+        # Create an emulator (AVD) for a given `api` number and `device` model
+        #
+        # @param [Integer] api The Android API level to use for this AVD
+        # @param [String] device The Device Model to use for this AVD. Valid values can be found using `avdmanager list devices`
+        # @param [String] name The name to give for the created AVD. Defaults to `<device>_API_<api>`.
+        # @param [String] sdcard The size of the SD card for this device. Defaults to `512M`.
+        #
+        # @return [String] The device name (i.e. either `name` if provided, or the derived `<device>_API_<api>` if provided `name` was `nil``)
+        #
+        def create_avd(api:, device:, system_image: nil, name: nil, sdcard: '512M')
+          package = system_image || system_image_package(api: api)
+          device_name = name || "#{device.gsub(' ', '_').capitalize}_API_#{api}"
+
+          UI.message("Creating AVD `#{device_name}` (#{device}, API #{api})")
+
+          Actions.sh(
+            @tools.avdmanager, 'create', 'avd',
+            '--force',
+            '--package', package,
+            '--device', device,
+            '--sdcard', sdcard,
+            '--name', device_name
+          )
+
+          UI.success("AVD `#{device_name}` successfully created.")
+
+          device_name
+        end
+
+        # Launch the emulator for the given AVD, then return the emulator serial
+        #
+        # @param [String] name name of the AVD to launch
+        # @param [Boolean] cold_boot if true, will do a cold boot, if false will try to use a previous snapshot of the device
+        # @param [Boolean] wipe_data if true, will wipe the emulator (i.e. reset the user data image)
+        #
+        # @return [String] emulator serial number corresponding to the launched AVD
+        #
+        def launch_avd(name:, cold_boot: true, wipe_data: true)
+          port = '5554'.freeze
+          serial = "emulator-#{port}"
+
+          shut_down_emulators!(serials: [serial]) # To ensure we can launch one on the port 5554 (as it's hardcoded for simplicity)
+
+          UI.message("Launching emulator for #{name}")
+
+          params = ['-avd', name, '-port', port]
+          params << '-no-snapshot' if cold_boot
+          params << '-wipe-data' if wipe_data
+
+          command = [@tools.emulator, *params].shelljoin + ' >/dev/null 2>/dev/null &'
+          UI.command(command)
+          system(command) # We can't use Actions.sh here because it doesn't handle `&` to run the process in the background :/
+
+          UI.message('Waiting for device to finish booting...')
+          Actions.sh(@tools.adb, '-s', serial, 'wait-for-device')
+          retry_loop(time_between_retries: BOOT_WAIT, timeout: BOOT_TIMEOUT, description: 'waiting for device to finish booting') do
+            Actions.sh(@tools.adb, '-s', serial, 'shell', 'getprop', 'sys.boot_completed').chomp == '1'
+          end
+
+          UI.success("Emulator #{name} successfully booted as `#{serial}`.")
+
+          serial
+        end
+
+        # @return [Array<Fastlane::Helper::AdbDevice>] List of currently booted emulators
+        #
+        def running_emulators
+          helper = Fastlane::Helper::AdbHelper.new(adb_path: @tools.adb)
+          helper.load_all_devices.select { |device| device.serial.include?('emulator') }
+        end
+
+        # Trigger a shutdown for all running emulators, and wait until there is no more emulator running.
+        #
+        # @param [Array<String>] serials List of emulator serials to shut down. Will shut down all of them if `nil`.
+        #
+        def shut_down_emulators!(serials: nil)
+          UI.message("Shutting down #{serials || 'all'} emulator(s)...")
+
+          emulators_list = running_emulators.map(&:serial)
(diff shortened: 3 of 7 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
```diff
@@ -0,0 +1,162 @@
...
+        # @param [Boolean] cold_boot if true, will do a cold boot, if false will try to use a previous snapshot of the device
+        # @param [Boolean] wipe_data if true, will wipe the emulator (i.e. reset the user data image)
+        #
+        # @return [String] emulator serial number corresponding to the launched AVD
+        #
+        def launch_avd(name:, cold_boot: true, wipe_data: true)
+          port = '5554'.freeze
+          serial = "emulator-#{port}"
+
+          shut_down_emulators!(serials: [serial]) # To ensure we can launch one on the port 5554 (as it's hardcoded for simplicity)
```
AliSoftware: Known limitation of the current implementation: the port is hardcoded to `5554` — which is why we need to ensure we shut down existing emulators if any first — because it ended up not being that trivial to make it dynamic.

If I let `emulator` choose the ports by itself (instead of explicitly providing them via `-port` flag on line 77), which would ideally be a more flexible alternative, I then couldn't find an easy and consistent/safe enough way to then retrieve the port that was used. I considered looking at the device list before and after to compare and detect the new one, but:
 - There's a small risk that the new emulator disappears from `adb devices -l` for a brief moment while it's booting and setting itself up during booting sequence
 -  That's not a strong guarantee because the user might also have other emulators that are in the process of booting or shutting down in parallel, which would lead to some race condition on checking the device list to compare
 - I've also took a quick look at if there were some command that would allow me to get a list of AVD names <-> adb ports used, which maybe we could use to try and match the AVD name with its port.
    - I've found some ideas around looping on each emulator, and call `adb -s emulator-#{port} emu avd name` command on each to get the name of the AVD…
    - …but that supposes the emulator has already finished booting (and we can't use `adb -s emulator-#{port} wait-for-device` to wait for the AVD we just launched to finish being ready… because we need to know the serial and thus the port to call that one… so catch-22)
    - …and we could consider looping over all the emulators found in `adb devices -l`, and call `adb -s emulator-#{port}` on each… but if we happen to do that loop before the AVD we just launched hasn't finished booting and connecting to `adb`, that won't find it and we'd have to do the loop all over again (using our `retry_loop` helper, probably), and calling `adb emu avd name` all over again… doesn't feel very efficient nor stable?

So in the end, I'm open to any ideas on how to improve flexibility here, or ways to find out the serial of an AVD we just booted without risking falling into the cracks and race condition / timing traps highlighted above. (@ParaskP7  any ideas or `adb` magic commands you know could help?)

---

The impact of this known limitation of hard-coding the port for now is that we won't be able to run screenshot tests on the same machine in parallel (as those would try to boot multiple emulators). If we find a working solution to make it possible for it to be dynamic (and thus allow screenshot parallelism), we can always add the feature in a subsequent PR.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
```diff
@@ -0,0 +1,162 @@
...
+
+            # NOTE: Alternative way of shutting down emulator would be to call the following command instead, which shuts down the emulator more gracefully:
+            # `adb -s #{e} shell reboot -p` # In case you're wondering, `-p` is for "power-off"
+            # But this alternate command:
+            #  - Requires that `-no-snapshot` was used on boot (to avoid being prompted to save current state on shutdown)
+            #  - Disconnects the emulator from `adb` (and thus disappear from `adb devices -l`) for a short amount of time,
+            #    before reconnecting to it but in an `offline` state, until `emulator` finally completely quits and it disappears
+            #    again (for good) from `adb devices --list`.
+            # This means that so if we used alternative, we couldn't really retry_loop until emulator disappears from `running_emulators` to detect
+            # that the shutdown was really complete, as we might as well accidentally detect the intermediate disconnect instead.
```
AliSoftware, line 159: Another area where I'd love this to be cleaner.

Especially since some old versions of the `emulator` binary end up triggering a CrashReporter alert on macOS when we send the `emu kill` command to the emulator to shut it down that way… and having a big alert on your OS every time and on every iteration of the loop that boots-then-shutdown an emulator for each locale and device that raw screenshots are taken in is… not ideal

Luckily, that CrashReporter alert I've seen in the past seems to have disappeared and not happening anymore with more recent versions of `emulator` (I'm currently using `emulator 31.3.10.0 (build_id 8807927)`), so it's probably safe to use `emu kill` after all instead of trying to be more clever with `adb shell reboot -p` 🤞

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
AliSoftware: EDIT: I've ended up fixing that one with latest commits on this PR after all 🎉

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
```diff
@@ -0,0 +1,198 @@
...
+        def find_serial(avd_name:)
+          running_emulators.find do |candidate|
+            command = [@tools.adb, '-s', candidate.serial, 'emu', 'avd', 'name']
+            UI.command(command.shelljoin)
+            candidate_name = Actions.sh(*command, log: false).split("\n").first.chomp
+            candidate_name == avd_name
+          end&.serial
+        end
+
+        # Trigger a shutdown for all running emulators, and wait until there is no more emulator running.
```
mokagio:
```suggestion
        # Trigger a shutdown for all running emulators, and wait until there is no more emulators running.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
```diff
@@ -0,0 +1,198 @@
...
+        BOOT_TIMEOUT = 60
+
+        SHUTDOWN_WAIT = 2
+        SHUTDOWN_TIMEOUT = 60
+
+        def initialize
+          @tools = Fastlane::Helper::Android::ToolsPathHelper.new
+        end
+
+        # Installs the system-image suitable for a given Android `api`, with `google_apis`, and for the current machine's architecture
```
mokagio: Nitpick: To make this consistent with the usage [below](https://github.com/wordpress-mobile/release-toolkit/pull/409/files#diff-d271fcb6de300236feaec1c6b889596caf09c69201f3f8d89a692cd0496dd8ddR170)


```suggestion
        # Installs the system-image suitable for a given Android `api`, with Google APIs, and for the current machine's architecture
```

Personally, I prefer "Google APIs" over `google_apis` because the latter makes me think of a method parameter.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
```diff
@@ -0,0 +1,198 @@
...
+
+          # Wait until all emulators are killed
+          retry_loop(time_between_retries: SHUTDOWN_WAIT, timeout: SHUTDOWN_TIMEOUT, description: 'waiting for devices to shutdown') do
+            (emulators_list & running_emulators.map(&:serial)).empty?
+          end
+
+          UI.success('All emulators are now shut down.')
+        end
+
+        # Find the system-images package for the provided API, with Google APIs, and matching the current platform/architecture this lane is called from.
```
mokagio: Nitpick, to make it consistent with the usage [above](https://github.com/wordpress-mobile/release-toolkit/pull/409/files#diff-d271fcb6de300236feaec1c6b889596caf09c69201f3f8d89a692cd0496dd8ddR17)

```suggestion
        # Find the system-images package for the provided `api`, with Google APIs, and matching the current platform/architecture this lane is called from.
```

Personally, I prefer using `api` over API because it makes it clearer that we are referring to the method paratmer.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
```diff
@@ -0,0 +1,198 @@
...
+        #
+        # @note Results from this method are memoized, to avoid repeating calls to `sdkmanager` when querying for the same api level multiple times.
+        #
+        def system_image_package(api:)
+          @system_image_packages ||= {}
+          @system_image_packages[api] ||= begin
+            platform = `uname -m`.chomp
+            all_packages = `#{@tools.sdkmanager} --sdk_root=#{@tools.android_sdk_root} --list`
+            package = all_packages.match(/^ *(system-images;android-#{api};google_apis;#{platform}(-[^ ]*)?)/)&.captures&.first
+            UI.user_error!("Could not find system-image for API `#{api}` and your platform `#{platform}` in `sdkmanager --list`. Maybe Google removed it for download and it's time to update to a newer API?") if package.nil?
```
mokagio, line 183: "Maybe ... it's time to update ..." is an appropriate suggestion, but sets the precedent for adding similar suggestions to other error messages.

Is there a particular reason why you added it? Would users get less value from the tool if it wasn't there?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_tools_path_helper.rb
```diff
@@ -0,0 +1,59 @@
...
+  module Helper
+    module Android
+      # Helper to find the paths of common Android build and SDK tools on the current machine
+      # Based on `$ANDROID_SDK_ROOT` and the common relative paths those tools are installed in.
+      #
+      class ToolsPathHelper
+        attr_reader :android_sdk_root
+
+        def initialize(sdk_root: nil)
+          @android_sdk_root = sdk_root || ENV['ANDROID_HOME'] || ENV['ANDROID_SDK_ROOT'] || ENV['ANDROID_SDK']
```
mokagio: Where does `ANDROID_SDK` come from? I couldn't find it in the [Environment variables](https://developer.android.com/studio/command-line/variables#envar) documentation.

Also, the docs say that `ANDROID_SDK_ROOT` is deprecated. Should we avoid looking for it?

> `ANDROID_HOME`: Sets the path to the SDK installation directory. Once set, the value does not typically change, and can be shared by multiple users on the same machine. `ANDROID_SDK_ROOT`, which also points to the SDK installation directory, is deprecated. If you continue to use it, Android Studio and the Android Gradle plugin will check that the old and new variables are consistent.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_shutdown_emulator_action.rb
```diff
@@ -0,0 +1,48 @@
...
+      def self.details
+        description
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :serial,
+                                       env_name: 'FL_ANDROID_SHUTDOWN_EMULATOR_SERIAL',
+                                       description: 'The serial(s) of the emulators to shut down. If not provided (nil), will shut them all down',
+                                       type: Array,
```
mokagio, line 26: Given the type is `Array`, how would you feel about renaming the parameter so `:serials`?

Or, is there a convention in Fastlane to use the singular form because it supports passing a single value and converts it to an array of one element under the hood?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_tools_path_helper.rb
mokagio: By the way, two points:

1. It goes without saying, but, as evidenced by the fact that I haven't requested changes, I don't think this is a big issue. Happy for a followup PR _if_ you consider the suggestion.
2. _If_ implementing the suggestion, would it be useful to rename the various `android_sdk_root` to `android_home` for consistency?

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_create_avd_action.rb
```diff
@@ -0,0 +1,90 @@
...
+                                       default_value: nil),
+          FastlaneCore::ConfigItem.new(key: :sdcard,
+                                       env_name: 'FL_ANDROID_CREATE_AVD_SDCARD',
+                                       description: 'The size of the SD card to use for the AVD',
+                                       type: String,
+                                       optional: true,
+                                       default_value: '512M'),
+          FastlaneCore::ConfigItem.new(key: :system_image,
+                                       env_name: 'FL_ANDROID_CREATE_AVD_SYSTEM_IMAGE',
+                                       description: 'The system image to use (as used/listed by `sdkmanager`). Defaults to the appropriate system image given the API level requested and the current machine\'s architecture',
```
mokagio, line 66: Random question out of curiosity: Why did you go for an escaped `'` for the Saxon possessive as opposed to using double quotes in the strings?

mokagio (approved):
Leaving a tentative approval to unblock your work.

I've looked at the code and it looks good. 

I meant to test this by running some actions and I even started one, but I'm running out of time for the day.

image

mokagio:
As I submitted my review, I noticed my command had failed. The failure looks to me like an HTTP out of the control of the release toolkit. The error message it printed is worth sharing, because of it's stacktrace formatting:

image

image

mokagio:
> The failure looks to me like an HTTP out of the control of the release toolkit

I retried the action and this time it succeeded, which is a point in favor of my conjecture about the nature of the previous failure:

image

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
AliSoftware: Right. I used this wording for `google_apis` here because that's how it's specified in the system-images qualifier expected to be passed to `sdkmanager --install <qualifier>`. But I can see the rationale for consistency with below and with differentiating `Google APIs` as a name from `api` as a parameter (and also the image qualifier could be seen as an implementation details… even if I thought the explicitness of the package component specifier here could help explain better which package this method installed)

```
sdkmanager --list_installed
…
system-images;android-28;google_apis;arm64-v8a | …
…

File: lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_create_avd_action.rb (hunk 9 above)
Comment (AliSoftware) [reply lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_create_avd_action.rb:66]: No reason really. I think it's a matter of finding balance as to when to switch to double-quotes when it's only to avoid escaping apostrophes in the String. If it's only one to be escaped, it's probably fine to keep single-quotes to avoid needing string interpolation to kick in for nought, when it's more than one, it's easier to switch to double-quotes to avoid escaping each one? But I don't have a strict rule for it tbh.

Another alternative would be to use alternate delimiters that still don't do string interpolation — since Ruby supports multiple string delimiters, interpolated or not. For example `%(…)` instead of `'…'` is a good choice when it comes to not needing interpolation but still wanting to avoid escaping.
```suggestion
                                       description: %(The system image to use (as used/listed by `sdkmanager`). Defaults to the appropriate system image given the API level requested and the current machine's architecture),
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_shutdown_emulator_action.rb
AliSoftware: Yeah I actually wondered about this. Ended up using singular because I figured in 95% of cases the call site will only provide either `nil` or a single `serial`, and providing multiple serials (but a subset of all of them) would be a rather rare case.

But I guess using `serials` make more sense for consistency with the param being an `Array` after all. Will update.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
AliSoftware: Mostly because last time I god a `PANIC: Broken AVD system image` errors when working on screenshots automation and trying to use the system-images that were defined in our scripts, it took me a very long time to understand what was broken (ref: paqN3M-cM-p2#broken-avd-system-image), and get why it suddenly didn't work on my machine while it worked for other team members (who had the system image installed on their machine from them downloading it a long time ago before it got removed from Google's servers), before I finally understand that that the only way to fix that was to update the configuration.

Given Google apparently removes old system images from their download servers from time to time, this makes this screenshot automation lane able to work on machines that installed the system image from a previous download but fail on machines that try to download it for the first time, and make for that lane to work one day, but break if you try to re-run it later without any change, i.e. it depends on Google download servers changes that makes old successful builds (e.g. from CI) not always reproducible for that reason. That's why I thought this suggestion was important to mention in this case.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_tools_path_helper.rb
AliSoftware: Good question, I don't remember where I found that from tbh, but I think I [borrowed that from `fastlane` itself](https://github.com/fastlane/fastlane/search?q=ANDROID_SDK)

> Also, the docs say that ANDROID_SDK_ROOT is deprecated.

Oh, funny, I always thought `ANDROID_HOME` was the one being deprecated and `ANDROID_SDK_ROOT` was the new hotness in town.

Given what you linked in the documentation but also the order those env vars are checked by `fastlane` (and thus by this line here which I borrowed from there), I guess I'll have to rename all my env vars in my machine, `~/.zshrc`, and personal scripts I have around on my Mac (and maybe in other scripts in this repo like [this one](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/tools/rename_apk_aab.sh#L33), and check which ones we use on our CI for our Android repos too)

> Should we avoid looking for it? 

We still definitively need to keep looking for it, because I think me and many Android devs still use `ANDROID_SDK_ROOT` in many places in their config on their machine.

AliSoftware:
> Leaving a tentative approval to unblock your work.

🙇 

> I meant to test this by running some actions and I even started one, but I'm running out of time for the day.
> 
> image

@mokagio  Can you tell me what command you ran to do that test? I'm surprised that in your capture above we see `fastlane` asking you for values for the command. Did you e.g. create a custom `lane :test` in your `Fastfile` for testing this action and then ran it, which is why it asked? (as opposed to following the testing steps I mention and using the `screenshots` lane in my WPAndroid PR, which are not supposed to ask for the `device_model` and `api_level` values and are supposed to provide them explicitly during the call of this action)

AliSoftware:
> The error message it printed is worth sharing, because of it's stacktrace formatting:

Yeah that's a quirk of how `sdkmanager` includes `\r` (without `'n`) in the `stdout` to do this ASCII progress bar, and fail to introduce a `\n` in `stdout` before printing its error message when it fails. And then `fastlane`'s `UI.shell_error!` prints the last "line" (as delimited by `\n`, rightfully) where the error appeared, hence including those extra `\r` that were not separated by a `\n` from the actual error message. I guess we could try and fix fastlane's `UI.shell_error!` to split around `\r` too though technically that would be a bit incorrect to do that for all cases… and we won't be able to fix the issue at its core (i.e. in `sdkmanager` itself), so… I guess the error in those cases will have to include those extra `[ASCII progress bar ===]\r` pseudo-lines in its output 🙃

mokagio:
@AliSoftware 

> @mokagio Can you tell me what command you ran to do that test?

```
bundle exec fastlane run android_create_avd
```

> I'm surprised that in your capture above we see fastlane asking you for values for the command.

Given I called the action with no parameters, that should be the expected behavior, right?

AliSoftware:
> @AliSoftware
> 
> > @mokagio Can you tell me what command you ran to do that test?
> 
> ```
> bundle exec fastlane run android_create_avd
> ```
> 
> > I'm surprised that in your capture above we see fastlane asking you for values for the command.
> 
> Given I called the action with no parameters, that should be the expected behavior, right?

Ah right, if that's how you called it, then yes, totally 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_tools_path_helper.rb
AliSoftware: Fixed in 7cba9a00eb16cfdba8193e5e664f313b79e2dd8e

Then applied a similar fix in 97d32333ad679942d9d8d594139cd8ed04da8cee on other actions that predates this PR (`android_send_app_size_metrics`) but had the same incorrect assumption (that `ANDROID_HOME` was the deprecated one, while `ANDROID_SDK_ROOT` is the deprecated one and `ANDROID_HOME` is the newest)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_shutdown_emulator_action.rb
AliSoftware: Fixed in 7c152d3705fa050b5f50196a60a7c8bc4000441e (then updated call site in https://github.com/wordpress-mobile/WordPress-Android/commit/b6346ab0eaf45891c8c66e47301b6529a93807b8

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_emulator_helper.rb
mokagio: Fair 👌

AliSoftware:
Merging this without waiting for someone else with M1 to test after all, to be able to move forward with this project.

PR #411, opened 2022-09-20
Require Fastlane `~> 2.210` to ensure compatibility with Xcode 14

mokagio:
What is says in the title.

It might be unnecessary at this point, since all iOS clients have been updated to use that Fastlane version, but it doesn't hurt to encode the requirement in the toolkit for future-proofing.


spencertransier (approved):
This looks good to me! 

Just to confirm, any project using the Release Toolkit would need to update to at least Fastlane 2.210.0 after upgrading to whichever Release Toolkit version this PR gets included in, correct?

mokagio:
> Just to confirm, any project using the Release Toolkit would need to update to at least Fastlane 2.210.0 after upgrading to whichever Release Toolkit version this PR gets included in, correct?

Yes. When running `bundle update fastlane-plugin-wpmreleasetoolkit`, the Fastlane gem installed in the repo will update accordingly.

PR #418, opened 2022-10-11
Improve resilience of `ios_lint_localizations` duplicates check against various file formats

### LINKED ISSUE #405: iOS Localization Linting can't check duplicate keys in XML-formatted strings files
It crashes with `Invalid character '<' found on line 1, col 1`.

We probably need to detect the XML format and process it differently.

AliSoftware:
The objective of this PR is mainly to fix #405

While at it, I also did a couple of other small related resilience improvements:

 - 44d643f: Moved the fixture files for `ios_lint_localizations_spec` into a common directory, for better organization of the fixtures
 - 578b36e: Added a test for the retry mechanism, which then highlighted a bug with that mode (yay TDD!) which I then fixed in 0b74402778a65b8ad57dfaf23ac2f2621b6c0fbc.
 - 25276c1: Fix file validation helper so that it can support non-UTF8 files (i.e. especially support for UTF-16)
 - b5986e9: Warn & skip non-ASCII plists from the "duplicate keys" checks — which is what really addresses #405


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -49,18 +48,21 @@ def self.report(violations:, base_lang:)
...
+
+          file_type = Fastlane::Helper::Ios::L10nHelper.strings_file_type(path: path)
+          if file_type == :text
+            duplicates = Fastlane::Helper::Ios::StringsFileValidationHelper.find_duplicated_keys(file: path)
+            duplicate_keys[language] = duplicates.map { |key, value| "`#{key}` was found at multiple lines: #{value.join(', ')}" } unless duplicates.empty?
+          else
+            UI.warning <<~WRONG_FORMAT
+              File `#{path}` is in #{file_type} format, while finding duplicate keys only make sense on files that are in ASCII-plist format.
+              Since your files are in #{file_type} format, you should probably disable the `check_duplicate_keys` option from this `#{self.action_name}` call.
+            WRONG_FORMAT
```
AliSoftware: As suggested by @leandroalonso in https://github.com/wordpress-mobile/release-toolkit/issues/405#issuecomment-1267270031

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+ - Improve resilience of the `ios_lint_localizations` action to support UTF16 files, and to warn and skip files in XML format when trying to detect duplicate keys on `.strings` files. [#418]
```
AliSoftware, line 17: This PR number reminded me of [some old April's fools joke](https://knowyourmeme.com/memes/http-error-418-im-a-teapot)… 🫖 😂

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -2,12 +2,11 @@ module Fastlane
...
       def self.run(params)
-        violations = Hash.new([])
+        violations = nil
 
         loop do
-          # If we did `violations = self.run...` we'd lose the default value for missing key being `[]` that we set above with `Hash.new`.
-          # We want that default value so that we can use `+=` when adding the duplicate keys violations below.
-          violations = violations.merge(self.run_linter(params))
+          violations = self.run_linter(params)
+          violations.default = [] # Set the default value for when querying a missing key
```
AliSoftware, line 9: ℹ️  This change was made because of the issue that was revealed by the unit test for the retry mechanism I added in 578b36e16e37f6fe07a9305edea1806b974bc2a5 (yay unit tests! 🎉 )
Only after writing the unit test did I make it go from red to green, via ☝️  those changes in 0b74402778a65b8ad57dfaf23ac2f2621b6c0fbc (yay TDD! 🎉 😄 )

The reason of the issue was that the previous implementation retained the old violations from each loop iteration between retries — which means that if violations had been found on the first try and the action offered the user to manually fix them and retry… then the subsequent retry would not start its now lint with empty set of violations but instead with still remembering the violations of the previous run… and thus would not make the linting pass even if the user properly manually-fixed the file in-between the retries!

Since the reason for the use of `violations = violations.merge(…)` in the old implementation was only to retain the default value returned by the Hash on missing keys, I instead addressed that in the new implementation just using `violations.default = []` which has the same outcome (but without requiring setting that default value on `.new`)

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_strings_file_validation_helper.rb
```diff
@@ -70,8 +70,8 @@ def self.find_duplicated_keys(file:)
 
           state = State.new(context: :root, buffer: StringIO.new, in_escaped_ctx: false, found_key: nil)
 
-          File.readlines(file).each_with_index do |line, line_no|
-            line.chars.each_with_index do |c, col_no|
+          File.readlines(file, mode: 'rb:BOM|UTF-8').each_with_index do |line, line_no|
+            line.encode('UTF-8').chars.each_with_index do |c, col_no|
```
AliSoftware: ℹ️  This `mode: 'rb:BOM|UTF-8'` basically means "[r]ead the file as [b]inary (because if it's UTF-16 or multi-byte we want to treat it as such), and for the text encoding (`:`) to use while reading, use the `BOM` if present (e.g. to detect if it's encoded in `UTF-16BE` or `UTF-16LE`), falling back (`|`) to `UTF-8` if there is no Byte-Order-Mark.

I've already used this technique and notation (documented clearly in the `IO.read`/`File.read`/… ruby documentation) in other places of the codebase, and this basically allows to read the file using the text encoding it's saved as, instead of assuming and hardcoding ASCII or UTF-8.

Then, as we read the lines (still encoded using the text encoding the file was saved as) and iterate over their characters, we need to convert the string to use the `UTF-8` encoding, because later in the code (line 88), we will do some `RegExp.match?`, and we can't use different encodings between the `RegExp` itself (which are UTF-8 by default when defined as constants in the codebase like here) and the string that it tries to match, hence us ensuring the line and characters being iterated are converted to `UTF-8`. If we didn't do that conversion, but were reading an UTF-16 `.strings` file, we'd get an exception like `Encoding::CompatibilityError — incompatible encoding regexp match (UTF-8 regexp with UTF-16 string)`

jkmassel: > I've already used this technique and notation (documented clearly in the IO.read/File.read/… ruby documentation) in other places of the codebase, and this basically allows to read the file using the text encoding it's saved as, instead of assuming and hardcoding ASCII or UTF-8.

I wonder if it'd be worth making a helper like `ios_strings_file_lines(path)` that wraps this approach? Not a blocker, just a thought 🙂

jkmassel (approved):

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_strings_file_validation_helper.rb
AliSoftware: Yeah, maybe 🤔 . Good thought

I only used this `rb:BOM|UTF-8` technique with `File.readlines` in two places — the other being in `ios_l10n_helper::merge_strings` method, where it's used similarly but slightly differently due to an `unless` condition. And the only other place I've used that magic `mode` was with `File.read` not `File.readlines`. So not sure how that would be mutializable even for just the two instances where I use it with `File.readlines`, but could keep that idea in the back of my head especially if we start needing that one with `File.readlines` in more occasions 🤔

AliSoftware: @jkmassel Addressed in e6a05751f6330af3ca3e13a9f85f1ce9580f18f1

PR #420, opened 2022-10-20
Make explicit the use of the parameter `GITHUB_TOKEN` when use `create_release`

raafaelima:
# What does this solve?

The issue is about CI failing at the moment of doing a GitHub release, The reason for the failure was a 401 from the GitHub API. This happened because neither GHHELPER_ACCESS nor GITHUB_TOKEN was available in the environment at runtime, As stated in [Issue#416](https://github.com/wordpress-mobile/release-toolkit/issues/416).  

# How was it solved?

The solution proposed was to change the implementation to make the token mandatory at the `github_client level`, and make the `create_release` receive an optional argument that will serve as the de facto GitHubToken.  In addition, create a fallback mechanism to the `create_release` function to use the environment variable `GITHUB_TOKEN` when the parameter passed at the function is not present.

A more verbose API in the actions by adding parameters for the tokens they expect will make it more difficult to forget their use of them. Making them invisible to our users can cause errors like the one on the issue.

# Testing Instructions 

Given this is a helper change affecting the way we use the GitHub client, the only E2E way to do it is by triggering the Github release using either a Fastlane lane or Gradle function with the new parameter set. Also, those scenarios are covered by unit testing

The expected behaviors are:
- The release succeeds if you pass the token explicitly or have set it on the environment variables
- The release fails if neither the token is passed at the create_release nor is present at the env variables.

### Example of Lane with the new githubtoken param set

```
create_release(
    repository: GHHELPER_REPO,
    version: version,
    release_notes_file_path: File.join(PROJECT_ROOT_FOLDER, 'WordPress', 'Resources', 'release_notes.txt'),
    release_assets: archive_zip_path.to_s,
    prerelease: options[:beta_release],
    githubtoken: GITHUB_TOKEN
)
```

PR #423, opened 2022-10-25
[WIP] Add Unit Tests for Actions that use GithubHelper

raafaelima:
## Dependency

This PR Depends on [PR#420](https://github.com/wordpress-mobile/release-toolkit/pull/420) 

# What does this solve?

As some of the actions that were changed on the [PR#420](https://github.com/wordpress-mobile/release-toolkit/pull/420) didn't have unit tests to check the config items and the use now mandatory `GITHUB_TOKEN`, were created new UnitTests to cover those new cases and assure that the original behavior of them are not changing.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
@@ -13,7 +13,7 @@ module Fastlane
         milestone = Fastlane::Helper::GithubHelper.get_milestone(repository, milestone_title)
         UI.user_error!("Milestone #{milestone_title} not found.") if milestone.nil?
 
-        Fastlane::Helper::GithubHelper.github_client().update_milestone(repository, milestone[:number], state: 'closed')
+        github_helper.update_milestone(repository: repository, number: milestone[:number], options: { state: 'closed' })
       end
 
       def self.description
--- a/spec/android_download_file_by_version_spec.rb
+++ b/spec/android_download_file_by_version_spec.rb
@@ -1,0 +1,79 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::AndroidDownloadFileByVersionAction do
+  describe 'run' do
+    let(:test_repo) { 'repo-test/project-test' }
+    let(:test_tag) { 'release/10.0' }
+    let(:test_version) { '10.0' }
+    let(:test_file) { 'test-file.xml' }
+    let(:test_folder) { 'test-folder' }
+    let(:test_token) { 'GITHUB_TOKEN' }
+    let(:github_helper) { class_double(Fastlane::Helper::GithubHelper) }
+
+    before do
+      allow(github_helper).to receive(:download_file_from_tag) # .and_return(nil)
+      allow(Fastlane::Helper::GithubHelper).to receive(:new).and_return(github_helper)
+      allow(Fastlane::Helper::Android::VersionHelper).to receive(:get_library_version_from_gradle_config).and_return(test_version)
+    end
+
+    it 'includes the github_token' do
+      expect(Fastlane::Helper::GithubHelper).to receive(:new).with(github_token: test_token)
+      described_class.run(mock_params)
+    end
+
+    it 'calls the download_file_from_tag' do
+      expect(github_helper).to receive(:download_file_from_tag).with(
+        repository: test_repo,
+        tag: test_tag,
+        file_path: test_file,
+        download_folder: test_folder
+      )
+
+      described_class.run(mock_params)
+    end
+
+    def mock_params
+      {
+        repository: test_repo,
+        github_release_prefix: 'release/',
+        file_path: test_file,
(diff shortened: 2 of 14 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
```diff
@@ -13,7 +13,7 @@ def self.run(params)
         milestone = Fastlane::Helper::GithubHelper.get_milestone(repository, milestone_title)
```
AliSoftware, line 13: I'm not sure what happened here, because this was properly fixed [in the other PR here](https://github.com/wordpress-mobile/release-toolkit/pull/420/files#diff-d5ec57b5e4681c99bd812b87182ed927eb4682ec339549c81ddf3fc82ad0aa22L13-R14) but still uses `get_milestone` as a class method in this PR.

I think it would make more send for this PR to target your `raafaelima:fix/github-token-required-at-github-helper` branch from the other PR — to make the two PRs stack on top of each other. That will require you to update/rebase this PR regularly as you add commits to the other PR — or backmerge that `raafaelima:fix/github-token-required-at-github-helper` branch, and its new commits you'd add, into this branch, to keep it up-to-date — so that this PR only contains what's new on top of the other one.

Alternatively, if you manage to make this PR completely independant of the other PR (i.e. if it only adds specs on actions for which the other PR didn't change their public API, meaning that the changes you're making in this PR don't depend on the one made on the other PR), then you can keep have this PR target trunk. But in that case I wouldn't expect the kind of diff on line 16 below to be part of this PR?

spec/android_download_file_by_version_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def mock_params
+      {
+        repository: test_repo,
+        github_release_prefix: 'release/',
+        file_path: test_file,
+        download_folder: test_folder,
+        github_token: test_token
+      }
+    end
```
AliSoftware, line 43: 💄  nit: this could actually have been a `let(:mock_params) do…end` instead

spec/android_download_file_by_version_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    it 'has the ConfigItem github_release_prefix' do
+      expect(described_class.available_options).to include(an_object_having_attributes(key: :github_release_prefix))
+    end
+
+    it 'has the ConfigItem github_token' do
+      expect(described_class.available_options).to include(an_object_having_attributes(key: :github_token))
+    end
+  end
```
AliSoftware, line 78: Mmmh I like the addition of more unit tests, but tbh I'm not convinced those are particularly useful ones.
They are kind of self-testing, and as soon as we might want to add a new `ConfigItem` (potentially optional and with default value, which would make it non-API-breaking) to the action in the future to add some flexibility, we'll have to update the unit tests to reflect… exactly the same. It's a bit like testing `expect(a).to eq(1)` right after `a=1` all over again 😅 . It's not that important, for the **behavior** of the action under test to be correct, that the number of `ConfigItems` of the action is checked to be exactly `7`, for example.

I think it's more useful when the unit test ensures the result of the action, i.e. the output at the end of the pipeline, based on the input parameters, to test the behavior of the action itself, rather than just testing trivialities at input. 

Like, it's more useful to test that when you pass a specific value to the `github_token` parameter when calling `described_action.run`, that token is the one being used in the REST API call / `Octokit::Client` that is made by the action (instead of being deduced by the helper via the value of an `ENV` var, like before) — which is what you did on [your test on lines 19-22](https://github.com/wordpress-mobile/release-toolkit/pull/423/files#diff-9ab3e8e35d6b48ba7b7484e5fc91ba5f4e78334c4c7e96074f22df068bcf5fc5R19-R22), rather than just to test that the action has a `ConfigItem` with a `github_token` key like you do here.

spec/close_milestone_action_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    it 'calls the update_milestone' do
+      expect(github_helper).to receive(:update_milestone).with(
+        repository: test_repository,
+        number: 1234,
+        options: { state: 'closed' }
+      )
+
+      described_class.run(mock_params)
+    end
```
AliSoftware, line 35: That's a good test, but I wonder if the thing we want to test shouldn't be that we ultimately call the right GitHub API (well, in our case, the right `Octokit::Client` method).

The purpose of a unit test is to test ultimate behavior independently of the implementation details. So we want to test that a given call of our helper methods ends up calling the final endpoint, regardless of the internal implementation of the action, and if we use an intermediate helper class to achieve our goal of if the action ends up calling `Octokit::Client` directly without using a helper, or whatnot. The end goal is that we call the right REST API / `Octokit::Client` method with the right parameters, so that the expected end result and action on GitHub happens in the end.

Given this, I'd suggest to rewrite your unit tests (this one and the others that have the same approach so far) to test the ultimate behavior independently of the implementation detail of the action using a helper under the hood or not.

PR #424, opened 2022-11-03
Add Hash-splat operator to GithubHelper methods

raafaelima:
# What does this solve?

As [discussed here](https://github.com/wordpress-mobile/release-toolkit/pull/420#discussion_r1012086322) at PR #420. This PR aims to add the hash conversion operator (aka **) to the methods on GithubHelper that use the named param `options:` in their signature. With this, we hope to improve the legibility and make those methods calls cleaner.

# Good, but what is this Hash Operator?

This operator will allow the call site of those methods to just provide the `options:` variable/parameter directly as if they were keywords (kinda "flattening the Hash into keyword parameters") without having to wrap those options into `{ … } ` curly braces to make them a Hash. So the call site would look like this: 

### Instead of
```ruby
github_helper.update_milestone(repository: repository, number: milestone[:number], options: { title: mile_title })
```

### We'd have:
``` ruby
# Inside `update_milestone`'s implementation, the variable/parameter `options` would be a Hash equal to `{ title: mile_title }`
github_helper.update_milestone(repository: repository, number: milestone[:number], title: mile_title)
```

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/close_milestone_action.rb
@@ -15,7 +15,7 @@ module Fastlane
 
         UI.user_error!("Milestone #{milestone_title} not found.") if milestone.nil?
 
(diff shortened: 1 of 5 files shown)


AliSoftware (approved):
Nice 👍 
:shipit:

Issue #416, opened 2022-10-06
Stop accessing `GITHUB_TOKEN` under the hood in `github_helper`


mokagio:
I just experienced a CI failure while my automation attempted to create a GitHub release for an internal app. The reason for the failure was a 401 from the GitHub API. This happened because neither `GHHELPER_ACCESS` nor `GITHUB_TOKEN` were available in the environment at runtime.

https://github.com/wordpress-mobile/release-toolkit/blob/24d6859170948737120f6ef12662536bee27ffe4/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb#L13-L14

While I agree that it's convenient to not have to pass tokens around when calling helper methods, by accessing them under the hood we make them invisible to our consumers, which can lead to issues such as the one I just experienced. 

I think we'd be better off incurring a more verbose API in the actions by adding parameters for all the tokens they expect, rather than risking CI failures like this one. 

Case in point, I had already experience a CI failure because of a missing token and grepped the `Fastfile` of that project for any `ENV` access that would reveal the need for others. Had `create_release` required an explicit token, I would have seen it and added it to our CI environment before retrying my deployment build.

spencertransier:
This issue was resolved by https://github.com/wordpress-mobile/release-toolkit/pull/420 last year.

Issue #278, opened 2021-06-17
Address the use of the legacy api of ConfigItem constructor (aka `is_string` vs `type`).


JavonDavis:
@AliSoftware highlighted in https://github.com/wordpress-mobile/release-toolkit/pull/258/files#r652892737 that this api is deprecated and there is a more modern version we should adopt. This should be a quick fix but tracking it here.

PR #425, opened 2022-11-03
Add Unit tests for milestone actions

raafaelima:
# What does this solve?
With the changes/improvements made on PR #420, we discover that some actions were not properly covered by unit tests. On this PR we start adding them, starting with the milestone actions. 

# What does PR #420 was about?

As stated in issue #416,  The CI was failing at the moment of doing a GitHub release, The reason for the failure was a 401 from the GitHub API. This happened because neither `GHHELPER_ACCESS` (that is not valid anymore) nor `GITHUB_TOKEN` was available in the environment at runtime. 

We solved this by turning the `GITHUB_TOKEN` into a mandatory env variable/parameter to run those actions and changing the `GithubHelper` class to receive this token at their initialization, so we have a centralized entry-point at the GithubAPI.

# Any Dependencies?
As this work is impacted by #426, Please, _DO NOT_ merge this before it. Thanks! 😄

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
@@ -11,6 +11,10 @@ module Fastlane
 
         github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
         last_stone = github_helper.get_last_milestone(repository)
+
+        UI.user_error!('No milestone found on the repository.') if last_stone.nil?
+        UI.user_error!("Milestone #{last_stone[:title]} has no due date.") if last_stone[:due_on].nil?
+
         UI.message("Last detected milestone: #{last_stone[:title]} due on #{last_stone[:due_on]}.")
         milestone_duedate = last_stone[:due_on]
         milestone_duration = params[:milestone_duration]
@@ -49,19 +53,19 @@ module Fastlane
                                        env_name: 'GHHELPER_NEED_APPSTORE_SUBMISSION',
                                        description: 'True if the app needs to be submitted',
                                        optional: true,
-                                       is_string: false,
+                                       type: Boolean,
                                        default_value: false),
           FastlaneCore::ConfigItem.new(key: :milestone_duration,
                                        env_name: 'GHHELPER_MILESTONE_DURATION',
                                        description: 'Milestone duration in number of days',
                                        optional: true,
-                                       is_string: false,
+                                       type: Integer,
                                        default_value: 14),
           FastlaneCore::ConfigItem.new(key: :number_of_days_from_code_freeze_to_release,
                                        env_name: 'GHHELPER_NUMBER_OF_DAYS_FROM_CODE_FREEZE_TO_RELEASE',
                                        description: 'Number of days from code freeze to release',
                                        optional: true,
-                                       is_string: false,
+                                       type: Integer,
                                        default_value: 14),
           Fastlane::Helper::GithubHelper.github_token_config_item,
         ]
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb
@@ -71,7 +71,7 @@ module Fastlane
                                        description: 'The GitHub milestone',
                                        optional: false,
                                        default_value: true,
-                                       is_string: false),
+                                       type: Boolean),
           Fastlane::Helper::GithubHelper.github_token_config_item,
         ]
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb
@@ -15,7 +15,6 @@ module Fastlane
         UI.user_error!("Milestone #{milestone_title} not found.") if milestone.nil?
 
         mile_title = milestone[:title]
-        puts freeze
         if freeze
           # Check if the state needs changes
           if is_frozen(milestone)
--- a/spec/close_milestone_action_spec.rb
+++ b/spec/close_milestone_action_spec.rb
@@ -1,0 +1,105 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::CloseMilestoneAction do
+  describe 'initialize' do
+    let(:test_token) { 'Test-GithubToken-1234' }
+    let(:mock_params) do
+      {
+        repository: 'test-repository',
+        milestone: '10'
+      }
+    end
+    let(:client) do
+      instance_double(
+        Octokit::Client,
+        list_milestones: [{ title: '10.1' }],
+        update_milestone: nil,
+        user: instance_double('User', name: 'test'),
+        'auto_paginate=': nil
+      )
+    end
+
+    before do
+      ENV['GITHUB_TOKEN'] = nil
+      mock_params[:github_token] = nil
+      allow(Octokit::Client).to receive(:new).and_return(client)
+    end
+
+    it 'properly passes the environment variable `GITHUB_TOKEN` all the way to Octokit::Client' do
+      ENV['GITHUB_TOKEN'] = test_token
+      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+      run_described_fastlane_action(mock_params)
+    end
+
+    it 'properly passes the parameter `:github_token` all the way to Octokit::Client' do
+      mock_params[:github_token] = test_token
+      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+      run_described_fastlane_action(mock_params)
+    end
+
+    it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` enviroment variable if both are present' do
+      ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
+      mock_params[:github_token] = test_token
+      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+      run_described_fastlane_action(mock_params)
+    end
+
+    it 'prints an error if no `GITHUB_TOKEN` environment variable nor parameter `:github_token` is present' do
+      expect { run_described_fastlane_action(mock_params) }.to raise_error(FastlaneCore::Interface::FastlaneError)
+    end
+  end
+
+  describe 'get_milestone' do
+    let(:test_repository) { 'test-repository' }
+    let(:test_milestone) { '10' }
+    let(:mock_params) do
+      {
+        repository: test_repository,
+        milestone: test_milestone,
+        github_token: 'Test-GithubToken-1234'
+      }
+    end
+    let(:client) do
+      instance_double(
+        Octokit::Client,
+        list_milestones: [{ title: '10.1' }],
+        update_milestone: nil,
+        user: instance_double('User', name: 'test'),
+        'auto_paginate=': nil
+      )
+    end
+
+    it 'properly passes the repository all the way down to the Octokit::Client to list the milestones' do
+      allow(Octokit::Client).to receive(:new).and_return(client)
+      expect(client).to receive(:list_milestones).with(test_repository)
+      run_described_fastlane_action(mock_params)
+    end
+  end
+
+  describe 'update_milestone' do
+    let(:test_repository) { 'test-repository' }
+    let(:test_milestone_number) { '1234' }
+    let(:mock_params) do
+      {
+        repository: test_repository,
+        milestone: '10',
+        github_token: 'Test-GithubToken-1234'
+      }
+    end
+    let(:client) do
+      instance_double(
+        Octokit::Client,
+        list_milestones: [{ title: '10.1', number: test_milestone_number }],
+        update_milestone: nil,
+        user: instance_double('User', name: 'test'),
+        'auto_paginate=': nil
+      )
+    end
+
+    it 'properly passes the parameters all the way down to Octokit::Client' do
+      allow(Octokit::Client).to receive(:new).and_return(client)
+      expect(client).to receive(:update_milestone).with(test_repository, test_milestone_number, { state: 'closed' })
+      run_described_fastlane_action(mock_params)
+    end
+  end
+end
--- a/spec/close_milestone_action_spec.rb
+++ b/spec/close_milestone_action_spec.rb
@@ -1,105 +1,86 @@
 require 'spec_helper'
 
 describe Fastlane::Actions::CloseMilestoneAction do
-  describe 'initialize' do
-    let(:test_token) { 'Test-GithubToken-1234' }
-    let(:mock_params) do
-      {
-        repository: 'test-repository',
-        milestone: '10'
-      }
-    end
-    let(:client) do
-      instance_double(
-        Octokit::Client,
-        list_milestones: [{ title: '10.1' }],
-        update_milestone: nil,
-        user: instance_double('User', name: 'test'),
-        'auto_paginate=': nil
-      )
-    end
+  let(:test_repository) { 'test-repository' }
+  let(:test_token) { 'Test-GithubToken-1234' }
+  let(:test_milestone) do
+    { title: '10.1', number: '1234' }
+  end
+  let(:client) do
+    instance_double(
+      Octokit::Client,
+      list_milestones: [test_milestone],
+      update_milestone: nil,
+      user: instance_double('User', name: 'test'),
+      'auto_paginate=': nil
+    )
+  end
 
-    before do
+  before do
+    allow(Octokit::Client).to receive(:new).and_return(client)
+  end
+
+  it 'properly passes the environment variable `GITHUB_TOKEN` to Octokit::Client' do
+    ENV['GITHUB_TOKEN'] = test_token
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_action_without_key(:github_token)
+  end
+
+  it 'properly passes the parameter `:github_token` to Octokit::Client' do
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_described_fastlane_action(default_params)
+  end
+
+  it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` environment variable if both are present' do
+    ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_described_fastlane_action(default_params)
+  end
+
+  it 'properly passes the repository and milestone to Octokit::Client to update the milestone as closed' do
+    expect(client).to receive(:update_milestone).with(test_repository, test_milestone[:number], state: 'closed')
+    run_described_fastlane_action(default_params)
+  end
+
+  it 'raises an error when the milestone is not found or does not exist' do
+    allow(client).to receive(:list_milestones).and_return([])
+    expect { run_described_fastlane_action(default_params) }.to raise_error(FastlaneCore::Interface::FastlaneError, 'Milestone 10.1 not found.')
+  end
+
+  describe 'Calling the Action validates input' do
+    it 'raises an error if no `GITHUB_TOKEN` environment variable nor parameter `:github_token` is present' do
       ENV['GITHUB_TOKEN'] = nil
-      mock_params[:github_token] = nil
-      allow(Octokit::Client).to receive(:new).and_return(client)
+      expect { run_action_without_key(:github_token) }.to raise_error(FastlaneCore::Interface::FastlaneError, "No value found for 'github_token'")
     end
 
-    it 'properly passes the environment variable `GITHUB_TOKEN` all the way to Octokit::Client' do
-      ENV['GITHUB_TOKEN'] = test_token
-      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
-      run_described_fastlane_action(mock_params)
+    it 'raises an error if no `GHHELPER_REPOSITORY` environment variable nor parameter `:repository` is present' do
+      expect { run_action_without_key(:repository) }.to raise_error(FastlaneCore::Interface::FastlaneError, "No value found for 'repository'")
     end
 
-    it 'properly passes the parameter `:github_token` all the way to Octokit::Client' do
-      mock_params[:github_token] = test_token
-      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
-      run_described_fastlane_action(mock_params)
+    it 'raises an error if no `GHHELPER_MILESTONE` environment variable nor parameter `:milestone` is present' do
+      expect { run_action_without_key(:milestone) }.to raise_error(FastlaneCore::Interface::FastlaneError, "No value found for 'milestone'")
     end
 
-    it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` enviroment variable if both are present' do
-      ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
-      mock_params[:github_token] = test_token
-      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
-      run_described_fastlane_action(mock_params)
-    end
-
-    it 'prints an error if no `GITHUB_TOKEN` environment variable nor parameter `:github_token` is present' do
-      expect { run_described_fastlane_action(mock_params) }.to raise_error(FastlaneCore::Interface::FastlaneError)
+    it 'raises an error if `milestone:` parameter is passed as Integer' do
+      expect { run_action_with(:milestone, 10) }.to raise_error "'milestone' value must be a String! Found Integer instead."
     end
   end
 
-  describe 'get_milestone' do
-    let(:test_repository) { 'test-repository' }
-    let(:test_milestone) { '10' }
-    let(:mock_params) do
-      {
-        repository: test_repository,
-        milestone: test_milestone,
-        github_token: 'Test-GithubToken-1234'
-      }
-    end
-    let(:client) do
-      instance_double(
-        Octokit::Client,
-        list_milestones: [{ title: '10.1' }],
-        update_milestone: nil,
-        user: instance_double('User', name: 'test'),
-        'auto_paginate=': nil
-      )
-    end
-
-    it 'properly passes the repository all the way down to the Octokit::Client to list the milestones' do
-      allow(Octokit::Client).to receive(:new).and_return(client)
-      expect(client).to receive(:list_milestones).with(test_repository)
-      run_described_fastlane_action(mock_params)
-    end
+  def run_action_without_key(key)
+    run_described_fastlane_action(default_params.except(key))
   end
 
-  describe 'update_milestone' do
-    let(:test_repository) { 'test-repository' }
-    let(:test_milestone_number) { '1234' }
-    let(:mock_params) do
-      {
-        repository: test_repository,
-        milestone: '10',
-        github_token: 'Test-GithubToken-1234'
-      }
-    end
-    let(:client) do
-      instance_double(
-        Octokit::Client,
-        list_milestones: [{ title: '10.1', number: test_milestone_number }],
-        update_milestone: nil,
-        user: instance_double('User', name: 'test'),
-        'auto_paginate=': nil
-      )
-    end
+  def run_action_with(key, value)
+    values = default_params
+    values[key] = value
+    run_described_fastlane_action(values)
+  end
 
-    it 'properly passes the parameters all the way down to Octokit::Client' do
-      allow(Octokit::Client).to receive(:new).and_return(client)
-      expect(client).to receive(:update_milestone).with(test_repository, test_milestone_number, { state: 'closed' })
-      run_described_fastlane_action(mock_params)
-    end
+  def default_params
+    {
+      repository: test_repository,
+      milestone: test_milestone[:title],
+      github_token: test_token
+    }
   end
 end
--- a/spec/close_milestone_action_spec.rb
+++ b/spec/close_milestone_action_spec.rb
@@ -1,10 +1,16 @@
 require 'spec_helper'
+require 'shared_examples_for_actions_with_github_token'
 
 describe Fastlane::Actions::CloseMilestoneAction do
   let(:test_repository) { 'test-repository' }
   let(:test_token) { 'Test-GithubToken-1234' }
   let(:test_milestone) do
     { title: '10.1', number: '1234' }
+  end
+  let(:default_params) do
+    { repository: test_repository,
+      milestone: test_milestone[:title],
+      github_token: 'Test-GithubToken-1234' }
   end
   let(:client) do
     instance_double(
@@ -20,24 +26,7 @@ describe Fastlane::Actions::CloseMilestoneAction do
     allow(Octokit::Client).to receive(:new).and_return(client)
   end
 
-  it 'properly passes the environment variable `GITHUB_TOKEN` to Octokit::Client' do
-    ENV['GITHUB_TOKEN'] = test_token
-    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
-    run_action_without_key(:github_token)
-  end
-
-  it 'properly passes the parameter `:github_token` to Octokit::Client' do
-    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
-    run_described_fastlane_action(default_params)
-  end
-
-  it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` environment variable if both are present' do
-    ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
-    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
-    run_described_fastlane_action(default_params)
-  end
-
-  it 'properly passes the repository and milestone to Octokit::Client to update the milestone as closed' do
+  it 'closes the expected milestone on the expected repository' do
     expect(client).to receive(:update_milestone).with(test_repository, test_milestone[:number], state: 'closed')
     run_described_fastlane_action(default_params)
   end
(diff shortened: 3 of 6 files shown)


spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+        expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+        run_described_fastlane_action(mock_params)
+      end
+
+      it 'prints an error if no `GITHUB_TOKEN` environment variable nor parameter `:github_token` is present' do
+        expect { run_described_fastlane_action(mock_params) }.to raise_error(FastlaneCore::Interface::FastlaneError)
+      end
+    end
+
+    context 'with default parameters' do
```
raafaelima: Self-Review: I had some doubts if I should use the context here to test the parameter's default values at the initialization, although it seems descriptive enough as it uses the same base setup, I don't know if is the pattern that you folks commonly use. So, let me know if this can be done other way 😄

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+        mock_params[:number_of_days_from_code_freeze_to_release] = nil
+        ENV['GHHELPER_NUMBER_OF_DAYS_FROM_CODE_FREEZE_TO_RELEASE'] = nil
+        expect(glithub_helper).to receive(:create_milestone).with(
+          anything,
+          anything,
+          anything,
+          anything,
+          default_code_freeze_days,
+          anything
+        )
```
raafaelima: Self-Review: The rationale here to use the `anything` keyword, is that we only care about the default value of the parameter being passed to the helper. As with the other parameters, we test reaching the `OktokitClient` at another test. Again, lemme know if you feel there is a cleaner way to do this.

spec/setfrozentag_action_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+      allow(Octokit::Client).to receive(:new).and_return(client)
+    end
+
+    it 'properly passes parameters all the way down to the Octokit::Client' do
+      expect(client).to receive(:list_milestones).with(test_repository)
+      run_described_fastlane_action(mock_params)
+    end
+  end
+
+  describe 'update_milestone' do
```
raafaelima: Self-Review: This is also a point where I have doubts about the pattern used across the app, I leave here the name of the method that Oktokit should call to pass the correct data. I was wondering at this moment if I should instead have done as:

```ruby
describe 'run' do #or executeAction (naming is hard here)
   it 'update_milestone properly passes parameters all the way down to the Octokit::Client'
     ...
   end
end

```

spec/close_milestone_action_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        list_milestones: [{ title: '10.1' }],
+        update_milestone: nil,
+        user: instance_double('User', name: 'test'),
+        'auto_paginate=': nil
+      )
+    end
+
+    before do
+      ENV['GITHUB_TOKEN'] = nil
+      mock_params[:github_token] = nil
```
AliSoftware: I _think_ that `let` declarations are re-evaluated for each "example" (aka `it` test case), so it shouldn't be necessary to reset `mock_params[:github_token]` to `nil` before each there.

Besides, I'm not a fan of modifying what is _conceptually_ a constant here. Instead of modifying `mock_params`, I'd suggest you to build a new Hash merging those `mock_params` with any additional values when you need those in your test cases below.

spec/close_milestone_action_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      ENV['GITHUB_TOKEN'] = test_token
+      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+      run_described_fastlane_action(mock_params)
+    end
+
+    it 'properly passes the parameter `:github_token` all the way to Octokit::Client' do
+      mock_params[:github_token] = test_token
+      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+      run_described_fastlane_action(mock_params)
+    end
```
AliSoftware: So, as per my suggestion above, what I'd do instead here would be something like:
```suggestion
    it 'properly passes the parameter `:github_token` all the way to Octokit::Client' do
      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
      run_described_fastlane_action(mock_params.merge({github_token: test_token}))
    end
```
Or, because Ruby's `Hash#merge` method uses a similar trick to what you used in #424 to allow last-parameter options to be provided by a list of additional keyword parameters instead of a formal Hash, you can also omit the `{}` in `merge` and write it like this:
```suggestion
    it 'properly passes the parameter `:github_token` all the way to Octokit::Client' do
      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
      run_described_fastlane_action(mock_params.merge(github_token: test_token))
    end
```

Or alternatively, use the "Hash-splat" operator in reverse here, making it expand the `mock_params` Hash as if its keys and values were provided as a list of named parameters directly:
```suggestion
    it 'properly passes the parameter `:github_token` all the way to Octokit::Client' do
      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
      run_described_fastlane_action(**mock_params, github_token: test_token)
    end
```

---

Whatever option you use, the point is to avoid modifying `mock_params` because even if Ruby technically allows you to do it, it should be considered as a constant (hence the `let` keyword convention) in principle (and modifying it in-place makes it harder to reason about imho).

PS: If you adopt this approach everywhere you need to adjust `mock_params`, you might consider renaming that constant to `let(:default_params)` (or `let(:common_params)`?) instead of `let(:mock_params)`, to make it clear that this is the "baseline of the params commonly used in all the test cases"?

spec/close_milestone_action_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    it 'prints an error if no `GITHUB_TOKEN` environment variable nor parameter `:github_token` is present' do
+      expect { run_described_fastlane_action(mock_params) }.to raise_error(FastlaneCore::Interface::FastlaneError)
+    end
+  end
+
+  describe 'get_milestone' do
```
AliSoftware: I'm a bit confused here because this announces that the test cases in that group will test `get_milestone`, but I don't see anything related to `get_milestone` in the `it` test cases below 🤔

spec/close_milestone_action_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        user: instance_double('User', name: 'test'),
+        'auto_paginate=': nil
+      )
+    end
+
+    it 'properly passes the repository all the way down to the Octokit::Client to list the milestones' do
+      allow(Octokit::Client).to receive(:new).and_return(client)
+      expect(client).to receive(:list_milestones).with(test_repository)
+      run_described_fastlane_action(mock_params)
+    end
```
AliSoftware: If this is really supposed to be testing the actual `CloseMilestoneAction` described action (since it calls `run_described_fastlane_action`), and is supposed to be about `get_milestone` (since that's what the `describe` group is announcing on line 52 above), then why is the expectation on `list_milestones`?

Tbh I feel like the organization of those tests (i.e. the split you used for the `describe` groups) is a bit confusing. I'm not sure it makes sense to have a whole `describe 'get_milestone'` and `describe 'update_milestone'` etc, and  each with only one `it` test case; besides the `describe` and the test they contain don't always seem to be consistent or to make the best of sense to me?

Since the goal here is to test that the action has the expected end result, regardless of the intermediate steps it needs to achieve this internally, I'd instead suggest to focus the unit test on exactly that, i.e. call `run_described_fastlane_action` with a given repo and milestone, and test that the right method of `Octokit::Client` gets called, with the correct parameters. Or that the action errors if we pass it incorrect input parameters.

So basically, I think the only useful test we really need for the happy path is [the one on line 99](https://github.com/wordpress-mobile/release-toolkit/pull/425/files#diff-1827e91d8a0925e31df233dd5a19e7287cd73a8a728a7e97c005f49011033c81R99) (which probably then doesn't really need its own `describe` just for a single `it`).

spec/create_new_milestone_action_spec.rb
AliSoftware: Use of `context` looks fine to me 👍  We could also go with `context 'when using default parameters' do`

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+        run_described_fastlane_action(mock_params)
+      end
+
+      it 'prints an error if no `GITHUB_TOKEN` environment variable nor parameter `:github_token` is present' do
+        expect { run_described_fastlane_action(mock_params) }.to raise_error(FastlaneCore::Interface::FastlaneError)
+      end
+    end
+
+    context 'with default parameters' do
+      let(:glithub_helper) do
```
AliSoftware:
```suggestion
      let(:github_helper) do
```

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+          default_milestone_duration,
+          anything,
+          anything
+        )
+        run_described_fastlane_action(mock_params)
+      end
+    end
+  end
+
+  describe 'get_last_milestone' do
```
AliSoftware: Again given this is the spec for `create_new_milestone_action_spec`, we should focus this test on testing the end result of the action, not the intermediate steps from internal details.

I'm not even sure when reading that test that I'd understand why we test `get_last_milestone` (isn't that the name of a separate and completely different action?) here.

---

If anything, what we want to test for the `CreateNewMilestoneAction` around `list_milestones` might be that the new milestone that it is about to create is based on the correct last milestone + the correct date computation from the due date of the last milestone and the day offsets. This is Unit Tests, so imho they should test the grand scheme of things from a public API point of view, not unit-test the details of the internal calls made in the implementation. So we should have a test like `it 'computes the date of the new milestone based on the last milestone and correct date offsets'` and that test should `allow(client).to receive(:list_milestones).and_return(mock_milestones_list)`, then `expect(client).to receive(:create_milestone).with(…the right params and especially the right due_on and description…)`, so that we unit-test that the `due_on` and other dates are correctly computed from the expected latest milestone found in `mock_milestones_list` + the correct day offsets.
That's what we really want to test when unit-testing the public API and the general behavior of the action from a black-box perspective.

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+      )
+    end
+
+    it 'properly passes the parameters all the way down to the Octokit::Client' do
+      comment = "Code freeze: November 14, 2022\nApp Store submission: November 28, 2022\nRelease: November 28, 2022\n"
+      options = { due_on: '2022-11-14T12:00:00Z', description: comment }
+      allow(Octokit::Client).to receive(:new).and_return(client)
+      expect(client).to receive(:create_milestone).with(test_repository, test_milestone_number, options)
+      run_described_fastlane_action(mock_params)
+    end
```
AliSoftware: This is the really useful unit test we need when testing the `CreateNewMilestoneAction` as a black box: pass some input parameters, and test that the `create_milestone` method of the `Octokit::Client` is called with the correct and expected parameters and final dates.

Except that there are so many more possible inputs to cover using that test case, especially:
 - Test that the computed dates (in `due_on` and `description`) are correct and the computation of the new dates were based on the right milestone (i.e. when there are multiple milestones returned by `list_milestones`, with different `due_on` dates, ensure the final dates for the new milestone is still correct)
 - Test the case where no milestones were returned
 - Test the case when we pass different offsets
 - Test error cases when we pass inconsistent or invalid offsets, if relevant
 - Test the case when the last milestone is just before a DST change, to ensure our date math takes that into account
 - …

Those are the use cases we need to cover, and they should basically all have the same structure as this `it` unit test above (i.e. their goal is always to `expect(client).to receive(:create_milestone).with(the-correct-params)` given the params passed to `run_described_fastlane_action`), but just with different inputs (different list of returned milestones, existing milestines with different due dates or with missing due dates or whatnot, case of need_appstore_submission true vs false, case of different day offsets, …) and thus different corresponding expected parameters passed to the client.

spec/setfrozentag_action_spec.rb
AliSoftware: Yeah as commented above, I'm having a hard time following the logic of how you split and organized the tests tbh.

Here since the goal is to unit test the `SetfrozentagAction`, why do we even have a `describe 'update_milestone'` (which relates to an API method here, while `describe` is usually more to group together multiple unit tests that test one common thing but stressing it in multiple ways)?

What we want to test in `setfrozentag_action_spec` is that, given various different inputs passed to the action (different parameters passed to `run_described_fastlane_action(…)`), the final outcome is what's expected — i.e. the final method call on the `Octokit::Client` that updates the milestone with the frozen emoji updates the right milestone, and with the right expected title. And, that if we pass invalid parameters or encounter an error or corner case (e.g. we're asking the action to `setfrozentag(milestone: '10.1', …)` but no milestone with title `10.1` exists), it handles it properly/nicely instead of doing something unexpected (like modifying another milestone or crashing badly).

AliSoftware (commented):
I have to admit that I have a hard time understanding the logic behind how you organized your tests (the way you decided to split your `describe` and `it`, why you only have one `it` unit test per `describe` while `describe` are usually to group multiple related tests together, …).

Besides, I don't think those tests are really testing what really matters, and they don't seem to cover the different cases (of validating the behavior of the actions under different values of input parameters and contexts) and edge cases (missing milestones, more than one milestone, tricky dates, …) that those tests would be useful to protect us against.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
```diff
@@ -49,19 +53,19 @@ def self.available_options
                                        env_name: 'GHHELPER_NEED_APPSTORE_SUBMISSION',
                                        description: 'True if the app needs to be submitted',
                                        optional: true,
-                                       is_string: false,
+                                       type: Boolean,
```
raafaelima, line 56: Self-Review: As the `is_string` is [deprecated on the ConfigItems](https://www.rubydoc.info/gems/fastlane_core/0.48.0/FastlaneCore%2FConfigItem:initialize), I replace them with the type of variable that we want to receive.  This also applies to the other places where I did that change.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
```diff
@@ -11,6 +11,10 @@ def self.run(params)
 
         github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
         last_stone = github_helper.get_last_milestone(repository)
+
+        UI.user_error!('No milestone found on the repository.') if last_stone.nil?
+        UI.user_error!("Milestone #{last_stone[:title]} has no due date.") if last_stone[:due_on].nil?
```
raafaelima, line 16: Self-Review: I had mixed feelings about putting this validation here, however as those two parameters are required to execute the action and without them, we receive a generic runtime error, I decided to put this here. Let me know if you think this is overkill.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb
```diff
@@ -71,7 +71,7 @@ def self.available_options
                                        description: 'The GitHub milestone',
                                        optional: false,
                                        default_value: true,
-                                       is_string: false),
+                                       type: Boolean),
```
raafaelima, line 74: Same as [here](https://github.com/wordpress-mobile/release-toolkit/pull/425/files#r1013529527)

raafaelima:
Indeed I was too focused on the tests of `GITHUB_TOKEN` and too biased about using the existent tests of `GithubHelper` to guide me, that I end up using the same patterns (like the describe with method names). I made some changes and improve the legibility and organization of those tests, I hope now I have reached the right point.

spec/create_new_milestone_action_spec.rb
raafaelima: Although I agree that those tests should be made as they are indeed valuable. Shouldn't all this validation about the date offsets, DST, and general due date/description cases that you mention be handled by the Unit tests from GithubHelper instead of the ones from the Action itself? 🤔

spec/setfrozentag_action_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  it 'does not freeze the milestone if is already frozen' do
+    allow(client).to receive(:list_milestones).and_return([{ title: '10.2 ❄️', number: '1234' }])
+    expect(client).not_to receive(:update_milestone)
+    run_action_with(:milestone, '10.2 ❄️')
+  end
+
+  it 'does not freeze the milestone if :freeze parameter is false' do
+    expect(client).to receive(:update_milestone).with(test_repository, test_milestone[:number], title: '10.1')
+    run_action_with(:freeze, false)
+  end
```
raafaelima: Self-Review: This for me is an odd behavior of this action If the `:freeze` parameter is false, it does not add the ❄️ symbol to the milestone title, as expected. However, it keeps calling the `update_milestone` to do an "unnecessary" update at that milestone.

Shouldn't in this case, the `update_milestone` not be called, following the [same logic](https://github.com/wordpress-mobile/release-toolkit/blob/15c5b990df99b33300765178aeced18441a91f88/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb#L21-L24) in place when the milestone is already frozen?

AliSoftware: The call is _NOT_ unnecessary, as it will remove any existing ❄️  emoji from a frozen milestone title if it contains one — i.e. update milestone `20.9 ❄️ ` to `20.9`. This is used when we finalize a release at the end of a sprint, just before closing the milestone (and freezing the next one on the next code freeze).

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
AliSoftware: Putting those here makes sense to me.

spec/close_milestone_action_spec.rb
```diff
@@ -1,105 +1,86 @@
...
+  it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` environment variable if both are present' do
+    ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_described_fastlane_action(default_params)
+  end
 
-    it 'prints an error if no `GITHUB_TOKEN` environment variable nor parameter `:github_token` is present' do
-      expect { run_described_fastlane_action(mock_params) }.to raise_error(FastlaneCore::Interface::FastlaneError)
-    end
+  it 'properly passes the repository and milestone to Octokit::Client to update the milestone as closed' do
```
AliSoftware:
```suggestion
  it 'closes the expected milestone on the expected repository' do
```

spec/close_milestone_action_spec.rb
```diff
@@ -1,105 +1,86 @@
...
-        Octokit::Client,
-        list_milestones: [{ title: '10.1' }],
-        update_milestone: nil,
-        user: instance_double('User', name: 'test'),
-        'auto_paginate=': nil
-      )
-    end
+  it 'raises an error when the milestone is not found or does not exist' do
+    allow(client).to receive(:list_milestones).and_return([])
+    expect { run_described_fastlane_action(default_params) }.to raise_error(FastlaneCore::Interface::FastlaneError, 'Milestone 10.1 not found.')
```
AliSoftware, line 36: 👍

spec/close_milestone_action_spec.rb
```diff
@@ -1,105 +1,86 @@
...
-    it 'properly passes the parameters all the way down to Octokit::Client' do
-      allow(Octokit::Client).to receive(:new).and_return(client)
-      expect(client).to receive(:update_milestone).with(test_repository, test_milestone_number, { state: 'closed' })
-      run_described_fastlane_action(mock_params)
+    it 'raises an error if `milestone:` parameter is passed as Integer' do
+      expect { run_action_with(:milestone, 10) }.to raise_error "'milestone' value must be a String! Found Integer instead."
     end
   end
+
+  def run_action_without_key(key)
```
AliSoftware: nit:
```suggestion
  def run_action_without(key:)
```

spec/close_milestone_action_spec.rb
```diff
@@ -1,105 +1,86 @@
...
+    run_described_fastlane_action(default_params.except(key))
+  end
+
+  def run_action_with(key, value)
+    values = default_params
+    values[key] = value
+    run_described_fastlane_action(values)
+  end
```
AliSoftware: Nit: why not make this more flexible by allowing it to take a `**options`-style Hash as parameter so you can not only have a nicer call site, but also one that looks like it uses named parameters?

```suggestion
  def run_action_with(**keys_and_values)
    params = default_params.merge(keys_and_values)
    run_described_fastlane_action(params)
  end
```
Should allow the call site to look like:
```ruby
run_action_with(milestone: '10')
```

spec/close_milestone_action_spec.rb
```diff
@@ -1,105 +1,86 @@
...
+
+  def default_params
+    {
+      repository: test_repository,
+      milestone: test_milestone[:title],
+      github_token: test_token
+    }
+  end
```
AliSoftware: Ah, here it is! When I saw `default_params` mentioned above in the code, I expected it to be a `let(:default_params)` and thus looked for it alongside all the other `let` declarations at the top, instead of here at the bottom 🙃 

Any reason why this is a method here?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
AliSoftware: This is great! It's even a small step towards addressing https://github.com/wordpress-mobile/release-toolkit/issues/278 that we opened a while ago about this 🙂

spec/close_milestone_action_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  it 'properly passes the parameter `:github_token` to Octokit::Client' do
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_described_fastlane_action(default_params)
+  end
+
+  it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` environment variable if both are present' do
+    ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_described_fastlane_action(default_params)
+  end
```
AliSoftware: It would be nice and imho makes sense to group those three tests into a parent `describe 'GitHub Token is properly passed to the client'` or something like that.

One could even consider [finding a way to DRY those 3 tests so that we can easily reuse them in the tests of all the actions that rely on that](https://relishapp.com/rspec/rspec-core/v/3-12/docs/example-groups/shared-examples) `github_token` shared `ConfigItem`, since those three tests will likely always be the same implementation and logic on all those affected actions.

spec/create_new_milestone_action_spec.rb
AliSoftware: > Shouldn't all this validation about the date offsets, DST, and general due date/description cases that you mention be handled by the Unit tests from GithubHelper instead of the ones from the Action itself? 🤔

That's a good point, indeed.

Thought the implementation of this action (and its split of the code between the action itself and the helper) is apparently currently at an odd place, because some of the date math is done in the action (namely `newmilestone_duedate = (milestone_duedate.to_datetime.next_day(milestone_duration).to_time).utc`), while other parts of the date math is done in `GithubHelper#create_milestone` itself (namely the computation of `submission_date` and `release_date` based on `days_until_submission` and `number_of_days_from_code_freeze_to_release`)…

We should clean that up at some point (the API of that `CreateNewMilestoneAction` is a bit odd anyway and we hope to modernize it at one point to make the `ConfigItem` it declares have better names and API… but that's another story for another time) and once we do, we'd probably have all the date math be done in the action and not in the helper (which would just receive the already-computed dates as input then), which would then make testing the date math in the `create_new_milestone_action_spec` the next sensible thing.

---

That being said, I see you added some tests about this in lines 30–57 since 👍

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+        expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+        run_described_fastlane_action(default_params)
+      end
+
+      it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` environment variable if both are present' do
+        ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
+        expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+        run_described_fastlane_action(default_params)
+      end
+    end
```
AliSoftware: I like how you grouped the three cases that all relate to checking the passing of the token to the client in a dedicated `context` group. Any reason why didn't you do the same for `close_milestone_action_spec` for consistency? (See also my other comment with suggestion to DRY that across action specs)

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+
+  it 'raises an error when the due date of milestone does not exists' do
+    allow(client).to receive(:list_milestones).and_return([{ title: '10.1', number: '1234' }])
+    expect { run_described_fastlane_action(default_params) }.to raise_error(FastlaneCore::Interface::FastlaneError, 'Milestone 10.1 has no due date.')
+  end
+
+  it 'raises an error when the milestone is not found or does not exist on the repository' do
+    allow(client).to receive(:list_milestones).and_return([])
+    expect { run_described_fastlane_action(default_params) }.to raise_error(FastlaneCore::Interface::FastlaneError, 'No milestone found on the repository.')
+  end
```
AliSoftware: Nice :+1:

nit: how about grouping those in `describe 'date computation is correct'` (for the first 3) and `describe 'raises error when it can't find last milestone'` (or maybe `context 'when last milestone cannot be used'`?) (for the last 2) for example?

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+          get_last_milestone: test_milestone,
+          create_milestone: nil
+        )
+      end
+
+      before do
+        allow(Fastlane::Helper::GithubHelper).to receive(:new).and_return(github_helper)
+      end
+
+      it 'uses default value when neither `GHHELPER_NUMBER_OF_DAYS_FROM_CODE_FREEZE_TO_RELEASE` environment variable nor parameter `:number_of_days_from_code_freeze_to_release` is present' do
```
AliSoftware, line 84: Gosh we really need to refactor the API of this action to use nicer and more comprehensible / consistent `ConfigItem` parameter names for that one… those are quite a mouthful currently 😅 

One day… in another PR… in the far future… maybe…

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+        default_code_freeze_days = 14
+        expect(github_helper).to receive(:create_milestone).with(
+          anything,
+          anything,
+          anything,
+          anything,
+          default_code_freeze_days,
+          anything
+        )
```
AliSoftware, line 93: I wonder if it wouldn't be the occasion to make this `GithubHelper#create_milestone` method start using named parameters instead of positional ones — that are even harder to identify in test cases like this amongst the other `anything` placeholders. That should then allow us to use `.with(hash_including(code_freeze_days: 14))` instead of having to use all those `anything` placeholders there.

But that means a small refactoring of all the call sites though (which might be outside of this PR's scope… but depending on how big a change it would be—not sure there are that many call sites, the `create_new_milestone` action might even be the only one?—, that could still be worth it so we can then have both the call sites and those unit tests be way nicer…
I'll let you evaluate how much changes it would warrant to do this, and decide if it's worth it or not.

spec/create_new_milestone_action_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+    run_described_fastlane_action(values)
+  end
+
+  def default_params
+    {
+      repository: test_repository,
+      need_appstore_submission: false,
+      github_token: test_token
+    }
+  end
```
AliSoftware: Same suggestions as for the other spec — `run_action_without(key:)`, `run_action_with(**additional_keys_and_values)`, and `let(:default_params)`

spec/setfrozentag_action_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  it 'properly passes the parameter `:github_token` all the way to Octokit::Client' do
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_described_fastlane_action(default_params)
+  end
+
+  it 'prioritizes `:github_token` parameter over `GITHUB_TOKEN` environment variable if both are present' do
+    ENV['GITHUB_TOKEN'] = 'Test-EnvGithubToken-1234'
+    expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+    run_described_fastlane_action(default_params)
+  end
```
AliSoftware: Should be able to DRY those, or at least group them in a `describe` given how all 3 are related

spec/close_milestone_action_spec.rb
raafaelima: That confused me a bit tbh 🤔. 

Because, If I put those in a `let(:default_params)` and then remove params from it, like I'm doing on the `run_action_without(key)`, as `let` means that this is a constant by convention and as you [stated here](https://github.com/wordpress-mobile/release-toolkit/pull/425#discussion_r1013252381), that [although is allowed is not right](https://github.com/wordpress-mobile/release-toolkit/pull/425#discussion_r1013253747) and should be avoided. 

I thought if I put the parameters again as a `let` and then do those operations, I will be breaking the conventions again. 😅

AliSoftware: `except` is not a mutation method tho (unlike `[]`)

spec/setfrozentag_action_spec.rb
raafaelima: Yeah, I didn't notice that behavior. 🤔 
I add a test on it in any case, about removing any existing ❄️ emoji from a frozen milestone. 😄

spec/close_milestone_action_spec.rb
raafaelima: Yeah, I see that you suggested that above, but forget to change it on the newest commit, sorry for that, Already change it 👍

spec/create_new_milestone_action_spec.rb
raafaelima: From what I see, the only places that will be affected are the `GithubSpec` and the `create_milestone_action/Spec`, I agree that this method should have the named params to "force" it to be better legible and cleaner. 

However, TBH I do not feel that I should do it on this PR, as it is a bit out of scope (although is a small change). So, I'll open another PR to deal with this, and update these tests here once this new PR is merged 😄

spec/create_new_milestone_action_spec.rb
raafaelima: Yeah! It will be more organized 😄

spec/create_new_milestone_action_spec.rb
raafaelima: I wonder if we can DRY those too using the [shared_context](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-context), or if this will be too overkill, as we also need to have the `default_params` defined on another place, to share the operations with those two methods 🤔

spec/create_new_milestone_action_spec.rb
raafaelima: Update: This is addressed on PR #426 😄

spec/setfrozentag_action_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  it 'does not change a milestone that is already frozen' do
+    allow(client).to receive(:list_milestones).and_return([{ title: '10.2 ❄️', number: '1234' }])
+    expect(client).not_to receive(:update_milestone)
+    run_action_with(milestone: '10.2 ❄️')
+  end
+
+  it 'does not change an unfrozen milestone if :freeze parameter is false' do
+    expect(client).to receive(:update_milestone).with(test_repository, test_milestone[:number], title: '10.1')
+    run_action_with(freeze: false)
+  end
```
AliSoftware, line 57: 👍

AliSoftware (dismissed):
Thanks for addressing the feedback!

This is looking good so far 👍 
I'd love to get @mokagio 's thoughts on it before merging if they have time.

spec/shared_examples_for_actions_with_github_token.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+RSpec.shared_examples 'github_token_initialization' do
+  let(:test_token) { 'Test-GithubToken-1234' }
+
+  describe 'GitHub Token is properly passed to the client' do
+    it 'properly passes the environment variable `GITHUB_TOKEN` to Octokit::Client' do
+      ENV['GITHUB_TOKEN'] = test_token
+      expect(Octokit::Client).to receive(:new).with(access_token: test_token)
+      run_action_without(:github_token)
```
mokagio, line 10: This shared example relies on the spec including it (`include_examples 'github_token_initialization'`) defining a `run_action_without` method and a `default_params`. The only way a consumer would find out about this would be by running the tests.

E.g. if we rename `run_action_without` to something else in `setfrozentag_action_spec`, we get:

```
Failures:

  1) Fastlane::Actions::SetfrozentagAction initialize GitHub Token is properly passed to the client properly passes the environment variable `GITHUB_TOKEN` to Octokit::Client
     Failure/Error: run_action_without(:github_token)

     NoMethodError:
       undefined method `run_action_without' for #<RSpec::ExampleGroups::FastlaneActionsSetfrozentagAction::Initialize::GitHubTokenIsProperlyPassedToTheClient:0x0000000109277748>
       Did you mean?  run_action_with
     Shared Example Group: "github_token_initialization" called from ./spec/setfrozentag_action_spec.rb:60
     ...
```

In Swift, we'd do something like constraining this to tests that conform to a certain `protocol`. There's no such thing in vanilla Ruby that I'm aware of.

There might be a way to update `run_described_fastlane_action` to add it the capability to do `_with` and `_without`, but that wouldn't solve the implicit dependency on `default_params` 🤔

At this point, the only thing I can think of is to document the requirement as a comment at the start of the file. What do you think?

AliSoftware: shared examples can also contain metadata, that you can then pass as additional params when you call `include_examples`. I think that would be a better way to pass `default_params`.

As for `_with` and `_without` variants, I wonder if it's worth relying on those existing in the implementation of `shared_examples` vs just directly using `default_params.except(…)` and `default_params.merge(…)` so that we don't rely on those `_with[out]` methods existing

raafaelima: > shared examples can also contain metadata, that you can then pass as additional params when you call include_examples. I think that would be a better way to pass default_params.

Indeed, when I try to use the shared examples passing the `default_params` as metadata, the RSpec blocked me from doing so. The error that I receive is: As we have the `default_params` as a `:let` variable it is not available on an example group (e.g. a `describe` or `context` block). It is only available within individual examples (e.g. `it` blocks). 

In that case, I don't know what is the best alternative here, if is to keep the shared examples relying on the existence of a variable named `default_params` and document this as @mokagio suggested. Or to [redundant use another `:let` variable](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples#providing-context-to-a-shared-group-using-a-block) to pass that metadata as in this example 🤔 

```ruby
  describe 'initialize' do
    include_examples 'github_token_initialization' do
      let(:params) { default_params }
    end
  end
```

> As for _with and _without variants, I wonder if it's worth relying on those existing in the implementation of shared_examples vs just directly using default_params.except(…) and default_params.merge(…) so that we don't rely on those _with[out] methods existing

Yeah, I have that "dilemma" when I was implementing this, I was first defaulting to those approaches of having the `merge` and `except` instead of relying on the `_with` and `_without` functions to exist. However, as I didn't find a good way to pass the params to inside the shared models, I end up following the same pattern that I was in the examples and relying on the existence of those methods as I do not want to duplicate them inside the shared examples.

raafaelima: Also, there is this third option to do it, which is cleaner than the other one, but I find it a little bit confusing the way that it used the `send` with the variable name (from a perspective of a person who does not has so much knowledge of ruby 😅) 

``` ruby
shared_examples 'github_token_initialization' do |params:|
  let(:params) { send(params) }
  ....
end

# spec/test.rb
describe 'initialize' do
 include_examples 'github_token_initialization', params: :default_params
end
```

But again, although I feel that this third option is cleaner, I have no strong option about which one is the right one that we should use here.

AliSoftware: Tbh I'm not a fan of using `send(…)`, both because it's lesser known, but also because it feels like a trick more than a proper official way to do this (the times I've used it was mostly to workaround some limitations rather than to implement something cleanly).

What about using metadata, but workaround the limitation that `let` is only visible in example and not groups… by declaring `default_params` as a function (`def`) rather than a `let`? I know that's what you had in an earlier iteration of the PR before I gave you some feedback that `let` is usually better suited for constant values, and that's still true, but if that is a blocker for us using it in `include_examples`' metadata, then I think it's ok to go back to `def` for that case then. At least I'd personally prefer this over using `send` 🙃

raafaelima: Yeah, I find the use of `send` kinda odd too 🙃. 

However,  even if I change `default_params` to a `def` function, I still receive the same error from the RSpec as using it as a `let` variable. 

```ruby
def default_params
{ 
  repository: 'test-repository',
  milestone: '10.1',
  github_token: 'Test-GithubToken-1234' 
}
end

#When I call
describe 'initialize' do
 include_examples 'github_token_initialization', default_params
end

# I receive the same error as the let
Failure/Error: include_examples 'github_token_initialization', default_params
  `default_params` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).
```

-------------------------

The only way that seems to do the job (although is not a DRY solution), without using the `send`, Is to repeat the declaration of the parameters at the call of the `shared_examples` like this:

``` ruby
describe 'initialize' do
  include_examples 'github_token_initialization',  { repository: 'test-repository', milestone: '10.1', github_token: 'Test-GithubToken-1234' }
end
```

Let me know what you think about this suggestion 😄

PR #426, opened 2022-11-06
Add named params and yard docs to `create_milestone` method at `GithubHelper`

raafaelima:
# What does this solve?

With the changes/improvements made on PR #420 and the creation of the unit tests being done on PR #425, we realized that the lack of yard documentation and named parameters were making the tests unclear and too verbose in the area of `CreateMilestoneAction`.  So, as [pointed on the PR #425 review](https://github.com/wordpress-mobile/release-toolkit/pull/425#discussion_r1014342778), we decided to open this new PR to address these issues that were found.

# How do the yard docs look now?

Glad you asked 😄 

Screenshot 2022-11-07 at 14 59 52

# Dependency
This work is part of the one started on #425. 

So, here at #426 we are addressing the changes at the API of `GithubHelper#create_milestone`, and there on #425 are the unit tests. As pointed on the #425 PR description, **_this should be merged first_**, as it affects the work done there.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
@@ -18,7 +18,14 @@ module Fastlane
         newmilestone_number = Fastlane::Helper::Ios::VersionHelper.calc_next_release_version(last_stone[:title])
         number_of_days_from_code_freeze_to_release = params[:number_of_days_from_code_freeze_to_release]
         UI.message("Next milestone: #{newmilestone_number} due on #{newmilestone_duedate}.")
-        github_helper.create_milestone(repository, newmilestone_number, newmilestone_duedate, milestone_duration, number_of_days_from_code_freeze_to_release, params[:need_appstore_submission])
+        github_helper.create_milestone(
+          repository: repository,
+          newmilestone_number: newmilestone_number,
+          newmilestone_duedate: newmilestone_duedate,
+          newmilestone_duration: milestone_duration,
+          number_of_days_from_code_freeze_to_release: number_of_days_from_code_freeze_to_release,
+          need_submission: params[:need_appstore_submission]
+        )
       end
 
       def self.description
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
@@ -11,20 +11,26 @@ module Fastlane
 
         github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
         last_stone = github_helper.get_last_milestone(repository)
+
         UI.message("Last detected milestone: #{last_stone[:title]} due on #{last_stone[:due_on]}.")
+
         milestone_duedate = last_stone[:due_on]
         milestone_duration = params[:milestone_duration]
         newmilestone_duedate = (milestone_duedate.to_datetime.next_day(milestone_duration).to_time).utc
         newmilestone_number = Fastlane::Helper::Ios::VersionHelper.calc_next_release_version(last_stone[:title])
         number_of_days_from_code_freeze_to_release = params[:number_of_days_from_code_freeze_to_release]
+        # If there is a review process, we want to submit the binary 3 days before its release
+        # Using 3 days is mostly for historical reasons where we release the apps on Monday and submit them on Friday.
+        days_until_submission = params[:need_appstore_submission] ? (number_of_days_from_code_freeze_to_release - 3) : milestone_duration
+
         UI.message("Next milestone: #{newmilestone_number} due on #{newmilestone_duedate}.")
+
         github_helper.create_milestone(
           repository: repository,
-          newmilestone_number: newmilestone_number,
-          newmilestone_duedate: newmilestone_duedate,
-          newmilestone_duration: milestone_duration,
-          number_of_days_from_code_freeze_to_release: number_of_days_from_code_freeze_to_release,
-          need_submission: params[:need_appstore_submission]
+          title: newmilestone_number,
+          due_date: newmilestone_duedate,
+          days_until_submission: days_until_submission,
+          days_until_release: number_of_days_from_code_freeze_to_release
         )
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
@@ -19,8 +19,8 @@ module Fastlane
         newmilestone_duedate = (milestone_duedate.to_datetime.next_day(milestone_duration).to_time).utc
         newmilestone_number = Fastlane::Helper::Ios::VersionHelper.calc_next_release_version(last_stone[:title])
         number_of_days_from_code_freeze_to_release = params[:number_of_days_from_code_freeze_to_release]
-        # If there is a review process, we want to submit the binary 3 days before its release
-        # Using 3 days is mostly for historical reasons where we release the apps on Monday and submit them on Friday.
+        # Because of the app stores review process, we submit the binary 3 days before the intended release date.
+        # Using 3 days is mostly for historical reasons, for a long time, we've been submitting apps on Friday and releasing them on Monday.
         days_until_submission = params[:need_appstore_submission] ? (number_of_days_from_code_freeze_to_release - 3) : milestone_duration
 
         UI.message("Next milestone: #{newmilestone_number} due on #{newmilestone_duedate}.")
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -74,23 +74,16 @@ module Fastlane
       # Creates a new milestone
       #
       # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
-      # @param [String] newmilestone_number The name of the milestone we want to create (e.g.: `16.9`)
-      # @param [Time] newmilestone_duedate milestone due date (e.g. `2022-10-22T12:00:00Z`)
-      # @param [Integer] newmilestone_duration Number of days that a milestone extents
-      # @param [Integer] number_of_days_from_code_freeze_to_release Number of days from code freeze to release
-      # @param [Boolean] need_submission The app needs to be submitted?
-      #  if `true`, will subtract 3 days from the `:number_of_days_from_code_freeze_to_release`.
-      #  if `false`, will use the days of `:newmilestone_duration`
+      # @param [String] title The name of the milestone we want to create (e.g.: `16.9`)
+      # @param [Time] due_date milestone due date (e.g. `2022-10-22T12:00:00Z`)
+      # @param [Integer] days_until_submission Number of days until submission
+      # @param [Integer] days_until_release Number of days from code freeze to release
       #
-      def create_milestone(repository:, newmilestone_number:, newmilestone_duedate:, newmilestone_duration:, number_of_days_from_code_freeze_to_release:, need_submission:)
-        # If there is a review process, we want to submit the binary 3 days before its release
-        #
-        # Using 3 days is mostly for historical reasons where we release the apps on Monday and submit them on Friday.
-        days_until_submission = need_submission ? (number_of_days_from_code_freeze_to_release - 3) : newmilestone_duration
-        submission_date = newmilestone_duedate.to_datetime.next_day(days_until_submission)
-        release_date = newmilestone_duedate.to_datetime.next_day(number_of_days_from_code_freeze_to_release)
+      def create_milestone(repository:, title:, due_date:, days_until_submission:, days_until_release:)
+        submission_date = due_date.to_datetime.next_day(days_until_submission)
+        release_date = due_date.to_datetime.next_day(days_until_release)
         comment = <<~MILESTONE_DESCRIPTION
-          Code freeze: #{newmilestone_duedate.to_datetime.strftime('%B %d, %Y')}
+          Code freeze: #{due_date.to_datetime.strftime('%B %d, %Y')}
           App Store submission: #{submission_date.strftime('%B %d, %Y')}
           Release: #{release_date.strftime('%B %d, %Y')}
         MILESTONE_DESCRIPTION
@@ -107,9 +100,9 @@ module Fastlane
         #
         # This is a bug in the GitHub API, not in our date computation logic.
         # To solve this, we trick it by forcing the time component of the ISO date we send to be `12:00:00Z`.
-        options[:due_on] = newmilestone_duedate.strftime('%Y-%m-%dT12:00:00Z')
+        options[:due_on] = due_date.strftime('%Y-%m-%dT12:00:00Z')
         options[:description] = comment
-        client.create_milestone(repository, newmilestone_number, options)
+        client.create_milestone(repository, title, options)
       end
 
       # Creates a Release on GitHub as a Draft
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -80,9 +80,9 @@ module Fastlane
       # @param [Integer] days_until_release Number of days from code freeze to release
       #
       def create_milestone(repository:, title:, due_date:, days_until_submission:, days_until_release:)
-        UI.user_error!('days_until_release must be greater than zero.') if days_until_release <= 0
-        UI.user_error!('days_until_submission must be greater than zero.') if days_until_submission <= 0
-        UI.user_error!('days_until_release must be greather than days_until_submission') if days_until_submission >= days_until_release
+        UI.user_error!('days_until_release must be greater than zero.') unless days_until_release > 0
+        UI.user_error!('days_until_submission must be greater than zero.') unless days_until_submission > 0
+        UI.user_error!('days_until_release must be greater or equal to days_until_submission.') unless days_until_release >= days_until_submission
 
         submission_date = due_date.to_datetime.next_day(days_until_submission)
         release_date = due_date.to_datetime.next_day(days_until_release)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -75,11 +75,15 @@ module Fastlane
       #
       # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
       # @param [String] title The name of the milestone we want to create (e.g.: `16.9`)
-      # @param [Time] due_date milestone due date (e.g. `2022-10-22T12:00:00Z`)
-      # @param [Integer] days_until_submission Number of days until submission
+      # @param [Time] due_date Milestone due date—which will also correspond to the code freeze date
+      # @param [Integer] days_until_submission Number of days from code freeze to submission to the App Store / Play Store
       # @param [Integer] days_until_release Number of days from code freeze to release
       #
       def create_milestone(repository:, title:, due_date:, days_until_submission:, days_until_release:)
+        UI.user_error!('days_until_release must be greater than zero.') if days_until_release <= 0
+        UI.user_error!('days_until_submission must be greater than zero.') if days_until_submission <= 0
+        UI.user_error!('days_until_release must be greather than days_until_submission') if days_until_submission >= days_until_release
+
         submission_date = due_date.to_datetime.next_day(days_until_submission)
         release_date = due_date.to_datetime.next_day(days_until_release)
         comment = <<~MILESTONE_DESCRIPTION
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -76,7 +76,7 @@ module Fastlane
       # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
       # @param [String] newmilestone_number The name of the milestone we want to create (e.g.: `16.9`)
       # @param [Time] newmilestone_duedate milestone due date (e.g. `2022-10-22T12:00:00Z`)
-      # @param [Integer] newmilestone_duration Milestone duration in number of days
+      # @param [Integer] newmilestone_duration Number of days that a milestone extents
       # @param [Integer] number_of_days_from_code_freeze_to_release Number of days from code freeze to release
       # @param [Boolean] need_submission The app needs to be submitted?
       #  if `true`, will subtract 3 days from the `:number_of_days_from_code_freeze_to_release`.
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -80,8 +80,8 @@ module Fastlane
       # @param [Integer] days_until_release Number of days from code freeze to release
       #
       def create_milestone(repository:, title:, due_date:, days_until_submission:, days_until_release:)
-        UI.user_error!('days_until_release must be greater than zero.') unless days_until_release > 0
-        UI.user_error!('days_until_submission must be greater than zero.') unless days_until_submission > 0
+        UI.user_error!('days_until_release must be greater than zero.') unless days_until_release.positive?
+        UI.user_error!('days_until_submission must be greater than zero.') unless days_until_submission.positive?
         UI.user_error!('days_until_release must be greater or equal to days_until_submission.') unless days_until_release >= days_until_submission
 
         submission_date = due_date.to_datetime.next_day(days_until_submission)
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -71,7 +71,18 @@ module Fastlane
         last_stone
       end
 
-      def create_milestone(repository, newmilestone_number, newmilestone_duedate, newmilestone_duration, number_of_days_from_code_freeze_to_release, need_submission)
+      # Creates a new milestone
+      #
+      # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
+      # @param [String] newmilestone_number The name of the milestone we want to create (e.g.: `16.9`)
+      # @param [Time] newmilestone_duedate milestone due date (e.g. `2022-10-22T12:00:00Z`)
+      # @param [Integer] newmilestone_duration Milestone duration in number of days
+      # @param [Integer] number_of_days_from_code_freeze_to_release Number of days from code freeze to release
+      # @param [Boolean] need_submission The app needs to be submitted?
+      #  if `true`, will subtract 3 days from the `:number_of_days_from_code_freeze_to_release`.
+      #  if `false`, will use the days of `:newmilestone_duration`
+      #
+      def create_milestone(repository:, newmilestone_number:, newmilestone_duedate:, newmilestone_duration:, number_of_days_from_code_freeze_to_release:, need_submission:)
         # If there is a review process, we want to submit the binary 3 days before its release
         #
         # Using 3 days is mostly for historical reasons where we release the apps on Monday and submit them on Friday.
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -71,7 +71,18 @@ def get_last_milestone(repository)
...
+      # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
+      # @param [String] newmilestone_number The name of the milestone we want to create (e.g.: `16.9`)
+      # @param [Time] newmilestone_duedate milestone due date (e.g. `2022-10-22T12:00:00Z`)
+      # @param [Integer] newmilestone_duration Number of days that a milestone extents
+      # @param [Integer] number_of_days_from_code_freeze_to_release Number of days from code freeze to release
+      # @param [Boolean] need_submission The app needs to be submitted?
+      #  if `true`, will subtract 3 days from the `:number_of_days_from_code_freeze_to_release`.
+      #  if `false`, will use the days of `:newmilestone_duration`
+      #
+      def create_milestone(repository:, newmilestone_number:, newmilestone_duedate:, newmilestone_duration:, number_of_days_from_code_freeze_to_release:, need_submission:)
```
AliSoftware: The parameter names seems a bit of a mouthful to me, given we already know from the context (of them being parameters of `create_milestone`) that those are about new milestones.

For example, what about naming the parameters just `title`, `due_date`, `duration`, and `days_until_release`?

Though a probably even nicer solution, especially because `duration` and `need_submission` are only used once and are kind of exclusive (i.e. either we pass `need_submission: false` and then `duration` doesn't matter, or we pass `need_submission: true` and that's only when `duration` is used…), we should probably combine the goal of the two parameters into a single one there, which would be `days_until_submission`, and let the caller pass to this value.

Given that, I suggest the following new API for this helper method:

```ruby
      def create_milestone(repository:, title:, due_date:, days_until_submission:, days_until_release:)
```

Then:
 - For the time being, the call site in `create_new_milestone_action` would just do the computation on its end, i.e. pass `params[:need_appstore_submission] ? number_of_days_from_code_freeze_to_release - 3 : milestone_duration` for the new `days_until_submission` parameter of this `GithubHelper#create_milestone`, carrying over the "Using 3 days mostly for historical reasons" logic from lines 88-89 for now
 - In the future, when we will [consider updating the API of the `create_new_milestone` action itself](https://github.com/wordpress-mobile/release-toolkit/pull/425#discussion_r1014337960) one day, we would then be able to update its `ConfigItem` (in a breaking-change way) to have the callers also directly provide the `days_until_submission` and `days_until_release` values in their `Fastfile`'s call sites instead of `duration` and/or `need_submission`

But for now, that change would make for a nicer and easier to read and understand API for the internal `GithubHelper#create_milestone`, all while preparing for the future of the planned refactor of the corresponding action and public API.

AliSoftware (commented):
I was a bit surprised that this PR didn't include the necessary changes to `create_new_milestone_action_spec.rb` on [those lines](https://github.com/wordpress-mobile/release-toolkit/pull/425/files#diff-6a06ad6e895c39263bf5df09b4f9077bda8328119f69623974cfcd3864de701dR86-R93).

Then I realized it was because that PR was targeting `trunk` (as opposed to being built on top of your `raafaelima:add/unit-tests-for-milestone-actions` branch from #425… which I guess you didn't do because you wouldn't have been able to open a PR in our repo otherwise, but instead only be able to open that PR on your fork?).

Would have been nice if you had mentioned that limitation / aspect in the description of the PR — or alternatively kept the PR as Draft (and commented on why) while you were waiting for #425 to land before un-drafting it — so that we wouldn't be surprised. Especially given that regardless if we merge #425 or this #426 first, you will have to adjust the tests in other one before we land it anyway (and as CI doesn't run on fork'd PRs by default for security reasons, it would be easy to accidentally merge both of your PRs into our `trunk` and miss that there would have been broken tests because of the PRs conflicting / depending on one another)

raafaelima:
> I was a bit surprised that this PR didn't include the necessary changes to `create_new_milestone_action_spec.rb` on [those lines](https://github.com/wordpress-mobile/release-toolkit/pull/425/files#diff-6a06ad6e895c39263bf5df09b4f9077bda8328119f69623974cfcd3864de701dR86-R93).
> 
> Then I realized it was because that PR was targeting `trunk` (as opposed to being built on top of your `raafaelima:add/unit-tests-for-milestone-actions` branch from #425… which I guess you didn't do because you wouldn't have been able to open a PR in our repo otherwise, but instead only be able to open that PR on your fork?).

Yep, Is the same case that we had about merging the #420 first, and then opening the unit tests because I can't stack PRs out of the trunk, because I'm on a fork, and my branch did not exists on the main repo, to stack other branchs on top of it.

> Would have been nice if you had mentioned that limitation / aspect in the description of the PR — or alternatively kept the PR as Draft (and commented on why) while you were waiting for #425 to land before un-drafting it — so that we wouldn't be surprised. Especially given that regardless if we merge #425 or this #426 first, you will have to adjust the tests in other one before we land it anyway (and as CI doesn't run on fork'd PRs by default for security reasons, it would be easy to accidentally merge both of your PRs into our `trunk` and miss that there would have been broken tests because of the PRs conflicting / depending on one another)

About that, I have made the [dependency commend at #425 description](https://github.com/wordpress-mobile/release-toolkit/pull/425#issue-1434817144) but forget to add it here to double-cross the dependency information to review. My Bad! Already fixing it. Thanks! 😄

raafaelima:
Just throwing some thoughts 🤔 💭 

That limitation of mine about not being able to open stacked PRs is leading to some confusion and, I'm not sure what is the best path here tbh 😵‍💫 

1. If it is to wait for PR#425 to be merged and then update PR#426, to fix the unit tests that will break with the change of parameters. So, therefore this should be a draft PR until the merge of PR#425 happens.

2. if we should do the opposite (as was my initial thought), and review and merge this PR#426 first, then go to PR#425, and fix the unit tests at `create_new_release_action` that will fail because of PR#426 changes.

As I can't go with the ideal scenario (the stacked PRs) I really do not have a strong option on this, 
I'd like to hear your thoughts on it @AliSoftware 😄

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
raafaelima: For sure, those parameters that are in place are big! 
I'll address those changes, thanks for the input 😄

AliSoftware:
> About that, I have made the https://github.com/wordpress-mobile/release-toolkit/pull/425#issue-1434817144 but forget to add it here to double-cross the dependency information to review. My Bad! Already fixing it. Thanks! 😄

I totally missed that comment, very likely because you added it after the fact and I didn't re-read the whole description on my subsequent reviews of #425 after I read it on the first pass 😅 This is why I didn't realize your plan was to land 426 first, and amend 425 afterwards (which is fine), rather than the other way around as I initially thought.

In any case, it's always a good habit to put the information in both, because anyone who might review your PR #426 might not be aware of / not have read or seen PR #425.

AliSoftware:
> That limitation of mine about not being able to open stacked PRs is leading to some confusion and, I'm not sure what is the best path here tbh 😵‍💫

Yeah, it's not an ideal situation due to you working on a fork, I acknowledge that!

Given your plan was, as I now see you mentioned by updating your description of #425 and given the limitation of the fork, to merge this #426 one _first_, and _then_ rebase and update #425 accordingly, I think it's ok to go with that then. I added the "Do Not Merge" label (that we use for such occasion) to #425 to indicate that it's pending this #426 to land first then.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -71,15 +71,19 @@ def get_last_milestone(repository)
...
-        # Using 3 days is mostly for historical reasons where we release the apps on Monday and submit them on Friday.
-        days_until_submission = need_submission ? (number_of_days_from_code_freeze_to_release - 3) : newmilestone_duration
-        submission_date = newmilestone_duedate.to_datetime.next_day(days_until_submission)
-        release_date = newmilestone_duedate.to_datetime.next_day(number_of_days_from_code_freeze_to_release)
+      # Creates a new milestone
+      #
+      # @param [String] repository The repository name, including the organization (e.g. `wordpress-mobile/wordpress-ios`)
+      # @param [String] title The name of the milestone we want to create (e.g.: `16.9`)
+      # @param [Time] due_date milestone due date (e.g. `2022-10-22T12:00:00Z`)
+      # @param [Integer] days_until_submission Number of days until submission
```
```suggestion
      # @param [Time] due_date milestone due date—which will also correspond to the code freeze date
      # @param [Integer] days_until_submission Number of days from code freeze to submission to the App Store / Play Store
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_new_milestone_action.rb
```diff
@@ -11,14 +11,27 @@ def self.run(params)
...
+
         UI.message("Last detected milestone: #{last_stone[:title]} due on #{last_stone[:due_on]}.")
+
         milestone_duedate = last_stone[:due_on]
         milestone_duration = params[:milestone_duration]
         newmilestone_duedate = (milestone_duedate.to_datetime.next_day(milestone_duration).to_time).utc
         newmilestone_number = Fastlane::Helper::Ios::VersionHelper.calc_next_release_version(last_stone[:title])
         number_of_days_from_code_freeze_to_release = params[:number_of_days_from_code_freeze_to_release]
+        # If there is a review process, we want to submit the binary 3 days before its release
+        # Using 3 days is mostly for historical reasons where we release the apps on Monday and submit them on Friday.
```
mokagio: What do you think of this rewording?

```suggestion
        # Because of the app stores review process, we submit the binary 3 days before the intended release date.
        # Using 3 days is mostly for historical reasons, for a long time, we've been submitting apps on Friday and releasing them on Monday.
```

spec/github_helper_spec.rb
```diff
@@ -234,14 +234,14 @@ def get_milestone(milestone_name:)
     end
 
     def create_milestone(need_submission:, milestone_duration:, days_code_freeze:)
+      days_until_submission = need_submission ? (days_code_freeze - 3) : milestone_duration
```
AliSoftware: Since this file is focused on testing the helper, and the helper doesn't have knowledge of the concept of `need_submission` anymore, I think we should adapt the `github_helper_spec` accordingly, to write the `it` tests based on the new parameters (i.e. don't test "has correct dates (with|without) submission" cases, but test things like "computes the correct dates for (a standard case | dates crossing DST boundaries | inconsistent offsets | …)".

Also begs the question what the behavior should do if we provide inconsistent inputs (invalid date, negative values for `days_until_*`, `days_until_release < days_until_submission`, …?).
Maybe it's ok to let that to the caller's responsibility and not check (i.e. keep the current behavior), at least for now; but adding tests to the action like here would also be the best time to raise this kind of questions (after all, if we try to follow in the footsteps of TDD, we should write the tests to describe the behaviors and use cases we want first, and then we implement them to match, not the other way around).

raafaelima: > Also begs the question what the behavior should do if we provide inconsistent inputs (invalid date, negative values for days_until_*, days_until_release < days_until_submission, …?). [.....]

That's a question that I keep asking myself 🤔 💭 

Using as an example a case where I have more days from submission than from release (i.e `days_until_submission: 14`, `days_until_release: 3`), The submission date will be waaay after the release date. In this case, should we raise a `FastlaneError`?

```ruby
# Milestone with due_date = '2022-10-20', days_until_submission = 14 and days_until_release = 3
Code freeze: October 20, 2022
App Store submission: November 03, 2022
Release: October 23, 2022
```

--------------------------

Also, in the cases that we have negative values for these same parameters  (i.e `days_until_submission: -14`, `days_until_release: -14`), the Ruby DateApi at `param.to_datetime.next_day(..)` just subtracts days from the `due_date`, although is allowed, it is another inconsistency. So, negative values or zero should be allowed on these parameters, in the first place?

```ruby
# Milestone with due_date = '2022-10-20', days_until_submission = -14 and days_until_release = -14
Code freeze: October 20, 2022
App Store submission: October 06, 2022
Release: October 06, 2022
```
-----------------------------

Also, assuming that when we refer to `Release`, we are talking about the date when this version will be available to our users, and by `AppStore Submission`, is the date we submit the code to be reviewed and alpha/beta tested. 

What is the expected behavior of the dates when creating milestones?

- The `Code Freeze` date should always be before `AppStore Submission` and `Release` dates?
- The `Code Freeze` date is always before `AppStore Submission`, or those can also be equal?
- The `Release` date is always after `AppStore Submission`, or those can also be equal?
- Should this be validated at all?

-----------------------------

Validating dates is always tricky, I just feel that I should confirm my assumptions before starting anything 😅 

I also feel that I should write a mini-P2 to document those changes, just in case anyone passes through here in the future and needs to do any refactorings. 🗒️

AliSoftware: Yep writing a P2 seems a good idea.

As for the cases you mention above, those seems to be good ones; I think we should add those, then use `UI.user_error!` in the implementation for when those cases are encountered.

AliSoftware: > What is the expected behavior of the dates when creating milestones?

Every sprint we:
 - Do the code freeze for version N on Monday, which corresponds to us cutting a `release/*` branch from `trunk`, adding ❄️  to the N milestone, pushing some new adjustments commits in the `release/*` (like freezing the strings to send to translation, bumping the version numbers, …), then create a first beta that we send to beta testers
 - Then we let that period of beta test on the code-frozen beta run for almost a whole sprint (while developers are working on `trunk` on version N+1 in parallel), doing new betas from the `release/*` branch during that period if there are last-minute betafixes to land
 - On the Friday at the end of the Sprint, once the beta-test period is over and we hopefully beta-fixed anything that was urgent to fix during that period, we submit the final build to the App Store / Play Store
 - On the next Monday, providing that Apple/Google have approved the build we submit over the weekend, we start the rollout of that build to end users, and start a new sprint (code freeze of version N+1; wash, rince, repeat)

This process is also described in [this old blog post](https://mobile.blog/2019/05/06/wordpress-mobile-apps-the-heartbeat-release-process/), though it might be a tad outdated now.

Nowadays some of our apps still have 2-weeks sprints, which means when we do a code freeze on Day D+0 (Monday, start of sprint), we then submit for review on Day D+11 (i.e. second Friday), hope it gets approved over the weekend (it usually is), and release on Day D+14 (next Monday), same day we start a new code freeze. Some other apps use 1-week sprints, which means code freeze at D+0, submission at D+4, release at D+7.

AliSoftware:
@raafaelima FYI, the CI failure is due to the following failure when running the tests (`bundle exec rspec`):

```
1) FastlaneCore::Helper::GithubHelper create_milestone computes the correct dates when the due date is on the verge of a DST day change
  | Failure/Error: expect(client).to receive(:create_milestone).with(test_repo, test_milestone_number, options)
  |  
  | #<InstanceDouble(Octokit::Client) (anonymous)> received :create_milestone with unexpected arguments
  | expected: ("repo-test/project-test", "10.0", {:description=>"Code freeze: June 19, 2022\nApp Store submission: June 21, 2022\nRelease: June 22, 2022\n", :due_on=>"2022-06-19T12:00:00Z"})
  | got: ("repo-test/project-test", "10.0", {:description=>"Code freeze: June 18, 2022\nApp Store submission: June 20, 2022\nRelease: June 21, 2022\n", :due_on=>"2022-06-18T12:00:00Z"})
  | Diff:
  | @@ -1,6 +1,6 @@
  | ["repo-test/project-test",
  | "10.0",
  | {:description=>
  | -   "Code freeze: June 19, 2022\nApp Store submission: June 21, 2022\nRelease: June 22, 2022\n",
  | -  :due_on=>"2022-06-19T12:00:00Z"}]
  | +   "Code freeze: June 18, 2022\nApp Store submission: June 20, 2022\nRelease: June 21, 2022\n",
  | +  :due_on=>"2022-06-18T12:00:00Z"}]


```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -75,11 +75,15 @@ def get_last_milestone(repository)
...
-      # @param [Time] due_date milestone due date (e.g. `2022-10-22T12:00:00Z`)
-      # @param [Integer] days_until_submission Number of days until submission
+      # @param [Time] due_date Milestone due date—which will also correspond to the code freeze date
+      # @param [Integer] days_until_submission Number of days from code freeze to submission to the App Store / Play Store
       # @param [Integer] days_until_release Number of days from code freeze to release
       #
       def create_milestone(repository:, title:, due_date:, days_until_submission:, days_until_release:)
+        UI.user_error!('days_until_release must be greater than zero.') if days_until_release <= 0
+        UI.user_error!('days_until_submission must be greater than zero.') if days_until_submission <= 0
+        UI.user_error!('days_until_release must be greather than days_until_submission') if days_until_submission >= days_until_release
```
AliSoftware: It feels to me that this would be more readable if the conditions matched the messages described in the `user_error!`, rather than being the negative of those ("must be greater than zero" --> use a `days_until-release > 0` condition), otherwise it needs your brain to think what the negative should be to ensure that the `if` is correct and matching the message.

In fact, for the 3rd case I think the condition and messages are incorrect, because it could be valid to consider that the submission and the release could be on the same day (that's not a common/used setup for any of our apps currently, but might be one day, at least it shouldn't be invalid imho)

```suggestion
        UI.user_error!('days_until_release must be greater than zero.') unless days_until_release > 0
        UI.user_error!('days_until_submission must be greater than zero.') unless days_until_submission > 0
        UI.user_error!('days_until_release must be greater or equal to days_until_submission') unless days_until_release >= days_until_submission
```

spec/github_helper_spec.rb
```diff
@@ -217,31 +217,77 @@ def get_milestone(milestone_name:)
...
+      create_milestone(due_date: due_date, days_until_submission: 4, days_until_release: 7)
     end
 
-    it 'has the correct dates to code freeze with submission' do
-      comment = "Code freeze: October 22, 2022\nApp Store submission: October 22, 2022\nRelease: October 25, 2022\n"
-      options = { due_on: '2022-10-22T12:00:00Z', description: comment }
+    it 'computes the correct dates when the due date is on the verge of a DST day change' do
+      # The PST to PDT (DST) change is made on the second Sunday in March and finishes on the first Sunday in November
+      Time.zone = 'Pacific Time (US & Canada)'
+      due_date = '2022-06-18T23:00:00Z'.to_time
```
AliSoftware: > \# The PST to PDT (DST) change is made on the second Sunday in March and finishes on the first Sunday in November

Then that means that the date you used (`2022-06-18` is not even close to a DST change then, since it's neither close to March nor November… so this actually tests the opposite case?

spec/github_helper_spec.rb
```diff
@@ -217,31 +217,77 @@ def get_milestone(milestone_name:)
       allow(Octokit::Client).to receive(:new).and_return(client)
     end
 
-    it 'has the correct dates to code freeze without submission' do
-      comment = "Code freeze: October 22, 2022\nApp Store submission: November 15, 2022\nRelease: October 25, 2022\n"
-      options = { due_on: '2022-10-22T12:00:00Z', description: comment }
+    it 'computes the correct dates for standard period' do
+      due_date = '2022-10-20T08:00:00Z'.to_time.utc
```
AliSoftware: This example date feels dangerously close to the actual DST change date zone—e.g. in Europe, DST starts on the last Sunday of March and ends **on the last Sunday of October**, so it happened on Oct 30 this year. Which is still outside the Oct 20–Oct 27 date range that is used in this example, but still very close.

spec/github_helper_spec.rb
```diff
@@ -217,31 +217,77 @@ def get_milestone(milestone_name:)
...
+        due_date = '2022-10-20T08:00:00Z'.to_time.utc
+        expect { create_milestone(due_date: due_date, days_until_submission: 14, days_until_release: 3) }
+          .to raise_error(FastlaneCore::Interface::FastlaneError, 'days_until_release must be greather than days_until_submission')
+      end
+
+      it 'raises an error if submission date is equal to release date' do
+        due_date = '2022-10-20T08:00:00Z'.to_time.utc
+        expect { create_milestone(due_date: due_date, days_until_submission: 1, days_until_release: 1) }
+          .to raise_error(FastlaneCore::Interface::FastlaneError, 'days_until_release must be greather than days_until_submission')
+      end
```
AliSoftware: This should be a valid case imho (see other comment)

spec/github_helper_spec.rb
mokagio: I wanted to check the PST to PDT details and landed on Wikipedia. Here's [the link](https://en.wikipedia.org/wiki/Pacific_Time_Zone#:~:text=Effective%20in%20the%20U.S.%20in,the%20first%20Sunday%20in%20November.) for future reference, just in case.

spec/github_helper_spec.rb
```diff
@@ -217,31 +217,77 @@ def get_milestone(milestone_name:)
...
+      options = {
+        due_on: '2022-06-18T12:00:00Z',
+        description: "Code freeze: June 18, 2022\nApp Store submission: June 20, 2022\nRelease: June 21, 2022\n"
+      }
+
+      expect(client).to receive(:create_milestone).with(test_repo, test_milestone_number, options)
+      create_milestone(due_date: due_date, days_until_submission: 2, days_until_release: 3)
+    end
+
+    context 'with input validation' do
```
mokagio: Out curiosity, why did you choose to use a `context` here?

spec/github_helper_spec.rb
raafaelima: Yeah, I was a bit confused if this should be or not be a valid case, Initially, I think it was, but as [you explained here](https://github.com/wordpress-mobile/release-toolkit/pull/426#discussion_r1016860739) about the process, it feels that the submissions always need to be before release, and not fall in the same day. I'll change that test to be a valid case, thanks for the input. 😄

spec/github_helper_spec.rb
raafaelima: I thought that will be more organized and improve the readability to have the input validations under the context, instead of having them grouped under the main describe block.  

Although, now that you ask that question, it made me think if this is really useful at all 🤔  
As I do not have any additional behavior to input on those methods, the context here was just misused. 😄

spec/github_helper_spec.rb
raafaelima: I had problems testing this properly, to be honest. It confuses me that the fact that PST/PDT changes are made at 2 am, Also for some reason, the date to not change when I use the 14 march (the actual DST change date) as my date, even forcing the timezone to be at PST. So I just used any date inside the DST changes, given the date that was put was on UTC.

raafaelima: Due to the Ci failure, I think that this is not the proper way to enforce the use of a specific timezone (although it passes here on my local test run). I'll try to build it differently, I was researching other ways to do this and found this one, which is wrapping the test on the timezone change block:

```ruby
Time.use_zone('Europe/London') do
  due_date = Time.zone.parse('2022-03-27 23:00:00')
  # ...
end
```

TBH, I'm not sure if this is the correct way to do it either, as there are a lot of different answers on how to do this (most of them on rails), on StackOverflow. If you folks have suggestions on how to deal with this, I would love to hear @AliSoftware @mokagio 😄 

Note: I'll use the DST from Europe this time, that makes more sense in my head than the American ones. 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
raafaelima: Thanks for the feedback, I was not aware of the `unless` keyword, indeed is more legible that way. 😄

spec/github_helper_spec.rb
AliSoftware: I think Gio's question was more about why using `context` vs `describe` to group the tests. Functionally that doesn't make a difference, but semantically, `context` is more for `when [in situation X]` or `with [given use case or situation]`

mokagio: @raafaelima @AliSoftware, a bit of both 😄 

Why grouping the input validation? (You answered that already) And why with `context` vs `describe`?

It is my understanding that `context` is an alias for `describe` but I can't find the code where that happens. The closer I could get to was the DSL [source code](https://github.com/rspec/rspec-core/blob/ea8554afd1a2b63677c6593059fa8f2476181deb/lib/rspec/core/dsl.rb) which is very meta, and the [documentation](https://rubydoc.info/gems/rspec-core/RSpec/Core/DSL) which implies they are.

In particular, the example from the docs use `context` with a "when ..." like Olivier suggested above. 

```ruby
RSpec.describe "something" do
  context "when something is a certain way" do
    it "does something" do
      # example code goes here
    end
  end
end
```

That pattern matches my experience with and mental model of `describe` vs `context` and it's why seeing this code tickled my interest and made me ask the question 😄

AliSoftware: Some additional doc / reference about `context` vs `describe`:
 - https://rspec.rubystyle.guide/#example-group-naming
 - https://rspec.rubystyle.guide/#context-descriptions
 - https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspeccontextmethod

spec/github_helper_spec.rb
AliSoftware: From the top of my head I don't remember what's the right way to do it, though `Time.use_zoen(…) do…end` feels like a good lead. In any case, I'd recommend picking a date that you know is before _both_ Europe and Pacific DST change dates, and use an offset for the date addition that is large enough to end up after _both_ DST change dates as well. E.g. a date at start of Oct and an offset of 60 days, or something like that, to be sure to cover both cases anyway, even if `Time.use_zone` should do the trick.

AliSoftware:
Note that Rubocop is reporting the two following codestyle offenses:

```
$ bundle exec rubocop
Inspecting 164 files
.....................................................................................................C..............................................................

Offenses:

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb:83:80: C: [Correctable] Style/NumericPredicate: Use days_until_release.positive? instead of days_until_release > 0.
        UI.user_error!('days_until_release must be greater than zero.') unless days_until_release > 0
                                                                               ^^^^^^^^^^^^^^^^^^^^^^
lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb:84:83: C: [Correctable] Style/NumericPredicate: Use days_until_submission.positive? instead of days_until_submission > 0.
        UI.user_error!('days_until_submission must be greater than zero.') unless days_until_submission > 0
                                                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^
```

raafaelima:
Thanks for the catch, fixed on [the last commit](https://github.com/wordpress-mobile/release-toolkit/pull/426/commits/d87c03ccf738ab4fd7d0818c9d8d5c5b508cd687)

spec/github_helper_spec.rb
```diff
@@ -217,31 +217,140 @@ def get_milestone(milestone_name:)
...
+      create_milestone(due_date: due_date, days_until_submission: 1, days_until_release: 1)
     end
 
-    def create_milestone(need_submission:, milestone_duration:, days_code_freeze:)
+    it 'computes the correct dates when the due date is on the verge of a DST day change' do
+      # Europe DST starts on the last Sunday of March, and ends on the last Sunday of October
+      Time.use_zone('Europe/London') do
+        due_date = Time.zone.parse('2022-03-27 23:00:00Z')
+        options = {
+          due_on: '2022-03-28T12:00:00Z',
```
AliSoftware, line 250: Nice edge case of having `2022-03-27 23:00:00Z` being exactly `2022-03-28 00:00:00 +0100` because that day is the exact day of the DST change, when London has switched to `+0100`… while for your next test case on line 259, the `2022-03-26 23:00:00Z` date still corresponds to `2022-03-26 23:00:00 +0000` because this is just before the DST change when London were still on `+0000`…

Took me a while to figure that out though and to triple-check that this difference between the `due_date` (`2022-03-27`) and the `due_on` (`2022-03-28`) was the correct and expected behavior here, and not an incorrect value due to our date math being buggy or whatnot. Phew!

Could maybe benefit from some small code comment (here and on line 259) to give more context about those tricky details?

AliSoftware (approved):
Looking good :shipit:
Thanks for adding so many test cases around timezone boundaries 👍

spec/github_helper_spec.rb
raafaelima: For sure! 😄 

Besides the comments, I also added another test to cover the scenario of the `due_date` is not close enough to the end of the day to exist a date change (Line 259) and also add a new case, to test what happens if you have a date close to the day change, but not in DST (Line 278).

spec/github_helper_spec.rb
```diff
@@ -242,6 +242,9 @@ def get_milestone(milestone_name:)
     it 'computes the correct dates when the due date is on the verge of a DST day change' do
       # Europe DST starts on the last Sunday of March, and ends on the last Sunday of October
       Time.use_zone('Europe/London') do
+        # March 27th, 2022 is the exact day that London switches to the DST (+1h)
+        # If the due date is too close to the next day, a day change will happen
+        # So, 2022-03-27 23:00:00Z will be exactly 2022-03-28 00:00:00 +0100 at the DST change
```
AliSoftware, line 247: ❤️

mokagio:
> Thanks for adding so many test cases around timezone boundaries 👍

+1 – Thanks @raafaelima

PR #427, opened 2022-11-09
Allow `upload_to_s3` not to crash if file exists

AliSoftware:
As discussed in Slack p1667975573803849/1667390691.763629-slack-CC7L49W13, we want to allow `upload_to_s3` to be able to just log the error if the file already exists in S3, to make it a soft error (instead of a hard one that makes the lane crash and fail).

This will especially be useful so that we can make the "Installable Build" (Android app repos) and "Upload to S3" (Android lib repos) steps in CI not fail when users rebuild a whole CI pipeline—that failed on other steps but already uploaded the Installable Build to S3 in the first run—from scratch.

## To Test

 - Validate that the new tests/spec added in this PR on CI do pass ✅ 
 - Check that [the PR in WPAndroid I opened to test this in action](https://github.com/wordpress-mobile/WordPress-Android/pull/17442#discussion_r1018050599) demonstrates the use case properly ✅

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -24,10 +24,10 @@ module Fastlane
           message = "File already exists in S3 bucket #{bucket} at #{key}"
           if params[:skip_if_exists]
             UI.important("#{message}. Skipping upload.")
+            return key
           else
             UI.user_error!(message)
           end
-          return key
         end
 
         UI.message("Uploading #{file_path} to: #{key}")
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -27,7 +27,7 @@ module Fastlane
           else
             UI.user_error!(message)
           end
-          return
+          return key
         end
 
         UI.message("Uploading #{file_path} to: #{key}")
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -20,7 +20,15 @@ module Fastlane
           key = [file_name_hash, key].join('/')
         end
 
-        UI.user_error!("File already exists in S3 bucket #{bucket} at #{key}") if file_is_already_uploaded?(bucket, key)
+        if file_is_already_uploaded?(bucket, key)
+          message = "File already exists in S3 bucket #{bucket} at #{key}"
+          if params[:skip_if_exists]
+            UI.important("#{message}. Skipping upload.")
+          else
+            UI.user_error!(message)
+          end
+          return
+        end
 
         UI.message("Uploading #{file_path} to: #{key}")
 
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -20,7 +20,15 @@ def self.run(params)
...
 
-        UI.user_error!("File already exists in S3 bucket #{bucket} at #{key}") if file_is_already_uploaded?(bucket, key)
+        if file_is_already_uploaded?(bucket, key)
+          message = "File already exists in S3 bucket #{bucket} at #{key}"
+          if params[:skip_if_exists]
+            UI.important("#{message}. Skipping upload.")
+          else
+            UI.user_error!(message)
+          end
+          return key
```
mokagio: What do you think of moving the `return` in the successful branch?

```suggestion
            UI.important("#{message}. Skipping upload.")
            return key
          else
            UI.user_error!(message)
          end
```

Since `UI.user_error!` will crash, it might be misleading to see the `return` after it in the code when looking at it top to bottom.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -101,6 +109,13 @@ def self.available_options
             default_value: true,
             type: Boolean
           ),
+          FastlaneCore::ConfigItem.new(
+            key: :skip_if_exists,
+            description: 'If the file already exists in the S3 bucket, just log the error instead of failing with `user_error!`',
```
mokagio: Nitpick: "the error" is not introduced in the sentence before. What do you think of

```suggestion
            description: 'If receiving an error from S3 because the file already exists in the bucket, just log the error instead of failing with `user_error!`',
```

Not the best, to be honest. It's challenging to be succinct while also informative in these short descriptions. So, happy for this to be discarded if it doesn't seem like a clear improvement.

spec/upload_to_s3_spec.rb
```diff
@@ -158,5 +158,24 @@ def stub_s3_response_for_file(key, exists: true)
         end.to raise_error(FastlaneCore::Interface::FastlaneError, "File already exists in S3 bucket #{test_bucket} at #{expected_key}")
       end
     end
+
+    it 'just logs if the file already exists on S3 and skip_if_exists:true' do
```
mokagio: Have you considered using a context?

```ruby
context 'if the file already exists on S3' do
  it 'fails when skip_if_exists:false' do
    ...
  end

  it 'logs a message without failing if skip_if_exists:true' do
    ...
  end
```

It's six on one, half a dozen on the other. I like how the `context` maps to a branch in the possible code paths but it's a slippery slope towards the pyramid of doom... I think this code base is managed by a team that's disciplined enough not to abuse the construct, still, not a clear cut decision for me ⚖️

mokagio (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: Actually S3 doesn't return an error: it's not like we try to upload and check if S3 rejected us. What we do is we do a HEAD request first to check if the file exists (checking if it returned a content size > 0 or not), and then do the upload only if it doesn't. So mentioning "if S3 returned an error" here would be misleading I think.

spec/upload_to_s3_spec.rb
AliSoftware: Addressed in 964795991609f28fc303a00c4045695d0e60da2f

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: @mokagio Improved in dd723fa. using this wording instead:

> If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`

Let me know what you think 🙂

jkmassel (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
mokagio: :shipit:

PR #429, opened 2022-11-21
Promo Screenshots action improvements

AliSoftware:
## What

This PRs makes improvements around the `promo_screenshots` action

- Add option to check that needed fonts are installed
  - This parses the config file, then find all the stylesheets used in it, then all the `font-family` attributs in those, and finally uses a small Swift script and AppKit to check if the font is installed and activated.
  - In modern macOS, it's not enough to copy/check if a font is in `~/Library/Fonts` to be sure that it's available at runtime, as one can deactivate fonts in FontBook, hence the need for that swift script.
  - There is sadly no built-in CLI to manipulate / install / activate fonts in macOS (at least none that I could find). Only potential candidate was `atsutil`, but it doesn't really cover our needs, hence why I relied on writing my own swift script in the end…
- Allow the config file to be written in YAML
  - so we can use comments as well as YAML anchors
  - Which is particularly relevant for the config file given the number of repetitions of device names and file paths to txt files in different parts of the config\
  - JSON will still be supported because it's a subset of YAML anyway.
- Fix/Update `screenshots-compositor.md` documentation

> **Note**: I didn't take the time to add unit tests because this is just a quick fix for which I don't want to spend time on. In fact, the screenshots automation project is supposed to be on hold since last week, and I'm just finishing up and committing what I had in my WIP to avoid letting it rot only on my machine, before moving on to other priorities ASAP.

---

## What's next

Once I land https://github.com/wordpress-mobile/WordPress-Android/pull/17505, I hope to create another PR on top to point to this release-toolkit, then convert the promo config files from JSON to YAML, so we can start making use of `# comments` officially, as well as DRYing those config files using YAML anchors, to render those config files more readable.

I'm not sure I'll have time to do that soon tho, as I now have other priorities taking over this project, so I'm just leaving a note here as improvement ideas in WPAndroid for later.


lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -32,17 +32,62 @@ def initialize
...
-      def read_json(configFilePath)
+      def read_config(configFilePath)
         configFilePath = resolve_path(configFilePath)
 
         begin
-          return JSON.parse(open(configFilePath).read)
-        rescue
-          linter = JsonLint::Linter.new
-          linter.check(configFilePath)
-          linter.display_errors
```
AliSoftware, line 43: I considered replacing the `jsonlint` gem with `yamllint` gem, but then found out that `yamllint` is not maintained anymore since a while, and trying to `bundle install` it leads to multiple deprecation warnings about it relying on an old `trollop` gem.

I figured that since this was the only place where we used `JsonLint` and linting, and given the config files are not that often modified anyway, and if they are it would likely be with a dedicated IDE, so the most likely mistakes won't be syntax errors in the YAML, but incorrect keys or structure used compared to the keys that are expected… so I considered that it wasn't worth the effort to keep a syntax linter after all.

CHANGELOG.md
```diff
@@ -10,11 +10,13 @@ _None_
 
 ### New Features
 
-_None_
+- `promo_screenshots` now checks that the fonts—referenced via `font-family` in all the stylesheets referenced in the config file—are installed before starting, and prompt to install them if they are not. This check is enabled by default now but can be disabled/skipped if it causes any issue. [#429]
+- `promo_screenshots` now supports config files to be written in `YAML` in addition to still supporting `JSON`. [#429]  
+  Note that while this is technically not a breaking change—because YAML is a superset of JSON—the old way of parsing the JSON file had a loophole that incorreclty allowed `// comments` in the JSON file (despite this not being allowed in the JSON standard). If your config files relied on this bug and used `// comments`, you will have to replace them with `# comments` instead so that they can be parsed correctly by the YAML parser.
```
AliSoftware: I hesitated to put that item under Breaking Change.

Technically speaking, the breaking change is only true if we relied on a bug in the ruby JSON parser ignoring `//` comments, while it should never have according to the JSON spec, so it's not a breaking of the official API contract we had so far, and clients which were using `// comment` were relying on this glitch/bug not on something official from our API contract.

In practice though, I know that most of our `screenshots.json` config files did include `// comments` before each entry to document which screenshot that entry was for… so it will still be breaking and require changes in most of our clients adopting this versions and which relied on that bug… So should that mean a `7.0.0` bump after all?

ParaskP7: Minor (🔍): A small type here, see: `incorreclty` -> `incorrectly`

AliSoftware: Nice catch!
```suggestion
  Note that while this is technically not a breaking change—because YAML is a superset of JSON—the old way of parsing the JSON file had a loophole that incorrectly allowed `// comments` in the JSON file (despite this not being allowed in the JSON standard). If your config files relied on this bug and used `// comments`, you will have to replace them with `# comments` instead so that they can be parsed correctly by the YAML parser.
```

ParaskP7 (commented):
👋 @AliSoftware !

After reading the description, reading your comments and reviewing the code multiple times, without knowing how to best test this, I don't think I am the best to review it after all. Thus, I think it is better that someone else proceed with a proper review. Apologies for not being that useful to you at this instance... 😑

CHANGELOG.md
AliSoftware: Actually, instead of all that, I just pushed a442ee8 which still uses `JSON.parse` if the file has a `.json` extension, and use `YAML.load_file` otherwise (i.e. for `.yml` and `.yaml`). This way, we will still be lenient and support `// comments` if the file is in `.json`, making sure this won't be a breaking change _even_ for clients that relied on that glitch and ruby parser leniency.

jkmassel (approved):

PR #431, opened 2022-11-30
Bump the VM image used to Xcode 14.1

AliSoftware:
What it says on the tin.

Not that we require a recent Xcode to build the toolkit per se; but using a more recent one will mostly avoid having our CI Mac hosts having to download an older xcode VM (which made [my last build for my recent other PR take 1h to build](https://buildkite.com/automattic/release-toolkit/builds/816#0184c8e4-1b51-4aad-b247-630d1ada21a4)), and instead use one that most other repos use — and thus which is more likely to be cached in our Mac hosts, at least for the upcoming months or so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ _None_
 
 ### Internal Changes
 
-_None_
+- Update the CI image used to build this project to use `xcode-14.1`. [#431]
 
 ## 6.0.0
 
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   env: &xcode_image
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ _None_
 
 ### Internal Changes
 
-_None_
+- Update the CI image used to build this project to use `xcode-14.1`. [#431]
```
mokagio, line 21: Does this set a precedent for adding an entry here every time we bump the CI image? I fear it might amount to more noise than signal. What do you think?

mokagio (approved):

CHANGELOG.md
AliSoftware: Yeah, you might be right. I'm ok to remove it.

[EDIT] Oh, forgot I had auto-merge enabled on this PR so it's already merged… oh well.

PR #433, opened 2022-12-05
Add is_draft parameter to create_release_action

rynaardb:
## What does it do?

Adds a new `is_draft` named parameter to the `create_release_action`, allowing callers to specify whether the GitHub release should be created as a draft or not. The `is_draft` named parameter will default to `true` to make it backward compatible.


AliSoftware (commented):
Hey @rynaardb , nice to see this in order to shape more automation for Tumblr 🙂 

Note though that your PR only modifies the helper, which is internal implementation not exposed to the clients of the release-toolkit. You will thus need to also:
 - Update `actions/common/create_release_action.rb`, to add the corresponding `ConfigItem` to expose that parameter to the consumer (i.e. to the caller from `Fastfile`) and pass that parameter to the helper call site
 - Ideally also update the unit tests (aka specs), especially `spec/github_helper_spec.rb`, to add a unit test to validate this `ConfigItem` is properly passed to the (stubbed) GitHub client's call site (our unit tests for this part and this helper and action are not super exhaustive so far, so feel free to extend the coverage there)
 - Also, don't forget to add a `CHANGELOG.md` entry  to this PR 🙂

rynaardb:
> 

Thanks a lot for the feedback @AliSoftware! Much appreciated. This is my first time contributing to the repo, so not surprising to see that I missed a few important steps. I'm going to address them all shortly!

AliSoftware:
You're welcome :)

That makes me think that we could maybe do some quick-win improvements in that area:
 - Maybe add a `.github/PULL_REQUEST.md` PR template, with a small checklist
    - [ ] Add an entry in the appropriate section of the `CHANGELOG.md`
 - We could also consider adding a `Danger` check to validate that a PR contains changes in the `CHANGELOG.md` file and remind the author to add an entry to it otherwise (except if the PR is trivial and doesn't warrant such entry)

You might also be interested to read some P2s our past trialees wrote about their first experience working on the release-toolkit, where they shared some insights: paaHJt-3Rw-p2 (+ followup posts linked at the end)

rynaardb:
> You're welcome :)
> 
> That makes me think that we could maybe do some quick-win improvements in that area:
> 
> * Maybe add a `.github/PULL_REQUEST.md` PR template, with a small checklist
>   
>   * [ ]  Add Unit Tests (aka `specs/*_spec.rb`) if applicable
>   * [ ]  Run `bundle exec rubocop` to test for code style violations and recommendations
>   * [ ]  Run `bundle exec rspec` to run the whole test suite and ensure all your tests pass
>   * [ ]  Add an entry in the appropriate section of the `CHANGELOG.md`
> * We could also consider adding a `Danger` check to validate that a PR contains changes in the `CHANGELOG.md` file and remind the author to add an entry to it otherwise (except if the PR is trivial and doesn't warrant such entry)
> 
> You might also be interested to read some P2s our past trialees wrote about their first experience working on the release-toolkit, where they shared some insights: paaHJt-3Rw-p2 (+ followup posts linked at the end)

I think the pull request template is a great idea! Let me know what you think about https://github.com/wordpress-mobile/release-toolkit/pull/434. Happy to refine it more if needed.

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ _None_
 
 - Update the CI image used to build this project to use `xcode-14.1`. [#431]
 
+## 6.2.0
```
AliSoftware: ⚠️  We should not update the version of the CHANGELOG header when adding new entries to the CHANGELOG, because we don't know when we will do a new version of the release-toolkit, which means we also don't know if there will be breaking changes added to the CHANGELOG by future PRs before we do the next release, which would impact the version bump differently according to SemVer.

For tht reason, you should only add your entry under the existing `### New Features` subsection that is under `Trunk` (which is the section for yet-unreleased changes). In that case, replacing the `_None_` line with yours.

It is only when we will decide to do a new release of the release-toolkit and trigger the release process for it (`bundle exec rake new_release`), that we (well, the `rake` the script, really) will determine which version to bump to (according to semver rules and content of the subsections under `Trunk` section at that moment), create the release with appropriate version, and create a new empty `Trunk` section for the next PRs after it.

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ _None_
...
+
+### New Features
+
+- Add a `is_draft` parameter to the `create_release` action to specify whether the release should be created as a draft.
```
AliSoftware: 💄  Nit: As you can see in the other entries for inspiration, we like to include the PR number in brackets for each entry, to make it easier to find and go read the details of what the changes for that entry were about if one wants to.

```suggestion
- Add a `is_draft` parameter to the `create_release` action to specify whether the release should be created as a draft. [#433]
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
```diff
@@ -83,6 +83,12 @@ def self.available_options
                                        optional: true,
                                        default_value: false,
                                        is_string: false),
+          FastlaneCore::ConfigItem.new(key: :is_draft,
+                                       env_name: 'GHHELPER_CREATE_RELEASE_IS_DRAFT',
+                                       description: 'True if this is a draft release',
```
AliSoftware: 💄  Nit:
```suggestion
                                       description: 'True to create the GitHub release as a draft (instead of publishing it immediately)',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
```diff
@@ -83,6 +83,12 @@ def self.available_options
...
+                                       optional: true,
+                                       default_value: true,
+                                       is_string: false),
```
AliSoftware: 💄  Nit: `is_string` is the legacy way of declaring a non-string parameter. We still have many `ConfigItem` that are using `is_string:` for legacy reasons, but we hope to get rid of them all to migrate to use the more modern `type:` attribute instead (see https://github.com/wordpress-mobile/release-toolkit/issues/278)

While it's outside of this PR to change all the existing `is_string` attributes to the more modern `type:` instead, it would be nice to at least use `type:` for any newly introduced `ConfigItem`:
```suggestion
                                       type: Boolean),
```

spec/github_helper_spec.rb
```diff
@@ -418,7 +418,8 @@ def create_release(assets: [])
         target: test_target,
         description: test_description,
         assets: assets,
-        prerelease: false
+        prerelease: false,
+        is_draft: true
```
AliSoftware: 🧪  I see you added the `is_draft` parameter to the call site, but not yet added any new Unit Tests to validate that different values of `is_draft` in the calls to `GithubHelper#create_release` will indeed result in `client.create_release` and GitHub API calls that contains the proper `draft` parameter too.

Would you like to give it a try? I think that, by taking inspiration from some of the existing tests, adding 2 small tests (one to test the case of `is_draft: true` and another for `is_draft: false`) to ensure that passing those values will indeed result in the `client` be called with the expected parameter… too shouldn't be too difficult, but let me know if you need some help with that.

CHANGELOG.md
rynaardb: Thanks, it wasn't obvious how we deal with the CHANGELOG entries. But your explanation helped me a lot to understand how the process works. Moved the entry to the correct section in https://github.com/wordpress-mobile/release-toolkit/pull/433/commits/30fd7b4fb27ec951a10e3bb6c37bbee9bd777e9e.

CHANGELOG.md
rynaardb: Address in https://github.com/wordpress-mobile/release-toolkit/pull/433/commits/30fd7b4fb27ec951a10e3bb6c37bbee9bd777e9e

spec/github_helper_spec.rb
rynaardb: I've added the unit test in 173f1c264167fbf54573a6f2e0d691c0ae3a29a9. I'm relatively new to Ruby's world of unit testing. So, please feel free to point out any mistakes or if I can make any additional improvements.

spec/github_helper_spec.rb
```diff
@@ -410,15 +410,28 @@ def create_milestone(due_date:, days_until_submission:, days_until_release:)
       create_release(assets: [test_assets])
     end
 
-    def create_release(assets: [])
+    it 'creates a draft release if is_draft is set to true' do
```
AliSoftware, line 413: What wpmobilebot complains about here is a code style violation that you can also see locally by running `bundle exec rubocop`:
```
Offenses:

spec/github_helper_spec.rb:398:5: C: RSpec/RepeatedExample: Don't repeat examples within an example group.
    it 'has the correct options' do ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/github_helper_spec.rb:413:5: C: RSpec/RepeatedExample: Don't repeat examples within an example group.
    it 'creates a draft release if is_draft is set to true' do ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Basically this "example" (which is what Rspec calls a Unit Test case) has the same implementation as the `has the correct options` example set a couple of lines above. Since your new one is more precise in wording and balancing the other one you added about "when is_draft set to false", I'd suggest to delete the one on lines 398 and keep yours which has a nice symetry with the `false` case.

spec/github_helper_spec.rb
```diff
@@ -410,15 +410,28 @@ def create_milestone(due_date:, days_until_submission:, days_until_release:)
...
+      options = { body: test_description, draft: true, name: test_tag, prerelease: false, target_commitish: test_target }
+      expect(client).to receive(:create_release).with(test_repo, test_tag, options)
+      create_release
```
AliSoftware: Totally open question here, but I'm wondering if we shouldn't explicitly pass `create_release(is_draft: true)` in that call site on that example to make it explicity that is the case we are testing. Though I could also see benefits in testing that the behavior is correct with the default values, like you did here. So I don't mind either way.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -119,14 +119,15 @@ def create_milestone(repository:, title:, due_date:, days_until_submission:, day
       # @param [String] description The text to use as the release's body / description (typically the release notes)
       # @param [Array<String>] assets List of file paths to attach as assets to the release
       # @param [TrueClass|FalseClass] prerelease Indicates if this should be created as a pre-release (i.e. for alpha/beta)
+      # @param [TrueClass|FalseClass] is_draft Indicates if this should be created as a draft release (defaults to true)
       #
-      def create_release(repository:, version:, target: nil, description:, assets:, prerelease:)
+      def create_release(repository:, version:, target: nil, description:, assets:, prerelease:, is_draft: true)
```
AliSoftware: I wonder if we should set a default value for `is_draft` in the helper here, instead of letting the responsibility of the default value to be set by the Action (via the `ConfigItem`'s `default_value: true` which already provides this) that uses that helper. That is something that is often up for debate, and I'm not sure we've settled for an answer on what's prefered here.

Personally I'd advocate for not providing a default value in the helper in that case… mostly because it then forces you to search for all the call sites calling that helper internally to check when you'd need to pass an explicit value for `is_draft`. And that would maybe have helped you realize that, while you updated the signature of that `create_release` method in the helper, you indeed forgot to update the main call site for this helper method, that is done by the `create_release_action` (i.e. the public-facing action called by users from their `Fastfile`) — see other comment

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
```diff
@@ -83,6 +83,12 @@ def self.available_options
                                        optional: true,
                                        default_value: false,
                                        is_string: false),
+          FastlaneCore::ConfigItem.new(key: :is_draft,
```
AliSoftware, line 88: You added that parameter (aka `ConfigItem`) to the Fastlane action… but didn't make use of it in the implementation (inside the `self.run` method of the `CreateReleaseAction` class) when actually calling the helper method on line 25 that makes use of it… so technically this parameter will currently be unused and ignored by the action's implementation 😓

spec/github_helper_spec.rb
AliSoftware: Those tests seems good to me 👍  Great job!

_PS: In an ideal word, we'd not only add tests for the internal helper class (`github_helper.rb`) like there, but also for the `CreateReleaseAction` Fastlane action (which is what is exposed as public action for consumption when users call `create_release` from their Fastfile). But we have zero tests for this action at the moment (only tests for the helper it wraps), so maybe this PR is not the place to start adding a brand new `create_release_action_spec.rb` file and tests to that action, and we'll keep that for later (though… writing _those_ action unit tests would have made you notice that you forgot to use the new `ConfigItem` in the action's implementation, so… 😅 )_

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_action.rb
rynaardb: Ahh yeah, thanks for catching that! I completely missed adding the rest of the implementation. I've added it in a2c79346b43c0187f8f1f166941dc99808d32fdd

spec/github_helper_spec.rb
rynaardb: > Though I could also see benefits in testing that the behavior is correct with the default values, like you did here

This was my thinking too.  But, I don't have a strong opinion about this either.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: > I wonder if we should set a default value for is_draft in the helper here, instead of letting the responsibility of the default value to be set by the Action (via the ConfigItem's default_value: true which already provides this) that uses that helper

Another advantage of not providing a default value for `is_draft:` parameter in this helper's method is that there would then be no risk of accidentally having them out of sync with the default value provided by the `ConfigItem` of the action.

Like, imagine if in the future we decide to change the `default_value` of the `ConfigItem(key: :is_draft, …)` in the `CreateReleaseAction` from `true` to `false`… but then forget to also change the default value of the `is_draft:` parameter in that helper function in concert… we'd start having inconsistencies in what default value we consider, depending on if we look at it from the context of the public API / fastlane `Action`, or from the POV of the helper, and that could become confusing.

I'd say better have a single source of truth for the default value, and that source of truth being the `ConfigItem` of the `CreateReleaseAction` (it being the public entry point for this)

spec/github_helper_spec.rb
rynaardb: > though… writing those action unit tests would have made you notice that you forgot to use the new ConfigItem in the action's implementation

Yep, this would have made the missing implementation obvious. I can definitely see the benefit of added tests for the Fastlane actions too.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
rynaardb: Yep! I agree with your reasoning. I also think it makes more sense to set the default value on the `ConfigItem` of the action. I've made the necessary change in 31ba06154133509fbfa34dc5c49b626ce3c08922.

spec/github_helper_spec.rb
jkmassel: `whynotboth.gif`

AliSoftware: > `whynotboth.gif`

@jkmassel Because [this](https://github.com/wordpress-mobile/release-toolkit/pull/433#discussion_r1041046769) 😂 

_(TL;DR: we removed `is_draft` having a default value at helper level, to only keep the default value at action/`ConfigItem` level and avoid multiple sources of truth)_

rynaardb: Yep, as @AliSoftware mentioned, this is no longer relevant. Also see 31ba06154133509fbfa34dc5c49b626ce3c08922

rynaardb:
@AliSoftware Please let me know if I still need to address anything in order to get a 👍

AliSoftware (approved):

PR #439, opened 2023-01-19
Add Mac support to Release Toolkit actions

spencertransier:
## What does it do?
Many of the Release Toolkit actions only included iOS in the available platforms that could be used with that action. With Day One Mac, most of those actions can be used on Mac as well as iOS. This PR adds Mac as a supported platform for all the relevant actions so Day One Mac can use them as well. I went ahead and updated all the relevant actions even if Day One Mac isn't currently using them, just for the sake of not having to make a new Release Toolkit release in the future in case more of them end up being needed. 

For the two actions I changed in the `common` folder, I switched them from `[:ios, :android].include?(platform)` to `true` to match the other actions in `common`. 

For the actions in the `ios` folder, I changed them from `platform == :ios` to `[:ios, :mac.include?(platform)`. A few of them had already been updated, I assume for Simplenote Mac. 

I know it may not make sense to have Mac support in Release Toolkit actions that are specifically labeled "iOS". Because Day One Mac and Simplenote Mac are the only Automattic apps that might even potentially use them right now, I think it makes sense to just add Mac support for these actions rather than duplicating them into a separate `mac` folder. Especially because all of those actions would be just be exact copies of the code within the iOS actions. 

## How to Test
1. In an app like WordPress-iOS, open the Gemfile and change it from this: 

```
# gem 'fastlane-plugin-wpmreleasetoolkit', git: 'git@github.com:wordpress-mobile/release-toolkit', branch: 'trunk'
gem 'fastlane-plugin-wpmreleasetoolkit', '~> 6.0'
```

to this: 

```
gem 'fastlane-plugin-wpmreleasetoolkit', git: 'git@github.com:wordpress-mobile/release-toolkit', branch: 'add/mac-support-in-ios-actions'
# gem 'fastlane-plugin-wpmreleasetoolkit', '~> 6.0'
```

2. Run `bundle install` to get the changes from this PR's branch
3. Add this to the Fastfile: 

```
platform :mac do
  lane :test_mac_support do
    ios_get_app_version
  end
end
```
4. Run `bundle exec fastlane mac test_mac_support`
5. That command should successfully run. Before this PR, it would give an error about Mac not being a supported platform for that action. 

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_betabuild_prechecks.rb
@@ -91,7 +91,7 @@ module Fastlane
       end
 
       def self.is_supported?(platform)
-        platform == :ios
+        [:ios, :mac.include?(platform)
       end
     end
   end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_betabuild_prechecks.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_betabuild_prechecks.rb
@@ -91,7 +91,7 @@ module Fastlane
       end
 
       def self.is_supported?(platform)
-        [:ios, :mac.include?(platform)
+        [:ios, :mac].include?(platform)
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- Added Mac support to all `common` actions and any relevant `ios` actions [#439]
 
 ### Bug Fixes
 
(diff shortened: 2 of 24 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_betabuild_prechecks.rb
```diff
@@ -91,7 +91,7 @@ def self.authors
       end
 
       def self.is_supported?(platform)
-        platform == :ios
+        [:ios, :mac.include?(platform)
```
AliSoftware:
```suggestion
        [:ios, :mac].include?(platform)
```

AliSoftware: You actually got the same typo everywhere, so I'm going to let you do the find-and-replace rather than commenting on each 😉

AliSoftware (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_betabuild_prechecks.rb
spencertransier: @AliSoftware Yes, thank you! 😅 🤦‍♂️ Hey, it's a draft still 😜

AliSoftware: Ah whoops 😅  Received a notification when you added us as reviewer, and didn't think twice nor noticed it was still Draft before going straight to review, sorry 😄

spencertransier: No, thank you for jumping on it so quickly! I feel like GitHub shouldn't notify until it's ready for a review instead of still being a draft.

AliSoftware (approved):

Issue #372, opened 2022-06-14
Default `skip_glotpress` to `true` in `ios_bump_version_release`


mokagio:
With the [new localization tooling](https://github.com/orgs/wordpress-mobile/projects/105) we no longer rely on a local `download_metadata.swift` script for metadata localization. To account for this behavior, the `ios_bump_version_release` action has a `skip_glotpress` parameter. Both [WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/1022e0c0dec6a4f1225b8b8896f5dc7a4987bc86/fastlane/lanes/release.rb#L19) and [WooCommerce iOS](https://github.com/woocommerce/woocommerce-ios/pull/7081) are now using the new tooling and setting `true` for that configuration.

Given that not using `download_medata.swift` is the desired behavior going forward, I think we should also make `skip_glotpress` default to `true`. **Alternatively, we could wholesale remove it and ship a new major version.**

iangmaia:
Picking up this issue 🙋‍♂️

PR #443, opened 2023-01-23
Remove `skip_glotpress` and related cleanup

iangmaia:
**EDIT:**
The initial PR had a few other related changes, but independent, so we decided to break it down into smaller PRs built on top of this one. See **Related PRs**.

### 1️⃣ Removes the `skip_glotpress` parameter from the `ios_bump_version_release` action
Fixing #372, this PR goes on the route of completely removing the `skip_glotpress` config item, given the latest localisation tooling doesn't rely on a local `download_metadata.swift` in the hosted project anymore.

### 2️⃣ Adds basic tests to `ios_bump_version_release`
In order to test the changes, a suite of basic unit tests were added to the `ios_bump_version_release` action.

## Related PRs
- https://github.com/iangmaia/release-toolkit/pull/2 -- includes the changes from @mokagio's [old PR](https://github.com/wordpress-mobile/release-toolkit/pull/370) and builds on top of this one.

### DIFF
--- a/spec/ios_bump_version_release_spec.rb
+++ b/spec/ios_bump_version_release_spec.rb
@@ -1,0 +1,46 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::IosBumpVersionReleaseAction do
+  let(:default_branch) { 'my_new_branch' }
+  let(:ensure_git_action_instance) { double() }
+  let(:versions) { ['6.30'] }
+  let(:next_version) { '6.31.0.0' }
+  let(:next_version_short) { '6.31' }
+
+  describe 'creates the release branch, bumps the app version and commits the changes' do
+    before do
+      allow(Fastlane::Action).to receive(:other_action).and_return(ensure_git_action_instance)
+      allow(ensure_git_action_instance).to receive(:ensure_git_branch).with(branch: default_branch)
+
+      allow(Fastlane::Helper::GitHelper).to receive(:checkout_and_pull).with(default_branch)
+      allow(Fastlane::Helper::GitHelper).to receive(:create_branch).with("release/#{next_version_short}", from: default_branch)
+
+      allow(Fastlane::Helper::Ios::VersionHelper).to receive(:get_version_strings).and_return(versions)
+      allow(Fastlane::Helper::Ios::VersionHelper).to receive(:update_xc_configs).with(next_version, next_version_short, nil)
+    end
+
+    it 'does the fastlane deliver update' do
+      skip_deliver = false
+
+      expect(Fastlane::Helper::Ios::VersionHelper).to receive(:update_fastlane_deliver).with(next_version_short)
+      expect(Fastlane::Helper::Ios::GitHelper).to receive(:commit_version_bump).with(include_deliverfile: !skip_deliver, include_metadata: false)
+
+      run_described_fastlane_action(
+        skip_deliver: skip_deliver,
+        default_branch: default_branch
+      )
+    end
+
+    it 'skips the fastlane deliver update properly' do
+      skip_deliver = true
+
+      expect(Fastlane::Helper::Ios::VersionHelper).not_to receive(:update_fastlane_deliver)
+      expect(Fastlane::Helper::Ios::GitHelper).to receive(:commit_version_bump).with(include_deliverfile: !skip_deliver, include_metadata: false)
+
+      run_described_fastlane_action(
+        skip_deliver: skip_deliver,
+        default_branch: default_branch
+      )
+    end
+  end
+end
--- a/spec/ios_bump_version_release_spec.rb
+++ b/spec/ios_bump_version_release_spec.rb
@@ -2,15 +2,15 @@ require 'spec_helper'
 
 describe Fastlane::Actions::IosBumpVersionReleaseAction do
   let(:default_branch) { 'my_new_branch' }
-  let(:ensure_git_action_instance) { double() }
   let(:versions) { ['6.30'] }
   let(:next_version) { '6.31.0.0' }
   let(:next_version_short) { '6.31' }
 
   describe 'creates the release branch, bumps the app version and commits the changes' do
     before do
-      allow(Fastlane::Action).to receive(:other_action).and_return(ensure_git_action_instance)
-      allow(ensure_git_action_instance).to receive(:ensure_git_branch).with(branch: default_branch)
+      other_action_mock = double()
+      allow(Fastlane::Action).to receive(:other_action).and_return(other_action_mock)
+      allow(other_action_mock).to receive(:ensure_git_branch).with(branch: default_branch)
 
       allow(Fastlane::Helper::GitHelper).to receive(:checkout_and_pull).with(default_branch)
       allow(Fastlane::Helper::GitHelper).to receive(:create_branch).with("release/#{next_version_short}", from: default_branch)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-_None_
+- Remove the `skip_glotpress` parameter from the `ios_bump_version_release` action [#443]
 
 ### New Features
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_release.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_release.rb
@@ -21,10 +21,6 @@ module Fastlane
         Fastlane::Helper::GitHelper.create_branch(@new_release_branch, from: default_branch)
         UI.message 'Done!'
 
-        UI.message 'Updating glotPressKeys...' unless params[:skip_glotpress]
-        update_glotpress_key unless params[:skip_glotpress]
-        UI.message 'Done' unless params[:skip_glotpress]
-
         UI.message 'Updating Fastlane deliver file...' unless params[:skip_deliver]
         Fastlane::Helper::Ios::VersionHelper.update_fastlane_deliver(@new_short_version) unless params[:skip_deliver]
         UI.message 'Done!' unless params[:skip_deliver]
@@ -35,7 +31,7 @@ module Fastlane
 
         Fastlane::Helper::Ios::GitHelper.commit_version_bump(
           include_deliverfile: !params[:skip_deliver],
-          include_metadata: !params[:skip_glotpress]
+          include_metadata: false
         )
 
         UI.message 'Done.'
@@ -55,16 +51,11 @@ module Fastlane
 
       def self.available_options
         [
-          FastlaneCore::ConfigItem.new(key: :skip_glotpress,
-                                       env_name: 'FL_IOS_CODEFREEZE_BUMP_SKIPGLOTPRESS',
-                                       description: 'Skips GlotPress key update',
-                                       is_string: false, # true: verifies the input is a string, false: every kind of value
-                                       default_value: false), # the default value if the user didn't provide one
           FastlaneCore::ConfigItem.new(key: :skip_deliver,
                                        env_name: 'FL_IOS_CODEFREEZE_BUMP_SKIPDELIVER',
                                        description: 'Skips Deliver key update',
-                                       is_string: false, # true: verifies the input is a string, false: every kind of value
-                                       default_value: false), # the default value if the user didn't provide one
+                                       type: Boolean,
+                                       default_value: false),
           FastlaneCore::ConfigItem.new(key: :default_branch,
                                        env_name: 'FL_RELEASE_TOOLKIT_DEFAULT_BRANCH',
                                        description: 'Default branch of the repository',
@@ -104,15 +95,6 @@ module Fastlane
         UI.message("New internal version: #{@new_version_internal}") unless ENV['INTERNAL_CONFIG_FILE'].nil?
         UI.message("New short version: #{@new_short_version}")
         UI.message("Release branch: #{@new_release_branch}")
-      end
-
-      def self.update_glotpress_key
-        dm_file = ENV['DOWNLOAD_METADATA']
-        if File.exist?(dm_file)
-          sh("sed -i '' \"s/let glotPressWhatsNewKey.*/let glotPressWhatsNewKey = \\\"v#{@new_short_version}-whats-new\\\"/\" #{dm_file}")
-        else
-          UI.user_error!("Can't find #{dm_file}.")
-        end
       end
     end
   end


lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -15,14 +15,13 @@ module GitHelper
...
         #
         # @param [Bool] include_deliverfile If true (the default), includes the `fastlane/Deliverfile` in files being commited
-        # @param [Bool] include_metadata If true (the default), includes the `fastlane/download_metadata.swift` file and the `.pot` file (which typically contains an entry or release notes for the new version)
+        # @param [Bool] include_metadata If true (the default), includes the `.pot` file (which typically contains an entry or release notes for the new version)
         #
         def self.commit_version_bump(include_deliverfile: true, include_metadata: true)
-          files_list = [File.join(ENV['PROJECT_ROOT_FOLDER'], 'config', '.')]
+          files_list = [File.join(env_project_root(), 'config', '.')]
           files_list.append File.join('fastlane', 'Deliverfile') if include_deliverfile
           if include_metadata
```
iangmaia: I left this branch for the param `include_metadata` as at first I wasn't sure whether it was being used, but now looking at it and considering our recent discussions, I believe this can also go away and pushed a commit removing it.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -45,29 +40,19 @@ def self.localize_project
...
-        #
-        def self.update_metadata
-          Action.sh("cd #{get_from_env!(key: 'PROJECT_ROOT_FOLDER')} && ./Scripts/update-translations.rb")
-
-          Fastlane::Helper::GitHelper.commit(message: 'Update translations', files: strings_files, push: false)
-
-          Action.sh('cd fastlane && ./download_metadata.swift')
+        def self.strings_files
+          project_root = env_project_root()
+          project_name = env_project_name()
```
AliSoftware: When I saw `env_project_name` aka the value of `PROJECT_NAME` env var it made me wonder if we were still using that as well as I thought we got rid of it…

A quick look at the usages of `def self.strings_files`'s usages made me see that its only call sites were the `commit_version_bump` above and the `GitHelper::localize_project` that is a couple of lines above it. And `localize_project` is already marked in its YARD doc header block as being deprecated/to remove in next major, and is a method that called `./Scripts/localize.py`, which we got rid of during the L10n Project Thread that also got rid of `update-translations.rb` and `download_metadata.swift` and all that.

So in the end I think we could also get rid of the `localize_project` that was calling `localize.py`, and if that means it was the only one left that called `strings_files`, get rid of that, and by cascading effect we could probably also get rid of `env_project_name` as well…

AliSoftware: All that being said, I think the PR is already big as it is, and already addressing too many things at once (4 big items in your PR description) — and it might be better for you to split the work into separate PRs stacked on top of one another to address those step by step (and make the review easier)?

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -15,15 +15,10 @@ module GitHelper
         # @env PROJECT_NAME The name of the directory containing the project code (especially containing the Resources/ subfolder)
         #
         # @param [Bool] include_deliverfile If true (the default), includes the `fastlane/Deliverfile` in files being commited
-        # @param [Bool] include_metadata If true (the default), includes the `fastlane/download_metadata.swift` file and the `.pot` file (which typically contains an entry or release notes for the new version)
         #
-        def self.commit_version_bump(include_deliverfile: true, include_metadata: true)
-          files_list = [File.join(ENV['PROJECT_ROOT_FOLDER'], 'config', '.')]
+        def self.commit_version_bump(include_deliverfile: true)
```
AliSoftware: Thought: at that point I'm also wondering if the `commit_version_bump` is going to be useful very long.

I'm not sure we have many projects left that contain the version number as part of their `Deliverfile` (and that thus gets bumped there) instead of the more modern way we have of doing it in client projects that don't involve the version to be in `Deliverfile` anymore.
If we get rid of that too in a subsequent PR very soon, we won't need the `include_deliverfile` either, which will make that helper method kinda pointless if the remaining of its implementation is just "commit ./config/* files" (and that path to `./config/*.xcconfig` should not be assumed/hardcoded in that helper anyway…)

Definitively not for this PR because its scope is already large enough, but to keep in mind in a subsequent cleanup after this lands (or stacked on top)

AliSoftware (commented):
@iangmaia Thanks for the PR. Took a quick look but then got pre-empted by another task so couldn't finish my review, so I'll just submit the few comments / thoughts I wrote but note that this is not a full review yet 😅  I'll try to look into it in more details tomorrow or the next day.

I liked that you added tests, though I wonder if they might not be testing more the implementation details than the final results? Though it's a tricky balance to find, also because most of the actions you're testing involve git operations (including push) and we definitively want to mock them. But still, I wondered what your thoughts were on that.

Also it feels like this PR packs many different things. I don't mind having a single PR to add the tests and do the changes, that makes sense, but I wonder if some of the small changes in that PR unrelated to the action removal—e.g. the small refactorings like `env_project_root()` instead of `ENV['PROJDCT_ROOT_FOLDER']` etc—couldn't have been extracted in a separate PR stacked/built on top of this one, to not mix refactorings and removal of deprecated code.

Again, that's sometimes a tough balance to find because if the refactorings are very small renaming that doesn't warrant a dedicated PR just for so little. But my first impression when I read the PR and the PR description was "wow, 4 different things packed in one single PR" and then the diff shown by GitHub doesn't always help (depending on how it chooses to decide where to split the diffs of added/removed lines) + me being EOD already… probably didn't help 😄  Yet, if you think splitting in the changes in separate PR is doable and makes sense to you, then please do.

_And of course, given the breakcrumbs of comments I left already (about `Deliverfile` and maybe more of the code also being obsolete), we will likely have subsequent PRs to also clean some more anyway, so…_

iangmaia:
@AliSoftware thanks for the feedback, looking forward to the rest but this already gives me a few things to think about 🙂 

> I liked that you added tests, so I wonder if they might not be testing more the implementation details than the final results? Though it's a tricky balance to find, also because most of the actions you're testing involve git operations (including push) and we definitively want to mock them. But still, I wondered what your thoughts were on that.

It's indeed a tricky balance to find. I actually thought about this after sending the PR. My general line of thought, in this case, was that the dependencies could have their own unit tests and that these were unit tests for the method/action logic (checking the different scenarios, error cases, how data was going out to dependencies, etc) as opposed to something closer to an integration test (e.g. focused more on final results with all or mostly the real dependencies).
I think there is an argument to even have both, but often choices have to be made 🙂 

> Also it feels like this PR packs many different things. I don't mind having a single PR to add the tests and do the changes, that makes sense, but I wonder if some of the small changes in that PR unrelated to the action removal—e.g. the small refactorings like `env_project_root()` instead of `ENV['PROJDCT_ROOT_FOLDER']` etc—couldn't have been extracted in a separate PR stacked/built on top of this one, to not mix refactorings and removal of deprecated code.

To me each change kinda led to the next one and they're all related (but not strictly dependent, as you point out) -- this small refactoring of `ENV['PROJDCT_ROOT_FOLDER']` -> `env_project_root()`, for example, was made with the intent of making it easier to unit test `commit_version_bump()` from `ios_git_helper.rb`.
So it was like "remove skip_glotpress"  led to -> "remove everything else referring to download_metadata.swift, adding tests for commit_version_bump() " led to -> "Remove /Scripts/update-translations.rb and the related action".
But yes I agree, they could have been split for an easier reviews 👍 

> Again, that's sometimes a tough balance to find because if the refactorings are very small renaming that doesn't warrant a dedicated PR just for so little. But my first impression when I read the PR and the PR description was "wow, 4 different things packed in one single PR" and then the diff shown by GitHub doesn't always help (depending on how it chooses to decide where to split the diffs of added/removed lines) + me being EOD already… probably didn't help 😄 Yet, if you think splitting in the changes in separate PR is doable and makes sense to you, then please do.
> _And of course, given the breakcrumbs of comments I left already (about `Deliverfile` and maybe more of the code also being obsolete), we will likely have subsequent PRs to also clean some more anyway, so…_

Yes, when I checked the comment on Deliverfile, this made even more sense. I agree it's a difficult balance sometimes, and different teams / companies may have different "thresholds" on what is considered a big PR or not. As I said, in the beginning I thought "everything is related, so fine to keep it here", but the fact that they are independent changes with different goals is a good argument to split them into different PRs, IMO.

So I'm now thinking of keeping changes 1 & 2 here in this PR (so the main work fixing the issue is here) and 3 & 4 & any further cleanups can go to their own PRs.
Since this is clear, I'll just do it now to then make it easier already for everyone to continue on their review (this discussion is also gonna help with the fix for the next issue).

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
iangmaia: As I said [here](https://github.com/wordpress-mobile/release-toolkit/pull/443#issuecomment-1402626367), totally agree. I can send a separate PR for cleaning it up as well while I'm at it, I think it would be nice to remove all the references to the deprecated localisation code when we release a new version.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
iangmaia: Got it! Yes, definitely for a subsequent PR. This is something more obviously in another domain as we're not talking about removing deprecated localisation related code, but anyway, relates to the affected files, thanks for pointing it out 👍

mokagio:
_If_ there's time left in the trial, it might be fun to remove `Deliver` file from the apps still using it. [WooCommerce iOS is one of them](https://github.com/woocommerce/woocommerce-ios/blob/e7e167702bf4e220d32f1570910aa5a2caaa10a6/fastlane/Deliverfile).

mokagio:
In regards to PR size and scope

> I agree it's a difficult balance sometimes, and different teams / companies may have different "thresholds" on what is considered a big PR or not. As I said, in the beginning I thought "everything is related, so fine to keep it here", but the fact that they are independent changes with different goals is a good argument to split them into different PRs, IMO.

It's a bit of a science and a bit of an art, isn't it?

I find that in the context of a highly-distributed, async-first team like ours, it's best to err on the size of smaller PRs, ideally independent and parallel ones.

When the feedback cycle on a PR can take a full day because of the relative location of author and reviewer, having a single PR with various items in it might result in all the items being blocked from merging because of an issue found in only one of them. Splitting them allows unrelated changes not to block each other.

spec/ios_bump_version_release_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  let(:default_branch) { 'my_new_branch' }
+  let(:ensure_git_action_instance) { double() }
+  let(:versions) { ['6.30'] }
+  let(:next_version) { '6.31.0.0' }
+  let(:next_version_short) { '6.31' }
+
+  describe 'creates the release branch, bumps the app version and commits the changes' do
+    before do
+      allow(Fastlane::Action).to receive(:other_action).and_return(ensure_git_action_instance)
+      allow(ensure_git_action_instance).to receive(:ensure_git_branch).with(branch: default_branch)
```
AliSoftware: Nit: given how `other_action` works in fastlane internally—i.e. it's a sort of _proxy object_, to call any other built-in Fastlane action from a fastlane plugin's implementation—I think it'd make more sense to call the mock just "`other_action`" or `other_action_mock` or something, rather than name the object returned by the `other_action` function call to be related to what method will be called on it later.

```suggestion
      other_action = double()
      allow(Fastlane::Action).to receive(:other_action).and_return(other_action)
      allow(other_action).to receive(:ensure_git_branch).with(branch: default_branch)
```

AliSoftware (approved):
Overall looks good, nice isolated PR, and adding tests that we were crucially missing 👍 

I guess the next step will be to continue the cleanup by getting back on [Gio's Draft PR you found earlier](https://github.com/wordpress-mobile/release-toolkit/pull/370), then lastly check there are no other leftovers (of the our old way of relying on scripts that were expected to be in the client repo) to clean up after all that.

I'll let @mokagio confirm (here on or the P2 where you asked the question) if the blocker for his Draft PR back in the day (PocketCasts-iOS) is now lifted and let him tell you if it makes sense for you to take over his Draft PR 🙂

AliSoftware:
> My general line of thought, in this case, was that the dependencies could have their own unit tests and that these were unit tests for the method/action logic (checking the different scenarios, error cases, how data was going out to dependencies, etc) 

I like that rationale, great answer, and I agree 👍

> So I'm now thinking of keeping changes 1 & 2 here in this PR (so the main work fixing the issue is here) and 3 & 4 & any further cleanups can go to their own PRs.

Sounds like a good plan 👍

iangmaia:
Thanks for your comments, @mokagio.
> _If_ there's time left in the trial, it might be fun to remove `Deliver` file from the apps still using it. [WooCommerce iOS is one of them](https://github.com/woocommerce/woocommerce-ios/blob/e7e167702bf4e220d32f1570910aa5a2caaa10a6/fastlane/Deliverfile).

That would be interesting, yes, I hope there will be some time left 😬 

> It's a bit of a science and a bit of an art, isn't it?
> 
> I find that in the context of a highly-distributed, async-first team like ours, it's best to err on the size of smaller PRs, ideally independent and parallel ones.
> 
> When the feedback cycle on a PR can take a full day because of the relative location of author and reviewer, having a single PR with various items in it might result in all the items being blocked from merging because of an issue found in only one of them. Splitting them allows unrelated changes not to block each other.

Exactly, very good point -- larger PRs can potentially take much longer to be merged due to eventual issues found in the review, specially in an async fashion.

spec/ios_bump_version_release_spec.rb
iangmaia: Good point 👍

mokagio:
> I'll let @mokagio confirm (here on or the P2 where you asked the question) if the blocker for his Draft PR back in the day (PocketCasts-iOS) is now lifted and let him tell you if it makes sense for you to take over his Draft PR 🙂

Pocket Casts iOS [no longer uses the legacy APIs](https://github.com/Automattic/pocket-casts-ios/pull/399).

@iangmaia I don't think you can push to #370 directly not being a contributor. Maybe it's easier to cherry-pick those changes and make a new PR from there. Or, given #370 is showing conflicts, only look at the diff as a starting point.

PR #445, opened 2023-01-27
Add configuration item for .xcconfig file in `ios_get_app_version`

### LINKED ISSUE #380: Add configuration item for `xcconfig` with version info to `ios_get_app_version`
Currently, the action relies on the value being set in an environment variable, `PUBLIC_CONFIG_FILE`. We ought to make it explicit and ditch this easy to miss configuration.

iangmaia:
## What does it do?
This PR fixes #380 by adding a configuration item in the action `ios_get_app_version`, to be used instead of the environment variable `ENV['PUBLIC_CONFIG_FILE']`.

### 1️⃣ Adds the parameter `public_version_xcconfig_file` as a ConfigItem to the action `ios_get_app_version`.
The new parameter will be used to extract the app version, without relying on environment variables and deprecating the method that still does.
In order to to that, I've created a parameterised copy of the method `Ios::VersionHelper.get_public_version()`, which I called `Ios::VersionHelper.get_xcconfig_public_version`, also marking the original version as deprecated, and eventually we should stop using it.

Please note that there are still many uses in the codebase of the `ENV['PUBLIC_CONFIG_FILE']` (as well as `ENV['INTERNAL_CONFIG_FILE']`), not only directly in the actions but by their dependencies. These should be addressed in follow up Pull Requests.

### 2️⃣ Adds basic tests to `ios_get_app_version`
Basic tests to the action functionality. These tests are gonna be extended in the next PR, where I'll also improve some of the related code in `Ios::VersionHelper`.

## Related PRs
- [This PR](https://github.com/iangmaia/release-toolkit/pull/1) improves the code for the .xcconfig parsing and adds new tests to `ios_get_app_version`

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_app_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_app_version.rb
@@ -24,7 +24,7 @@ module Fastlane
         [
           FastlaneCore::ConfigItem.new(
             key: :public_version_xcconfig_file,
-            env_name: 'FL_IOS_GET_APP_VERSION_PUBLIC_VERSION_XCCONFIG_FILE',
+            env_name: 'FL_IOS_PUBLIC_VERSION_XCCONFIG_FILE',
             description: 'Path to the .xcconfig file containing the public app version',
             type: String,
             optional: false
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_app_version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_app_version.rb
@@ -4,9 +4,8 @@ module Fastlane
       def self.run(params)
         require_relative '../../helper/ios/ios_version_helper'
 
-        UI.user_error!('You need to set at least the PUBLIC_CONFIG_FILE env var to the path to the public xcconfig file') unless ENV['PUBLIC_CONFIG_FILE']
-
-        Fastlane::Helper::Ios::VersionHelper.get_public_version
+        public_version_xcconfig_file = params[:public_version_xcconfig_file]
+        Fastlane::Helper::Ios::VersionHelper.get_xcconfig_public_version(xcconfig_file: public_version_xcconfig_file)
       end
 
       #####################################################
@@ -22,7 +21,15 @@ module Fastlane
       end
 
       def self.available_options
-        # Define all options your action supports.
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :public_version_xcconfig_file,
+            env_name: 'FL_IOS_GET_APP_VERSION_PUBLIC_VERSION_XCCONFIG_FILE',
+            description: 'Path to the .xcconfig file containing the public app version',
+            type: String,
+            optional: false
+          ),
+        ]
       end
 
       def self.output
--- a/spec/ios_get_app_version_spec.rb
+++ b/spec/ios_get_app_version_spec.rb
@@ -1,0 +1,41 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::IosGetAppVersionAction do
+  describe 'getting the public app version from the provided .xcconfig file' do
+    it 'parses the xcconfig file format correctly and gets the public version' do
+      xcconfig_mock_content = <<~CONTENT
+        // a comment
+        VERSION_SHORT=6
+        VERSION_LONG=6.30.0
+      CONTENT
+
+      allow(File).to receive(:exist?).and_return(true)
+
+      expect_version(xcconfig_mock_content: xcconfig_mock_content, expected_version: '6.30')
+    end
+
+    it 'parses the xcconfig file format correctly and gets the public hotfix version' do
+      xcconfig_mock_content = <<~CONTENT
+        VERSION_SHORT=6
+        // a comment
+        VERSION_LONG=6.30.1
+      CONTENT
+
+      allow(File).to receive(:exist?).and_return(true)
+
+      expect_version(xcconfig_mock_content: xcconfig_mock_content, expected_version: '6.30.1')
+    end
+
+    def expect_version(xcconfig_mock_content:, expected_version:)
+      xcconfig_mock_file_path = File.join('mock', 'file', 'path')
+
+      allow(File).to receive(:open).with(xcconfig_mock_file_path, 'r').and_yield(StringIO.new(xcconfig_mock_content))
+
+      version_result = run_described_fastlane_action(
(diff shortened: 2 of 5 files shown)


spec/ios_get_app_version_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      allow(File).to receive(:exist?).and_return(true)
+
+      expect_version(xcconfig_mock_content: xcconfig_mock_content, expected_version: '6.30.1')
+    end
+
+    def expect_version(xcconfig_mock_content:, expected_version:)
+      xcconfig_mock_file_path = File.join('mock', 'file', 'path')
+
+      allow(File).to receive(:open).with(xcconfig_mock_file_path, 'r').and_yield(StringIO.new(xcconfig_mock_content))
```
AliSoftware, line 32: I really like the use of `StringIO` here to mock the `File.open` with a simple IO object; but that also means that we assume implementation details (i.e. that the action will use `File.open` to read the `.xcconfig` file, as opposed to any other mean to read that file, like `File.read` or using the `xcodeproj` gem and `Xcodeproj::Config.new` (which, tbh, is what `read_from_config_file` should really have been using all that time… ohhh legacy code…)

As such, I'd suggest to try not to rely on the knowledge of implementation details of how the action reads the file, and instead create the temp file in a temp folder (you have some helper methods to help you with that in `spec_helper.rb`) so that the test will still work even if we update the implementation details of how the `.xcconfig` file is read in the future.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_app_version.rb
```diff
@@ -22,7 +21,15 @@ def self.details
       end
 
       def self.available_options
-        # Define all options your action supports.
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :public_version_xcconfig_file,
+            env_name: 'FL_IOS_GET_APP_VERSION_PUBLIC_VERSION_XCCONFIG_FILE',
```
AliSoftware: I anticipate that, when we will add a similar `ConfigItem` to more actions (in future PRs), we'll likely also name those `public_version_xcconfig_file`, while they will all expect the same `.xcconfig` file path each and won't really benefit from having a _separate/different_ ENV var for each separate action. Given that, what do you think about using an env var name that would be a bit more generic (and action-agnostic), like `FL_IOS_PUBLIC_VERSION_XCCONFIG_FILE`… that would _also_ fit the other actions when those will end up also adopting such `ConfigItem` later?

spec/ios_get_app_version_spec.rb
iangmaia: Ah, I totally agree, with this there's a big assumption on implementation details 👍 
I'll change it, thanks for the suggestion!

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_app_version.rb
iangmaia: Very good point! This env var will definitely be required on several actions and it refers to exactly the same thing -- I was trying to think of reasons to have different values between actions, but I don't think this is something that can (or even should) happen in practice.
I'll change the `env_name` to a more generic one and, when I touch another action later, I'll use the same name 👍

spec/ios_get_app_version_spec.rb
iangmaia: Also, if you agree, I'll make this change on the [improvement PR](https://github.com/iangmaia/release-toolkit/pull/1) I made on top of this, using `Xcodeproj::Config`.

AliSoftware: Works for me 👍

iangmaia: Done [here](https://github.com/iangmaia/release-toolkit/pull/1/commits/02a99f3f822de360db6587ce97ecf19fb25b2f0e)

spec/ios_get_app_version_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+describe Fastlane::Actions::IosGetAppVersionAction do
+  describe 'getting the public app version from the provided .xcconfig file' do
+    it 'parses the xcconfig file format correctly and gets the public version' do
+      xcconfig_mock_content = <<~CONTENT
+        // a comment
+        VERSION_SHORT=6
+        VERSION_LONG=6.30.0
+      CONTENT
+
+      allow(File).to receive(:exist?).and_return(true)
```
mokagio, line 12: I started commenting about this `allow` being a bit too loose and duplicated, but I see that https://github.com/iangmaia/release-toolkit/pull/1 extracted it in the `expect_version` helper (great idea).  I'll comment there.

mokagio: https://github.com/iangmaia/release-toolkit/pull/1/files#r1094033726

AliSoftware (approved):
Looking good, assuming we land https://github.com/iangmaia/release-toolkit/pull/1 on top of it as planned, to refine/improve the way the xcconfig file is read in the way you did in that other PR)

⚠️  Note that I'm still waiting for us to land my unrelated PR https://github.com/wordpress-mobile/release-toolkit/pull/449 in trunk and do a new release before starting to land your PRs, because I'm hoping we can make an initial (almost-)non-breaking release of the toolkit first, before starting to land your breaking-change ones.

AliSoftware:
> ⚠️ Note that I'm still waiting for us to land my unrelated PR https://github.com/wordpress-mobile/release-toolkit/pull/449 in trunk and do a new release before starting to land your PRs, because I'm hoping we can make an initial (almost-)non-breaking release of the toolkit first, before starting to land your breaking-change ones.

Actually, I just created and pushed a new branch called `iangmaia/trial`, currently pointing to the exact same commit as our current `trunk`… so that if you re-target your PR here to this `iangmaia/trial` branch instead of `trunk`, we could start landing it (and then other PRs of yours) there, and move forward with https://github.com/iangmaia/release-toolkit/pull/1 and all the other PRs you created on top of those… without us having to wait for me to land my PR and do a release first. I don't know why I didn't think of this before 😅

AliSoftware:
Dammit, now we have a conflict on the `CHANGELOG.md` that will need fixing before I can re-trigger CI and we can finally merge 😅 🙂

iangmaia:
> Dammit, now we have a conflict on the `CHANGELOG.md` that will need fixing before I can re-trigger CI and we can finally merge 😅 🙂

Done!

AliSoftware:
CI took its time (Friday evening UTC is usually quite a busy period 😅 ) but finally went thru, all green and merged!

PR #447, opened 2023-01-30
Remove legacy localisation script references and actions

iangmaia:
## What does it do?
This is a clean up removing the remainder references to the legacy localisation tooling, following up #443.

It contains:

- Removal of the actions `ios_localize_project` and `ios_update_metadata`.
- Removal of the methods `localize_project`, `update_metadata` and `strings_files` on `Ios::GitHelper`
- Removal of the parameter `include_metadata` from the method `Ios::GitHelper.commit_version_bump(include_deliverfile:)`

## Related PRs
- **Note**: I originally created this PR in my fork https://github.com/iangmaia/release-toolkit/pull/2, now I re-created it here given #443, the PR this one was based on, has been merged.
- This PR contains @mokagio's [draft PR](https://github.com/wordpress-mobile/release-toolkit/pull/370), which can be closed if this is merged.
- Based on the changes done here, [this PR](https://github.com/iangmaia/release-toolkit/pull/3) will remove functionality related to the obsolete `Deliverfile`.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 - Remove the `skip_glotpress` parameter from the `ios_bump_version_release` action [#443]
+- Remove the actions `ios_localize_project` and `ios_update_metadata` [#447]
```
AliSoftware: Nit: English wording
```suggestion
- Remove the `ios_localize_project` and `ios_update_metadata` actions [#447]
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
```diff
@@ -9,67 +9,18 @@ module GitHelper
...
-        # @param [Bool] include_metadata If true (the default), includes the `fastlane/download_metadata.swift` file and the `.pot` file (which typically contains an entry or release notes for the new version)
         #
-        def self.commit_version_bump(include_deliverfile: true, include_metadata: true)
-          files_list = [File.join(ENV['PROJECT_ROOT_FOLDER'], 'config', '.')]
+        def self.commit_version_bump(include_deliverfile: true)
+          files_list = [File.join(get_from_env!(key: 'PROJECT_ROOT_FOLDER'), 'config', '.')]
           files_list.append File.join('fastlane', 'Deliverfile') if include_deliverfile
-          if include_metadata
-            files_list.append File.join('fastlane', 'download_metadata.swift')
-            files_list.append File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'Resources', ENV['APP_STORE_STRINGS_FILE_NAME'])
```
AliSoftware, line 25: Seeing this, I wonder if it would be worth it to document somewhere the changes potentially needed in client projects when migrating to the next major version that will contain this change.

We don't really have a "Migration Guides" documentation so far in release-toolkit (because most of our past breaking changes were usually things we already adopted in all client projects before getting rid of them in the toolkit). But for all these changes we're making here, given how it can be easy to lose track, maybe that would be worth it to make a list of things to check when migrating? (Not sure where, maybe in a `docs/MigrationGuide.md` file for example?)

---

Typical examples of things we could remind people to check on their client projects when migrating it to the future new toolkit:
 - Ensure that any call to `ios_bump_version_release` in their Fastfile already passed `skip_glotpress: true`. If they passed `false` or didn't provide a value (`false` being the default for this `ConfigItem`), that means they'll have to ensure (1) they don't use `download_metadata.swift` anymore, or if not, it'd be high time for them to migrate to the new tooling instead [link to my Project Thread maybe], and (2) that they don't rely on this `ios_bump_version_release` for commiting the `.po`/`.pot` file
 - Note that they can delete the `ENV['APP_STORE_STRINGS_FILE_NAME']` from their `Fastfile`, if they still have it defined, because it's not used by anything anymore.

---

Mentioning all this because, while reviewing your PR, I wanted to check the status of WordPress-iOS to validate this wouldn't break it, and while I confirmed that [the only place where it calls `ios_bump_version_release` it also passes `skip_glotpress: true`](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/fastlane/lanes/release.rb#L19)—so we're good there—I also noticed that [we kept the declaration of that now-unused env var](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/fastlane/Fastfile#L59) while it is not needed anymore and would make for a good occasion of a quick cleanup.

AliSoftware (approved):
The diff looks good to me, and I checked the various call sites of the helper methods that you deleted and confirm that they were not used elsewhere outside of the actions you also removed. Great cleanup!! 🧹 🧹 🧹 

Dropped an idea about starting a quick migration guide for all the things we're breaking in those various cleanups you're doing. But isn't a blocker for this to land.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb
iangmaia: > Seeing this, I wonder if it would be worth it to document somewhere the changes potentially needed in client projects when migrating to the next major version that will contain this change.

This is a nice idea! It's a good point on cases where there isn't necessarily a breaking change but still the client projects could do a clean up on unnecessary things they previously had to do, or to implement a different behaviour than they'd previously have.
I think the examples you gave for this PR sound like a good starting point.

What about a `MIGRATION.md` file in the root, next to the `CHANGELOG.md`?
The way I see it, the structure and the process would be very similar to the change log, with sections for each version containing each a bullet list of items that require attention during the migration from the previous version. Then all PRs making such changes would include something in the list.
What do you think? I can kick it off for this PR, it shouldn't take long :-)

AliSoftware:
I've created and pushed a new branch `iangmaia/trial` that points to the same commit as what `trunk` currently points to. If you re-target your PR to this branch, we can land this PR right now (and when we'll be ready to start landing breaking changes like this one in `trunk`, we'll just land all the PRs of yours that landed in `iangmaia/trial` back to `trunk` in one swoop) 🙂

PR #448, opened 2023-01-31
Update `docs/Release Notes.md`

AliSoftware:
While wandering around in the repo and seeing the `docs/` folder, I found that old `docs/Release Notes.md` file there which happened to be quite outdated, so I figured I might as well make a quick update to it while I was around.

## What does it do?

Updates the `docs/Release Notes.md` file with more up-to-date documentation.


docs/Release Notes.md
```diff
@@ -1,17 +1,19 @@
...
+4. That editorialized copy is then used to update the `metadata/release-notes.txt` file, which will contain the "nice" release notes for the current version.
+5. The `{ios|an}_update_metadata_source` action is then used to update the `{App|Play}StoreStrings.po` file with the content in `step 4`. That file will later be picked up by a cron job and uploaded to the appropriate GlotPress project dedicated to the release notes and metadata of each app, so that our translators can start translating the release notes.
+6. At the end of the beta testing, the `gp_downloadmetadata` action is used to download the translation for the release notes and the other relevant metadata from GlotPress, and update the `fastlane/metadata/{locale}/release_notes.txt` files in fastlane's dedicated folder—ready to be used by `deliver`/`supply` which will take care of using the Apple and Google APIs to upload them in App Store Connect / Play Store Console. 
 
 _Note:_ All these actions need some not-trivial configuration, so it’s usually best to add related lanes in the project and use them to invoke the actions.
```
mokagio:
```suggestion
_Note:_ All these actions need some not-trivial configuration, so it’s usually best to add dedicated lanes in the project and use them to invoke the actions.
```

🤷‍♂️ ?

mokagio (approved):
Thank you for taking the time to update this.

However, I wonder whether we should keep this around or remove it? This is bound to become outdated again, no?

I can see the value in documenting how the actions the plugin offers should be used together for the release process, but maybe we'd be best served to elect one project per platform as the gold standard and say "refer to `project` to see how the actions are used together to run a release"?

AliSoftware:
Yeah tbh I wondered the same. I'd be ok to delete that doc too; though when we have a newcomer to the release-toolkit (like a trialee or a new hire) it's nice to have at least a bird's eye view of the overall process documented right in the toolkit?

mokagio (approved):
>  I'd be ok to delete that doc too; though when we have a newcomer to the release-toolkit (like a trialee or a new hire) it's nice to have at least a bird's eye view of the overall process documented right in the toolkit?

Yeah, that's true. At that point, reading a few lines in a doc would be faster than navigating the `Fastfile` of an app.

The worst that can happen here is that the document becomes out-of-date again. Updating it when the time comes might be a good occasion to revisit our process, too.

PR #452, opened 2023-02-12
Add `MIGRATION.md` migration guide

iangmaia:
## What does it do?
This PR proposes the usage of a migration guide in the form of the file `MIGRATION.md`.
The idea is that a guide like this can be specially useful to provide more context for a successful migration between major releases of the `release-toolkit`, giving more information about deprecations, clean-ups that the clients can do, how to deal with breaking changes and any other adaptations that the hosting projects need to implement when migrating to a new version.

A process to update the `MIGRATION.md` file can be very similar to the one followed by `CHANGELOG.md`:
- PRs containing changes that would require attention during a migration can highlight them in the appropriate subsection of the `## Migrating to Trunk` section
- Once the new major version is released, `## Migrating to Trunk` can be changed to `## Migrating to 7.0.0`, for example
- A new empty `## Migrating to Trunk` section is added to contain the updates for the next version

I've initially organised the migration topics into two main subsections, `Considerations for breaking changes` and `Clean-ups`, which I believe will be common cases; more subsections can be added or a different grouping can be done as we see fit.

[This](https://github.com/wordpress-mobile/release-toolkit/pull/447#discussion_r1092386301) is the discussion I had with @AliSoftware that originated this PR.

## Related PRs
- This PR was originally created on my fork; for the original review and discussion, please check https://github.com/iangmaia/release-toolkit/pull/4


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,4 +5,5 @@
 - [ ] Run `bundle exec rubocop` to test for code style violations and recommendations
 - [ ] Add Unit Tests (aka `specs/*_spec.rb`) if applicable
 - [ ] Run `bundle exec rspec` to run the whole test suite and ensure all your tests pass
-- [ ] Make sure you added an entry in [the `CHANGELOG.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md#trunk) to describe your changes under the approprioate existing `###` subsection of the existing `## Trunk` section.
\ No newline at end of file
+- [ ] Make sure you added an entry in [the `CHANGELOG.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md#trunk) to describe your changes under the appropriate existing `###` subsection of the existing `## Trunk` section.
```
iangmaia, line 8: @AliSoftware: I noticed a small typo in the `.github/PULL_REQUEST_TEMPLATE.md` ("approprioate"), and while editing it I realised it would be nice to mention the `MIGRATION.md` here as well. Please let me know if you think this makes sense, otherwise I can just revert it.

AliSoftware (commented):
Monday morning, new week, fresh mind, and now I'm wondering if having a separate document is worth it vs including those instructions directly in the `CHANGELOG.md` as a dedicated paragraph 🤔  @wordpress-mobile/apps-infrastructure  what do you think?

On one hand, I was the one to initially suggest a Migration Guide, and I like having a dedicated document for how to adopt breaking changes in client repos, On the other hand, having a separate document would me a separate document to maintain and update, having to update our release scripts to ensure it has been updated when we are about to release a major version, and having people to know about the document when they adopt a new version so that they remember to check the separate file to know the instructions. So I kinda think I might have changed my mind and would maybe prefer to put this all in the `CHANGELOG.md` after.

But this also depends on how much we want to write in such Migration Guide. If those migrations end up having a lot of things to write about and containing long paragraphs, then it's likely worth a separate document. If it's smaller content, maybe not?

Also, other OSS repos [like Alamofire](https://github.com/Alamofire/Alamofire/tree/master/Documentation) have a separate document for each major version bump — as opposed to have a single document with sections for each version. Could be worth considering such an option. But that being said, they have way more things to write about in each of those migration guides, because each of their major version bumps was significant, while in our cases we are more likely (at least at this stage of the toolkit's life) to do major version bumps and small breaking changes more often rather than hold them for a while a do a big refactoring and major bump with a lot of breaking changes packed in a single release, so… 🤔

iangmaia:
> Monday morning, new week, fresh mind, and now I'm wondering if having a separate document is worth it vs including those instructions directly in the `CHANGELOG.md` as a dedicated paragraph 🤔

I had the exact same thought when I first heard it: this could just be part of the `CHANGELOG.md`. But as I thought more about it and what else it could cover, the context it could provide, I also felt that this could clutter the change log a lot. To me, they are also for different audiences: the migration document focus is primarily external, while the change log audience is external but also internal, so it tends to be a compact list of changes, as opposed to an extensive document.

It's not one of the primary aims of the document and maybe it is a bit of wishful thinking from my side, but I also think of it as a communication tool between the Toolkit and the various clients, potentially providing more details on the reasoning behind changes, what's their goal and long term view, so the users can also be better prepared on their side (for example in the case of not depending anymore on Deliverfile or on env. vars).

> having a separate document would me a separate document to maintain and update, having to update our release scripts to ensure it has been updated when we are about to release a major version, and having people to know about the document when they adopt a new version so that they remember to check the separate file to know the instructions. So I kinda think I might have changed my mind and would maybe prefer to put this all in the `CHANGELOG.md` after.

To avoid extra effort from the beginning, we could also consider starting with a section in the change log, see how it goes, and then move to a separate document if we think it makes sense, as this would be less disruptive than the other way around 🤔 

> But this also depends on how much we want to write in such Migration Guide. If those migrations end up having a lot of things to write about and containing long paragraphs, then it's likely worth a separate document. If it's smaller content, maybe not?

Exactly my thinking. I'm still somewhat on the fence on this one. My first instinct is that there will always be more context and useful information to add and that this could clutter the change log, which should be a short bullet list of the main changes for each version.

> Also, other OSS repos [like Alamofire](https://github.com/Alamofire/Alamofire/tree/master/Documentation) have a separate document for each major version bump — as opposed to have a single document with sections for each version. Could be worth considering such an option. 

Ah, this can be an interesting approach. As you also said, to me it makes more sense for them as they have way more things to write about and a more complex migration procedure.

mokagio:
@iangmaia 

> To avoid extra effort from the beginning, we could also consider starting with a section in the change log, see how it goes, and then move to a separate document if we think it makes sense, as this would be less disruptive than the other way around 🤔

👍 Go for it

iangmaia:
Replaced by https://github.com/wordpress-mobile/release-toolkit/pull/490

PR #453, opened 2023-02-15
Release 7.0.0 into trunk

AliSoftware:
New version 7.0.0.

💭  🏷️  Be sure to create a GitHub Release and tag once this PR gets merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@
 
 ### Breaking Changes
 
+_None_
+
+### New Features
+
+_None_
+
+### Bug Fixes
+
+_None_
+
+### Internal Changes
+
+_None_
+
+## 7.0.0
+
+### Breaking Changes
+
 - Remove the `skip_glotpress` parameter from the `ios_bump_version_release` action [#443]
 
 ### New Features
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,20 +6,34 @@
...
 
 ### Bug Fixes
 
 _None_
 
 ### Internal Changes
 
+_None_
+
+## 7.0.0
```
mokagio, line 23: Out of scope: Interesting that the header the automation generates here do not include links to the `/release/<version>` page.

The template I suggested in the various Apple library repos had that link because I though it was handy. Evidently, it's not a must have feature, otherwise we would have felt the pain of not having it in this library's CHANGELOG.

In fact, the only memories I have of those changelog version links are of _fixing missing or wrongly copy-pasted instances_!

Maybe we'd be better served to omit it elsewhere, too. One seldom needs to open the release page when all the key information is in the CHANGELOG already. Given that, the maintenance savings over time should amply pay off for the extra friction of manually finding a release page every now and then, for example by typing `/release/<version>` in the browser URL bar.

What do you think?

mokagio (approved):

CHANGELOG.md
AliSoftware: Yep same reasoning as you, agree 💯 with pros vs cons. And I think that's probably the reason why I didn't bother add those links in the first place when introducing the toolkit's CHANGELOG… and why I stopped doing them for my personal OSS repos as well

PR #463, opened 2023-04-21
Use `git_branch_name_using_HEAD` instead of `git_branch`

oguzkocer:
## What does it do?

We have been using [git_branch](https://docs.fastlane.tools/actions/git_branch/#git_branch) helper action to check if we are on a release branch during complete code freeze & finalize release steps. This has been working as expected in our local environments. However, the return value of this helper can be modified by environment variables such as `GIT_BRANCH` or `BUILDKITE_BRANCH`. That means, if we switch the git branch during a CI step, it'll not return the value we are interested in.

There is a `Fastlane::Actions.git_branch_name_using_HEAD` helper method we can use, which is exactly what we expect. `git_branch` method actually [wraps its implementation around this helper](https://github.com/fastlane/fastlane/blob/08e50939de32aa411e1dc2d113aae8247b324dc2/fastlane/lib/fastlane/helper/git_helper.rb#L126-L128) and adds the extra capability of [modifying the return value with environment variables](https://github.com/fastlane/fastlane/blob/08e50939de32aa411e1dc2d113aae8247b324dc2/fastlane/lib/fastlane/helper/git_helper.rb#L125).

We have 2 different approaches to address this issue.

1. As implemented in 918027ba922db73e3d003bf3f507a881579c7c4b, we can directly use `Fastlane::Actions.git_branch_name_using_HEAD` everywhere we need. The main issue with that is, we need to add a disclaimer about why we are using this method instead of `git_branch`.
2. As implemented in 76696048fe0032369fad07d871336a19e4bb0e0e, we can add a method to `git_helper`. I like this better because we can have a longer explanation and contain it in a single place. I think this also makes it more likely that we'll use it instead of `git_branch` when we need to use it somewhere new - although neither option can guarantee that.


spencertransier (commented):
> I like this better because we can have a longer explanation and contain it in a single place. I think this also makes it more likely that we'll use it instead of git_branch when we need to use it somewhere new - although neither option can guarantee that.

I also like the https://github.com/wordpress-mobile/release-toolkit/commit/76696048fe0032369fad07d871336a19e4bb0e0e better. It feels cleaner and more self-explanatory, removing the need for the long comment each time the Fastlane helper is called directly. 

(Not approving since auto-merge is enabled and I figured I'd let folks with better understanding weigh in here if needed)

AliSoftware:
> I also like the https://github.com/wordpress-mobile/release-toolkit/commit/76696048fe0032369fad07d871336a19e4bb0e0e better. It feels cleaner and more self-explanatory, removing the need for the long comment each time the Fastlane helper is called directly.

Agreed.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_finalize_prechecks.rb
```diff
@@ -6,8 +6,10 @@ def self.run(params)
 
         require_relative '../../helper/ios/ios_version_helper'
         require_relative '../../helper/ios/ios_git_helper'
+        require_relative '../../helper/git_helper'
 
-        UI.user_error!('This is not a release branch. Abort.') unless other_action.git_branch.start_with?('release/')
+        current_branch = Fastlane::Actions.git_branch_name_using_HEAD
```
crazytonyli: Should `current_git_branch` be used here too?

crazytonyli (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_finalize_prechecks.rb
oguzkocer: 🤦  Yes it should be, thank you for catching that!

Addressed in f256ae5925ebe77a372d0fa330e4ef692a5c5d6b.

crazytonyli (approved):

PR #464, opened 2023-04-22
Update `octokit` to `5.6.1`

### LINKED ISSUE #412: Update OctoKit
We should update the OctoKit to the latest version – the 5.0 release just drops old ruby versions, which we've already done in our own 5.0 release, so it's not a breaking change.

spencertransier:
## What does it do?

This PR updates [Octokit](https://github.com/octokit/octokit.rb) to `5.6.1`. I tried updating it to the latest, `6.1.1`, but there is a dependency issue because Danger doesn't support upgrading Octokit past `5.x`. At least this gets us much further along. I also updated Danger to `9.3.0` because Danger `8.x` doesn't support Octokit above `4.x`. 

Rubocop was updated due to those dependency changes, which introduced a bunch of new violations. Nearly all of them made sense to me after researching them, but there are a few I have questions about. I'll comment on those down in the code. You can see Rubocop's offense list in [this Buildkite run](https://buildkite.com/automattic/release-toolkit/builds/972#0187a70b-732c-467b-a6d6-1264d472ccd1). 

Fixes #412 

- [X] Make sure you added an entry in [the `CHANGELOG.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md#trunk) to describe your changes under the approprioate existing `###` subsection of the existing `## Trunk` section.


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ _None_
 - Add "Mobile Secrets" to `configure_update` current branch message to clarify which repo it's referring to. [#455]
 - `buildkite_trigger_build` now prints the web URL of the newly scheduled build, to allow you to easily open it via cmd-click. [#460]
 - Add the branch information to the 'This is not a release branch' error that's thrown from complete code freeze lane. [#461]
+- Update `octokit` to `5.6.1` and `danger` to `9.3.0`. These are both major version changes, but as the dependency is internal-only, it shouldn't be a breaking change for clients. [#464]
```
spencertransier: Please correct me if I am wrong about this not being a breaking change for clients!

Dangerfile
```diff
@@ -32,7 +32,7 @@ end
 
 # Lint with Rubocop and report violations inline in GitHub
 github.dismiss_out_of_range_messages # This way, fixed violations should go
-renaming_map = (git.renamed_files || []).map { |e| [e[:before], e[:after]] }.to_h # when files are renamed, git.modified_files contains old name, not new one, so we need to do the convertion
+renaming_map = (git.renamed_files || []).to_h { |e| [e[:before], e[:after]] } # when files are renamed, git.modified_files contains old name, not new one, so we need to do the convertion
```
spencertransier, line 35: The [Rubocop docs say this](https://docs.rubocop.org/rubocop/cops_style.html#stylemaptohash) about this change: 

> This cop is unsafe, as it can produce false positives if the receiver is not an Enumerable.

I believe the usage here is an enumerable, but please let me know if I'm wrong so that this isn't changed incorrectly.

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -61,4 +60,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rubocop-require_tools', '~> 0.1.2'
   spec.add_development_dependency 'rubocop-rspec', '2.3.0'
   spec.add_development_dependency 'simplecov', '~> 0.16.1'
+  spec.metadata['rubygems_mfa_required'] = 'true'
```
spencertransier, line 63: This was automatically added, but sounds like a bigger change in our process that is outside the scope of this PR.

I did a bit of reading in the RubyGem docs here about this specific line: https://guides.rubygems.org/mfa-requirement-opt-in/

While it would increase security, I'm guessing that we'd need to have a discussion about how and if to implement MFA for our gem publishing.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_translations_action.rb
```diff
@@ -48,7 +48,7 @@ def self.available_options
             env_name: 'FL_DOWNLOAD_TRANSLATIONS_RES_DIR',
             description: "The path to the Android project's `res` dir (typically the `<project name>/src/main/res` directory) containing the `values-*` subdirs",
             type: String,
-            default_value: "#{ENV['PROJECT_NAME']}/src/main/res"
+            default_value: "#{ENV.fetch('PROJECT_NAME', nil)}/src/main/res"
```
spencertransier: From what I've read, `ENV.fetch` will produce an error if the environment variable isn't found, whereas `ENV[]` will just pass in `nil` and not produce an error. The default of `nil` was automatically added by Rubocop, but does that negate some of the benefit of using `ENV.fetch` if it still defaults to `nil` when the variable isn't found? 

Rubocop doesn't say a default is needed when using `ENV.fetch`, but I left it in there for each of the `ENV.fetch` corrections so I could ask about it here.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
```diff
@@ -45,10 +45,10 @@ def self.run(params)
       # @return [Hash<String, Array<String>>] The hash listing the keys to extract for each target file
       #
       def self.keys_list_per_target_file(original_files)
-        original_files.map do |original_file|
+        original_files.to_h do |original_file|
           keys = Fastlane::Helper::Ios::L10nHelper.read_strings_file_as_hash(path: original_file).keys
           [original_file, keys]
-        end.to_h
+        end
```
spencertransier, line 51: Same question here as https://github.com/wordpress-mobile/release-toolkit/pull/464/files#r1174292121

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -240,7 +240,7 @@ def self.download_from_glotpress(res_dir:, glotpress_project_url:, glotpress_fil
           attributes_to_copy = %w[formatted] # Attributes that we want to replicate into translated `string.xml` files
           orig_file = File.join(res_dir, 'values', 'strings.xml')
           orig_xml = File.open(orig_file) { |f| Nokogiri::XML(f, nil, Encoding::UTF_8.to_s) }
-          orig_attributes = orig_xml.xpath('//string').map { |tag| [tag['name'], tag.attributes.select { |k, _| attributes_to_copy.include?(k) }] }.to_h
+          orig_attributes = orig_xml.xpath('//string').to_h { |tag| [tag['name'], tag.attributes.select { |k, _| attributes_to_copy.include?(k) }] }
```
spencertransier, line 243: Same question here as https://github.com/wordpress-mobile/release-toolkit/pull/464/files#r1174292121

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_tools_path_helper.rb
```diff
@@ -10,7 +10,7 @@ class ToolsPathHelper
         attr_reader :android_home
 
         def initialize(sdk_root: nil)
-          @android_home = sdk_root || ENV['ANDROID_HOME'] || ENV['ANDROID_SDK_ROOT'] || ENV['ANDROID_SDK']
+          @android_home = sdk_root || ENV.fetch('ANDROID_HOME', nil) || ENV.fetch('ANDROID_SDK_ROOT', nil) || ENV.fetch('ANDROID_SDK', nil)
```
spencertransier: Rubocop's automatic change here was 

`@android_home = sdk_root || ENV['ANDROID_HOME'] || ENV['ANDROID_SDK_ROOT'] || ENV.fetch('ANDROID_SDK', nil)` 

I wasn't sure if it just missed the first two `ENV[]` instances, so I manually added `ENV.fetch` to those two. Was that the correct thing to do?

Dangerfile
AliSoftware: TIL that `Enumerable#to_h` could take a block, and thus be used directly instead of `Enumerable#map` followed by `#to_h`. Nice.

And yes, here `git.renamed_files || [])` is an `Enumerable` so we should be good.

fastlane-plugin-wpmreleasetoolkit.gemspec
AliSoftware: Yeah, especially since our gem publishing is automated by CI every time we create a new tag on this repo, and I doubt this process would work well with a OTP being asked during that process 😓

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_translations_action.rb
AliSoftware: Yes, I believe that `ENV.fetch('PROJECT_NAME')` would crash is the env var was not set, while `ENV.fetch('PROJECT_NAME', nil)` would return nil if the env var was not set… just like `ENV['PROJECT_NAME']` would. So in this particular condition, I'm not sure I see the benefit of `ENV.fetch('PROJECT_NAME', nil)` over `ENV['PROJECT_NAME']`.

If anything, what would probably make sense if `PROJECT_NAME` was not set, would be to use the `./` current directory as a fallback instead. So I'd instead maybe suggest `ENV.fetch('PROJECT_NAME', '.')` instead?

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
AliSoftware: Same, `original_files` seems to be an `Enumerable` to me there so we should be ok.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
AliSoftware: Same. I'm pretty sure the return value of `xpath` is an `Enumerable`, so 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_tools_path_helper.rb
AliSoftware: Yeah seems like a strange change from rubocop. Especially since `ENV.fetch(key, nil)` is exactly the same as `ENV[key]` (see [doc](https://apidock.com/ruby/Hash/fetch))…

If it were me I'd just keep `ENV[…]` for all the cases in that line. But rubocop will then likely complain so…
Another option is to rewrite this using something like:

```ruby
@android_home = sdk_root || ENV.values_at('ANDROID_HOME', 'ANDROID_SDK_ROOT', 'ANDROID_SDK'].compact.first
```

But I'm not sure that better… 🤷

CHANGELOG.md
AliSoftware: For Danger, this is correct, because the version of `danger` you changed is in the `Gemfile` of this repo, so this is for the gems that are used to run this repo itself. For example, we use Danger in the `release-toolkit` repo itself to warn about missing CHANGELOG entry in the `release-toolkit`'s own PRs and such.

It would only be a change impacting clients if you modified the `.gemspec`, which defines the dependencies that are needed for this gem when it's included in clients. Danger is not a gem that is used but the implementation of the `release-toolkit` itself, but by the CI testing the release-toolkit's own PRs.

---

For Octokit otoh, you did change the version to a major version in the `.gemspec` itself. So this means that this will impact clients. Indeed, `octokit` is part of the gems needed for `release-toolkit`'s own implementation, and for the code of `release-toolkit` itself to run when integrated in clients. In particular, many of our `release-toolkit` actions' implementations use `octokit` to communicate with the GitHub API (e.g. `create_new_milestone_action` and such).

In this context, the update of Octokit to a new major version is NOT an internal change only but is a Breaking Change for clients. Depending on the breaking changes between version 4.x and 5.x of octokit, we also need to make sure that all the changes listed in Octokit's breaking changes don't break our code, and that our code and all the `release-toolkit` actions which are using `octokit` in their implementation don't use old methods and code that was valid in Octokit 4 but is not valid in Octokit 5 due to a breaking change in Octokit.

AliSoftware (changes requested):
Most changes look ok to me, except that the Octokit version bump, being a dependency of the gem itself (listed in the `.gemspec`), should be listed as a Breaking Change for clients of the `release-toolkit`.

And we also need to double check that all the changes between Octokit `4.x` and Octokit `5.x` don't make any of our usages of Octokit in the `release-toolkit`'s own code… break (e.g if an API of Octokit that we were using has changed or been removed since). Which I didn't check myself personally, but would still need to be audited by someone to confirm first.

fastlane-plugin-wpmreleasetoolkit.gemspec
spencertransier: I added a Rubocop exception for this until we decide whether to pursue it further. https://github.com/wordpress-mobile/release-toolkit/pull/464/commits/cdfa2baf90f94d34ec32b83d029020bc4715f9c0

Dangerfile
spencertransier: Thank you for double-checking my assumption 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_extract_keys_from_strings_files.rb
spencertransier: Thank you for double-checking my assumption 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
spencertransier: Thank you for double-checking my assumption 👍

CHANGELOG.md
spencertransier: Responding here to https://github.com/wordpress-mobile/release-toolkit/pull/464#discussion_r1175610504. 

@AliSoftware The only [breaking change in `5.0.0`](https://github.com/octokit/octokit.rb/releases/tag/v5.0.0) was dropping all versions of Ruby older than 2.7. Since we're already at 2.7.4 everywhere, this shouldn't actually break anything in our client implementations of the Release Toolkit.

spencertransier: I moved the `octokit` upgrade to the "Breaking Changes" section: https://github.com/wordpress-mobile/release-toolkit/pull/464/commits/d2a1e86604e42d71d344cb0181c52690e54e827d

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_translations_action.rb
spencertransier: I decided to keep these as `ENV.fetch`'s instead of reverting to `ENV[]` or setting defaults (at least for now). My thinking is that if one of these environment variables isn't set, we wouldn't necessarily want a default value to be swapped in. That seems like it could lead to unexpected changes when running the toolkit actions.

We can also set a default other than `nil` or remove the default altogether in the future if we think that's the correct decision for a particular environment variable. 

I would say that maybe we should remove the default and have the commands crash if a variable isn't set, but I think that's a discussion worth having outside this PR.

spencertransier: @AliSoftware what do you think about that reasoning?

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_tools_path_helper.rb
spencertransier: Same as https://github.com/wordpress-mobile/release-toolkit/pull/464#discussion_r1175891328 and the other instances of `ENV.fetch` in this PR. I think that may be a better approach than needing to add a new Rubocop exception.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_translations_action.rb
AliSoftware: Problem is the default value is evaluated even if you provide one when the action is parsed by fastlane. Simply asking fastlane for the documentation of the action (`fastlane action name_of_an_action`) would crash if one of the env var was not set.

spencertransier: @AliSoftware But not if we leave the defaults set as `nil`, right? It should act how `ENV[]` currently acts and just do nothing. 

When I ran these lines: 
```
lane :test_env do    
    puts ENV['EMPTY_CONSTANT_1']
    puts ENV.fetch('EMPTY_CONSTANT_2', nil)
    puts ENV.fetch('EMPTY_CONSTANT_3')
  end
```

It didn't error until hitting the `EMPTY_CONSTANT_3` line. So if we do have an env var missing but the default value is set as `nil`, it should perform exactly like `ENV[]` works now if an env var is missing.

spencertransier: Sorry if I'm going around in circles 😄

AliSoftware: Sorry, by "keeping the default value", I meant "keeping the `default_value` parameter of the `ConfigItem`", not "keeping the fallback value of `fetch`".

> I decided to keep these as ENV.fetch's instead of reverting to ENV[] or setting defaults (at least for now). My thinking is that if one of these environment variables isn't set, we wouldn't necessarily want a default value to be swapped in.

If you use `ConfigItem.new(…, default_value: "#{ENV.fetch('PROJECT_NAME')}/foo/bar"), …)` then the action will crash even when you just ask fastlane to print its documentation using `fastlane action android_download_translations`, because it will try to print the `default_value` of the action parameter as part of the documentation, and since `ENV.fetch` doesn't have a fallback (2nd parameter) it will crash if the env var doesn't exist. We definitively don't want just asking the lane documentation to crash.

So TL;DR: `ENV.fetch(key)` without a fallback value is not something we should use inside a `default_value` of a `ConfigItem`.

Even in other contexts, where we'd consider the env var to be required, I feel it would be better to fail more gracefully than with a generic crash, like use a comprehensive `UI.user_error!("You need to define a value for the 'PROJECT_NAME' env var")` than having `ENV.fetch(key)` crash with an obscure message that would make it look like a bug in the toolkit.

mokagio:
> except that the Octokit version bump, being a dependency of the gem itself (listed in the .gemspec), should be listed as a Breaking Change for clients of the release-toolkit.

@AliSoftware would listing it as a breaking change also require us to publish a new major build? Unless we expose Octokit via release_toolkit in a way that breaks with this upgrade, my understanding is that a new major version should not be necessary.

Quoting from the [SemVer guidelines](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api):

> **What should I do if I update my own dependencies without changing the public API?**
>
> That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. We would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.

AliSoftware:
> @AliSoftware would listing it as a breaking change also require us to publish a new major build? Unless we expose Octokit via release_toolkit in a way that breaks with this upgrade, my understanding is that a new major version should not be necessary.

Very good point, my bad, you're right indeed. Let's move it back to internal changes 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_translations_action.rb
spencertransier: @AliSoftware 

> `ENV.fetch(key)` without a fallback value is not something we should use inside a `default_value` of a `ConfigItem`

Ah okay, that makes sense. Thank you for helping me understand! 

I rolled back the `ENV.fetch` changes to `ENV[]` and added a Rubocop exception. Hopefully this will all be moot soon anyway as we work towards not using as many env vars in these actions 🙂

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ _None_
 - Add "Mobile Secrets" to `configure_update` current branch message to clarify which repo it's referring to. [#455]
 - `buildkite_trigger_build` now prints the web URL of the newly scheduled build, to allow you to easily open it via cmd-click. [#460]
 - Add the branch information to the 'This is not a release branch' error that's thrown from complete code freeze lane. [#461]
+- Update `octokit` to `5.6.1` This is a major version bump from version `4.18`, but is not a breaking change for the Release Toolkit because it doesn't change any public APIs for clients. [#464]
+- Update `danger` to `9.3.0`. This is an internal-only change and is not a breaking change for clients. [#464]
```
AliSoftware, line 25: 👌

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -14,8 +14,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.7'
 
-  spec.files         = Dir['lib/**/*'] + %w[README.md LICENSE]
-  spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
+  spec.files = Dir['lib/**/*'] + %w[README.md LICENSE]
```
AliSoftware, line 17: Interesting, not sure why rubocop made us remove `spec.test_files` here… is it because that gemspec attribute has now been a deprecated by rubygems or something? 🤔 
(I indeed don't see `test_files` documented in https://guides.rubygems.org/specification-reference/ 🤔 )

AliSoftware (approved):

fastlane-plugin-wpmreleasetoolkit.gemspec
spencertransier: @AliSoftware Yep, that's it. This is the rule that removed the line: 

https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecdeprecatedattributeassignment

PR #465, opened 2023-04-23
Update `buildkite-test_collector` to `2.2.0`

spencertransier:
## What does it do?

This PR replaces the `rspec-buildkite-analytics` gem with `buildkite-test_collector` (it's the same gem but Buildkite renamed it a while ago) and updates it to `2.2.0`. This is an internal-only change and is not a breaking change for clients.

I followed the breaking change update steps from the `1.0.0` and `2.1.0` releases. 

I'm merging into `update/octokit` because there are a lot of dependency changes there and I thought this would be a good way to avoid conflicts on those. 

## How to Test

You can verify that the Test Analytics report for this PR successfully shows up in Buildkite: https://buildkite.com/organizations/automattic/analytics/suites/release-toolkit/runs/85caafb1-865e-4eaa-8a1d-c720a2957c43

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,4 +9,4 @@ gem 'codecov', require: false
 gem 'webmock', require: false
 gem 'yard'
 
-gem 'buildkite-test_collector', '~> 2.1'
+gem 'buildkite-test_collector', '~> 2.2'
--- a/Gemfile
+++ b/Gemfile
@@ -9,4 +9,4 @@ gem 'codecov', require: false
 gem 'webmock', require: false
 gem 'yard'
 
-gem 'rspec-buildkite-analytics', '~> 0.8.1'
+gem 'buildkite-test_collector', '~> 2.1'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ _None_
 - `buildkite_trigger_build` now prints the web URL of the newly scheduled build, to allow you to easily open it via cmd-click. [#460]
 - Add the branch information to the 'This is not a release branch' error that's thrown from complete code freeze lane. [#461]
 - Update `octokit` to `5.6.1` and `danger` to `9.3.0`. These are both major version changes, but as the dependency is internal-only, it shouldn't be a breaking change for clients. [#464]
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -9,4 +9,4 @@ gem 'codecov', require: false
 gem 'webmock', require: false
 gem 'yard'
 
-gem 'rspec-buildkite-analytics', '~> 0.8.1'
+gem 'buildkite-test_collector', '~> 2.1'
```
oguzkocer: Would it be worth changing this to `2.2` since the gem in the lock file is ending up with `2.2`?

oguzkocer (approved):
The changes match the [documentation](https://buildkite.com/docs/test-analytics/ruby-collectors#rspec-collector) and the test analytics report is showing up for the build.

> I'm merging into update/octokit because there are a lot of dependency changes there and I thought this would be a good way to avoid conflicts on those.

Unless there is a rush to get this merged, I think it'd be better to wait for the `update/octokit` branch to get merged to `trunk` first. This will automatically update the base branch for this PR to `trunk`. This is so that the PR for `update/octokit` won't have any irrelevant code changes in it.

Another merge strategy in these situations - although I don't think it's necessary in this one - is to create a temporary branch such as `update/octokit-and-buildkite-test_collector` and merge both PRs into it. And then, once everything is ready a new PR can merge the `update/octokit-and-buildkite-test_collector` to `trunk`. This strategy is generally useful in cases where there are many changes in different PRs that all go together and wouldn't make much sense to merge alone. It closely resembles a feature flag and we used it often when we worked on projects where there were a lot of breaking changes and we needed to test all of them together before merging to `trunk`.

Gemfile
spencertransier: Good point, updated in https://github.com/wordpress-mobile/release-toolkit/pull/465/commits/a06257eed46daa53078e90377021ca7240b5dfcf

spencertransier:
@oguzkocer Thank you for the merge tips! I will wait to merge this until the `octokit` PR is merged to `trunk`.

PR #466, opened 2023-04-24
Add `ios_register_ci_machine` action

crazytonyli:
## What does it do?

Register the current macOS machine in the Apple Developer Portal, using [the fastlane built-in `register_device` action](https://docs.fastlane.tools/actions/register_device/).

In our macOS image building process, we have a manual step to register a new VM image's hardware uuid. This action is created to replace that manual step.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_register_ci_machine.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_register_ci_machine.rb
@@ -1,0 +1,71 @@
+module Fastlane
+  module Actions
+    class IosRegisterCiMachineAction < Action
+      def self.run(params)
+        return unless other_action.is_ci
+
+        require 'socket'
+        require 'open3'
+
+        hostname = Socket.gethostname
+                         .delete_suffix('.local')
+                         .delete_suffix('.shared')
+
+        device_uuid, = Open3.capture2("system_profiler SPHardwareDataType | awk '/Hardware UUID/ { print $NF }'")
+        device_uuid.strip!
+        raise 'Failed to get device UUID' if device_uuid.nil?
+
+        UI.message "Registering device #{device_uuid} for CI image #{hostname}"
+
+        other_action.register_device(
+          name: "CI - #{hostname}",
+          udid: device_uuid,
+          platform: 'mac',
+          api_key_path: params[:api_key_path],
+          team_id: params[:team_id]
+        )
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        'Register the current CI machine as a new device in the Apple Developer Portal'
+      end
+
+      def self.details
+        <<~DETAILS
+          Register the current CI machine as a new device in the Apple Developer Portal.
+
+          As the action name suggests, this action is only meant to be run on a CI machine and does nothing if CI environment is not detected (via the `is_ci` action).
+        DETAILS
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :api_key_path,
+            description: 'The path of the App Store Connect API key file',
+            type: String,
+            optional: false
+          ),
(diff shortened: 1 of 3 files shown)


AliSoftware (commented):
I am not convinced that this is the place to implement such an action…

`release-toolkit`'s primary focus is to be used by client repos for iOS and Android and Mac apps. But this action should not be used from the `Fastfile` of one of our app, not even from the `Fastfile` of our Mac apps like DayOne or Simplenote, but instead used by `buildkite-ci` itself, only once, when creating a a VM.

Maybe this would make more sense to implement this as a `Fastfile` and lane in `buildkite-ci` itself directly? Otherwise it will get confusing why we need the whole `release-toolkit` (intended to manage recurring actions for a typical **app release** scenario) for creating a CI VM, and also could make it accidentally tempting to call this action from a `Fastfile` of a Mac app's repo.

_Except if that was actually the plan and this action was all supposed to be called from the Fastfiles or theDayOne Mac and Simplenote Mac repos, every time then run a `step` in their own `pipeline.yml` and need the host machine be registered? But if so, that would not be optimal as that would try to register the Mac every time the pipeline runs, instead of only once when the VM image is created…_

crazytonyli:
Oh good point. I thought this plugin can be used as a library to share some actions among our apps. I can certainly move this automation somewhere else.

> Except if that was actually the plan and this action was all supposed to be called from the Fastfiles or theDayOne Mac and Simplenote Mac repos, [...] that would not be optimal as that would try to register the Mac every time the pipeline runs, instead of only once when the VM image is created…

Yes, my intention is calling this action from the app pipelines. I don't intend to call this action during creating a VM. The idea is, the app pipeline should ensure the machine it runs on is registered (by calling this action in appropriate fastlane lanes), so that the actions the fastlane lane carries out latter (building the mac app, or running the mac app unit tests) can be performed successfully.

This registration is only required once, and is ideally only performed once. But I don't see a great risk in calling the `register_device` action multiple times, since it finishes pretty fast (in just a few seconds to make a couple of App Store Connect API calls) and there is no side-effect.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_register_ci_machine.rb
```diff
@@ -0,0 +1,71 @@
...
+        return unless other_action.is_ci
+
+        require 'socket'
+        require 'open3'
+
+        hostname = Socket.gethostname
+                         .delete_suffix('.local')
+                         .delete_suffix('.shared')
+
+        device_uuid, = Open3.capture2("system_profiler SPHardwareDataType | awk '/Hardware UUID/ { print $NF }'")
```
AliSoftware, line 14: Would probably be safer to use `-json` to have `system_profiler` output JSON instead of plain text, and parse that output with `JSON.parse`, rather than relying on `awk` and textual parsing?

AliSoftware:
Ah I see. If the plan was to indeed let each app's pipeline register the VM their build step is running on every time, then maybe this was the right place to implement this.

For this kind of CI-specific command (i.e. more tied to the CI setup and doing stuff with our CI machines, than tied to the app release scenario itself), `a8c-ci-toolkit-buildkite-plugin` could in theory be a better place… but that's `bash`, while you need `fastlane`'s `register_device` action to implement this… so maybe it wasn't a bad idea to implement it here after all.

---

Another concern I had at first with having this in the `release-toolkit` is we have to make sure this action would only be called as part of a Buildkite pipeline, and never be called locally, otherwise the developer's Mac would get registered in ASC accidentally (using the assumptions you have in this action that this was a CI machine, like the name given to the device). But… that was before I realized that you already added a `return unless other_action.is_ci?` check at the very top of the action, so we'd be good on that front after all.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_register_ci_machine.rb
crazytonyli: TIL!

PR #467, opened 2023-04-24
Add `android_generate_apk_from_aab` action

rynaardb:
## What does it do?

Adds a new `android_generate_apk_from_aab` action to generate an APK from a given AAB.

This action can be useful in cases where you already have an AAB and just need the APK from it without having to build the APK from the source.


lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb, line 1: Although the action is very much Android-specific, I've decided to add the action under a new `bundletool` folder instead of the existing `android` folder to make it easier to navigate the project and to avoid a long action name with the added `android_` prefix and with the assumption that we will add additional `bundletool`-related actions in the future.

If we feel strongly about having it under `android` and prefixed with `android_`, let me know then I'll happily change it.

AliSoftware: The `android` namespace is per convention with out the project is structured though. I'd advocate against creating a `bundletool/` subfolder here that would not respect the rest of the nomenclature.

Even though I agree that our nomenclature and folder structure isn't the best at the moment and has accumulated a bit of legacy, such a change in the structure should be done during a larger refactor to reorganize the whole actions in a better folder structure, and especially one that follows the conventions of a Ruby gem (which are similar conventions to the ones used by Android/Java packages, with one folder per module/namespace etc…), which should be done for all actions and in a consistent way based on modules/namespaces.

---

As for the name of the action, imho the use of `bundletool` for this action is an implementation detail. So I'd personally drop that from the name.

Also, the fact that it can generate a signed APK should be optional and dropped from the name as well, so that if we don't want a signed APK, or simply can't create one because we don't have release signing keys. For example our Jetpack app used Managed Code Signing (see paqN3M-Hk-p2), which means only Google has the release signing key and we don't; but we are still interested in generating potentially non-signed APKs for prototype builds and such).

As a result, I'd suggest having a name like `android_generate_univeral_apk` or `android_generate_apk_from_aab` instead.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        sh('bundletool', 'build-apks',
+           '--mode', 'universal',
+           '--bundle', aab_path,
+           '--output-format', 'DIRECTORY',
+           '--output', apk_output_path,
+           '--ks', keystore_path,
+           '--ks-pass', keystore_password,
+           '--ks-key-alias', keystore_key_alias,
+           '--key-pass', signing_key_password)
```
AliSoftware: It would be nice if those four code-signing-related parameters were optional, so that it would also be possible to generate non-signed APKs as well as signed APKs depending on the needs (and if we have the signing keys at all, like for Jetpack or any other app using "Google Managed Code Signing").

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
```diff
@@ -0,0 +1,104 @@
+module Fastlane
+  module Actions
+    class BundletoolGenerateUniversalSignedApkAction < Action
+      def self.run(params)
+        aab_path = params[:aab_path]
```
AliSoftware: Given the way we typically generate the `.aab` to be used here is by calling the `gradle` action in our Fastfile first, it would be nice to default [to `SharedValues::GRADLE_AAB_OUTPUT_PATH`](https://docs.fastlane.tools/actions/gradle/#lane-variables) here if the user didn't provide an `aab_path` explicitly during invocation, so that it uses the value provided by the previous invocation of the `gradle` action if any,

You can take some inspiration from [the official `upload_to_play_store` fastlane action here](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/upload_to_play_store.rb#L19-L29) which uses a similar trick so that when `upload_to_play_store` is called just after calling `gradle`, you don't have to pass a value for those parameters explicitly and instead it will use the output values of the previous `gradle` command.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
```diff
@@ -0,0 +1,104 @@
...
+            type: String,
+            optional: false,
+            default_value: nil
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :keystore_path,
+            env_name: 'BUNDLETOOL_KEYSTORE_PATH',
+            description: 'The path to the keystore file',
+            type: String,
+            optional: false,
```
AliSoftware: imho this (and the other remaining `ConfigItems`) should be optional and we should allow the action to generate a non-signed APK if the caller doesn't provide any of the `keystore_*` parameters.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
```diff
@@ -0,0 +1,104 @@
...
+        rescue StandardError
+          UI.user_error!('bundletool is not installed. Please install it using the instructions at https://developer.android.com/studio/command-line/bundletool')
+          raise
+        end
+
+        sh('bundletool', 'build-apks',
+           '--mode', 'universal',
+           '--bundle', aab_path,
+           '--output-format', 'DIRECTORY',
+           '--output', apk_output_path,
```
AliSoftware: I just tested the command, and the `--output` provided here is supposed to be a path to a directory, not the APK file.

For example, I tried the command on an `.aab` and passing `--output-format DIRECTORY --output foo/bar.apk` and it created the folders `foo` and `foo/bar.apk`, and generated a `foo/bar.apk/universal.apk` inside it.

We need more logic in this action to handle this
 - We should create a temp dir (`mktmpdir`) and make `bundletool` generate the out directory structure in this temp dir
 - Then, we move the `#{tmpdir}/universal.apk` file to the destination provided by the caller in `apk_output_path`
 - To go even further, if the `apk_output_path` is an existing `Dir.is_directory?`, or does not exist yet but the string ends with a `/`, then use `File.join(apk_output_path, File.basename(aab_path, '.aab) + '.apk')` as the destination to move the generated `universal.apk` to once `bundletool` is done. And otherwise, we consider that the `apk_output_path` is a path to a file and not a dir, and we use it directly as the destination to rename/`mv` the generated `.apk` to.

In any case, we should not keep the code as is because the `.apk` generated by `bundletool` would end up always be named `universal.apk` and be in a **subdirectory** provided by the `apk_output_path`, even if from a user of the action, it's likely that people will expect `apk_output_path` to be the path to the final `.apk`, and that will likely create a folder with extension `.apk` instead of getting the intended behavior.

AliSoftware (changes requested):

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :aab_path,
+            env_name: 'BUNDLETOOL_AAB_PATH',
+            description: 'The path to the AAB file',
+            type: String,
+            optional: false,
+            default_value: nil
```
AliSoftware: Nitpick: you could also add a `verify_block` here that would check that the path provided exists and is a regular file (and not a folder). You can get some inspiration from [this](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb#L103) for example.

Same for most of the other `ConfigItems` that expect an existing file (like the ones for `keystore_*`)—though those should have their `verify_block` accept either an existing file… or a `nil` value, if we want to allow those to be optional (see other comment)

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
```diff
@@ -0,0 +1,104 @@
...
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :apk_output_path,
+            env_name: 'BUNDLETOOL_APK_OUTPUT_PATH',
+            description: 'The path to the output directory where the APK will be generated',
```
AliSoftware: Should this be the path to the output directory, or the path to the generated `.apk` (including the `.apk` name)? By just reading the parameter name `apk_output_path` I'd expect it to potentially be the path to the `.apk` itself, and I think it should support that as well (see other comment above)

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: I guess the name `apk_output_path` might be a bit misleading in this context. I'll rename it to `apk_output_dir_path` instead to avoid confusion.

As you've already noted, the APK generated by `bundletool` will always be named `universal.apk`, so it's really just about the path to the directory where the APK will be generated. Maybe I should also mention it in the description of the parameter's `ConfigItem`.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Good call! I was trying to think of a use case where we would want to generate a non-signed APK but didn't think of apps using Google's Managed Code Signing. So yes, making these optional makes a lot of sense.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Sure, no problem. I'll move the action to the `android` namespace and rename it accordingly.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
AliSoftware: Given that we'll likely need to use the generated `.apk` (and thus need its path) in the following code of the lane calling that action (e.g. to upload it as artifact or upload it to Google Play or attach it to the GitHub Release, etc), we need to have this name be explicit at some point so that the code following this action knows what path to use to reference the generated file.

imho our options are:

1. Keep the path at which `bundletool` generates the `.apk` and the fact that it's always named `universal.apk`, and document it here in the action's `ConfigItem` description.
   - I'm not a fan of this because that means that if we wanted to e.g. rely on the `artifacts_paths: ['**/*.aab', '**/*.apk']` key in our pipeline to upload all AABs and APKs, the artefacts would appear as `app-base-release.aab` but `mydir/universal.apk`, making it inconsistent and also not clear what flavor and build type the APK was from.
   - Also that means we won't have a choice in the name of the generated APK
2. Make the `apk_output_path` be the path of the APK to be generated, not the path to the folder.
   - That simply means have `bundletool` use a temp dir for `--output #{tmpdir}`, then call `FileUtils.mv("#{tmpdir}/universal.apk", apk_output_path)` to move it to the expected destination path
   - Feels more flexible, and allows the caller to choose the exact path for the generated APK instead of having the file name forced on them.
3. Be even more flexible and detect if the `apk_output_path` is an existing folder, and if so use the same basename as the aab for the output apk
   - i.e. same as option 2, but just do something like `apk_output_path = File.join(apk_output_path, "#{File.basename(aab_path, '.aab')}.apk") if Dir.directory?(apk_output_path)` to adjust the `apk_output_path` before doing the final `FileUtils.mv`.

My preference would be for option 3, which is the most flexible all while not being that complicated to implement.
_(And TBH, if we went with option 1, I'm pretty sure many of us, for the various app repos we have, will be tempted to manually implement what option 2 or 3 in the calling `Fastfile` of each repo directly. So we might as well DRY this behavior directly into the action, right?)_

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+      def self.generate_command(aab_file_path, apk_output_file_path, keystore_path, keystore_password, keystore_key_alias, signing_key_password)
+        command = "bundletool build-apks --mode universal --bundle #{aab_file_path} --output-format DIRECTORY --output #{apk_output_file_path} "
+        code_sign_arguments = "--ks #{keystore_path} --ks-pass #{keystore_password} --ks-key-alias #{keystore_key_alias} --key-pass #{signing_key_password} "
+        move_and_cleanup_command = "&& mv #{apk_output_file_path}/universal.apk #{apk_output_file_path}_tmp && rm -rf #{apk_output_file_path} && mv #{apk_output_file_path}_tmp #{apk_output_file_path}"
+
+        # Attempt to code sign the APK if a keystore_path parameter is specified
+        command += code_sign_arguments unless keystore_path.nil?
+
+        # Move and rename the universal.apk file to the specified output path and cleanup the directory created by bundletool
+        command += move_and_cleanup_command
```
rynaardb: I've tried several approaches to deal with how `bundletool` handles the `--output` option and how it always defaults to the `universal.apk` filename, but I ran into all sorts of limitations and unnecessary complexity. Appending the move and cleanup commands as I've done here is the least complicated approach, especially with how I've structured the tests.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Addressed in 16e5050557dcad2312d1b7db19c7b1ddfa71119b

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Addressed in 16e5050557dcad2312d1b7db19c7b1ddfa71119b

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Thanks, that's a great suggestion! I've implemented it in 16e5050557dcad2312d1b7db19c7b1ddfa71119b

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Addressed in 16e5050557dcad2312d1b7db19c7b1ddfa71119b

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Please see my comment [here](https://github.com/wordpress-mobile/release-toolkit/pull/467#discussion_r1184174852). It deals with the `--output` option and how `bundletool` handles it. I've basically implemented what you described in option 3 above but using a slightly different approach.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Addressed in 16e5050557dcad2312d1b7db19c7b1ddfa71119b

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
rynaardb: Addressed in 16e5050557dcad2312d1b7db19c7b1ddfa71119b

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
+module Fastlane
+  module Actions
+    class AndroidGenerateApkFromAabAction < Action
+      def self.generate_command(aab_file_path, apk_output_file_path, keystore_path, keystore_password, keystore_key_alias, signing_key_password)
+        command = "bundletool build-apks --mode universal --bundle #{aab_file_path} --output-format DIRECTORY --output #{apk_output_file_path} "
```
AliSoftware: Using a string to build the command that you'll invoke via `sh` is generally bad practice and should be avoided if possible to use explicitly separate parameters on the `sh` call instead.

This is because when you pass a command as a string to `sh(…)`, that means that it will involve shell interpolation and interpretation (i.e. having a shell interpret the string as a command and parameters, interpolating variables and spliting the command via spaces and handling piping and all) which leads to a lot of potential side effects (like issues with file paths containing spaces or special characters).

I'd instead recommend to build the list of arguments as an `Array`, and then pass that array when invoking `sh('bundletool', *arguments)` down below. See suggestion in https://github.com/wordpress-mobile/release-toolkit/pull/467#discussion_r1185132848

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+        code_sign_arguments = "--ks #{keystore_path} --ks-pass #{keystore_password} --ks-key-alias #{keystore_key_alias} --key-pass #{signing_key_password} "
```
AliSoftware: I see there's no protection or nice-error-handling for the case where the caller would provide a `keystore_path` but not provide all of the other arguments (e.g. say the caller provides all but the `keystore_key_alias`).

This will likely lead to `bundletool` erroring with a unpleasant stacktrace and error, instead of us printing a nicer error message, or instead of us only passing to `bundletool` the parameters that are passed to the action by the user.

---

Unless if it's valid to pass a keystore path without a keystore password/alias/key password? But in any case it would make more sense to me to add each individual parameter if they're passed, rather than making all of those four conditioned on a single one?

Something like this (assuming you follow [my suggestion from this other comment](https://github.com/wordpress-mobile/release-toolkit/pull/467#discussion_r1185132848)):
```ruby
UI.user_error!('You provided key/keystore parameters without providing a keystore path to apply them to') if keystore_path.nil? && [keystore_password, keystore_key_alias, signing_key_password].any? { |v| !v.nil? }

arguments += ['--ks', keystore_path] unless keystore_path.nil?
arguments += ['--ks-pass', keystore_password] unless keystore_password.nil?
arguments += ['--ks-key-alias', keystore_key_alias] unless keystore_key_alias.nil?
arguments += ['--key-pass', signing_key_password] unless signing_key_password.nil?
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+        move_and_cleanup_command = "&& mv #{apk_output_file_path}/universal.apk #{apk_output_file_path}_tmp && rm -rf #{apk_output_file_path} && mv #{apk_output_file_path}_tmp #{apk_output_file_path}"
```
AliSoftware: If you prefer to go that route (i.e. using `mv` and a temp file) then I'd suggest to do that with native Ruby methods, namely `FileUtils.mv` and friends, rather than doing that using `sh` (which would involve spawning a shell, interpreting the command—which would fail to work as you expect here if that path have spaces, by the way— and running the move via terminal command, instead of doing the move directly from Ruby).

That being said, I'm not a fan of this solution, especially because that `rm -rf #{apk_output_file_path}` can delete an entire existing **directory** where I might have lots of things in it. For example say that the user accidentally pass a folder path (e.g. `artifacts/`) when calling the action (instead of the full path to the final `.apk` like you expect in the first place), then your `rm -rf` command will remove the entire `artifacts/` folder then move the `universal.apk` into a file named `artifacts`.

Sure, the caller is expected to provide a path to the APK not to the folder, but the fact that they misused the action led to destructive action, which is not good. Besides, we should also have some check to protect us from the user passing a folder rather than the APK path in the first place if that's how we want the action to work.

---

To solve all this, see my suggestion in https://github.com/wordpress-mobile/release-toolkit/pull/467#discussion_r1185132848

spec/android_generate_apk_from_aab_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    allow(File).to receive(:file?).with(apk_output_file_path).and_return('mocked file data')
+    allow(File).to receive(:file?).with(keystore_path).and_return('mocked file data')
+  end
+
+  let(:aab_file_path) { 'path/to/app.aab' }
+  let(:apk_output_file_path) { 'path/to/app.apk' }
+  let(:keystore_path) { 'path/to/keystore' }
+  let(:keystore_password) { 'keystore_password' }
+  let(:keystore_key_alias) { 'keystore_key_alias' }
+  let(:signing_key_password) { 'signing_key_password' }
```
AliSoftware: To make sure you can handle tricky cases, you might want to add some spaces and special characters to those dummy paths and values, so that your tests can validate that they are properly escaped (if you pass a string to `sh` that will go thru shell interpolation, which I definitively recommend **against** anyway) and handled too.

spec/android_generate_apk_from_aab_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    code_sign_arguments = "--ks #{keystore_path} --ks-pass #{keystore_password} --ks-key-alias #{keystore_key_alias} --key-pass #{signing_key_password} "
+    move_and_cleanup_command = "&& mv #{apk_output_file_path}/universal.apk #{apk_output_file_path}_tmp && rm -rf #{apk_output_file_path} && mv #{apk_output_file_path}_tmp #{apk_output_file_path}"
+
+    # Append the code signing arguments
+    command += code_sign_arguments unless keystore_path.nil?
+
+    # Append the move and cleanup command
+    command += move_and_cleanup_command
+    return command
+  end
```
AliSoftware: The fact that this is a copy/paste of the actual implementation means that you're basically testing nothing. You're testing that the `generate_command` from your `AndroidGenerateApkFromAabAction` implementation will return the same string than this `generate_command` which is mostly a copy/paste of the same code. So this is like testing that `1+1 == 1+1`.

If you re-implement the logic using `sh('bundletool', *arguments)`—instead of `sh(command_string)`—as I already suggested above, you can instead write your tests in a similar fashion that how we test it [here](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/firebase_account_spec.rb) for example.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
AliSoftware: That approach still has issues, not only with its potential to be destructive of entire folders on your filesystem if the user passes the wrong output path, but also not working if paths have spaces.

I'd suggest to instead generate the APK in a temp folder and move it to destination once done, which imho is simpler, avoids issues with spaces and special characters in paths if you do all of it using native Ruby commands instead of `sh('mv …')` calls, and avoids risk of deleting folders accidentally:

```ruby
Dir.mktmpdir('a8c-release-toolkit-generate-apk-') do |tmpdir|
  arguments = [
    'build-apks',
    '--mode', 'universal',
    '--bundle', aab_file_path,
    '--output-format', 'DIRECTORY',
    ' --output', tmpdir
  ]
  # … additional logic to add each keystore parameter to `arguments` array if provided …

  # Run Bundletool to generate the universal APK in the temp dir
  sh('bundletool', *arguments)

  # Move generated file to destination once done
  FileUtils.mv( File.join(tmpdir, 'universal.apk'), apk_output_file_path )
end
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+            description: 'The path to the AAB file. If not speicified, the action will attempt to read from the lane context using the `SharedValues::GRADLE_ALL_AAB_OUTPUT_PATHS` and `SharedValues::GRADLE_AAB_OUTPUT_PATH` keys',
+            type: String,
+            optional: true,
+            default_value: nil,
+            verify_block: proc { |p| UI.user_error!("AAB path `#{p}` is not a valid file path.") unless File.file?(p) }
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :apk_output_file_path,
+            env_name: 'ANDROID_APK_OUTPUT_PATH',
+            description: 'The output path where the APK file will be generated. The directory will be created if it does not yet exist',
```
AliSoftware: This needs to be updated.

The behavior that you implemented so far (the one which adds `mv` command at the end of `bundletool` invocation), expects this to be the path to the APK (and not a folder), and does not create the directory if it doesn't exist.

Even my rewriting of your behavior (using a `Dir.mktmpdir` + `arguments` array and `sh('bundletool', *arguments)`), which uses the exact same assumptions as your code expects the path to be to the APK and not to a folder, and doesn't create the intermediate folders if they don't exist. So even if you go with my suggestion, that description isn't true anymore.

And to be honest I prefer this parameter to behave like your or my latest implementation is doing, i.e. only support that path to be a path to the `.apk` (and not to a folder), and also expecting that the parent folder of that destination already exists. This is better practice to not create the intermediate folders if they don't exist imho, so that a typo in the path doesn't end up creating the file in an unexpected place.

```suggestion
            description: 'The path to the output APK file we want to generate',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+          ),
+        ]
+      end
+
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.is_supported?(platform)
+        true
```
AliSoftware: We might want to use `platform == 'android'` here instead of supporting all platforms?

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+            default_value: nil,
+            verify_block: proc { |p| UI.user_error!("AAB path `#{p}` is not a valid file path.") unless File.file?(p) }
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :apk_output_file_path,
+            env_name: 'ANDROID_APK_OUTPUT_PATH',
+            description: 'The output path where the APK file will be generated. The directory will be created if it does not yet exist',
+            type: String,
+            optional: false,
+            default_value: nil
```
AliSoftware: 1️⃣  If that `ConfigItem` is not optional (`optional: false`) because we want to make sure the user provides an output path, then it doesn't make sense to have a `default_value` there.

```suggestion
            optional: false
```

Alternatively, we could consider that it could be valid to not pass any value for that parameter and in such case have the action infer its value as being the same path as `aab_file_path` but with `.aab` extension replaced by `.apk`. (suggestion 2️⃣ ) below)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+          ),
```
AliSoftware: Suggestion only to apply if we want behavior 2️⃣  described above, namely allow that parameter to be omitted at call site and use the `aab_file_path` as a base path to infer the `apk_output_file_path` if not provided explicitly.

```suggestion
          FastlaneCore::ConfigItem.new(
            key: :apk_output_file_path,
            env_name: 'ANDROID_APK_OUTPUT_PATH',
            description: 'The path to the output APK file we want to generate. If not provided, will use the same basename as the `aab_file_path` but with an `.apk` extension',
            type: String,
            optional: true,
            default_value: nil
          ),
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+          aab_file_path = params[:aab_file_path]
+        end
+
+        # If no AAB file path was found, raise an error
+        if aab_file_path.nil?
+          UI.user_error!('No AAB file path was specified and none was found in the lane context. Please specify the `aab_file_path` parameter or ensure that the relevant build action has been run prior to this action.')
+          raise
+        end
+
+        apk_output_file_path = params[:apk_output_file_path]
```
AliSoftware: Suggestion only to apply if we want behavior 2️⃣  described above, namely allow that parameter to be omitted at call site and use the `aab_file_path` as a base path to infer the `apk_output_file_path` if not provided explicitly.

```suggestion
        apk_output_file_path = params[:apk_output_file_path] || File.join(File.dirname(aab_file_path), "#{File.basename(aab_file_path, '.aab')}.apk")
```
This will allow the action to use the same path and basename as the `aab_file_path` but with `.aab` extension replaced with `.apk` extension… if the user didn't provide an explicit value for `apk_output_file_path` at call site.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+        keystore_path = params[:keystore_path]
+        keystore_password = params[:keystore_password]
+        keystore_key_alias = params[:keystore_key_alias]
+        signing_key_password = params[:signing_key_password]
+
+        sh(generate_command(aab_file_path, apk_output_file_path, keystore_path, keystore_password, keystore_key_alias, signing_key_password))
+      end
```
AliSoftware, line 61: If we go with suggestion 2️⃣ , we might want to return the path the the generated `apk` (given that if `apk_output_file_path` wasn't provided at call site we'll infer it, but that means that if the call site needs that APK output path it's nice to be able to get it as return value)

```suggestion
        apk_output_file_path
      end
```

If we do that, we should also document it by implementing:
```ruby
def self.return_type
  :string
end

def self.return_value
  'The path to the APK that has been generated'
end
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
AliSoftware, line 95: Now that we decided to not create intermediate folders if they don't exist, we should add a check to ensure the parent dir exists. Otherwise we'd still have the action run `bundletool` but only fail afterwards while trying to move the file to the destination, instead of failing early.

```suggestion
            verify_block: proc { |p| UI.user_error!('The parent folder for the destination does not exist. Please create it first.') unless p.nil? || File.directory?(File.dirname(p)) }
          ),
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
```diff
@@ -0,0 +1,132 @@
...
+            default_value: nil
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :keystore_path,
+            env_name: 'ANDROID_KEYSTORE_PATH',
+            description: 'The path to the keystore file',
+            type: String,
+            optional: true,
+            default_value: nil,
+            verify_block: proc { |p| UI.user_error!("Keystore file path `#{p}` is not a valid file path.") unless File.file?(p) || p.nil }
```
AliSoftware: `p.nil` is not a method that exists. (1) The method you want to call is `p.nil?` with a question mark, which is how the method is called to test if a value is nil (it's a convention in Ruby for methods returning boolean results to end with a question mark) and (2) you want to test for nullability **before** trying to test for `File.file?` otherwise `File.file?(nil)` will be tested first and crash due to `nil` parameter.
```suggestion
            verify_block: proc { |p| UI.user_error!("Keystore file path `#{p}` is not a valid file path.") unless p.nil? || File.file?(p) }
```

AliSoftware (dismissed):
Your implementation is not resilient to paths with spaces and special characters, and spawns a shell interpreter for little benefit, instead of using `sh('bundletool', 'arg1', 'arg2')` pre-parsed syntax.

Besides, your logic with `mv` can lead to unwanted destructive action if the caller passes a path to a destination folder instead of a path to the APK.

I've suggested an alternate implementation below. If you adopt it, you'll have to adjust your unit tests accordingly too; I'd suggest you to add some more unit tests as well, to validate various combinations of parameters (like if the user is passing the `keystore_path` but not the other keystore parameters, or if we provide paths with spaces or `$` characters in them, test error handling, etc)

Left some other suggestions to improve the code here and there, feel free to review them.

lib/fastlane/plugin/wpmreleasetoolkit/actions/bundletool/bundletool_generate_universal_signed_apk_action.rb
AliSoftware: I've now fixed the two security issues (path with spaces not working and leading to potential parameter injection, and risk of deleting an entire destination directory due to unchecked `rm -rf`) in 0e801e1be5b514e1b8c8c7717881991e916021a0 (+ added some unit tests in b28c1f02ecfa2ec49bc777ee072b52dbd0276219 to validate those issues were resolved)

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
AliSoftware: I've solved this [using a slightly different but similar way that I suggested in the above comment](https://github.com/wordpress-mobile/release-toolkit/blob/1d566d1e235dad84f79783a82e70378e6cbb865a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb#L15-L33). This should fix the issues the previous code was having with path containing spaces + security issues about risk of parameter injection + of removing entire user directory. lmkwyt.

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
AliSoftware: > Alternatively, we could consider that it could be valid to not pass any value for that parameter and in such case have the action infer its value as being the same path as `aab_file_path` but with `.aab` extension replaced by `.apk`.

Now implemented in [0e801e1](https://github.com/wordpress-mobile/release-toolkit/pull/467/commits/0e801e1be5b514e1b8c8c7717881991e916021a0#diff-89bdefd7112342eae6bc39216c324717a41f958babd4374b1bc30e3a47a7c0b1R13) via this line: https://github.com/wordpress-mobile/release-toolkit/blob/1d566d1e235dad84f79783a82e70378e6cbb865a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb#L13

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_generate_apk_from_aab.rb
AliSoftware: Implemented via [this line](https://github.com/wordpress-mobile/release-toolkit/pull/467/files#diff-89bdefd7112342eae6bc39216c324717a41f958babd4374b1bc30e3a47a7c0b1R35) to return the value + [this action documentation](https://github.com/wordpress-mobile/release-toolkit/pull/467/files#diff-89bdefd7112342eae6bc39216c324717a41f958babd4374b1bc30e3a47a7c0b1R132-R138).

AliSoftware:
@rynaardb GitHub doesn't let me add you as reviewer for the PR (likely because you're the PR author and GitHub doesn't allow authors to review and approve their own PRs, I guess), but feel free to review my latest changes addressing what we discussed.

rynaardb:
@AliSoftware looks good. Once you approve the PR, I'll get it merged, or feel free also to do it.

CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ _None_
 ### New Features
 
 - Add `ios_get_build_number` action to get the current build number from an `xcconfig` file. [#458]
-
+- Add new `android_generate_apk_from_aab` action to generate an APK from a given AAB.
```
```suggestion
- Add new `android_generate_apk_from_aab` action to generate an APK from a given AAB. [#467]
```

AliSoftware (approved):

rynaardb:
@AliSoftware I've marked the PR as `Do Not Merge` since there seems to be an issue with the logic that checks whether `bundletool` is installed. See [this build](https://buildkite.com/automattic/tumblr-android/builds/15864#0187ffd1-a1fe-475c-a416-01979f7e1a11) for an example. It only happens on our Android CI build agents, with `bundletool` obviously installed. The behavior seems to be different when testing locally, as it works as expected. So I'll need a bit more time to investigate why that is and to ensure we haven't missed anything.

AliSoftware:
Good catch.

Idea (not checked): could the reason for the failure be that `bundletool` is installed… but not in the `$PATH` of those agents, and thus not found by `command -v bundletool`? 🤔

rynaardb:
> Good catch.
> 
> Idea (not checked): could the reason for the failure be that `bundletool` is installed… but not in the `$PATH` of those agents, and thus not found by `command -v bundletool`? 🤔

So here's the thing, it does actually finds the path when you run `command -v bundletool`:

```bash
root@buildkite-android-builder-66998c988b-4hnfl:/# command -v bundletool
/usr/local/bin/bundletool
```

But for some reason, the check reports something else. I've already spent a good amount of time looking at the build agents to ensure we haven't misconfigured something. And so far as I can tell, everything looks fine to me.

AliSoftware:
Mmmmh… confusing 🤔 

Trying to think of something but the only think I could think of would be if the shell context could be different when running `command -v bundletool` directly in your Terminal vs via fastlane and its `sh` method?
 - Like the Terminal uses one shell (say, `bash`) and fastlane uses another (say, `sh`)?
 - Or like the code that adds adds `bundletool` to the `PATH` is in the wrong RC file (e.g. `.bashrc` vs `.bash_profile`) which would explain why it's loaded when in an interactive terminal session but not when invoked directly?

One way of checking that this is due to different context could be to create a dummy lane that runs `sh('command -v bundletool')` and see if the result differs from when you invoke it in the Terminal. Especially maybe the error you get (and that our code `rescue` [here](https://github.com/wordpress-mobile/release-toolkit/pull/467/files#diff-89bdefd7112342eae6bc39216c324717a41f958babd4374b1bc30e3a47a7c0b1R7)) is something other than "executable not found" in practice, and not rescuing it in your lane test to let it crash and print will give you more insights as to what might be wrong?

---

Note: the fact that when I used a similar logic to test the presence of `bundletool` in https://github.com/wordpress-mobile/release-toolkit/pull/365, and that this `sh('command', '-v', 'bundletool')` call from my fastlane action worked when I tested it in https://github.com/wordpress-mobile/WordPress-Android/pull/16580 (although that was back when we were still using CircleCI, not Buildkite, but still)… makes me think that the reason it doesn't work here is most likely related to something we missed in the agent's setup/config (rather than something wrong in the ruby code logic, since the same logic worked on another runner back then)? 🤔 
The big question is what that missing agent configuration is of course… but at least that might help narrowing down the ideas to test? 🤞

PR #468, opened 2023-04-25
Ignore pipeline branch filters when we trigger Buildkite via API

oguzkocer:
## What does it do?

While working on [release-toolkit-demo](https://github.com/wordpress-mobile/release-toolkit-demo), I've had the following error when I trigger a Buildkite build via its API:

```
POST https://api.buildkite.com/v2/organizations/automattic/pipelines/release-toolkit-demo/builds: 422 - Branches have been disabled for this pipeline
```

According to [this answer](https://forum.buildkite.community/t/request-build-error-branches-have-been-disabled-for-this-pipeline/1463/2) in Buildkite forums, this is because the GitHub activity is turned off. We don't want our GitHub activity settings to affect the API triggers because those settings are mainly for the default pipeline whereas we can use the API to trigger a completely separate pipeline.

This PR adds the `ignore_pipeline_branch_filters` to the parameters for triggering **all** Buildkite builds via API. I don't think we need to make this optional as I can't foresee a situation where we specifically want to prevent an API trigger to go through. Having said that, I might be missing a use case and I am open to suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ _None_
 - Add "Mobile Secrets" to `configure_update` current branch message to clarify which repo it's referring to. [#455]
 - `buildkite_trigger_build` now prints the web URL of the newly scheduled build, to allow you to easily open it via cmd-click. [#460]
 - Add the branch information to the 'This is not a release branch' error that's thrown from complete code freeze lane. [#461]
+- Adds `ignore_pipeline_branch_filters=true` parameter to the API call triggering a Buildkite build [#468]
 
 ## 7.0.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ _None_
 - Add "Mobile Secrets" to `configure_update` current branch message to clarify which repo it's referring to. [#455]
 - `buildkite_trigger_build` now prints the web URL of the newly scheduled build, to allow you to easily open it via cmd-click. [#460]
 - Add the branch information to the 'This is not a release branch' error that's thrown from complete code freeze lane. [#461]
+- Adds `ignore_pipeline_branch_filters=true` parameter to the API call triggering a Buildkite build [#468]
```
oguzkocer, line 24: I don't think this needs to be in the breaking change section because it doesn't break any existing triggers. It only adds the possibility to trigger builds for projects where the GitHub activity is turned off.

Admittedly, `release-toolkit-demo` is a special case, as we have the GitHub activity turned on for all our other projects (at least as far as I know). However, I think it's best to address it now that we know about it and avoid any further debugging in case it happens in another special case.

AliSoftware (approved):
> I don't think we need to make this optional as I can't foresee a situation where we specifically want to prevent an API trigger to go through

Agreed.

PR #469, opened 2023-04-26
Replace all instances of `is_string` with `type`

### LINKED ISSUE #278: Address the use of the legacy api of ConfigItem constructor (aka `is_string` vs `type`).

spencertransier:
## What does it do?

This PR replaces all instances of the deprecated `is_string` with the appropriate `type` in Fastlane `ConfigItem`'s. 

According to the [Fastlane documentation](https://docs.fastlane.tools/advanced/actions/#boolean-parameters), `type: Boolean` shouldn't be used because "Ruby does not have a single class to represent a Boolean type". However, if you [search the Fastlane PRs](https://github.com/fastlane/fastlane/pulls?page=3&q=is%3Apr+%22type%3A+Boolean%22+is%3Aclosed) for `type: Boolean`, there are dozens of examples where `type: Boolean` was accepted and merged in. I went ahead and used `type: Boolean` in this PR for consistency and clarity. 

A few of these had already been replaced in #425 (search for `is_string`), but I decided to replicate those changes here since that fork hasn't been merged in yet and it's been nearly 6 months. 

Fixes https://github.com/wordpress-mobile/release-toolkit/issues/278


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ _None_
 - Update `danger` to `9.3.0`. This is an internal-only change and is not a breaking change for clients. [#464]
 - Replace `rspec-buildkite-analytics` with `buildkite-test_collector` (Buildkite renamed the gem) and update it to `2.2.0`. This is another internal-only change and is not a breaking change for clients. [#465]
 - Adds `ignore_pipeline_branch_filters=true` parameter to the API call triggering a Buildkite build [#468]
+- Replace all instances of `is_string` with `type` [#469]
```
spencertransier, line 28: Because these changes don't actually change any types (just specifies which one the `ConfigItem` really is), I believe this is only an internal change since no public APIs change.

Gemfile.lock
```diff
@@ -413,7 +413,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
-  buildkite-test_collector (~> 2.1)
+  buildkite-test_collector (~> 2.2)
```
spencertransier, line 416: I updated the `Gemfile` in #465 to `2.2` but forgot to commit the `bundle install` changes.

CHANGELOG.md
AliSoftware: Correct 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_validate_lib_strings_action.rb
```diff
@@ -41,17 +41,17 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :app_strings_path,
                                        description: 'The path of the main strings file',
                                        optional: false,
-                                       is_string: true),
+                                       type: String),
           FastlaneCore::ConfigItem.new(key: :libs_strings_path,
                                        env_name: 'CHECK_LIBS_STRINGS_PATH',
                                        description: 'The list of libs to merge',
```
AliSoftware: Would have been nice of us back then to be a bit more specific and better document what's expected here, as from what I see in the source code, this is not expected to just be an "array of strings" like one might think, but an "Array of Hashes", each hash in the array being expected to have certain specific keys.

Those keys are actually [documented in some YARD doc of the private method in the helper here](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L112-L118) but they are not documented in the action's own doc so that makes them hard to find and not being printed as part of the documentation when one would run `fastlane action an_validate_lib_strings`.

Up to you if you want to use the occasion of this PR to also update the description of this `ConfigItem` while at it, or do it in a separate PR, but in any case it would still be nice to fix at some point 🙂

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb
```diff
@@ -71,7 +71,7 @@ def self.available_options
                                        description: 'The GitHub milestone',
```
AliSoftware: Could be worth fixing this `ConfigItem`'s `description` while at it (seems like this was an accidental copy/paste from the previous one)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_hotfix.rb
```diff
@@ -43,19 +43,19 @@ def self.available_options
...
             description: 'The version to branch from',
-            is_string: true
+            type: String
           ),
           FastlaneCore::ConfigItem.new(
             key: :skip_deliver,
             env_name: 'FL_IOS_BUMP_VERSION_HOTFIX_SKIP_DELIVER',
             description: 'Skips Deliverfile key update',
-            is_string: false, # Boolean parameter
+            type: Boolean,
```
AliSoftware, line 58: Side note: I think this `ConfigItem` has been removed by Ian as part of his trial project in `trial/iangmaia`, so this might create a conflict with his branch down the line.

I wonder if we might not be better off finishing [the migration document that he started to document breaking changes from his branch](https://github.com/wordpress-mobile/release-toolkit/pull/452/files), then merging his branch at least, to avoid that branch staying open too long and drifting away and getting lots of conflicts down the line—the more we wait to merge it, the more conflicts we'll risk having with subsequent requests.

This is obviously outside of this PR's scope, but just wanted to drop a note about it here in case you might have time to tackle that one next before moving to other work on the release-toolkit. Including because part of his work in his branch also contains updated or additional unit tests for some actions, and if that's also what you planned to worked on next, might be better to land theirs first too…

AliSoftware (approved):
Nice work and cleanup! ❤️ 

Left a couple of side-notes for some typos that we could consider fixing in the ConfigItem's descriptions while at it, but otherwise :shipit:

lib/fastlane/plugin/wpmreleasetoolkit/actions/android/an_validate_lib_strings_action.rb
spencertransier: Added documentation here: https://github.com/wordpress-mobile/release-toolkit/pull/469/commits/1b6391c00d62bb2be8794583ff47d6cd37885fc7

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_hotfix.rb
spencertransier: Good idea, I will get working on that next 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/setfrozentag_action.rb
spencertransier: Fixed here: https://github.com/wordpress-mobile/release-toolkit/pull/469/commits/cb40b83a2c59034795d754980ab00bfe3949644d

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
```diff
@@ -55,19 +55,19 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :target_files,
                                        env_name: 'FL_DOWNLOAD_METADATA_TARGET_FILES',
                                        description: 'The hash with the path to the target files and the key to use to extract their content',
-                                       is_string: false),
+                                       type: Hash),
           FastlaneCore::ConfigItem.new(key: :locales,
                                        env_name: 'FL_DOWNLOAD_METADATA_LOCALES',
                                        description: 'The hash with the GlotPress locale and the project locale association',
-                                       is_string: false),
+                                       type: Hash),
```
oguzkocer, line 62: @spencertransier @AliSoftware Is this the correct type for this item? According to [this](https://github.com/wordpress-mobile/release-toolkit/blob/9f2d1920219001ca8613d87155f00ebfd69d9e45/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb#L21) we are checking if it's an `Array` or a `String`. I've also checked the [locales parameter we are passing from WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/blob/fbc44e5df55f064bf7a83602aa65d3aadbff4a4c/fastlane/lanes/localization.rb#L211) and Ruby is telling me that it's an `Array`.

I was trying to test the [auto-retry downloading metadata strings PR](https://github.com/wordpress-mobile/release-toolkit/pull/474) and I've hit an issue where `fastlane` is telling me that `'locales' value must be a Hash! Found Array instead`. I thought it was something I was doing wrong, but I think this change is the culprit.

AliSoftware: Ah, this is confusing indeed!

> According to [this](https://github.com/wordpress-mobile/release-toolkit/blob/9f2d1920219001ca8613d87155f00ebfd69d9e45/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb#L21) we are checking if it's an Array or a String.

The `params[:locales]` is expected to be an `Enumerable`, on which we call `params[:locales].each do |loc| …` to iterate over each `loc` element. And those individual array elements are what we check for them to be an `Array` or a `String`.

In practice:
 - if `params[:locales]` is a `Hash` (e.g. `{'fr'=>'fr-FR', 'de'=>'de-DE', 'en'=>'en-US'}`, then iterating over that `Hash` yields elements that each are 2-item arrays of the form `[key, value]` (e.g. `['fr', 'fr-FR']`). We thus then falls into the `if loc.is_a?(Array)` branch inside the loop.
    ```ruby
    irb> {'fr'=>'fr-FR', 'de'=>'de-DE'}.each { |loc| puts loc.inspect }
    ["fr", "fr-FR"]
    ["de", "de-DE"]
    ```
 - if `params[:locales]` is an `Array`, then it can be either an `Array` of `Strings` (`['fr', 'de', 'en']`) and fall into the `if loc.is_a?(String)` branch inside the loop… or be an `Array` of 2-element-Arrays (`[['fr', 'fr-FR'], ['de', 'de-DE'], ['en', 'en-US']`) and fall into the `if loc.is_a?(Array)` branch.

So in reality, this parameter can be either a `Hash` or an `Array<String>` or an `Array<[String,String]>` (the latter being what WPAndroid uses) 😓 

I think down the line [our long term plan was to replace this legacy `gp_download_metadata` with a more modern API](pdnsEh-jJ-p2), including because of this odd API around the parameters like `locales` here. But until then we're stuck with this API that is a bit confusing and too flexible to be consistent.

---

I guess the best option here would be to set `type: Enumerable` and better document the parameter in the `description`. That being said, I've never seen `type: Enumerable` being used for a `ConfigItem` (as it's more common to provide explicit types), so we better test that this more generic type would work and be supported by `ConfigItem`'s implementation. So I guess this would be the ideal time to write some specs for that action (even minimalistic ones for now, that would only focus on feeding different type of inputs for this parameter) to ensure `type: Enumerable` would work as expected (or what would need to be used instead if not).

oguzkocer: Could we temporarily revert the change and open a separate issue for this? Since it's a breaking issue we need to address it immediately and I unfortunately don't have the bandwidth to handle it properly.

I could also test if `type: Enumerable` works for WPAndroid and handle the hotfix and the releases if that's good enough for now.

cc @spencertransier In case you have more availability to follow up on this better than I can.

spencertransier: @oguzkocer I'll revert it and get the 7.1.2 release going.

spencertransier: Created separate issue for this here: https://github.com/wordpress-mobile/release-toolkit/issues/484

PR #471, opened 2023-05-08
Deprecate ios_final_tag and ios_clear_intermediate_tags

oguzkocer:
## What does it do?

This PR deprecates `ios_final_tag` & `ios_clear_intermediate_tags` actions. We don't believe these actions are used anymore, but just to be sure, we wanted to deprecate them first.

In order to deprecate the actions, here is a list of changes I've made:
* Added `:deprecated` category
* Added `deprecated_notes`
* Added a confirmation to the action noting that the action is deprecated - as suggested by @jkmassel 
* Added `(DEPRECATED)` prefix to the `details` & `description` of each action

Hopefully this ^ is the correct approach to deprecate these actions. Please let me know if I missed anything or if you have any suggestions.

## Checklist before requesting a review

- [x] Run `bundle exec rubocop` to test for code style violations and recommendations
- [x] Add Unit Tests (aka `specs/*_spec.rb`) if applicable
- [x] Run `bundle exec rspec` to run the whole test suite and ensure all your tests pass
- [x] Make sure you added an entry in [the `CHANGELOG.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md#trunk) to describe your changes under the approprioate existing `###` subsection of the existing `## Trunk` section.

PR #472, opened 2023-05-09
Remove git push commands after creating a new commit or branch

### LINKED ISSUE #308: Don't auto-push from actions
Following the principle of least surprise, the toolkit should never push changes – the developer should have to call something like `push_changes_to_remote` in order for that to happen.

oguzkocer:
Closes #308.

## What does it do?

This PR removes the `git push` commands after creating a new commit or a branch. The goal for this change is to leave the decision for pushing the changes to remote to the developer or the project configuration.

For projects where a developer manages the release in their local environment, this gives them the flexibility to check the diff before pushing. It also removes the inconsistency where some actions were pushing the changes to remote and others weren't. A minor drawback that needs to be taken into account is that we can't trigger a beta/final build in CI before we push the change. So, we'd potentially need to add a push command to the project's `Fastfile` before these triggers.

For projects where the release is managed in CI, this allows us to decide the best time to push the changes to remote which in most cases should be after we are done with all the necessary changes. If we only push the changes when everything is successful, there will be less changes to revert if something goes wrong. Note that there might still be changes that needs to be reverted, such as closing a milestone, because these actions are not related to git commits. However, with release management in CI, we'll likely not need to worry about these either.

---

The code changes in this PR are very straightforward. I removed all the `push` commands, the arguments them and any mention of it in the documentation. Since this is a breaking change which requires extra steps, I've also added a new `MIGRATION.md` file which has steps to migrate `release-toolkit` from `7.0.0` to `8.0.0`.

---

Note that the only remaining push command is in `create_tag`. I don't know if this action makes much sense without pushing to remote, because we don't really care about local tags. On the bright side, I don't think the actions that use this helper is being used at the moment. #471 proposes the deprecation of `ios_final_tag` & `ios_clear_intermediate_tags` actions and I intend to follow that up with another PR that proposes the deprecation of `android_tag_build`. This is because we no longer tag our builds and instead create a draft GitHub release which upon publishing will create the tag for us. So, we'll likely end up removing the `create_tag` helper as well.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,4 +1,6 @@
-###From `7.0.0` to `8.0.0`
+# Migration Instructions for Major Releases
+
+### From `7.0.0` to `8.0.0`
 
 We are no longer pushing to remote after creating a new commit or a branch. That means, developers need to manually push the changes or add push commands in the project's `Fastfile`. Most importantly, we can no longer immediately trigger beta/final builds after creating a new commit because the changes will not be in remote yet. If you want to keep the existing behavior, you'll need to add a push command before these triggers.
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,0 +1,5 @@
+###From `7.0.0` to `8.0.0`
+
+We are no longer pushing to remote after creating a new commit or a branch. That means, developers need to manually push the changes or add push commands in the project's `Fastfile`. Most importantly, we can no longer immediately trigger beta/final builds after creating a new commit because the changes will not be in remote yet. If you want to keep the existing behavior, you'll need to add a push command before these triggers.
+
+For example, in [WordPress-Android's `new_beta_release` lane](https://github.com/wordpress-mobile/WordPress-Android/blob/0c64cb84c256e004473e97d72b4ac6682ebc140b/fastlane/lanes/release.rb#L86), we download translations, bump the beta version and then trigger a new build in CI. After migrating to `8.0.0` of `release-toolkit`, we'll need to add [`push_to_git_remote`](https://docs.fastlane.tools/actions/push_to_git_remote/) command before this trigger to keep the existing behavior.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-_None_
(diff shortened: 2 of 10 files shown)


MIGRATION.md
```diff
@@ -0,0 +1,7 @@
+# Migration Instructions for Major Releases
+
+### From `7.0.0` to `8.0.0`
+
+We are no longer pushing to remote after creating a new commit or a branch. That means, developers need to manually push the changes or add push commands in the project's `Fastfile`. Most importantly, we can no longer immediately trigger beta/final builds after creating a new commit because the changes will not be in remote yet. If you want to keep the existing behavior, you'll need to add a push command before these triggers.
+
+For example, in [WordPress-Android's `new_beta_release` lane](https://github.com/wordpress-mobile/WordPress-Android/blob/0c64cb84c256e004473e97d72b4ac6682ebc140b/fastlane/lanes/release.rb#L86), we download translations, bump the beta version and then trigger a new build in CI. After migrating to `8.0.0` of `release-toolkit`, we'll need to add [`push_to_git_remote`](https://docs.fastlane.tools/actions/push_to_git_remote/) command before this trigger to keep the existing behavior.
```
mokagio, line 7: For reference, we discussed having a `MIGRATION.md` document during @iangmaia's trial, in https://github.com/wordpress-mobile/release-toolkit/pull/452

My understanding of the conclusion reached at the time was to put the information into the `CHANGELOG.md` to avoid having multiple file to keep track of and update.

It's a "tough" call. On the one hand, it's neat to move the info here to avoid cluttering the changelog. On the other, separating information might make it less discoverable. ⚖️ 

For what is worth, since you already implemented this I'm happy to keep it. It will be easy enough to move the content into `CHANGELOG.md` later on if we find it cumbersome or lossy to maintain both files.

oguzkocer: Thank you @mokagio! I missed this discussion.

I like having a separate file - as is probably clear from my proposal in the first place, but I am happy to move the instructions to `CHANGELOG` file. I just need some guidance on where exactly I should put it.

AliSoftware (approved):
Searched for the word `push` in the code and confirmed that all the remaining occurrences found are not related to `git push`, so seems that all of them have been removed.

MIGRATION.md file is also clear 👍 

:shipit:

oguzkocer:
Thank you for the review @AliSoftware! 🙇‍♂️

PR #474, opened 2023-05-10
Adds the option to auto-retry for downloading metadata from GlotPress

oguzkocer:
## What does it do?

This PR adds `auto_retry` option to `gp_downloadmetadataAction`. This is an important feature for release management in CI because we can't interact with the process.

The implementation is pretty basic. If we get `429: Too Many Requests` error, we wait for `20` seconds and we try again. If we auto retry `30` times, we'll stop auto retrying, but the manual retry will still work. When I manually retry, I typically do it in a few seconds, but in CI we have the luxury of waiting, so I randomly selected `20` seconds. For the max retries, I thought `10` minutes of total retrying would be a reasonable one. However, I am happy to change these numbers if you have any feedback.

@mokagio You might especially be interested in this PR as WPiOS release manager.

**To Test**

I found testing these changes to be a bit tricky because we don't know whether we'll get `429` from the server or not. So, I prepared a test branch `test/auto-retry-gp_downloadmetadata` which switches the `200` and `429` status codes, so we can reproduce the scenario easily. I also changed the `MAX_AUTO_RETRY_ATTEMPTS` to `3`, so we can easily test that as well.

* Checkout the `test/auto-retry-gp_downloadmetadata` **WordPress-Android** branch. This branch is already [setup to use the correct branch for testing this PR](https://github.com/wordpress-mobile/WordPress-Android/commit/c6c3f86bcb6fbb08e7a152e66174f0fae2c00184) and has [`auto_retry` enabled](https://github.com/wordpress-mobile/WordPress-Android/commit/38fb8d3c4457a9a1a98577e9a96d6b6b8d8885d2).
* Run `bundle exec fastlane download_metadata_strings`
* Verify the log `Received 429 for `ar`. Auto retrying in 20 seconds...` 4 times. First is the original attempt, and then 3 auto retry attempts (It's very unlikely but you might get a different locale than `ar`)
* Verify the log `Retry downloading `ar` after receiving 429 from the API? (y/n)`
* Abort the process - You can type `n` and then quickly do `ctrl+c`. Otherwise you might get stuck in a loop 😅 

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
@@ -10,12 +10,13 @@ module Fastlane
         UI.message "Locales: #{params[:locales].inspect}"
         UI.message "Source locale: #{params[:source_locale].nil? ? '-' : params[:source_locale]}"
         UI.message "Path: #{params[:download_path]}"
+        UI.message "Auto-retry: #{params[:auto_retry]}"
 
         # Check download path
         FileUtils.mkdir_p(params[:download_path])
 
         # Download
-        downloader = Fastlane::Helper::MetadataDownloader.new(params[:download_path], params[:target_files])
+        downloader = Fastlane::Helper::MetadataDownloader.new(params[:download_path], params[:target_files], params[:auto_retry])
 
         params[:locales].each do |loc|
           if loc.is_a?(Array)
@@ -68,6 +69,12 @@ module Fastlane
                                        env_name: 'FL_DOWNLOAD_METADATA_DOWNLOAD_PATH',
                                        description: 'The path of the target files',
                                        type: String),
+          FastlaneCore::ConfigItem.new(key: :auto_retry,
+                                       env_name: 'FL_DOWNLOAD_METADATA_AUTO_RETRY',
+                                       description: 'Whether to auto retry downloads after Too Many Requests error',
+                                       type: FastlaneCore::Boolean,
+                                       optional: true,
+                                       default_value: false),
         ]
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
@@ -74,7 +74,7 @@ module Fastlane
                                        description: 'Whether to auto retry downloads after Too Many Requests error',
                                        type: FastlaneCore::Boolean,
                                        optional: true,
-                                       default_value: false),
+                                       default_value: true),
         ]
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
@@ -4,11 +4,14 @@ require 'json'
 module Fastlane
   module Helper
     class MetadataDownloader
+      AUTO_RETRY_SLEEP_TIME = 20
+
       attr_reader :target_folder, :target_files
 
-      def initialize(target_folder, target_files)
+      def initialize(target_folder, target_files, auto_retry)
         @target_folder = target_folder
         @target_files = target_files
+        @auto_retry = auto_retry
         @alternates = {}
       end
 
@@ -112,8 +115,12 @@ module Fastlane
           UI.message("Received 301 for `#{locale}`. Following redirect...")
           download(locale, response.header['location'], is_source)
         when '429'
-          # We got rate-limited, offer to try again
-          if UI.confirm("Retry downloading `#{locale}` after receiving 429 from the API?")
+          # We got rate-limited, auto_retry or offer to try again with a prompt
+          if @auto_retry
+            UI.message("Received 429 for `#{locale}`. Auto retrying in #{AUTO_RETRY_SLEEP_TIME} seconds...")
+            sleep(AUTO_RETRY_SLEEP_TIME)
+            download(locale, response.uri, is_source)
+          elsif UI.confirm("Retry downloading `#{locale}` after receiving 429 from the API?")
             download(locale, response.uri, is_source)
           else
             UI.error("Abandoning `#{locale}` download as requested.")
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
@@ -5,6 +5,7 @@ module Fastlane
   module Helper
     class MetadataDownloader
       AUTO_RETRY_SLEEP_TIME = 20
+      MAX_AUTO_RETRY_ATTEMPTS = 30
 
       attr_reader :target_folder, :target_files
 
@@ -13,6 +14,7 @@ module Fastlane
         @target_files = target_files
         @auto_retry = auto_retry
         @alternates = {}
+        @auto_retry_attempt_counter = 0
       end
 
       # Downloads data from GlotPress, in JSON format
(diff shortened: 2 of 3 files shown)


AliSoftware:
As a side note just for your information as future directions (even tho this probably won't be implemented anytime soon given our other priorities):
We've been [working with i18n folks to nudge in the direction of using a single request to get a zip download of all locales of a GlotPress Project at once](https://i18n.wordpress.com/2022/05/13/batch-download-of-translations-from-glotpress/) instead of one request per locale (see also [this PR that adds the feature to GlotPress' .com instance](https://github.com/Automattic/Orbit/issues/307)), and hopefully that might allow us to reduce the risk of 409 errors.

Of course in the meantime, the auto-retry mechanism you propose here is still useful to get a workaround in ASAP, so the PR is still useful. But just figured I'd link to those ideas we've been brainstorming to ultimately get rid of those 409 altogether FYI :)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
```diff
@@ -68,6 +69,12 @@ def self.available_options
                                        env_name: 'FL_DOWNLOAD_METADATA_DOWNLOAD_PATH',
                                        description: 'The path of the target files',
                                        type: String),
+          FastlaneCore::ConfigItem.new(key: :auto_retry,
+                                       env_name: 'FL_DOWNLOAD_METADATA_AUTO_RETRY',
+                                       description: 'Whether to auto retry downloads after Too Many Requests error',
+                                       type: FastlaneCore::Boolean,
+                                       optional: true,
+                                       default_value: false),
```
AliSoftware: tbh I wouldn't mind having this set to `true`, given how useful it'll be to have this feature (even for repos that are still doing manual releases and not using release-on-CI) 🙃

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
```diff
@@ -4,12 +4,17 @@
 module Fastlane
   module Helper
     class MetadataDownloader
+      AUTO_RETRY_SLEEP_TIME = 20
+      MAX_AUTO_RETRY_ATTEMPTS = 30
```
AliSoftware, line 8: I wonder if it wouldn't make more sense to have the `auto_retry` ConfigItem be an `Integer` that would allow to set that max-retry-attempts value?

That way we could use a value like 30 (!) for CI because we really need to be sure to end up in success, while we could use a value like 3 for repos still doing manual release, because it'd be nice not to have to `UI.confirm` the retry manually every time this happens just on the first or second try, but if it fails consistently more than 3 times we might want to wait longer when confirming manually.

oguzkocer:
@AliSoftware Thank you for linking those. I was aware that this current action is likely to be deprecated, but I need a temporary solution to continue my progress in release management in CI.

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
```diff
@@ -112,8 +117,13 @@ def handle_glotpress_download(response:, locale:, is_source:)
           UI.message("Received 301 for `#{locale}`. Following redirect...")
           download(locale, response.header['location'], is_source)
         when '429'
-          # We got rate-limited, offer to try again
-          if UI.confirm("Retry downloading `#{locale}` after receiving 429 from the API?")
+          # We got rate-limited, auto_retry or offer to try again with a prompt
+          if @auto_retry && @auto_retry_attempt_counter <= MAX_AUTO_RETRY_ATTEMPTS
+            UI.message("Received 429 for `#{locale}`. Auto retrying in #{AUTO_RETRY_SLEEP_TIME} seconds...")
+            sleep(AUTO_RETRY_SLEEP_TIME)
```
AliSoftware, line 123: I wonder if we wouldn't benefit from a linear or exponential wait time between attempts instead of a fixed time?

e.g. `sleep(10 + @auto_retry_attempt_counter*5)` or something?

AliSoftware (commented):
Logic looks sound, but left some ideas; curious what you think.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
oguzkocer: I wouldn't mind it either, but that'd be somewhat of a breaking change. If others feel the same way about setting it to `true`, it's easy enough to change :)

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
oguzkocer: That's a cool idea. I could make that change if there is practical use for it for others or if you feel strongly that it is a better approach.

My understanding is that, this issue is mostly specific to WPAndroid & WPiOS and as you mentioned it's a temporary one. That's why I did this in the most straightforward way.

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
oguzkocer: I've considered this one, but in my experience, it's really not necessary. I think the server just checks the number of requests in a certain amount of time, but I could be wrong.

For what it's worth, I actually wanted to go with something shorter, but since it's in CI, it doesn't really matter if it takes 1 or 2 minutes longer. Funnily enough, I tried to create the `429` error today for quite some time and I couldn't get it even once.

Happy to apply the suggestion if you feel there is enough benefit.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
mokagio: +1 for setting to `true` for the convenience it would provide.

I can see how it would change the default behavior and be seen as a breaking change. However, given we are the only consumers, I think our threshold for when to introduce breaking changes is relatively low. I'd also be fine with this being a minor version bump, as updating to this version would not "break" anything.

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
mokagio: FWIW, I don't feel strongly about it.

mokagio:
Thank you for doing this @oguzkocer ! I think about auto-retry literally every time I run the lane, but never made the time for it.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/gp_downloadmetadata_action.rb
oguzkocer: Addressed in dc56ef0ca29e7977db8b29b358f9d352d052b00f.

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
```diff
@@ -4,12 +4,17 @@
...
+
       attr_reader :target_folder, :target_files
 
-      def initialize(target_folder, target_files)
+      def initialize(target_folder, target_files, auto_retry)
         @target_folder = target_folder
         @target_files = target_files
+        @auto_retry = auto_retry
         @alternates = {}
+        @auto_retry_attempt_counter = 0
```
AliSoftware, line 17: It just happened to me that [the `downloader` instance is reused at call site for each locale to be downloaded](https://github.com/wordpress-mobile/release-toolkit/pull/474/files#diff-1901c13107da3d09f269520d27db2427b6ceb5b09b76eaa9a443836a97c3fc85R19).

This means that if you only set the `@auto_retry_attempt_counter` in `initialize`, the counter will be shared across all usages of the downloader, and thus across all locales for which you call `download(…)` on.

Was this intentional to consider that `MAX_AUTO_RETRY_ATTEMPTS` be seen as "max 30 retries in total" as opposed to "max 30 retries per locale"? If not, I think we need to make sure `@auto_retry_attempt_counter` is reset to `0` every time the caller calls `download(…)` on a new locale.

For example, maybe you could move the current 3 LoC implementation of `download` to a private `try_download` method, then change `download(…)` implementation to now do `@auto_retry_attempt_counter = 0` + call `try_download(…)`?

oguzkocer: This was intentional. The only goal I had with this attempt counter is so that we wouldn't get into an infinite loop. I don't think we'll ever hit this limit unless there is an issue with the server or CI in which case, I thought it'd be best to stop and let the release manager handle it either manually or by retrying once the issue is resolved.

AliSoftware: SGTM 👍

AliSoftware (approved):

PR #486, opened 2023-05-24
Release 8.1.0 into trunk

oguzkocer:


AliSoftware (approved):
Side note: I'm surprised that the PR doesn't have [the auto-generated description that is usually provided when we run `rake new_release`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/Rakefile#L66), which made me wonder if you've been doing the preparation for that release (`CHANGELOG.md`, `version.rb` & `Gemfile.lock` changes) manually, instead of using that script… and thus made me wonder if you were aware of the script, to make your life easier for next time🙃

oguzkocer:
@AliSoftware I used the script and the PR description was there. I just removed it 😅 

It's a nice reminder for the author, but I didn't think it added any value and I have a thing for empty PR descriptions when appropriate 😅

PR #490, opened 2023-06-02
Update MIGRATION.md

iangmaia:
## What does it do?

Updates the `MIGRATION.md` file with instructions on how to handle the latest breaking changes.

## Related PRs
- This PR is based on https://github.com/wordpress-mobile/release-toolkit/pull/452 (which was created before the repo had a `MIGRATION.md`).

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,6 +1,21 @@
 # Migration Instructions for Major Releases
 
-### From `7.0.0` to `8.0.0`
+## Migrating to Trunk
+
+- Ensure that calls to `ios_bump_version_release` already passed `skip_glotpress: true`.
+In case of passing false as parameter or not providing a value (false being the default for this ConfigItem), you'll have to ensure that:
+  - `download_metadata.swift` isn't being used; if it is, it's a good time to migrate to the new tooling
+  - You're not relying on `ios_bump_version_release` for commiting the `.po/.pot` file
+- The deprecated actions `ios_localize_project` and `ios_update_metadata` were now completely removed. If your project is still using them, please use the new tooling instead.
+  - See `ios_generate_strings_file_from_code`, `ios_extract_keys_from_strings_files`, `ios_download_strings_files_from_glotpress` and `ios_merge_strings_files` for typical replacements.
+- The action `ios_get_app_version` now requires a parameter `public_version_xcconfig_file` with the public `.xcconfig` file path instead of relying on the environment variable `PUBLIC_CONFIG_FILE`. While the complete removal of this environment variable is our goal, at this point it is still required by other actions such as `ios_bump_version_release` and `ios_codefreeze_prechecks`.
+- The usage of a `Deliverfile` (including its `app_version`) is discouraged -- please use `upload_to_app_store` directly from your `Fastfile` instead. Therefore, the parameter `skip_deliver` from the actions `ios_bump_version_hotfix` and `ios_bump_version_release` has been removed.
+
+### Clean-ups
+
+- You can now delete the `ENV['APP_STORE_STRINGS_FILE_NAME']` from your Fastfile, as it isn't being used anymore.
+
+## From `7.0.0` to `8.0.0`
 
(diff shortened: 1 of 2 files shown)


MIGRATION.md
```diff
@@ -1,6 +1,21 @@
 # Migration Instructions for Major Releases
 
-### From `7.0.0` to `8.0.0`
+## Migrating to Trunk
+
+- Ensure that calls to `ios_bump_version_release` already passed `skip_glotpress: true`.
```
AliSoftware: From what I see in the CHANGELOG, this one landed in `trunk` via #443 already — and that's also probably why you removed it from line 9 of the CHANGELOG.md because it was already listed on line 80 under `7.0.0` changes, right?)

Which is confirmed if I look at the merge commit on GitHub to see which tags include that commit:
image

So that means that technically this should be part of the Migration from 6.x to 7.0, not from 8.0 to Trunk

MIGRATION.md
```diff
@@ -1,6 +1,21 @@
 # Migration Instructions for Major Releases
 
-### From `7.0.0` to `8.0.0`
+## Migrating to Trunk
```
AliSoftware: I'm wondering if we shouldn't already set this title to ``## Migrating from `8.x` to `9.x` ``, because, unlike the CHANGELOG for which we don't know what will be the next version number until we decide to do the release (depending on what PRs will have landed when we do the release and if there are breaking changes or minor changes or only bugfixes), for the Migration guide we know that it is specifically made to document what to change in client repos to account for **breaking changes**. So this section is guaranteed to be for a `9.x` release, not a `8.2.0` minor release or similar.

I'm mentioning this because, while for the `CHANGELOG.md` we have some automation that makes the section title be automatically updated by our `rake new_release` task to replace the mention of "Trunk" to the actual version number we're releasing and all… I don't think we've yet ported a similar logic for the `MIGRATION.md` file. Which means it would be really easy to do a new major release without remembering to manually update this section title to `9.0.0` at the time we release said `9.0.0` major version.

Maybe at some point we'll automate that, or maybe even add `Danger` to this repo to detect if the new CHANGELOG entry one added as part of a PR was under "Breaking Changes" and if so request that the `MIGRATION.md` file to be updated…? But in the meantime, I feel it'd be easier and ok to just use the already foreseen next major version in the title here.

AliSoftware (commented):
Since I'm going on vacation in a few hours, but also I think it would be nice to have someone else (cc @wordpress-mobile/apps-infrastructure), and external to the trial, to read those migration instructions to confirm if those instructions are clear enough to them — even if they haven't followed all the different PRs you made back in your branch during your trial and don't have as much context as we both do.

MIGRATION.md
iangmaia: > I'm wondering if we shouldn't already set this title to ## Migrating from `8.x` to `9.x` , because, unlike the CHANGELOG for which we don't know what will be the next version number until we decide to do the release (depending on what PRs will have landed when we do the release and if there are breaking changes or minor changes or only bugfixes), for the Migration guide we know that it is specifically made to document what to change in client repos to account for breaking changes. So this section is guaranteed to be for a 9.x release, not a 8.2.0 minor release or similar.
> 
> I'm mentioning this because, while for the CHANGELOG.md we have some automation that makes the section title be automatically updated by our rake new_release task to replace the mention of "Trunk" to the actual version number we're releasing and all… I don't think we've yet ported a similar logic for the MIGRATION.md file. Which means it would be really easy to do a new major release without remembering to manually update this section title to 9.0.0 at the time we release said 9.0.0 major version.

I had the same doubt -- I decided here to go for trunk just because the next version doesn't exist yet (and, theoretically, one could adopt trunk).  I agree it's just easier to use the next major version though, will change it!

MIGRATION.md
iangmaia: Ah, exactly, I updated the CHANGELOG but didn't realise this was also referring to it. Will change it, thanks for noticing!

jkmassel (approved):

PR #492, opened 2023-06-13
Support Ruby 3.2.2

crazytonyli:
## What does it do?

Ruby 3.2.2 support has been addressed in previous PRs (#493, #497, and #504). This PR updates the CI checks to ensure this library continues to be compatible with Ruby 3.2.2.

The "required ruby version" settings in the gemspec file stays unchanged—the library should stay compatible with 2.7.4.

I have updated the "Build and Test" CI step to run with both Ruby 3.2.2 and 2.7.4. Since there are no ruby source code changes in this PR, tests using Ruby 2.7.4 all passed as expected. However, there are some test failures when using Ruby 3.2.2. I'll look into those later.

I have also configured the rubocop to use 2.7 as the oldest Ruby version, so that it won't report issues like "please use this new Ruby syntax and API".

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   env: &xcode_image
-    IMAGE_ID: xcode-14.1
+    IMAGE_ID: xcode-14.3.1
   plugins:
   - &docker_plugin
     docker#v3.8.0:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -21,27 +21,39 @@ steps:
...
+        env: *xcode_image
+        plugins:
+          - automattic/a8c-ci-toolkit#2.15.0
+        agents:
+          queue: "mac"
+        matrix:
+          setup:
+            ruby:
+              - 2.7.4
+              - 3.2.2
```
AliSoftware, line 56: 👌

.rubocop.yml
```diff
@@ -6,6 +6,7 @@ require:
 # Enable new cops when we update rubocop to a newer version
 AllCops:
   NewCops: enable
+  TargetRubyVersion: 2.7
```
AliSoftware, line 9: Why not make rubocop enforce Ruby 3 rules right away? 🤔

AliSoftware (approved):

.rubocop.yml
crazytonyli: That means we'll have to change the code to use Ruby 3 syntax or API to address the new rubocop issues. I don't think we should use Ruby 3 right now, because the apps are still using 2.7. Once a new release-toolkit is released (which contains all the fixes for the Ruby 3 compatibility), I'll try to upgrade a few apps to use Ruby 3. If all go well, we can then changing this library to [require Ruby 3 and above](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/fastlane-plugin-wpmreleasetoolkit.gemspec#L15). That's when we can drop 2.7 and let rubocop enforce Ruby 3 rules.

crazytonyli:
The "build-and-test" status check shows as "Pending" because it's replaced by two "build and test" jobs. I'll merge this PR now and will update the branch protection rule to use the two new build-and-test status checks.

PR #493, opened 2023-06-13
Compare meta and metrics individually instead of the full JSON payload

crazytonyli:
## What does it do?

The app metrics JSON includes two array fields: meta and metrics. Here is [an example](https://github.com/wordpress-mobile/release-toolkit/blob/8.1.0/spec/test-data/app_size_metrics/android-metrics-payload-aab.json). In an unit test which checks the JSON payload generated by an app metrics fastlane action, it compares the full JSON payload with fixtures. This JSON string comparison [breaks on Ruby 3.2.2](https://buildkite.com/automattic/release-toolkit/builds/1148#0188b239-0551-4a5b-9e40-72aa13f3aef2/551-779).

When running with Ruby 3.2.2, the generated metrics array contains all the elements that are expected, but in an different order. The root cause is [this Dir.glob call](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_send_app_size_metrics.rb#L123) returns files in a different order, which then causes the `metrics` array elements in the JSON has a different order than expected.

I thought about changing the fastlane action to sort the `meta` and `metrics` array, to make the JSON payload deterministic. But I feel like it's not ideal to add an unnecessary functionality to the library for the sole purpose of fixing unit tests. After all, it's the unit tests that made an incorrect assumption that the elements in the metric array should be in certain order. But from the API point of view, the order doesn't really matter.

This PR updates related unit test assertions to compare individual `meta` and `metrics` properties as `Set`s, to ignore the elements order. They assertions should pass as long as what the expected meta and metrics are there, regardless of their places in the JSON array.

## Test Instructions

All is good if CI passes. You can take one step further to merge this commit to the `use-ruby-3.2.2` branch locally, and `bundle exec rspec ./spec/android_send_app_size_metrics_spec.rb` should pass too.

### DIFF
--- a/spec/android_send_app_size_metrics_spec.rb
+++ b/spec/android_send_app_size_metrics_spec.rb
@@ -60,10 +60,10 @@ describe Fastlane::Actions::AndroidSendAppSizeMetricsAction do
       expect(File).to exist(output_file)
       gzip_disabled = other_action_args[:use_gzip_content_encoding] == false
       generated_payload = gzip_disabled ? File.read(output_file) : Zlib::GzipReader.open(output_file, &:read)
-      # Compare the payloads as pretty-formatted JSON, to make the diff in test failures more readable if one happen
-      expect(JSON.pretty_generate(JSON.parse(generated_payload))).to eq(JSON.pretty_generate(expected_payload)), 'Decompressed JSON payload was not as expected'
-      # Compare the payloads as raw uncompressed data as a final check
-      expect(generated_payload).to eq(expected_payload.to_json)
+
+      generated_payload = JSON.parse(generated_payload, symbolize_names: true)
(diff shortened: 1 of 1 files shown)


spencertransier (approved):
> I thought about changing the fastlane action to sort the meta and metrics array, to make the JSON payload deterministic. But I feel like it's not ideal to add an unnecessary functionality to the library for the sole purpose of fixing unit tests. After all, it's the unit tests that made an incorrect assumption that the elements in the metric array should be in certain order. But from the API point of view, the order doesn't really matter.

I agree with that logic 👍

> You can take one step further to merge this commit to the use-ruby-3.2.2 branch locally, and bundle exec rspec ./spec/android_send_app_size_metrics_spec.rb should pass too.

All the `android_send_app_size_metrics_spec.rb` tests pass 👍

PR #495, opened 2023-06-15
Add `if_exists` config to `upload_to_s3`

mokagio:
## What does it do?

The idea for this comes from @crazytonyli's suggestion [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5856#discussion_r1230390566).

This supersedes `skip_if_exists`, but to maintain backward compatibility, the code ended up a bit convoluted. If the idea of a breaking change sounds worth it, I opened #496.

`it_exists` can assume three possible values, `:fail`, `:skip`, and `:replace`.

_Note:_ I haven't tried `:replace` IRL. I _assume_ it'll just work, but I could be wrong. I'll try it tomorrow by branching off https://github.com/wordpress-mobile/gutenberg-mobile/pull/5856 and pointing to this branch.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -134,6 +134,9 @@ module Fastlane
             key: :skip_if_exists,
             description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
             deprecated: 'Use if_exists instead',
+            verify_block: proc do
+              UI.important('The :skip_if_exists option is deprecated. Please use :if_exists instead.')
+            end,
             conflicting_options: [:if_exists],
             conflict_block: file_exists_conflicting_options_handler,
             optional: true,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -32,7 +32,7 @@ module Fastlane
                                  end
           end
 
-          case params[:if_exists].to_sym
+          case params[:if_exists]
           when :fail
             UI.user_error!(message)
           when :replace
@@ -133,13 +133,12 @@ module Fastlane
             key: :if_exists,
             description: 'What do to if the file file already exists in the S3 bucket. Possible values :skip, :replace, :fail. When set, overrides the deprecated skip_if_exists option',
             optional: true,
-            is_string: false,
+            type: Symbol,
             default_value: nil, # Using nil under the hood until we remove skip_if_exists
             verify_block: proc do |value|
               next if value.nil?
 
-              UI.user_error!("`if_exist` must be a symbol or convertible to a symbol, got #{value}") unless value.respond_to?(:to_sym)
-              UI.user_error!('`if_exist` must be one of :skip, :replace, :fail') unless %i[skip replace fail].include?(value.to_sym)
+              UI.user_error!('`if_exist` must be one of :skip, :replace, :fail') unless %i[skip replace fail].include?(value)
             end
           ),
         ]
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -132,7 +132,7 @@ module Fastlane
           ),
           FastlaneCore::ConfigItem.new(
             key: :skip_if_exists,
-            description: '[DEPRECATED: Use if_exists instead]. If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`. When if_exists is set, this option is ignored',
+            description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
             deprecated: 'Use if_exists instead',
             conflicting_options: [:if_exists],
             conflict_block: file_exists_conflicting_options_handler,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -125,6 +125,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(
             key: :skip_if_exists,
             description: '[DEPRECATED: Use if_exists instead]. If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`. When if_exists is set, this option is ignored',
+            deprecated: 'Use if_exists instead',
             optional: true,
             default_value: false,
             type: Boolean
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -134,9 +134,6 @@ module Fastlane
             key: :skip_if_exists,
             description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
             deprecated: 'Use if_exists instead',
-            verify_block: proc do
-              UI.important('The :skip_if_exists option is deprecated. Please use :if_exists instead.')
-            end,
             conflicting_options: [:if_exists],
             conflict_block: file_exists_conflicting_options_handler,
             optional: true,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -89,14 +89,6 @@ module Fastlane
       end
 
       def self.available_options
-        file_exists_conflicting_options_handler = proc do |other|
-          # Fail if the conflicting options are not the ones we expect.
-          # This can only happen if the config items are updated without updating the conflicting options.
-          UI.user_error!("Unexpected conflict with option #{other.key}") unless [:skip_if_exists, :if_exists].include?(other.key)
-
-          UI.important('When both :skip_if_exists and :if_exists are set, :skip_if_exists is ignored in favor of :if_exists')
-        end
-
         [
           FastlaneCore::ConfigItem.new(
             key: :bucket,
@@ -135,7 +127,9 @@ module Fastlane
             description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
             deprecated: 'Use if_exists instead',
             conflicting_options: [:if_exists],
-            conflict_block: file_exists_conflicting_options_handler,
+            conflict_block: proc do |option|
+              UI.user_error!("You cannot set both :#{option.key} and :skip_if_exists. Please only use :if_exists.")
+            end,
             optional: true,
             default_value: false,
             type: Boolean
@@ -144,7 +138,9 @@ module Fastlane
             key: :if_exists,
             description: 'What do to if the file file already exists in the S3 bucket. Possible values :skip, :replace, :fail. When set, overrides the deprecated skip_if_exists option',
             conflicting_options: [:skip_if_exists],
-            conflict_block: file_exists_conflicting_options_handler,
+            conflict_block: proc do |option|
+              UI.user_error!("You cannot set both :#{option.key} and :if_exists. Please only use :if_exists.")
+            end,
             optional: true,
             type: Symbol,
             default_value: nil, # Using nil under the hood until we remove skip_if_exists
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -89,6 +89,14 @@ module Fastlane
       end
 
       def self.available_options
+        file_exists_conflicting_options_handler = proc do |other|
+          # Fail if the conflicting options are not the ones we expect.
+          # This can only happen if the config items are updated without updating the conflicting options.
+          UI.user_error!("Unexpected conflict with option #{other.key}") unless [:skip_if_exists, :if_exists].include?(other.key)
+
+          UI.important('When both :skip_if_exists and :if_exists are set, :skip_if_exists is ignored in favor of :if_exists')
+        end
+
         [
           FastlaneCore::ConfigItem.new(
             key: :bucket,
@@ -126,6 +134,8 @@ module Fastlane
             key: :skip_if_exists,
             description: '[DEPRECATED: Use if_exists instead]. If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`. When if_exists is set, this option is ignored',
             deprecated: 'Use if_exists instead',
+            conflicting_options: [:if_exists],
+            conflict_block: file_exists_conflicting_options_handler,
             optional: true,
             default_value: false,
             type: Boolean
@@ -133,6 +143,8 @@ module Fastlane
           FastlaneCore::ConfigItem.new(
             key: :if_exists,
             description: 'What do to if the file file already exists in the S3 bucket. Possible values :skip, :replace, :fail. When set, overrides the deprecated skip_if_exists option',
+            conflicting_options: [:skip_if_exists],
+            conflict_block: file_exists_conflicting_options_handler,
             optional: true,
             type: Symbol,
             default_value: nil, # Using nil under the hood until we remove skip_if_exists
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
@@ -22,11 +22,25 @@ module Fastlane
 
         if file_is_already_uploaded?(bucket, key)
           message = "File already exists in S3 bucket #{bucket} at #{key}"
-          if params[:skip_if_exists]
+
+          # skip_if_exists is deprecated but to keep backward compatibility we still support it by reading it only if if_exists is not set.
+          if params[:if_exists].nil?
+            if params[:skip_if_exists]
+              UI.important("#{message}. Skipping upload.")
+              return key
+            else
+              UI.user_error!(message)
+            end
+          end
+
+          case params[:if_exists].to_sym
+          when :fail
+            UI.user_error!(message)
+          when :replace
+            UI.important("#{message}. Will replace with the given one.")
+          when :skip
             UI.important("#{message}. Skipping upload.")
             return key
-          else
-            UI.user_error!(message)
           end
         end
 
@@ -111,10 +125,23 @@ module Fastlane
           ),
           FastlaneCore::ConfigItem.new(
             key: :skip_if_exists,
-            description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
+            description: '[DEPRECATED: Use if_exists instead]. If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`. When if_exists is set, this option is ignored',
             optional: true,
             default_value: false,
             type: Boolean
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :if_exists,
+            description: 'What do to if the file file already exists in the S3 bucket. Possible values :skip, :replace, :fail. When set, overrides the deprecated skip_if_exists option',
+            optional: true,
+            is_string: false,
+            default_value: nil, # Using nil under the hood until we remove skip_if_exists
+            verify_block: proc do |value|
+              next if value.nil?
+
+              UI.user_error!("`if_exist` must be a symbol or convertible to a symbol, got #{value}") unless value.respond_to?(:to_sym)
+              UI.user_error!('`if_exist` must be one of :skip, :replace, :fail') unless %i[skip replace fail].include?(value.to_sym)
+            end
           ),
         ]
       end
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -111,11 +125,24 @@ def self.available_options
           ),
           FastlaneCore::ConfigItem.new(
             key: :skip_if_exists,
-            description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
+            description: '[DEPRECATED: Use if_exists instead]. If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`. When if_exists is set, this option is ignored',
             optional: true,
             default_value: false,
             type: Boolean
           ),
```
mokagio: I couldn't find a way to mark this as deprecated [in the docs](https://docs.fastlane.tools/advanced/actions/).

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -22,11 +22,25 @@ def self.run(params)
...
+
+          # skip_if_exists is deprecated but to keep backward compatibility we still support it by reading it only if if_exists is not set.
+          if params[:if_exists].nil?
+            if params[:skip_if_exists]
+              UI.important("#{message}. Skipping upload.")
+              return key
+            else
+              UI.user_error!(message)
+            end
+          end
```
AliSoftware: Instead of duplicating the implementation for the skip case both here and below, you could make the code simpler by replacing the deprecated param with the new param value and then let the rest of the implementation take its course as normal:
```suggestion
          # skip_if_exists is deprecated but we want to keep backward compatibility.
          params[:if_exists] = :skip if params[:skip_if_exists] && params[:if_exists].nil?
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -22,11 +22,25 @@ def self.run(params)
...
+
+          case params[:if_exists].to_sym
```
mokagio: I don't like that we need to "remember" to call `to_sym` here but there's no way to set the type as `Symbol`.

The docs suggest storing the value in a class property (instance [in the docs](https://docs.fastlane.tools/advanced/actions/#polymorphic-parameters), but we use `self.run`) but I don't know... we read everything else from `params` that it would seem odd to, again, have to "remember" to do that.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -111,11 +125,24 @@ def self.available_options
...
+            description: '[DEPRECATED: Use if_exists instead]. If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`. When if_exists is set, this option is ignored',
             optional: true,
             default_value: false,
             type: Boolean
           ),
+          FastlaneCore::ConfigItem.new(
+            key: :if_exists,
+            description: 'What do to if the file file already exists in the S3 bucket. Possible values :skip, :replace, :fail. When set, overrides the deprecated skip_if_exists option',
+            optional: true,
+            is_string: false,
```
AliSoftware: Given [the auto-convert mechanism baked into `ConfigItem`](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L278), you can use `Symbol` as a type and it will auto-convert strings to symbols for you (see my reply on your other comment).
```suggestion
            type: Symbol,
```
(And then you can remove the check in your `verify_block` about the value responding to `:to_sym`, to only keep the check about accepted values)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: Tbh when it comes to `ConfigItem` I always go [right to the source code to find which attributes are available](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L48-L55) and read the code comments, rather than relying solely on the HTML docs.

Looking at that source it seems there's a `deprecated` for `ConfigItem` we could use.
Likewise, we probably want to take advantage of the `conflicting_options` to make `skip_if_exists` and that new `if_exists` options mutually exclusive.

If you're in need of inspiration, I know that I've recently found out while working on Supply-related actions [that there's a similar case for those](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb#L68-L120) that might help you see how it's supposed to be used 🙂

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
AliSoftware: [According to the source code it should be possible to set `type: Symbol` though](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L278).

I think what is misleading in the docs is that you thought you needed polymorphic parameters, to allow both `String` and `Symbol` to be passed.
But it happens that for some types like `Symbol`, `Boolean` and `Array`, `ConfigItem` is able to auto-convert to the expected type when we pass just a `String` (I think this build-in auto-conversion was baked into `ConfigItem` mostly to make it easy to support calling actions directly from the Terminal via `fastlane run myaction option1:value1 option2:value2`)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
mokagio: Neat approach. Thanks!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -111,11 +132,28 @@ def self.available_options
           ),
           FastlaneCore::ConfigItem.new(
             key: :skip_if_exists,
-            description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
+            description: '[DEPRECATED: Use if_exists instead]. If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`. When if_exists is set, this option is ignored',
```
mokagio, line 135:
```suggestion
            description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`,
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -112,10 +133,27 @@ def self.available_options
           FastlaneCore::ConfigItem.new(
             key: :skip_if_exists,
             description: 'If the file already exists in the S3 bucket, skip the upload (and report it in the logs), instead of failing with `user_error!`',
+            deprecated: 'Use if_exists instead',
```
mokagio, line 128: I originally modified the `description` field, but that's unnecessary now that we have `deprecated`.

image

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -112,10 +133,30 @@ def self.available_options
...
+            verify_block: proc do
+              UI.important('The :skip_if_exists option is deprecated. Please use :if_exists instead.')
+            end,
```
mokagio: Turns out this was being overly cautious. Fastlane already prints a warning when using the deprecated option from within a lane. See [this build](https://buildkite.com/automattic/gutenberg-mobile/builds/6210#0188d238-ce13-46fc-9d4e-a1926ae75752/9688-9700):

image

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -22,11 +22,24 @@ def self.run(params)
...
+                                 else
+                                   :skip
+                                 end
+          end
+
+          case params[:if_exists]
+          when :fail
+            UI.user_error!(message)
+          when :replace
+            UI.important("#{message}. Will replace with the given one.")
```
mokagio, line 39: Also verified live, via [this](https://buildkite.com/automattic/gutenberg-mobile/builds/6221#0188d368-8586-4cb6-a484-0a565da0db65):

image

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -76,6 +89,14 @@ def self.details
       end
 
       def self.available_options
+        file_exists_conflicting_options_handler = proc do |other|
+          # Fail if the conflicting options are not the ones we expect.
+          # This can only happen if the config items are updated without updating the conflicting options.
+          UI.user_error!("Unexpected conflict with option #{other.key}") unless [:skip_if_exists, :if_exists].include?(other.key)
+
+          UI.important('When both :skip_if_exists and :if_exists are set, :skip_if_exists is ignored in favor of :if_exists')
```
AliSoftware: 1. I haven't tested this live, but doesn't fastlane already prints some error in case of `conflicting_options`? Do we really have to raise a `UI.user_error!` message ourselves, as opposed to letting fastlane do it? I didn't double-check in fastlane's implementation but I'd expect that `conflict_block` might only be useful in case you want to handle/resolve the conflict automatically or whatnot? (e.g. [looking at an example of other options using `conflicting_options`, I don't see them using an explicit `conflict_block`](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb#L68-L120))
2. `UI.user_error!` is a raising method (as highlighted by the convention of ending with `!`). As such, `UI.important` will never be executed, because the cause will have raised and crashed beforehand.

mokagio: > UI.user_error! is a raising method (as highlighted by the convention of ending with !). As such, UI.important will never be executed, because the cause will have raised and crashed beforehand.

Correct, but `UI.user_error!` will only run if the conflicting options is "unexpected" because of the `unless` clause. That is neither `:skip_if_exists` nor `:if_exists`.

mokagio: > [looking at an example of other options using conflicting_options, I don't see them using an explicit conflict_block](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb?rgh-link-date=2023-06-19T12%3A42%3A39Z#L68-L120)

[There are others that do](https://github.com/search?q=repo%3Afastlane%2Ffastlane%20conflict_block&type=code), e.g.:

```ruby
# sigh/lib/sigh/options.rb
#
conflicting_options: [:developer_id, :development],
conflict_block: proc do |option|
  UI.user_error!("You can't enable both :#{option.key} and :adhoc")
# ...                                     
conflicting_options: [:adhoc, :development],
conflict_block: proc do |option|
  UI.user_error!("You can't enable both :#{option.key} and :developer_id")
```

mokagio: @AliSoftware 

> doesn't fastlane already prints some error in case of conflicting_options?

Yes. I haven't tried it "live" either, but this can be verified running the tests without the `conflict_block`. We'll get failures like:

```
  2) Fastlane::Actions::UploadToS3Action uploading a file with invalid parameters when the file already exists on S3 logs a message without failing when if_exists is :skip, ignoring skip_if_exists
     Failure/Error: Fastlane::FastFile.new.parse(lane).runner.execute(:test)

     FastlaneCore::Interface::FastlaneError:
       Unresolved conflict between options: 'if_exists' and 'skip_if_exists'
```

The `conflict_block` as it's currently implemented will allow the conflicting options, print a warning, and defer to `:if_exists`.

I came to that implementation from my previous mental model that lacked the knowledge of Fastlane's support for conflicting options. I like that the behavior is flexible, but maybe is best if we act more strongly and do not allow conflicting options at all? After all, that's sort of what the setting is for.

What do you think?

AliSoftware: > Correct, but UI.user_error! will only run if the conflicting options is "unexpected" because of the unless clause. That is neither :skip_if_exists nor :if_exists.

🤦  If only I learned to scroll horizontally before commenting… 😅 

>  is best if we act more strongly and do not allow conflicting options at all? After all, that's sort of what the setting is for.

💯 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -22,11 +22,24 @@ def self.run(params)
...
           message = "File already exists in S3 bucket #{bucket} at #{key}"
-          if params[:skip_if_exists]
+
+          # skip_if_exists is deprecated but we want to keep backward compatibility.
+          if params[:if_exists].nil?
+            params[:if_exists] = if params[:skip_if_exists].nil? || params[:skip_if_exists] == false
+                                   :fail
+                                 else
+                                   :skip
+                                 end
```
AliSoftware, line 32: 💄 Nitpick: testing the opposite condition (`== true`) would make it shorter because we won't have to test both `nil` and `false`

```suggestion
            params[:if_exists] = if params[:skip_if_exists] == true
                                   :skip
                                 else
                                   :fail
                                 end
```
Or even
```suggestion
            params[:if_exists] = (params[:skip_if_exists] == true) ? :skip : :fail
```

Just a matter of taste though, so up to you.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -112,10 +125,31 @@ def self.available_options
...
             optional: true,
             default_value: false,
             type: Boolean
           ),
+          FastlaneCore::ConfigItem.new(
+            key: :if_exists,
+            description: 'What do to if the file file already exists in the S3 bucket. Possible values :skip, :replace, :fail. When set, overrides the deprecated skip_if_exists option',
+            conflicting_options: [:skip_if_exists],
+            conflict_block: proc do |option|
+              UI.user_error!("You cannot set both :#{option.key} and :if_exists. Please only use :if_exists.")
```
AliSoftware, line 142: 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
```diff
@@ -112,10 +125,31 @@ def self.available_options
...
+            end,
+            optional: true,
+            type: Symbol,
+            default_value: nil, # Using nil under the hood until we remove skip_if_exists
+            verify_block: proc do |value|
+              next if value.nil?
+
+              UI.user_error!('`if_exist` must be one of :skip, :replace, :fail') unless %i[skip replace fail].include?(value)
+            end
```
AliSoftware, line 151: 💡  I wonder if it would make it easier for future removal of `skip_if_exists` to:
 - Remove the `default_value: false` of `skip_if_exists` (effectively making that param default to `nil` if not provided explicitly)
 - Set the `default_value: :fail` for the new `if_exists` (effectively replicating the same default behavior as the one we already have with `skip_if_exists: false` being the default so far).
 - Remove the `next if value.nil?` test in the `verify_block` of this `ConfigItem`

That way we don't risk to forget to change this `default_value` of this `if_exists` in the future AND remove the `next if value.nil?` from its `verify_block` (that one I almost missed when writing that comment… so I suspect we could miss it later too) when we're ready to remove `skip_if_exists:` entirely 🤔

wdyt? (Not a blocker, up to you)

spec/upload_to_s3_spec.rb
AliSoftware, line 1: 💄 Nitpick

I wonder if it would be clearer to create:
 - A `context 'when if_exists is unspecified, fallback to skip_if_exists'` sub-block for grouping specs about `skip_if_exists:false`, `skip_if_exists:true`, `skip_if_exists unspecified` cases
 - And then another `context 'if_exists is provided explicitly'` sub-block for all the rest except the last case (the one testing the conflict)
 - Then keep the spec for testing the conflicting case on its own.

Mentioning this because:
 - While reviewing your PR and checking you covered all test cases, I caught myself keeping in my head the table of all possible combinations ([`nil`,`nil`], [`nil`,`true`], [`nil`, `false`], [`:skip`, `nil`], [`:fail`, `nil`], [`:replace`, `nil`], [`!nil`, `!nil`]) while reading thru. Making that more explicit by spliting in two `context` or `describe` could help
 - It would also make it easier to remove the tests when we decide to completely remove `skip_if_exists` in a future breaking version, I think?

spec/upload_to_s3_spec.rb
```diff
@@ -145,39 +145,141 @@ def stub_s3_response_for_file(key, exists: true)
...
+      end
+
+      it 'throws when if_exists is not one of the expected values' do
+        with_tmp_file(named: 'key') do |file_path|
+          expect do
+            run_described_fastlane_action(
+              bucket: test_bucket,
+              key: 'a8c-key1',
+              file: file_path,
+              if_exists: :invalid
```
AliSoftware, line 264: 💡  What do you think about also testing non-`Symbol` values, like:
 - `"fail"` (a String, which should pass and be auto-converted to `Symbol` by `ConfigItem` itself)
 - `"invalid"` (still a String, but that would make an invalid `Symbol`
 - `true` (Boolean`, which should fail and be invalid type)

(Or do you feel like adding those tests would inherently be us testing `ConfigItem#auto_convert` after all instead of our own code? 🤔 )

spec/upload_to_s3_spec.rb
```diff
@@ -145,39 +145,141 @@ def stub_s3_response_for_file(key, exists: true)
...
+        with_tmp_file(named: 'existing-key') do |file_path|
+          expect do
+            run_described_fastlane_action(
+              bucket: test_bucket,
+              key: 'existing-key',
+              file: file_path,
+              if_exists: :skip,
+              skip_if_exists: true
+            )
+          end.to raise_error(FastlaneCore::Interface::FastlaneError, 'You cannot set both :skip_if_exists and :if_exists. Please only use :if_exists.')
```
AliSoftware, line 280: 💄  Nitpick: wdyt about trying **_all_** non-nil combinations, instead of just a specific arbitrary one? 🤔 

```suggestion
          [true, false].product(%i[skip fail replace]).each do |skip_if_exists, if_exists|
            expect do
              run_described_fastlane_action(
                bucket: test_bucket,
                key: 'existing-key',
                file: file_path,
                if_exists: if_exists,
                skip_if_exists: skip_if_exists
              )
            end.to raise_error(FastlaneCore::Interface::FastlaneError, 'You cannot set both :skip_if_exists and :if_exists. Please only use :if_exists.')
          end
```

AliSoftware (approved):
Approving because none of my feedback is blocking, but left a couple of ideas and suggestions, so feel free to apply them (or not) before merging.

spec/upload_to_s3_spec.rb
mokagio: I had those tests for the custom implementation, then removed them once I moved to use `type: Symbol`. My rationale for removing the few examples I had is the one you highlighted:

> Or do you feel like adding those tests would inherently be us testing `ConfigItem#auto_convert` after all instead of our own code? 🤔

To be fair, I could argue the case for adding a couple of tests: Additional confidence for relatively cheap overhead. But, I wonder whether it would set the precedent for spending more and more time testing Fastlane code that we rely upon. 

I think I'd be more keen to add them if this was an action that we expect to use standalone via `bundle exec fastlane run upload_to_s3` but I don't think that's the case. All the usages I've seen have been within lanes.

spec/upload_to_s3_spec.rb
mokagio: Brilliant!

This, too, is a bit on the edge of ours vs Fastlane's behavior. However, I think the critical difference here is that for this behavior to take place we need to configure Fastlane explicitly, whereas in the non-`Symbol` conversion it was all automatic

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
mokagio: Oh yes! That `nil` default value was a trick I set to ensure backward compatibility before knowing about the conflicting options feature. Definitely worth going back a clean up.

spec/upload_to_s3_spec.rb
AliSoftware: > This, too, is a bit on the edge of ours vs Fastlane's behavior. However, I think the critical difference here is that for this behavior to take place we need to configure Fastlane explicitly,

Totally. In particular, I think that this test doesn't just verify that we have set up `conflicting_options` correctly, but also checks that we have decided — via the `conflict_block` customization — to reject _any_ combination of those. As opposed to e.g. allowing the not-so-conflicting options like `skip_if_exists: true, if_exists: :skip` or `skip_if_exists: false, if_exists: :fail` for example.

So this test is really to verify that we won't have any future regressions about "rejecting any non-nil combination", even if decide to change our implementation of `conflict_block` in the future 🙂

mokagio:
Thanks @AliSoftware for all the suggestions. I felt adopting them would have resulted in too big a change set, so I opened #500 for them. I'll merge this in the meantime.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
mokagio: 17c6d6dde135966e00a7be1374e7736e4f657bb6

spec/upload_to_s3_spec.rb
mokagio: ac5b8dc9313bb28143d7860091124378c76c579d

spec/upload_to_s3_spec.rb
mokagio: Addressed in 97999706685ce0ba933aec98b589cd759ee21438

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb
mokagio: Partially addressed in 723a782e99e13ac52e8ebbd434d2b0c9652b3a28

PR #501, opened 2023-06-22
WIP: Versioning Improvements

spencertransier:


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/version_action.rb
```diff
@@ -0,0 +1,92 @@
...
+
+        platform = params[:app_platform]
+        type = params[:version_type]
+        file_path = params[:version_file_path]
+        version = ''
+
+        case platform
+        when ':android'
+          case type
+          when 'alpha'
```
jkmassel, line 16: Let's try to pin down these `type` options so that they're consistent across platforms:

- release
- beta
- alpha
- build number

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/version_action.rb
```diff
@@ -0,0 +1,92 @@
...
+        [
+          FastlaneCore::ConfigItem.new(key: :app_platform,
+                                       env_name: 'FL_VERSION_FILE_PATH',
+                                       description: 'The platform of the requested version number. Options are `:mac`, `:ios`, or `:android`',
+                                       optional: false,
+                                       type: String,
+                                       default_value: Fastlane::Helper::LaneHelper.current_platform),
+          FastlaneCore::ConfigItem.new(key: :version_type,
+                                       env_name: 'FL_VERSION_TYPE',
+                                       description: 'The type of version or version part being requested. Examples include `release`, `alpha`, `build_number`',
```
jkmassel, line 67: Let's validate that the version type being passed is one of the set of correct values

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/version_action.rb
```diff
@@ -0,0 +1,92 @@
...
+        require_relative '../../helper/android/android_version_helper'
+        require_relative '../../helper/ios/ios_version_helper'
+
+        platform = params[:app_platform]
+        type = params[:version_type]
+        file_path = params[:version_file_path]
+        version = ''
+
+        case platform
+        when ':android'
```
jkmassel, line 14: Can we match this as a symbol instead of as a string?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/version_action.rb
```diff
@@ -0,0 +1,92 @@
...
+          case type
+          when 'alpha'
+            version = Fastlane::Helper::Android::VersionHelper.get_alpha_version
+          when 'public'
```
jkmassel, line 18: Once we figure out what these are, can we represent them as symbols as well?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/version_action.rb
```diff
@@ -0,0 +1,92 @@
+module Fastlane
+  module Actions
+    class VersionAction < Action
+      def self.run(params)
+        require_relative '../../helper/android/android_version_helper'
+        require_relative '../../helper/ios/ios_version_helper'
+
+        platform = params[:app_platform]
+        type = params[:version_type]
```
jkmassel, line 9: Do we need to have the type passed into this action? Can we assume there is just one version (prod, beta, alpha, etc) at any given time, and that they'll be in the same file, referenced by the same key, just in different git branches?

If we can find an exemplar of this, we can define the API based on it.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_bump_version_release.rb
```diff
@@ -12,7 +12,7 @@ def self.run(params)
         other_action.ensure_git_branch(branch: default_branch)
 
         # Create new configuration
-        @new_version = Fastlane::Helper::Ios::VersionHelper.bump_version_release
+        @new_version = Fastlane::Helper::Ios::VersionHelper.bump_version_release(params[:version_scheme])
```
jkmassel, line 15: Let's push the changes in this file off into a different PR

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb
```diff
@@ -32,6 +32,7 @@ module VersionHelper
         #
         def self.get_public_version
           version = get_release_version
+          return nil if version.nil?
```
jkmassel, line 35: Let's push the changes in this file off into a different PR

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_version_helper.rb
```diff
@@ -53,20 +53,28 @@ def self.get_public_version
...
         #
         # @param [String] version The current version that we want to increment
+        # @param [String] version_scheme The versioning scheme used by the app
         #
         # @return [String] The predicted next version, in the form of "X.Y".
         #         Corresponds to incrementing the minor part, except if it reached 10
         #         (in that case we go to the next major version, as decided in our versioning conventions)
         #
-        def self.calc_next_release_version(version)
+        def self.calc_next_release_version(version, version_scheme = 'marketing')
```
jkmassel, line 62: Let's push the changes in this file off into a different PR

lib/fastlane/plugin/wpmreleasetoolkit/helper/version_helper.rb
```diff
@@ -0,0 +1,30 @@
+module Fastlane
```
jkmassel, line 1: Let's push the changes in this file off into a different PR

spec/android_version_helper_spec.rb
```diff
@@ -99,6 +99,83 @@
     end
   end
 
+  describe 'calculates_the_next_release_base_version' do
```
jkmassel, line 102: Let's push the changes in this file off into a different PR

spec/ios_version_helper_spec.rb
```diff
@@ -0,0 +1,59 @@
+require_relative './spec_helper'
```
jkmassel, line 1: Let's push the changes in this file off into a different PR

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/version_action.rb
```diff
@@ -0,0 +1,92 @@
...
+                                       default_value: Fastlane::Helper::LaneHelper.current_platform),
+          FastlaneCore::ConfigItem.new(key: :version_type,
+                                       env_name: 'FL_VERSION_TYPE',
+                                       description: 'The type of version or version part being requested. Examples include `release`, `alpha`, `build_number`',
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :version_file_path,
+                                       env_name: 'FL_VERSION_FILE_PATH',
+                                       description: 'The file path where the version is stored. Usually `build.gradle` for Android or an `xcconfig` file for iOS and Mac',
+                                       optional: true,
```
jkmassel, line 73: WDYT about making this non-optional, and have it replace the environment variable in the `Fastfile`? The underlying implementation may still rely on an environment variable, but we could define it in the body of this action then destroy it once the action is complete – this would allow us to get our `Fastfile` looking how we want in a single change that won't break anything down the road, and the "badness" would be constrained to internals.

spec/version_action_spec.rb
```diff
@@ -0,0 +1,528 @@
+require 'spec_helper'
+
+describe Fastlane::Actions::VersionAction do
+  describe 'getting the build number from the provided .xcconfig file' do
+    context 'with an iOS app' do
+      it 'parses an xcconfig file with keys without spaces and returns the correct build number' do
```
jkmassel, line 6: This is more of an integration test, which is good, but we probably don't need test that `xcodeproj` properly reads the `xcconfig` file.

spec/version_action_spec.rb
```diff
@@ -0,0 +1,528 @@
...
+
+        expect_version(
+          mock_version_content: mock_version_content,
+          expected_version: nil,
+          app_platform: ':ios',
+          version_type: 'build_number'
+        )
+      end
+
+      it 'throws an error when the xcconfig file does not exist' do
```
jkmassel, line 66: Good test.

spec/version_action_spec.rb
```diff
@@ -0,0 +1,528 @@
...
+          run_described_fastlane_action(
+            version_file_path: 'file/not/found',
+            app_platform: ':ios',
+            version_type: 'build_number'
+          )
+        end.to raise_error(FastlaneCore::Interface::FastlaneError)
+      end
+    end
+
+    context 'with a Mac app' do
```
jkmassel, line 77: We probably don't need to repeat the tests for Mac vs iOS, we could probably extract the specifics into a shared setup though.

spec/version_action_spec.rb
```diff
@@ -0,0 +1,528 @@
...
+        allow(File).to receive(:exist?).and_return(true)
+        allow(File).to receive(:read).with('./version.properties').and_return(mock_version_content)
+
+        version_result = run_described_fastlane_action(
+          app_platform: ':android',
+          version_type: 'alpha',
+          version_file_path: './version.properties'
+        )
+
+        expect(version_result).to eq({ 'name' => 'alpha-222', 'code' => 1234 })
```
jkmassel, line 381: We should iron out the difference in return type between this and iOS – we should really return the same datatype for each.

spec/version_action_spec.rb
```diff
@@ -0,0 +1,528 @@
+require 'spec_helper'
```
jkmassel, line 1: It'd be great to see some input validation tests on this, as well as _possibly_ mocking out the `ios_version_helper` and `android_version_helper` to ensure we're passing the expected arguments. We want to avoid "testing that the code does what the code does", but those tests can have their place.

End-to-end tests also aren't bad necessarily, but if we only rely on them it'll make our test suite a little bit slower, and encode specific assumptions about the underlying implementation into this "controller" object's tests (which could then break if we were to change `android_version_helper`, which is often a sign of a bad test).

PR #503, opened 2023-06-30
Make create_release print & return the release URL

AliSoftware:
- Make the `create_release` action print a `UI.success` with the URL of the created GitHub Release (similar to how `create_pull_request` and `buildkite_trigger_build` do), so that it can be easily command-clickable from the Terminal to jump directly to it.
 - Make the `create_release` action return the URL too, so that it can be used at call site (e.g. to add a Buildkite annotation with it or send Slack messages linking to it etc)

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -134,6 +134,7 @@ module Fastlane
         assets.each do |file_path|
           client.upload_asset(release[:url], file_path, content_type: 'application/octet-stream')
         end
+        release[:url]
       end
 
       # Downloads a file from the given GitHub tag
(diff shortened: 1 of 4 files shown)


oguzkocer (approved):
Looks great! Thank you for this improvement @AliSoftware! 🙇‍♂️

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -134,6 +134,7 @@ def create_release(repository:, version:, description:, assets:, prerelease:, is
         assets.each do |file_path|
           client.upload_asset(release[:url], file_path, content_type: 'application/octet-stream')
         end
+        release[:url]
```
AliSoftware: I just realized [looking at the API docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) that the `url` field in the response looks like `https://api.github.com/repos/octocat/Hello-World/releases/1`, while what we want to return and print is the `html_url` field, which looks like `https://github.com/octocat/Hello-World/releases/v1.0.0`

```suggestion
        release[:html_url]
```

AliSoftware: Fixed in 0128b8dc

oguzkocer: Aahh.. Funnily enough, I actually knew about this, but until you pointed out, completely forgot 🤦 Thanks for catching that before the PR lands.

Issue #482, opened 2023-05-16
Should `fastlane` be public dependency?


mokagio:
Currently, `fastlane` is a development dependency to this gem:

https://github.com/wordpress-mobile/release-toolkit/blob/4df4c387126a94f93f1392a76bd210cc2c5c482e/fastlane-plugin-wpmreleasetoolkit.gemspec#L54

I haven't verified whether there's a rationale behind it or if it's the default behavior when Fastlane creates a plugin, but I think it would be more appropriate to make it a public dependency.

I run into this in setting up a new project. I though it would have been enough to add the plugin to get access to Fastlane, but that wasn't the case. I found that inconvenient, because I now had to specify two gems instead of one.

PR #504, opened 2023-07-03
Remove unused dependencies bigdecimal and activesupport

### LINKED ISSUE #289: Version clash for the `activesupport` & `bigdecimal` gems between toolkit and cocoapods-core

crazytonyli:
## What does it do?

I don't know much about the history of the two removed dependencies bigdecimal and activesupport, but they appear to not being used by this library.

There was this https://github.com/wordpress-mobile/release-toolkit/issues/289 issue documenting a version clash between this library and some app's gems. But I can't find what the app was, and don't know if that issue is still relevant. I'm happy to dig deeper if someone can provide more context around these two dependencies. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - Adds `if_exists` parameter to `upload_to_s3` action, with possible values `:skip`, `:fail`, and `:replace`. [#495]
 - The `create_release` action now prints and returns the URL of the created GitHub Release. [#503]
+- Removes two dependencies bigdecimal and activesupport. [#504]
 
 ### Bug Fixes
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,6 +5,7 @@ PATH
       buildkit (~> 1.5)
       chroma (= 0.2.0)
       diffy (~> 3.3)
+      fastlane (~> 2.213)
       git (~> 1.3)
       google-cloud-storage (~> 1.31)
       nokogiri (~> 1.11)
@@ -171,7 +172,7 @@ GEM
     faraday_middleware (1.2.0)
       faraday (~> 1.0)
     fastimage (2.2.6)
-    fastlane (2.212.2)
+    fastlane (2.213.0)
       CFPropertyList (>= 2.3, < 4.0.0)
       addressable (>= 2.8, < 3.0.0)
       artifactory (~> 3.0)
@@ -195,7 +196,7 @@ GEM
       json (< 3.0.0)
       jwt (>= 2.1.0, < 3)
       mini_magick (>= 4.9.4, < 5.0.0)
-      multipart-post (~> 2.0.0)
+      multipart-post (>= 2.0.0, < 3.0.0)
       naturally (~> 2.2)
       optparse (~> 0.1.1)
       plist (>= 3.1.0, < 4.0.0)
@@ -431,4 +432,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   2.4.4
+   2.4.13
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,8 +2,6 @@ PATH
   remote: .
   specs:
     fastlane-plugin-wpmreleasetoolkit (8.1.0)
-      activesupport (>= 6.1.7.1)
-      bigdecimal (~> 1.4)
       buildkit (~> 1.5)
       chroma (= 0.2.0)
       diffy (~> 3.3)
@@ -53,7 +51,6 @@ GEM
     aws-sigv4 (1.5.2)
       aws-eventstream (~> 1, >= 1.0.2)
     babosa (1.0.4)
-    bigdecimal (1.4.4)
     buildkit (1.5.0)
       sawyer (>= 0.6)
     buildkite-test_collector (2.3.1)
--- a/fastlane-plugin-wpmreleasetoolkit.gemspec
+++ b/fastlane-plugin-wpmreleasetoolkit.gemspec
@@ -24,7 +24,6 @@ Gem::Specification.new do |spec|
   # since this would cause a circular dependency
 
   # spec.add_dependency 'your-dependency', '~> 1.0.0'
-  spec.add_dependency 'activesupport', '>= 6.1.7.1'
   spec.add_dependency 'buildkit', '~> 1.5'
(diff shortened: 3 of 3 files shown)


jkmassel (dismissed):
This is ok, but in the past we'd included these in the gemspec because we needed a minimum version of both.

IMHO to accommodate that, we should pin the release toolkit to a more recent version of `fastlane`, which itself specifies these for us?

WDYT?

mokagio:
@jkmassel 
> IMHO to accommodate that, we should pin the release toolkit to a more recent version of `fastlane`, which itself specifies these for us?

See also https://github.com/wordpress-mobile/release-toolkit/issues/482

mokagio:
> There was this https://github.com/wordpress-mobile/release-toolkit/issues/289 issue documenting a version clash between this library and some app's gems. But I can't find what the app was, and don't know if that issue is still relevant.

The app was WordPress iOS and [the issue had to do with CocoaPods](https://github.com/wordpress-mobile/release-toolkit/issues/289#issuecomment-890688968).

From what I can gather from the issue – I had forgotten all about it – the problem had to do with CocoaPods depending on `activesupport` which in turn depended on `bigdecimal`. Looks like the `bigdecimal` dependency [is no longer there](https://github.com/CocoaPods/Core/blob/9320bf557741d7ae80eda4415bc6fb2b0b3a8b9c/Gemfile.lock) which makes me quite confident this removal will be fine.

Still, I triggered [a WordPress iOS build](https://buildkite.com/automattic/wordpress-ios/builds/15169#01891e7c-71d6-433e-bb9f-de0a4c1dab58) with this branch in CI to verify. `bundle install` worked on my machine ✅ 😅 

**Update:** The build failed, but the reason was [the breaking change](https://github.com/wordpress-mobile/release-toolkit/blob/2443d442d9913b2989fee71d9e1ab6e537a8d3ad/CHANGELOG.md#breaking-changes) from #445. `bundle install` worked fine 👍

crazytonyli:
@jkmassel @mokagio I don't see any issue in declaring fastlane as a dependency. Added in https://github.com/wordpress-mobile/release-toolkit/pull/504/commits/161dec9f78f067789b3dc3d12dcfe6f0e3f71640.

mokagio (approved):

AliSoftware:
I just spend 45mn trying to understand why fastlane couldn't find my actions while I was trying to test integrating my new actions from https://github.com/wordpress-mobile/release-toolkit/pull/505 into Tumblr-Android 😓 

At first I thought it was an issue with the actions I just added so I tried to bisect my changes… but finally I ended up looking at previous changes then narrowing it down to this PR being the culprit.

The issue is that if you point your `Gemfile` to any commit of `release-toolkit` that is past 4e11a60 (the commit corresponding to the merge of this PR into `trunk`), you will get the following error for any fastlane command:

```
[✔] 🚀 
[16:56:21]: Error loading plugin 'fastlane-plugin-wpmreleasetoolkit': cannot load such file -- active_support/all
+-----------------------------------+---------+---------------------------------------------------------------------------+
|                                                      Used plugins                                                       |
+-----------------------------------+---------+---------------------------------------------------------------------------+
| Plugin                            | Version | Action                                                                    |
+-----------------------------------+---------+---------------------------------------------------------------------------+
| fastlane-plugin-appcenter         | 1.11.0  | appcenter_fetch_version_number, appcenter_fetch_devices, appcenter_upload |
| fastlane-plugin-wpmreleasetoolkit | 8.1.0   | No actions found                                                          |
+-----------------------------------+---------+---------------------------------------------------------------------------+

[!] No actions were found while loading one or more plugins
    Please use `bundle exec fastlane` with plugins
    More info - https://docs.fastlane.tools/plugins/using-plugins/#run-with-plugins
```

I think we need to revert at the very least the removal of `activesupport` (i.e. re-add it to the plugin's dependencies) to be able to make it work again. @crazytonyli I'm just hoping that removal of `activesupport` that was done in this PR wasn't a requirement for the toolkit to support Ruby 3.2.2 or something?

crazytonyli:
Ah sorry. I think I may have used searched the wrong thing when searching for code that uses activetsupport. I probably searched `activesupport`, not `active_support`...

> I'm just hoping that removal of activesupport that was done in this PR wasn't a requirement for the toolkit to support Ruby 3.2.2 or something?

I think removing bigdecimal is a requirement, but not activesupport. I removed activesupport because the issue that I looked into (#289) updated both. I thought these two gems may have some sort of connection.

PR #505, opened 2023-07-18
Add actions to compute generated release notes (i.e. PRs list) between tags

AliSoftware:
# Why

These are features that I worked on as part of my Project Thread for Tumblr (Android: [#20409](https://github.tumblr.net/TumblrMobile/android2/pull/20409), [#20434](https://github.tumblr.net/TumblrMobile/android2/pull/20434), [#20451](https://github.tumblr.net/TumblrMobile/android2/pull/20451), [#20452](https://github.tumblr.net/TumblrMobile/android2/pull/20452); iOS: [#25398](https://github.tumblr.net/TumblrMobile/orangina/pull/25398), [#25420](https://github.tumblr.net/TumblrMobile/orangina/pull/25420), [#25459](https://github.tumblr.net/TumblrMobile/orangina/pull/25459)), to generate the list of PRs that went into an alpha/beta/RC build since the last build of the corresponding version.

Typically, we use that to generate a PR list (aka generated release notes) when we create a new alpha, beta or RC build, to:
 - List the PRs that landed between the build we just did on CI and the previous build of the same version
    <details><summary>Example use cases</summary>
    
    - Listing PRs between tag `builds/beta/30.2.0.104` we just built and the corresponding previous tag `builds/beta/30.2.0.103`…
    - …or between tag `builds/beta/30.2.0.106` we just build after our code freeze promoting alpha to beta; and the previous tag `builds/alpha/30.2.0.105` that happened before the code-freeze
    - …or between tag `builds/rc/30.2.0.110` we do during submission day and the previous beta that was tagged `builds/beta/30.2.0.109`
    </details>
  - Attach that list of PRs / generated release notes to the GitHub Release we generate during those builds ([example](https://github.tumblr.net/TumblrMobile/orangina/releases/tag/builds%2Fbeta%2F305005))
  - Use that list of PRs / generated release notes as the "changelog" text of the corresponding App Center build

# How

## 1️⃣  New action `find_previous_tag(pattern:)`

This uses `git describe` to find the last tag that was reachable from the current commit (but that is _not_ the current commit) and that matches the provided `pattern`.

e.g. `find_previous_tag(pattern: 'builds/*/30.5.*')` for the tag name convention used by Tumblr-Android, or `find_previous_tag(pattern: '30.5-rc-*')` for the convention used by e.g. Jetpack

💡  _This is typically useful to then call the next action below with the proper `previous_tag` parameter._

## 2️⃣  New action `get_prs_between_tags` (†)

This uses the `releases/generate-notes` GitHub API to generate for you the markdown list of PRs that landed between two tags.

💡  Note that the markdown list generated by GitHub [can be customized using `.github/release.yml`](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options) ([example of such a customization in Tumblr-Android](https://github.tumblr.net/TumblrMobile/android2/blob/master/.github/release.yml), which allows to generate [a nicely categorized PR list](https://github.tumblr.net/TumblrMobile/android2/releases/tag/builds%2Falpha%2F30.5.0.104) based on the PR labels)

(†) 💭 I'm open to proposals for a better name for this action, if you feel inspired 🙂 

# Testing

 - I've added the corresponding specs for both actions, so if CI goes green on those, this should be ok
 - I've [tested integrating those actions in Tumblr-Android here](https://github.tumblr.net/TumblrMobile/android2/compare/Celray...tooling/update-release-toolkit-changelog-helper) and confirmed the code was still properly generating the same changelog after the refactor now using those release-toolkit actions.


MIGRATION.md
```diff
@@ -6,10 +6,12 @@
   - See `ios_generate_strings_file_from_code`, `ios_extract_keys_from_strings_files`, `ios_download_strings_files_from_glotpress` and `ios_merge_strings_files` for typical replacements.
 - The action `ios_get_app_version` now requires a parameter `public_version_xcconfig_file` with the public `.xcconfig` file path instead of relying on the environment variable `PUBLIC_CONFIG_FILE`. While the complete removal of this environment variable is our goal, at this point it is still required by other actions such as `ios_bump_version_release` and `ios_codefreeze_prechecks`.
 - The usage of a `Deliverfile` (including its `app_version`) is discouraged -- please use `upload_to_app_store` directly from your `Fastfile` instead. Therefore, the parameter `skip_deliver` from the actions `ios_bump_version_hotfix` and `ios_bump_version_release` has been removed.
+- If you were using the `get_prs_list` action, you should consider using `get_prs_between_tags` instead (which relies on the GitHub API and a config file, rather than milestones)
```
iangmaia: From what I can see, `get_prs_list` was actually removed so there's not really an option to continue using it?

```suggestion
- If you were using the `get_prs_list` action, you should use `get_prs_between_tags` instead (which relies on the GitHub API and a config file, rather than milestones)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -137,6 +137,43 @@ def create_release(repository:, version:, description:, assets:, prerelease:, is
...
+      # @param [String] previous_tag The name of the git tag to compare to.
+      # @param [String] target_commitish The commit sha1 or branch name to use as the head for the comparison if the `tag_name` tag does not exist yet. Unused if `tag_name` exists.
+      # @param [String] config_file_path The path to the GitHub configuration file to use for generating release notes. Will use `.github/release.yml` by default if it exists.
+      #
+      # @return [String] The string returned by GitHub API listing PRs between `previous_tag` and current `tag_name`
+      # @raise [StandardError] Might raise if there was an error during the API call
+      #
+      def generate_release_notes(repository:, tag_name:, previous_tag:, target_commitish: nil, config_file_path: nil)
+        repo_path = Octokit::Repository.path(repository)
+        api_url = "#{repo_path}/releases/generate-notes"
```
iangmaia, line 154: TIL that there's such thing in the API, nice!

MIGRATION.md
AliSoftware: To me `use` suggests it's the only solution and also that it would work the same way (as in just switch one for the other). While they are in fact different, as `get_prs_list` was based on the list of PRs in a given milestone, not the PRs that have been merged between two tags.

While by saying `consider using` here I meant "either you still want a list of PRs and you might be interested in `get_prs_between_tags`… _but_ if you go that route you'll have to change the process used by your app's release process accordingly, to rely on tags not on milestones for that list… _or_ maybe this use of `get_prs_list` in your client app was just legacy and in fact you haven't been using its result anymore and could be a good occasion to drop it completely"

I considered just renaming `get_prs_list` you something like `get_prs_in_milestone` instead and keeping it. But also, to my knowledge, we've stopped using that action a while ago in our client repos, replacing it with a manually-maintained `RELEASE-NOTES.txt` (that allows developers to provide better context for when those are intended to be used as a base for the final user-facing release notes). 

So I'm pretty sure that no repo use it anymore, or if their Fastfile calls it the file it generates isn't used by anyone anymore, so they might as well remove their call to it. And then, maybe consider if `get_prs_between_tags` could be useful to them… but at potentially a completely different step in the process (like when creating GitHub Pre-Releases for beta builds that happen in the middle of sprints/milestones.

iangmaia: > To me use suggests it's the only solution and also that it would work the same way (as in just switch one for the other)

Indeed, to me that that would imply that the actions were interchangeable -- I didn't realise at first they weren't, thanks for the context.

> So I'm pretty sure that no repo use it anymore, or if their Fastfile calls it the file it generates isn't used by anyone anymore, so they might as well remove their call to it. And then, maybe consider if get_prs_between_tags could be useful to them… but at potentially a completely different step in the process (like when creating GitHub Pre-Releases for beta builds that happen in the middle of sprints/milestones.

If teams also started using a manually maintained `RELEASE-NOTES.txt`, what is the use case for the new action and who's going to use it? Perhaps this is what I'm missing from the `MIGRATION.md` note 🤔: if I use an older version, why should I change to / adopt the new action (and what to do if I use the old one or not)?

AliSoftware: Yeah good point. Short answer is that this `get_prs_between_tags` is more to generate internal release notes for App Center builds and GitHub Pre-Releases (while `get_prs_list` was initially aimed to generate the `RELEASE-NOTES.txt` during a whole sprint to hand over to our freelance writer). But I agree that this might not be clear in this MIGRATION.md paragraph; I'll try to reword this.

AliSoftware: @iangmaia I tried to improve the wording in 5f6bc24f02de17a8b91bfaa9b37c636138a07648. Let me know what you think.

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
AliSoftware: Yep, very useful one, and thankfully… because without it, it would have been quite a nightmare to correctly implement a similar automatic logic ourselves, especially for cases of complicated git tree graphs (see paaHJt-4Ne-p2).

Strangely there's no built-in shim/wrapper method [in the Octokit ruby Client](https://octokit.github.io/octokit.rb/Octokit/Client/Releases.html) for this API endpoint, hence us having to call the low-level `client.post(…)` method ourselves to use it 🙃  eh. 🤷

MIGRATION.md
iangmaia: Yes, it looks much clearer now 👍

iangmaia (approved):
lgtm 👍

PR #506, opened 2023-07-18
Fix `install_gems` cache for when using `RBENV_VERSION` usage on CI

AliSoftware:
## What

 - Adopt [this latest fix from `a8c-ci-toolkit`](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/70/files) so that our test step uses the cache based on the correct ruby version on CI for each build matrix run
 - Remove an old workaround of `gem install bundler` that is not necessary anymore, and made the `gem install bundler` command be run twice

## Testing

 - This CI build [shows that the cache key that was used for Ruby `3.2.2` now correctly uses `release-toolkit-Darwin-x86_64-ruby3.2.2-{hash}`](https://buildkite.com/automattic/release-toolkit/builds/1224#01896b22-7523-423b-b7f2-2792da06bc47/360-362), based on the value of `export RBENV_VERSION={{ matrix.ruby }}` (instead of `release-toolkit-x86_64-ruby2.7.4-{hash}` like it was using before this change)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,20 +1,24 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
 # Nodes with values to reuse in the pipeline.
 common_params:
   env: &xcode_image
     IMAGE_ID: xcode-14.3.1
   plugins:
-  - &docker_plugin
-    docker#v3.8.0:
-      image: &ruby_version "ruby:2.7.4"
-      propagate-environment: true
-      environment:
-        - "RUBYGEMS_API_KEY"
-  - &docker_plugin_with_danger_token
-    docker#v3.8.0:
-      image: *ruby_version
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,20 +1,24 @@
...
-      environment:
-        - "RUBYGEMS_API_KEY"
-  - &docker_plugin_with_danger_token
-    docker#v3.8.0:
-      image: *ruby_version
-      propagate-environment: true
-      environment:
-        - "DANGER_GITHUB_API_TOKEN"
+    - &ci_toolkit
+      automattic/a8c-ci-toolkit#2.18.2: ~
```
AliSoftware, line 9: This is the main change here — moving the plugin declaration from the step using it to it being at the top with the other anchors, + updating the version from `2.15.0` to `2.18.2`.

(The rest of the changes on this `common_params:` block are just fixing indentation)

jkmassel (approved):

PR #508, opened 2023-07-19
WIP: Add build number model

spencertransier:


lib/fastlane/plugin/wpmreleasetoolkit/models/build_number.rb
```diff
@@ -0,0 +1,38 @@
...
+        config = Xcodeproj::Config.new(@file_path)
+
+        UI.user_error!('The .xcconfig file doesn\'t have a BUILD_NUMBER configured') if config.attributes['BUILD_NUMBER'].nil?
+
+        config.attributes['BUILD_NUMBER']
+      end
+
+      def write_build_number_to_xcconfig(build_number)
+        if File.exist?(@file_path)
+          Action.sh("sed -i '' \"$(awk '/^BUILD_NUMBER/{ print NR; exit }' \"#{@file_path}\")s/=.*/=#{build_number}/\" \"#{@file_path}\"")
```
jkmassel, line 26: `awk` and `sed` behave very differently on macOS and Linux, and as a rule we should avoid calling out to `sh` unless we're puppeteering some other CLI tool we can't use any other way. 

I'd strongly suggest using the `xcodeproj` gem for writing to `xcconfig` files.

lib/fastlane/plugin/wpmreleasetoolkit/models/build_number.rb
```diff
@@ -0,0 +1,38 @@
+require 'xcodeproj'
+
+module Fastlane
+  module Model
+    class BuildNumber
```
jkmassel, line 5: This class doesn't model a build number so much as it does a file _containing_ a build number among other things (and seems very iOS-specific).

If you're trying to model a build number (and the operations that can be performed on it), you'll probably never want your model to be able to persist itself. That's generally true for anything that isn't modelling a file – if your class is called `ConfigurationFile`, it might make sense to have a `save` or `write` method, but for something that's consolidating computation like this it's probably not quite what you're after.

AliSoftware:
My two cents: personally the way I imagine the future API and models around versioning:

 - A `Version` model, which should just basically be a struct that holds 4 numbers (major, minor, patch, build)
 - Then, have separate classes dedicated to formatting of those versions, with different implementations depending on the needs and conventions of the project, but all implementing the same interface. Each formatter will know how to take a `Version` model, and produce:
    - A `full_version_name` (like `xx.y.z.n`, Tumblr-style; or `xx.y.z-rc-n`, WP-style); corresponding to the name used in git tags and GitHub Releases, for example, and as `versionName` for Android
    - A `prod_version` (like just `xx.y`), corresponding to the version we will use when submitting to App Store / Play Store, and how we refer to the current sprint's version as a whole during the sprint (as opposed to a specific build). Will also be used for `VERSION_SHORT` for iOS
    - And a `build_number`, which should be a generated Integer (like `1XXYYZZZNNN`, Tumblr-style), that will be used for `versionCode` for Android, `VERSION_LONG` for iOS

In other words, the way we transform a version model into a string should be separate from the model itself.

Similarly, for version bumping, we might want to support multiple different conventions. E.g. We could imagine having a class named `VersionManager` that would take a version file path + `minorMinValue` and `minorMaxValue` as parameters, and a formatter instance, and would use those to know how to read a version from file, bump the minor version (wrapping to next major as necessary), and save to file (we might need 2 subclasses of that parent class to suport both file formats— Android vs iOS — there)

In the end, a client project would then instantiate a `version_manager = Android::VersionManager` once and for all at the top of their `Fastfile`, passing it the right parameters to describe the _conventions_ used for versioning in this project as a whole (version file, formatter, bumping rules). Then in our lanes, you'd use that `version_manager` to get `Version` model object representing the current version, or get a new `Version` model representing the next version, asking it to save such `Version` objects to disk (to the version file), to format those `Version` objects as short or long printable version strings, etc…

---

Obviously all this is very opinionated and very subjective and only my two cents, so take this with a grain of salt and just as random ideas, and feel free to go another route or adapt with your ideas. I just figured that sharing those thoughts could help the collective brainstorming 🙂

PR #509, opened 2023-07-24
Release 9.0.0

AliSoftware:
New version 9.0.0.

_PR Author: Be sure to create a GitHub Release and tag once this PR gets merged._

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/version.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/version.rb
@@ -1,5 +1,5 @@
 module Fastlane
   module Wpmreleasetoolkit
-    VERSION = '8.1.0'.freeze
+    VERSION = '9.0.0'
   end
 end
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fastlane
   module Wpmreleasetoolkit
-    VERSION = '8.1.0'.freeze
+    VERSION = '9.0.0'
```
crazytonyli, line 5: Rubocop complainted about this removing `.freeze` change.

crazytonyli (approved):

lib/fastlane/plugin/wpmreleasetoolkit/version.rb
AliSoftware: Thanks for the catch. Fixed in [1b2ceb8](https://github.com/wordpress-mobile/release-toolkit/pull/509/commits/1b2ceb8b2682195bc4ce91d7a23a763f81d0eb53) by using the `frozen_string_literal: true` magic comment on the whole file, so that we don't have to update our `rake new_release` script that changes this `VERSION = '…'` line on every release and risk encountering the same issue on next release.

PR #511, opened 2023-09-04
Fix `prototype_build_details_comment_action` args

AliSoftware:
Fix which parameters are required or not for that action
 - Only `app_center_org_name` is required
 - All other parameters can be decided from the `lane_context` (assuming `appcenter_upload` was called before that lane)

The was never right, and never caused any trouble because all other repos that called this action have all provided an `app_display_name` explicitly)

@mokagio I'm targeting `release/9.0.1` because it seems there's still time to include this small fix into a patch release 🤞


AliSoftware:
Actually nvm, this is a bit more convoluted than that, because we don't want `app_center_org_name` to be required for the case when we use this action without `appcenter_upload` (especially for Android builds we upload to S3 instead of going thru App Center. So forget about this PR after all.

PR #513, opened 2023-09-18
Add/Expand actions to manipulate GitHub branch protections

AliSoftware:
This PR add/improve actions to manipulate GitHub branch protections:

# Why?

## Lock release branch after submission

As part of my project to improve Tumblr branching model (paaHJt-5ii-p2), I needed an action to lock a given branch (i.e. make it read-only), so that developers couldn't accidentally merge a PR into the release branch after the final build have already been triggered and the app have already been submitted for review.

 - In practice so far we've been doing this in Tumblr (which is a procedure we call "the lockdown" — paaHJt-4iY-p2) [via some ruby code that was duplicated in each Tumblr repo that we used while doing the branch dance for submission and code freeze](https://github.tumblr.net/TumblrMobile/orangina/blob/develop/fastlane/lib/helpers.rb#L283-L313). Once this PR lands, we plan to instead do this by setting `lock_branch: true` on the release branch
 - While we've only done this "lockdown" practice in Tumblr so far—as historically last-minute PRs were sadly more common in Tumblr (pbEOar-1c7-p2) than in other apps—I believe making the `release/*` branch read-only before triggering the final build that will be submitted for review **will be useful to other apps as well** and should actually be done for all our app.

## Set Required CI Checks on release branch

When we do the code freeze and create a new `release/*` branch, so far all our projects have only set the branch protection of those to "require 1 reviewer", but not setting any required CI checks on those `release/*` branch.

As a result, not only were people able to merge an approved PR even if it had red CI steps, but also we couldn't use the "Enable auto-merge" feature on PRs targeting `release/*` (because for that feature to be available, it needs at least one required check)

# How?

## `set_branch_protection`

<details><summary>This action was previously named <tt>addbranchprotection</tt> (without any `_`), so I renamed it for a nicer-looking name.</summary>

  - I kept `addbranchprotection` as a backwards-compatible but deprecated action, so that we don't need a new major version bump just yet though
  - In other words, calling `addbranchprotection(repo, branch)` when the branch wasn't protected yet will still work and do the same thing as before (except it will also print a deprecation notice) — since it will call the same code as `set_branch_protection(repo, branch)`, which will in turn set the same branch protection default options as what `addbranchprotection` did before this PR.
  - The only difference is that if the branch was _already_ protected, then before `addbranchprotection` would completely _overwrite_ the current branch protection settings to reduce them to the strict minimum defaults (i.e. "require 1 approval", and all other settings turned off)
</details>

The improved `set_branch_protection` action can now do much more though:
 - Set default branch protection settings (i.e. just "require a PR, with 1 approved review before merging") if the branch was not protected yet and no other parameter was passed (just like `addbranchprotection` did)
 - Set the default settings + just the settings provided as parameters if the branch was not protected yet and we pass explicit parameters (like `lock_branches: true` or `required_ci_checks: ['a', 'b', 'c']`)
 - Updates only the branch protection settings provided as parameters, while keeping the other settings unchanged, if the branch was already protected.

```
# If branch is not protected yet
set_branch_protection(repo, branch) # => protects the branch with the default settings (PR required, 1 approval)
set_branch_protection(repo, branch, allow_force_pushes: true) # => protects the branch with "PR required", "1 approval required" and "force pushes enabled"
set_branch_protection(repo, branch, required_approving_review_count: 3) # => protects the branch with "3 approval required"
set_branch_protection(repo, branch, required_approving_review_count: 3, required_ci_checks: ['buildkite/build', 'buildkite/test'], enforce_admins: true) # => protects the branch with "3 approval required", and the provided CI checks as required to go green before merge, even for admin users

# If branch is already protected
set_branch_protection(repo, branch) # => changes nothing, so it's basically a no-op
set_branch_protection(repo, branch, lock_branch: true) # => lock the branches, while keeping other existing branch protection settings (like required checks and so on) untouched
```

💡  In practice, this action and its improvements are intended especially to set `lock_branches: true` on the `release/*` branch just before we submit the final build to Apple/Google at the end of the sprint, so that people can't merge any last-minute PR into the `release/*` branch _after_ we've already submitted the build for review.

## `remove_branch_protection`

This action was previously named `removebranchprotection` (without any `_`). I just renamed it to `remove_branch_protection` for a nicer-looking name
 - I kept `removebranchprotection` as a backwards compatible but deprecated action, so that we don't need a new major version bump just yet though
 - I also simplified the implementation a bit, and added unit tests

## `copy_branch_protection(repo: from_branch:, to_branch:)`

This is a new action that copies existing branch protection onto a new branch.
 - If the `to_branch` was already protected, this will overwrite its branch protection settings completely with the values of the branch protection settings from the `from_branch`
 - If the `to_branch` was not protected yet, this will protect it with the same settings as the `from_branch`

💡  In practice, this action is intended especially to set up the branch protection settings of the `release/*` branch when we create it during code-freeze, by replicating the same settings on that new branch from the ones that are currently set on the `trunk` branch (including the required CI checks and number of approvals etc).

Currently in many app's `Fastfile` this is done by calling `setbranchprotection(repository: GHHELPER_REPO, branch: "release/#{new_version}")` at the end of the `code_freeze` lane. Once this lands, you'll be able to replace that call with `copy_branch_protection(repository: GHHELPER_REPO, from_branch: 'trunk', to_branch: "release/#{new_version|")`, and that will not only protect the newly created `release/*` branch, but also set the same required CI checks and every other settings as the ones set on `trunk`.

# Testing

Since I've added unit tests (specs) to all those actions, having the CI go green should be enough.

### DIFF
--- a/spec/copy_branch_protection_action_spec.rb
+++ b/spec/copy_branch_protection_action_spec.rb
@@ -37,7 +37,7 @@ describe Fastlane::Actions::CopyBranchProtectionAction do
     allow(client).to receive(:branch_protection).with(repo, from_branch).and_return(stub_response(existing_settings))
 
     new_settings = Fastlane::Helper::GithubHelper.branch_protection_api_response_to_normalized_hash(existing_settings)
-    expect(client).to receive(:protect_branch).with(repo, to_branch, **new_settings)
+    expect(client).to receive(:protect_branch).with(repo, to_branch, new_settings)
 
     run_described_fastlane_action(
       repository: repo,
@@ -77,7 +77,7 @@ describe Fastlane::Actions::CopyBranchProtectionAction do
     existing_settings = fixture('existing_branch_protection.json')
     allow(client).to receive(:branch_protection).with(repo, from_branch).and_return(stub_response(existing_settings))
     new_settings = Fastlane::Helper::GithubHelper.branch_protection_api_response_to_normalized_hash(existing_settings)
-    allow(client).to receive(:protect_branch).with(repo, to_branch, **new_settings).and_raise(Octokit::NotFound)
+    allow(client).to receive(:protect_branch).with(repo, to_branch, new_settings).and_raise(Octokit::NotFound)
 
     expect do
       run_described_fastlane_action(
--- a/spec/copy_branch_protection_action_spec.rb
+++ b/spec/copy_branch_protection_action_spec.rb
@@ -5,6 +5,7 @@ describe Fastlane::Actions::CopyBranchProtectionAction do
   let(:repo) { 'automattic/demorepo' }
   let(:from_branch) { 'trunk' }
   let(:to_branch) { 'release/12.3' }
+  let(:github_token) { 'stubbed-gh-token' }
   let(:client) do
     instance_double(
       Octokit::Client,
@@ -41,7 +42,8 @@ describe Fastlane::Actions::CopyBranchProtectionAction do
     run_described_fastlane_action(
       repository: repo,
       from_branch: from_branch,
-      to_branch: to_branch
+      to_branch: to_branch,
+      github_token: github_token
     )
   end
 
@@ -52,7 +54,8 @@ describe Fastlane::Actions::CopyBranchProtectionAction do
       run_described_fastlane_action(
         repository: repo,
         from_branch: from_branch,
-        to_branch: to_branch
+        to_branch: to_branch,
+        github_token: github_token
       )
     end.to raise_error(FastlaneCore::Interface::FastlaneError, "Branch `#{from_branch}` of repository `#{repo}` was not found.")
   end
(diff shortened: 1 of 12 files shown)


spec/copy_branch_protection_action_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    JSON.parse(File.read(path), symbolize_names: true)
+  end
+
+  def stub_response(hash)
+    stubs = Faraday::Adapter::Test::Stubs.new
+    agent = Sawyer::Agent.new 'http://release-toolkit.com/specs' do |conn|
+      conn.builder.handlers.delete(Faraday::Adapter::NetHttp)
+      conn.adapter :test, stubs
+    end
+    Sawyer::Resource.new(agent, hash)
```
AliSoftware, line 28: Borrowed directly [from the `sawyer` gem's own tests](https://github.com/lostisland/sawyer/blob/master/test/resource_test.rb#L7-L11)

jkmassel (approved):

PR #516, opened 2023-10-04
Share GlotPress 429 auto retry from ASC metadata to iOS strings download

mokagio:
## What does it do?

After benefitting from the GlotPress metadata downloads automatically retrying upon receiving a 429, it felt particularly painful, not to mention time consuming, to manually retry when the iOS `.strings` got a 429.

This PR extracts the "retry upon 429" logic in a dedicated `GlotpressDownloader` object, and updates the iOS `strings` and generic metadata download logic to use it.

I added as much test coverage as I could think of to guide my changes, but I didn't try to provide full coverage given the uncertain future of the metadata download component.


lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,52 @@
...
+        when '301' # Follow the redirect
+          UI.message("Received 301 for `#{response.uri}`. Following redirect...")
+          download(response.header['location'])
+        when '429' # We got rate-limited, auto_retry or offer to try again with a prompt
+          if @auto_retry
+            if @auto_retry_attempt_counter < @auto_retry_max_attempts
+              UI.message("Received 429 for `#{response.uri}`. Auto retrying in #{@auto_retry_sleep_time} seconds...")
+              sleep(@auto_retry_sleep_time)
+              @auto_retry_attempt_counter += 1
+              download(response.uri)
```
mokagio:
```suggestion
              download(response.uri, options)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,52 @@
...
+            else
+              UI.error("Abandoning `#{response.uri}` download after #{@auto_retry_attempt_counter} retries.")
+            end
+          elsif UI.confirm("Retry downloading `#{response.uri}` after receiving 429 from the API?")
+            download(response.uri)
```
mokagio:
```suggestion
            download(response.uri, options)
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -161,19 +161,34 @@ def self.generate_strings_file_from_hash(translations:, output_path:)
...
         # @param [String, IO] destination The path or `IO`-like instance, where to write the downloaded file on disk.
         #
-        def self.download_glotpress_export_file(project_url:, locale:, filters:, destination:)
+        def self.download_glotpress_export_file(
+          project_url:,
+          locale:,
+          filters:,
+          destination:,
+          autoretry: true,
+          autoretry_count: 0,
```
```suggestion
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/metadata_download_helper.rb
```diff
@@ -4,23 +4,24 @@
 module Fastlane
   module Helper
     class MetadataDownloader
-      AUTO_RETRY_SLEEP_TIME = 20
-      MAX_AUTO_RETRY_ATTEMPTS = 30
-
       attr_reader :target_folder, :target_files
 
-      def initialize(target_folder, target_files, auto_retry)
+      def initialize(target_folder, target_files, auto_retry, auto_retry_sleep_time = 20, auto_retry_max_attempts = 30)
```
mokagio:
```suggestion
      def initialize(target_folder, target_files, auto_retry, auto_retry_sleep_time = 20)
```

spec/glotpress_downloader_spec.rb
```diff
@@ -0,0 +1,116 @@
+require 'spec_helper'
+
+describe Fastlane::Helper::GlotpressDownloader do
+  describe 'downloading' do
+    context 'when auto retry is enabled' do
+      context 'when GlotPress returs a 200 code' do
+        it 'returns the response, without retrying' do
+          downloader = described_class.new(auto_retry: true)
+          fake_url = 'https://test.com'
```
mokagio, line 9: `fake_url` is misleading... `dummy_url` might be better?

spec/metadata_download_helper_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper'
+
+describe Fastlane::Helper::MetadataDownloader do
+  describe 'downloading from GlotPress' do
```
mokagio, line 4: This test would benefit with ensuring the files are saved.

.ruby-version
```diff
@@ -1 +1 @@
-2.7.4
\ No newline at end of file
+3.2.2
```
mokagio, line 1: This and the WebMock update were attempts to solve the tests having a missing `uri` in the stubbed response.

That turned out to be a bug in WebMock (see `spec_helper` diff). I decided to keep them because they don't seem to hurt, but I can remove this one if you think it should be a changed for a dedicated PR.

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -161,19 +161,33 @@ def self.generate_strings_file_from_hash(translations:, output_path:)
...
+          )
+
           # Set an unambiguous User Agent so GlotPress won't rate-limit us
           options = { 'User-Agent' => Wpmreleasetoolkit::USER_AGENT }
 
+          response = downloader.download(uri.to_s, options)
+
           begin
-            IO.copy_stream(uri.open(options), destination)
+            IO.copy_stream(StringIO.new(response.body), destination)
```
AliSoftware, line 188: iirc, the reason I initially used `IO.copy_stream` was because `uri.open` happened to be a stream in the first place, and as such this would allow memory optimization (i.e. to avoid having to read the whole, large file content into memory all at once only to write it to the destination file in one block, and instead use `copy_stream` to write data to the `destination` as it arrived from the socket).

Now that we don't use `uri.open` anymore, I'm not sure it makes sense to continue using `copy_stream` then wrapping the whole response body inside a `StringIO.new`. I think we can probably instead:
 - Either just use `File.write(destination, response.body)` directly (which means we lose the memory footprint optimization of write-as-you-receive that `copy_stream` had, but is it a big deal?)
 - Or try using something like `File.open(destination, mode: 'w') { |file| response.read_body(file) }` (which I didn't try, but based on [this doc](https://ruby-doc.org/stdlib-2.7.1/libdoc/net/http/rdoc/Net/HTTPResponse.html#method-i-read_body) I think that's how we'd be supposed to use it?) to still benefit from the "write-as-you-receive" behavior that would allow reducing the memory footprint for very large files.

wdyt?

spec/glotpress_downloader_spec.rb
AliSoftware: agreed

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,52 @@
...
+module Fastlane
+  module Helper
+    class GlotpressDownloader
+      AUTO_RETRY_SLEEP_TIME = 20
+      MAX_AUTO_RETRY_ATTEMPTS = 30
+
+      def initialize(
+        auto_retry: true,
+        auto_retry_sleep_time: 20,
+        auto_retry_max_attempts: 30
```
AliSoftware, line 12:
```suggestion
        auto_retry_sleep_time: AUTO_RETRY_SLEEP_TIME,
        auto_retry_max_attempts: MAX_AUTO_RETRY_ATTEMPTS
```

spec/metadata_download_helper_spec.rb
AliSoftware: I'm not sure I'm following, isn't it was those tests already do, by using an `in_tmp_dir` then testing `expect(File.read(destination_path_with_locale)).to eq(…)`?

spec/metadata_download_helper_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        metadata_downloader.download('en-AU', dummy_url, false)
+
+        destination_path_with_locale = File.join(tmp_dir, 'en-AU', destination_name)
+        expect(File.exist?(destination_path_with_locale)).to be true
+        # We also expect a trailing new line.
+        expect(File.read(destination_path_with_locale)).to eq("test metadata\n")
+      end
+    end
+
+    context 'when GlotPress returs a 429 code' do
```
AliSoftware, line 26:
```suggestion
    context 'when GlotPress returns a 429 code' do
```

spec/glotpress_downloader_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+          response = downloader.download(fake_url)
+
+          expect(response.code).to eq('200')
+          expect(response.body).to eq('OK')
+          # Make sure there was no retry
+          assert_requested(:get, fake_url, times: 1)
+        end
+      end
+
+      context 'when GlotPress returs a 429 code' do
```
AliSoftware, line 92:
```suggestion
      context 'when GlotPress returns a 429 code' do
```

spec/glotpress_downloader_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+            downloader.download(fake_url)
+
+            assert_requested(:get, fake_url, times: max_retries + 1) # the original request plus the retries
+          end
+        end
+      end
+    end
+
+    context 'when auto retry is disabled' do
+      context 'when GlotPress returs a 200 code' do
```
AliSoftware, line 76:
```suggestion
      context 'when GlotPress returns a 200 code' do
```

spec/glotpress_downloader_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+          response = downloader.download(fake_url)
+
+          expect(response.code).to eq('200')
+          expect(response.body).to eq('OK')
+          # Make sure there was no retry
+          assert_requested(:get, fake_url, times: 1)
+        end
+      end
+
+      context 'when GlotPress returs a 429 code' do
```
AliSoftware, line 22:
```suggestion
      context 'when GlotPress returns a 429 code' do
```

spec/glotpress_downloader_spec.rb
```diff
@@ -0,0 +1,116 @@
+require 'spec_helper'
+
+describe Fastlane::Helper::GlotpressDownloader do
+  describe 'downloading' do
+    context 'when auto retry is enabled' do
+      context 'when GlotPress returs a 200 code' do
```
AliSoftware, line 6:
```suggestion
      context 'when GlotPress returns a 200 code' do
```

iangmaia:
@mokagio I've realized we have https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb implemented on https://github.com/wordpress-mobile/release-toolkit/pull/667 -- perhaps when I implemented it I took yours as inspiration but I don't really remember 😂  Anyway, I guess we can close this PR?

mokagio:
> Anyway, I guess we can close this PR?

Seems like the case. Nice!

PR #531, opened 2023-12-29
Replace two GitHelper methods with Fastlane actions

spencertransier:
## What does it do?

This PR deprecates these two methods in the GitHelper module and replaces their uses with built-in Fastlane actions: 
- `update_submodules` replaced by `git_submodule_update`
- `ensure_on_branch!` replaced by `ensure_git_branch`

I'm deprecating the methods instead of removing them because we have used some of the GitHelper methods directly instead of through Release Toolkit actions. 

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
@@ -72,7 +72,7 @@ module Fastlane
       # @deprecated This method is going to be removed soon. Fastlane has a built-in `git_submodule_update` action that can be used instead.
       #
       def self.update_submodules
-        UI.message("DEPRECATED: The `update_submodules` Release Toolkit method will be removed soon. Please use Fastlane's `git_submodule_update` action instead.")
+        UI.deprecated("The `update_submodules` method will soon be removed from `release-toolkit`. Please use fastlane's `git_submodule_update` action instead.")
         Action.sh('git', 'submodule', 'update', '--init', '--recursive')
       end
 
@@ -219,7 +219,7 @@ module Fastlane
       #             See https://github.com/fastlane/fastlane/pull/21597 for more details.
       #
       def self.ensure_on_branch!(branch_name)
-        UI.message("DEPRECATED: The `ensure_on_branch!` Release Toolkit method will be removed soon. Please use Fastlane's `ensure_git_branch` action instead.")
+        UI.deprecated("The `ensure_on_branch!` method will soon be removed from `release-toolkit`. Please use fastlane's `ensure_git_branch` action (+ potentially set `FL_GIT_BRANCH_DONT_USE_ENV_VARS=true`) instead.")
         current_branch_name = Action.sh('git', 'symbolic-ref', '-q', 'HEAD')
         UI.user_error!("This command works only on #{branch_name} branch") unless current_branch_name.include?(branch_name)
       end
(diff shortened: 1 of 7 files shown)


AliSoftware:
iirc, there was a reason why we had our own `ensure_on_branch!` instead of `ensure_git_branch`, as _fastlane_'s implementation was doing something that we didn't want. I think it was something like taking some env vars (like `BUILDKITE_BRANCH` and such) into account for its check… which didn't work for release-on-CI where the env var still had the old value even after we cut a new branch during the process?) and that's why we re-implemented it ourselves without being affected by the special logic.

I think that now that @oguzkocer implemented a workaround in _fastlane_ in https://github.com/fastlane/fastlane/pull/21597, which has been shipped in fastlane 2.217.0, this is not a problem anymore… as long as your projects have updated to this version of fastlane or later, and also that you do set `FL_GIT_BRANCH_DONT_USE_ENV_VARS` in your CI environment.

If that's indeed what needs to be done to be ready to migrate, we should include a note about this in the `CHANGELOG.md` or in the `MIGRATION.md` file to document it, otherwise this would be breaking projects.

spencertransier:
@AliSoftware I updated the deprecation notice for `ensure_on_branch!` to add additional context about the Fastlane env var PR. When we're ready to remove it and make it a breaking change, I will add it to `MIGRATION.md`.

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -209,7 +212,14 @@ def self.branch_exists?(branch_name)
...
       # @raise [UserError] Raises a user_error! and interrupts the lane if we are not on the expected branch.
       #
+      # @deprecated This method is going to be removed soon. Fastlane has a built-in `ensure_git_branch` action that can be used instead.
+      #             After updating to Fastlane version `2.217.0` or later, the `FL_GIT_BRANCH_DONT_USE_ENV_VARS` environment
+      #             variable can be set to `true` to disable the use of environment variables for the `git_branch` action that
+      #             `ensure_git_branch` relies on. This will make `ensure_git_branch` work as expected in CI environments.
+      #             See https://github.com/fastlane/fastlane/pull/21597 for more details.
+      #
       def self.ensure_on_branch!(branch_name)
+        UI.message("DEPRECATED: The `ensure_on_branch!` Release Toolkit method will be removed soon. Please use Fastlane's `ensure_git_branch` action instead.")
```
AliSoftware: I wonder if we shouldn't also mention the env var fix not just in the `@deprecated` doc comment but also in this `UI.message` (which should probably be `UI.deprecated` instead, btw), for extra visibility?

```suggestion
        UI.deprecated("The `ensure_on_branch!` method will soon be removed from `release-toolkit`. Please use fastlane's `ensure_git_branch` action (+ potentially set `FL_GIT_BRANCH_DONT_USE_ENV_VARS=true`) instead.")
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -69,7 +69,10 @@ def self.checkout_and_pull(branch)
 
       # Update every submodule in the current git repository
       #
+      # @deprecated This method is going to be removed soon. Fastlane has a built-in `git_submodule_update` action that can be used instead.
+      #
       def self.update_submodules
+        UI.message("DEPRECATED: The `update_submodules` Release Toolkit method will be removed soon. Please use Fastlane's `git_submodule_update` action instead.")
```
```suggestion
        UI.deprecated("The `update_submodules` method will soon be removed from `release-toolkit`. Please use fastlane's `git_submodule_update` action instead.")
```

AliSoftware (approved):
Approved with wording tweak suggestions

PR #540, opened 2024-01-26
Update VSCode .editorconfig

AliSoftware:
Especially to make the IDEs (VSCode, RubyMine)'s auto-formatting-as-you-type better match our `rubocop` settings.


mokagio (approved):
I go in and out of using VSCode. The formatting-as-you-type with the EditorConfig plugin is definitely a reason to use it more consistently.

PR #544, opened 2024-02-05
Ensure `comment_on_pr` formats comments as Markdown

AliSoftware:
## Why?

The `comment_on_pr` action inserts a `<!-- REUSE_ID: … -->` hidden HTML comment in the comment body as a marker to identify comments and be able to update them (instead of creating new one) when appropriate.

But so far we didn't add any newlines after that marker comment. As a result, GitHub considered that, since the paragraph started with some HTML (that marker comment), the rest of the paragraph should also be interpreted as HTML, not Markdown.

This results in comments like [this one](https://github.tumblr.net/Tumblr/ios/pull/25462#issuecomment-649212) not being interpreted and formatted as Markdown.

## How?

The fix is simple: add newlines after the marker comment, so that it is considered as its own separate paragraph, thus allowing it to be interpreted as Markdown even if the first paragraph is a HTML comment. See [GitHub Flavored Markdown's spec on HTML blocks' end conditions](https://github.github.com/gfm/#html-blocks).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- Fixed `comment_on_pr` to allow first paragraph of the comment to still be interpreted as Markdown. [#544]
 
 ### Internal Changes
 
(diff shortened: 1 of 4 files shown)


iangmaia (approved):
I had also noticed it on Friday, but you beat me to fixing it 👍

PR #547, opened 2024-02-26
Refactor `update_pull_requests_milestone` to `update_assigned_milestone`

AliSoftware:
## What & Why

This updates the `update_pull_requests_milestone` action so it also includes issues when fetching the list of things to update the milestone of, not just PRs.

Since the action's name and parameters are too related with the "PR" wording, and we already have a breaking change in the `CHANGELOG` pending to be released (making the next version a major bump anyway), I've decided to use the occasion to rename the action and its parameters to make them more agnostic to the "PR" wording and be more generic:
 - Renamed the action from `update_pull_requests_milestone` to `update_assigned_milestone`
 - Renamed the `pr_numbers` parameter to `numbers`
 - Renamed the `pr_comment` parameter to `comment`


AliSoftware:
cc @mokagio FYI as I think you're the only one, besides Tumblr mobile repos, to have started using this action.

AliSoftware:
FYI: Also tested the renamed action with new parameter names manually live (from `fastlane run` in the command line) with our `release-toolkit-demo` repo as a guinea pig: Issues [44](https://github.com/wordpress-mobile/release-toolkit-demo/issues/44), [45](https://github.com/wordpress-mobile/release-toolkit-demo/issues/45) & [47](https://github.com/wordpress-mobile/release-toolkit-demo/issues/47), and PR [48](https://github.com/wordpress-mobile/release-toolkit-demo/issues/48)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_assigned_milestone_action.rb
```diff
@@ -74,22 +74,22 @@ def self.available_options
...
-          FastlaneCore::ConfigItem.new(key: :pr_numbers,
-                                       description: 'The PR numbers to update the milestone of',
+          FastlaneCore::ConfigItem.new(key: :numbers,
+                                       description: 'The PR and/or issue numbers to update the milestone of',
                                        optional: true,
                                        type: Array,
                                        conflicting_options: [:from_milestone]),
           FastlaneCore::ConfigItem.new(key: :from_milestone,
-                                       description: 'The version (milestone title\'s start) for which we want to update all open PRs of to a new milestone',
+                                       description: 'The version (milestone title\'s start) for which we want to update all open PRs and issues of to a new milestone',
```
iangmaia: Small suggestion trying to make it a bit easier to read:
```suggestion
                                       description: 'The version (start of the milestone title) for which we intend to update all open PRs and issues to a new milestone',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_assigned_milestone_action.rb
```diff
@@ -74,22 +74,22 @@ def self.available_options
...
                                        optional: true,
                                        type: String,
-                                       conflicting_options: [:pr_numbers]),
+                                       conflicting_options: [:numbers]),
           FastlaneCore::ConfigItem.new(key: :to_milestone,
-                                       description: 'The version (milestone title\'s start) for the new milestone to assign to the targeted PRs. Pass nil to unset the milestone',
+                                       description: 'The version (milestone title\'s start) for the new milestone to assign to the targeted PRs and issues. Pass nil to unset the milestone',
```
iangmaia:
```suggestion
                                       description: 'The version (start of the milestone title) for the new milestone to assign to the targeted PRs and issues. Pass nil to unset the milestone',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_assigned_milestone_action.rb
```diff
@@ -3,68 +3,68 @@
 
 module Fastlane
   module Actions
-    class UpdatePullRequestsMilestoneAction < Action
+    class UpdateAssignedMilestoneAction < Action
```
iangmaia, line 6: Again naming 😅
In one hand it is easy to first think "assigned to what?", but at the same time, I think you can only assign a milestone to PRs and issues? In any case, I wonder if being more explicit (e.g. `UpdatePullRequestsAndIssuesAssignedMilestoneAction`) wouldn't make it clearer.

AliSoftware: > I think you can only assign a milestone to PRs and issues?

Indeed.

> I wonder if being more explicit (e.g. UpdatePullRequestsAndIssuesAssignedMilestoneAction) wouldn't make it clearer

Yeah, but tbh the `pull_requests` in the original name already felt like a mouthful, so `update_pull_requests_and_issues_milestone` felt a bit too long as well…

But open to continue brainstorming on ideas if you have other suggestions 👍

iangmaia: And thinking a bit more, perhaps `Assigned` is redundant 🤔 

What about:
- `MilestoneUpdaterAction` / `MilestoneUpdateAction` / `UpdateMilestoneAction`
- `UpdatePullRequestsAndIssuesMilestoneAction`
- `UpdateMilestoneForPullRequestsAndIssuesAction`
- `MilestoneIssueUpdateAction` (considering PRs are also issues, though that might be less obvious at first glance)

iangmaia (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_assigned_milestone_action.rb
iangmaia: _[Approved the PR; feel free to merge it if you don't feel any of these or another variation improves the current name; I don't feel very strongly about it, but thought we could perhaps find a more fitting one]_

AliSoftware: - I thought about `upldate_milestone` / `milestone_update` but the problem with that naming is that it suggests that the action updates the milestone… as in the milestone's own title or due date, etc.
 - The 2nd and 3rd suggestion have the same problem as mentioned in above comment: they're a bit of a mouthful 😓 
 - I also thought about using just "issue" in the wording, considering that in the GitHub API parlance, pull requests are also issues (and the search API includes both issues and PRs when using `type:issue`)… But I felt like this was not obvious enough, especially as in common parlance when we talk about "GitHub Issues" we don't usually mean to include PRs. And as a result, I feel like including "issue" in the action name… might actually have the opposite effect of adding confusion, letting users of the action believe that it would only act on issues but not on PRs…

Boy, naming is hard 😅

oguzkocer: Found myself here from a "call for input" on Slack, but I only looked at this specific discussion. Hope my thoughts will prove useful!

---

I think the original name `UpdateAssignedMilestoneAction` is good and descriptive.

I agree with @AliSoftware that calling it `UpdateMilestoneAction` would be interpreted as updating the milestone itself, which means calling [this API endpoint](https://docs.github.com/en/rest/issues/milestones?apiVersion=2022-11-28#update-a-milestone) that GitHub also titles as `Update a milestone`.

The current action instead calls the [updating issue GitHub endpoint](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue), but it only handles the `milestone`. Since it's handling a single endpoint and there is really nothing else you can assign a milestone to, I think adding the `Assigned` is all we need.

Alternative could be `UpdateIssueMilestoneAction` or `UpdateMilestoneForIssueAction`, but I don't like these as much.

I also don't think the phrase `pull request` is necessary to include in the name. In GitHub, `pull request` is a type of issue, it's not a separate thing. When we do use `pull request`, I think we should do it because it only handles the `pull request` type, the same way this specific action only handles the `milestone` of the issue.

PR #557, opened 2024-04-10
Refactor `ios_check_beta_deps` action to make it more resilient & detect Pods referenced by commit/branch

### LINKED ISSUE #216: Make `ios_check_beta_deps` use `Podfile.lock`
## What

The action `ios_check_beta_deps` seems to only check the `Podfile` for beta pods.

We should instead check the `Podfile.lock` for those, so that the action would also detect transitive dependencies, and not just root/direct pods.

## Example

For example if we have `pod 'WordPressAuthenticator', '~> 1.34.0'` in our `Podfile` but don't have the pod `WordPressKit` (a transitive dependency of `WordPressAuthenticator`) _explicitly_ listed in the `Podfile`, this transitive dependency can end up being resolved to a beta version anyway when `WordPressAuthenticator`'s own dependencies are resolved, but the `ios_check_beta_deps` action won't catch that this transitive dependency is in beta and needs a new release.

## How

The `Podfile.lock` is a valid YAML file, so it's pretty easy to parse it in ruby and extract the pods that ends up having been resolved with a `*-beta` version:

```ruby
require 'yaml'
y = YAML.load_file('Podfile.lock')
pods_tree = y['PODS'] # An array containing a mix of `Hash<String, Array>` and `String` entries
resolved_pods = pods_tree.map do |p|
  # If the entry is a Hash, it only has one key – "PodName (x.y.z)" whose value is that pod's transitive dependencies' constraints
  return p.keys.first if p.is_a?(Hash)
  # If the entry is a String, this is a "PodName (x.y.z)" that doesn't have any transitive dependencies
  return p
end
beta_pods = resolved_pods.select { |p| p =~ /\(.*-beta\)/ } # filter to keep only the "PodName (x.y.z-beta)" entries
```

AliSoftware:
## What does it do?

> [!NOTE]
> See discussion in p1712630806132789-slack-CC7L49W13

 - Allow detection of Pods referenced by commit or branch (while previous implementation only detected pods whose version was `-beta`, but not pods referenced by commit/branch)
 - Make detection more resilient by using `Podfile.lock` instead of the `Podfile`, thus addressing https://github.com/wordpress-mobile/release-toolkit/issues/216
 - Make the action return the list of detected pods as well as the message, so that the caller can then decide to e.g. call `buildkite_annotate(message: result[:message])` to make the warning more visible in Buildkite.

## Testing

 - Updated the `Gemfile` of `woocommerce-ios` to point the `release-toolkit` to this branch
 - Ran `bundle install && bundle exec fastlane run ios_check_beta_deps`
    - I was checked out on [this wcios commit](https://github.com/woocommerce/woocommerce-ios/tree/601e819d28973b47f10a700d220976cf44c86c92) when I ran that, thus checking [this state of the `Podfile.lock`](https://github.com/woocommerce/woocommerce-ios/blob/601e819d28973b47f10a700d220976cf44c86c92/Podfile.lock)
 - Got the following output:
    image
 - Did the same test on `wordpress-ios` repo, and got a green light:
    image


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_check_beta_deps.rb
```diff
@@ -36,18 +61,45 @@ def self.details
...
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :report_branches,
+            description: 'Whether to report pods referenced by branch',
+            default_value: true,
+            type: Boolean
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :report_version_pattern,
+            description: 'Report any pod whose tag name or version constraint matches this Regex pattern. Set to empty string to ignore',
```
AliSoftware:
```suggestion
            description: 'Report any pod whose resolved version matches this Regex pattern. Set to empty string to ignore',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_check_beta_deps.rb
```diff
@@ -1,31 +1,56 @@
...
         require_relative '../../helper/ios/ios_git_helper'
 
-        beta_pods = []
-        File.open(params[:podfile]).each do |li|
-          beta_pods << li if li.match('^\s*\t*pod.*beta')
+        yaml = YAML.load_file(params[:lockfile])
+        non_stable_pods = {} # Key will be pod name, value will be reason for flagging
+
+        # Find pods referenced by commit and branch to a repo
+        yaml['EXTERNAL SOURCES'].each do |pod, options|
```
AliSoftware: I just realized that some `Podfile.lock` files might not have this section at all (if they don't have any external dependency), so better cover for the `nil` case.

```suggestion
        yaml['EXTERNAL SOURCES']&.each do |pod, options|
```

Ideally we should add a spec example to cover that too—[the one from Simplenote-macOS](https://github.com/Automattic/simplenote-macos/blob/trunk/Podfile.lock) can serve as a fixture for that case.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_check_beta_deps.rb
```diff
@@ -1,31 +1,56 @@
...
         UI.important(message)
+        { message: message, pods: non_stable_pods }
       end
 
       #####################################################
       # @!group Documentation
       #####################################################
 
+      ALL_PODS_STABLE_MESSAGE = 'All pods are pointing to a stable version. You can continue with the code freeze.'.freeze
+      NON_STABLE_PODS_MESSAGE = "Please create a new stable version of those pods and update the Podfile to the newly released version before continuing with the code freeze:\n".freeze
```
spencertransier: Why declare these constants here instead of at the top of the `run` method?

AliSoftware: We need those constant to be accessible from outside the code of `def self.run` because they are also referenced by our unit tests (<a href="https://github.com/wordpress-mobile/release-toolkit/pull/557/files#diff-133d41dbf229505918f2ce42db99fcb0f2a5171eeb424d5666e5b66e0ea46353R11">here</a> and <a href="https://github.com/wordpress-mobile/release-toolkit/pull/557/files#diff-133d41dbf229505918f2ce42db99fcb0f2a5171eeb424d5666e5b66e0ea46353R123">here</a>) 😉

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_check_beta_deps.rb
AliSoftware: @spencertransier I've fixed this (and added test coverage for it) via dc5992fd2fbe7e4459b134ce6b1f82805254635a + b1a13d9b0fde57223dc12f28ef28b175d614c239, so PR is now ready for review again 🙇

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_check_beta_deps.rb
AliSoftware: While we couldn't move them at the top of the `run` method, I've moved them at the top of the `class` definition in 25420dd374a398cdc2321a1a8dd1294ec22d481a, at least

spencertransier: Ah, got it. Thank you!

spencertransier (approved):
The refactor and tests look great to me. The logic makes sense and this will be a huge improvement to the code freeze automation. 👏

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 
 ### Breaking Changes
 
-_None_
+ - Make `ios_check_beta_deps` use the `Podfile.lock` instead of `Podfile` for its detection, and also be able to detect Pods referenced by commits and branches.
```
```suggestion
- Make `ios_check_beta_deps` use the `Podfile.lock` instead of `Podfile` for its detection, and also be able to detect Pods referenced by commits and branches.  
```

MIGRATION.md
```diff
@@ -1,5 +1,9 @@
 # Migration Instructions for Major Releases
 
+## From `10.0.0` to `11.0.0`
+
+ - The `ios_check_beta_deps` now uses the `Podfile.lock` instead of `Podfile` for its detection. If you called this action with an explicit `podfile: …` argument, you'll have to update the call to use `lockfile:` instead—or remove that argument completely from your call and rely on its default value being `Podfile.lock`.
```
mokagio:
```suggestion
- The `ios_check_beta_deps` now uses the `Podfile.lock` instead of `Podfile` for its detection. If you called this action with an explicit `podfile: …` argument, you'll have to update the call to use `lockfile:` instead—or remove that argument completely from your call and rely on its default value being `Podfile.lock`.
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_check_beta_deps.rb
```diff
@@ -1,25 +1,50 @@
+require 'yaml'
+
 module Fastlane
   module Actions
     class IosCheckBetaDepsAction < Action
+      ALL_PODS_STABLE_MESSAGE = 'All pods are pointing to a stable version. You can continue with the code freeze.'.freeze
+      NON_STABLE_PODS_MESSAGE = "Please create a new stable version of those pods and update the Podfile to the newly released version before continuing with the code freeze:\n".freeze
```
mokagio, line 7: Nitpick. These messages assume the action runs as part of the code freeze. That is true in practice, and the action name sort of suggests it, too. Still, I'd suggest removing the code freeze mention to make it agnostic.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_check_beta_deps.rb
```diff
@@ -1,25 +1,50 @@
...
+          non_stable_pods[pod] = 'branch' if params[:report_branches] && options.key?(:branch)
+        end
+
+        # Find pods whose resolved version matches the regex
+        version_pattern = params[:report_version_pattern]
+        unless version_pattern.nil? || version_pattern.empty?
+          regex = begin
+            Regexp.new(version_pattern)
+          rescue RegexpError
+            UI.user_error!("Invalid regex pattern: `#{version_pattern}`")
```
mokagio, line 28: Side note. I wanted to share appreciation for this kind of error handling where the things that's wrong is clearly reported in the error message. Recently, I've bump my head against some vague error messages; seeing this is encouraging. Thanks!

mokagio (approved):
With tests, too 👨‍🍳 💋

PR #563, opened 2024-04-19
Add action to download metadata from OneSky

AliSoftware:
## What

This is a WIP on migrating our code related to OneSky (the translation platform used by Tumblr apps) from Tumblr-iOS and Tumblr-Android's `Fastfiles` into the `release-toolkit`

This work has been paused as other priorities came in the way of its progress and now other projects are taking precedence, but still I wanted to push the existing code to keep track of the progress and be able to come back to it at a later time.

## Status

Right now the code for the `download_metadata_from_onesky` (which handles store metadata in `fastlane/metadata/*/*.txt`) should be working, and the only thing left for that action is to add more unit tests to cover more cases. So that action pretty close to be ready to ship.

Note that the initial plan was to also add `ios_download_translations_from_onesky` (to download app strings in `Localizable.strings`) and `android_download_translations_from_onesky` (for app strings in `values/strings.xml`) actions too. But in practice these could be added separately.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/models/locale_map.rb
```diff
@@ -0,0 +1,120 @@
...
+    #
+    def convert(locale:, from:, to:)
+      @locales.find { |lm| lm.send(from) == locale }&.send(to)
+    end
+
+    # The set of common locale codes mappings known to us
+    #
+    def self.default
+      # TODO: Add :android locale codes
+      # TODO: Add :glotpress locale codes
```
AliSoftware, line 95: 🎗️  Reminder to add values for android and glotpress too. Once we have values for GlotPress, we might also want to expose a `def self.mag16` convenience method too

lib/fastlane/plugin/wpmreleasetoolkit/models/locale_map.rb
```diff
@@ -0,0 +1,120 @@
...
+    # @param [Symbol] from The service/context to convert the locale from. Should be one of `LocaleMap.members`
+    # @param [Symbol] to The service/context to convert the locale to. Should be one of `LocaleMap.members`
+    #
+    def convert(locale:, from:, to:)
+      @locales.find { |lm| lm.send(from) == locale }&.send(to)
+    end
+
+    # The set of common locale codes mappings known to us
+    #
+    def self.default
```
AliSoftware, line 93: 💡  API idea: also provide a `def self.only(options)` method which will expect `options` to be a `Hash` with a single key which must be one of the `LocaleMap.members`, and a value being an `Array<String>`, so that we can call `LocalesMap.only(onesky: %w[fr pt-BR en es it])` to get a subset of the known locales.

We can then implement `def self.mag16` by just calling that `self.only` method with the 16 expected locales, and additionally make it easy for client apps to provide just the custom list of locales they support.

spec/download_metadata_from_onesky_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+            expect(File).not_to exist(File.join(tmp_dir, 'metadata', path.to_s))
+          else
+            expect(File).to exist(File.join(tmp_dir, 'metadata', path.to_s))
+            expect(File.read(File.join(tmp_dir, 'metadata', path.to_s))).to eq(content)
+          end
+        end
+      end
+    end
+
+    it 'works' do
```
AliSoftware, line 43: WIP. We need to add a lot more tests.

Including ones around composite keys (keys which contain `.` in their name, [which OneSky exports as a tree of keys and subkeys split around that `.`](https://github.com/wordpress-mobile/release-toolkit/pull/563/files#diff-332a57bea137650989cb66f84584f9a9bcb128bf42118e3f1c30d9c54e341827R65))

AliSoftware:
Closing this as it is no longer relevant now that OneSky is down and we're migrating to CrowdIn
(see https://linear.app/a8c/project/migrate-from-onesky-to-crowdin-704269ab5e79/overview)

Still cc @iangmaia in case you'd want to take inspiration for [the `LocaleMap` struct I started to introduced in that PR](https://github.com/wordpress-mobile/release-toolkit/pull/563/files#diff-8910842a4885e9d3dfd49c31d2f1317b66c49d02b81ad60fcb6d1866776f2613) back then to map our Mag16 locale codes to the different locale code formats (iOS/AppStore/Android/GooglePlay/CrowdIn/GlotPress…) so that we don't constantly repeat those mappings in each repo's `Fastfile`, as this could still be a useful idea to explore when you'll implement the CrowdIn actions in `release-toolkit` at some point as part of your project.

PR #570, opened 2024-07-11
Create Release Backmerge PR Action

iangmaia:
## What does it do?

This PR adds the `create_release_backmerge_pull_request` action.

Its goal is to facilitate the creation of Pull Requests from a release branch back to the main branch or, alternatively, to any currently ongoing version (e.g. in a case of a hotfix, there might be already new version(s) being released).

It standardize some common tasks such as creating the PR from an intermediate branch based on the release branch, assigning a milestone to the newly created PR, generating a illustrative PR body text and so on.

An example call looks like:
```ruby
create_release_backmerge_pull_request(
  repository: GITHUB_REPO,
  release_branch: 'release/30.6',
  default_branch: 'trunk',
  milestone_title: '30.6'
)
```

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -83,7 +83,11 @@ module Fastlane
       # @param milestone [String] the milestone to associate with the pull request.
       # @param reviewers [Array<String>] the individual reviewers for the pull request.
       # @param team_reviewers [Array<String>] the team reviewers for the pull request.
-      # @return [void]
+      # @param intermediate_branch_created_callback [Proc] A callback to call after having created the intermediate branch
+      #        to allow the caller to e.g. add new commits on it before the PR is created. The callback takes two parameters: the base branch and the intermediate branch
+      #
+      # @return [String] The URL of the created Pull Request, or `nil` if no PR was created.
+      #
       def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, labels:, milestone:, reviewers:, team_reviewers:, intermediate_branch_created_callback:)
         intermediate_branch = "merge/#{head_branch.gsub('/', '-')}-into-#{base_branch.gsub('/', '-')}"
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -11,28 +11,30 @@ module Fastlane
         repository = params[:repository]
         release_branch = params[:release_branch]
         default_branch = params[:default_branch]
-        target_branches_param = params[:target_branches]
+        target_branches = params[:target_branches]
         labels = params[:labels]
         milestone_title = params[:milestone_title]
+        reviewers = params[:reviewers]
+        team_reviewers = params[:team_reviewers]
 
         unless release_branch.start_with?('release/')
           UI.user_error!('`release_branch` must start with `release/`')
         end
 
-        if target_branches_param.include?(release_branch)
+        if target_branches.include?(release_branch)
           UI.user_error!('`target_branches` must not contain `release_branch`')
         end
 
         github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
         target_milestone = milestone_title.nil? ? nil : github_helper.get_milestone(repository, milestone_title)
 
-        target_branches = if target_branches_param.empty?
-                            determine_target_branches(release_branch.delete('release/'), default_branch)
-                          else
-                            target_branches_param
-                          end
+        final_target_branches = if target_branches.empty?
+                                  determine_target_branches(release_version: release_branch.delete('release/'), default_branch: default_branch)
+                                else
+                                  target_branches
+                                end
 
-        target_branches.map do |target_branch|
+        final_target_branches.map do |target_branch|
           Fastlane::Helper::GitHelper.checkout_and_pull(release_branch)
 
           create_backmerge_pr(
@@ -47,11 +49,10 @@ module Fastlane
         end
       end
 
-      def self.determine_target_branches(release_version, default_branch)
-        release_branches = Actions.sh('git', 'branch', '-r', '-l', 'origin/release/*').chomp
+      def self.determine_target_branches(release_version:, default_branch:)
+        release_branches = Actions.sh('git', 'branch', '-r', '-l', 'origin/release/*').chomp.split("\n")
 
         all_release_branches_versions = release_branches
-                                        .split("\n")
                                         .map { |branch| branch.match(%r{origin/release/([0-9.]*)})&.captures&.first }
                                         .compact
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -25,6 +25,10 @@ module Fastlane
         target_milestone = milestone_title.nil? ? nil : github_helper.get_milestone(repository, milestone_title)
 
         final_target_branches = if target_branches.empty?
+                                  unless source_branch.start_with?('release/')
+                                    UI.user_error!('`source_branch` must start with `release/`')
+                                  end
+
                                   determine_target_branches(release_version: source_branch.delete('release/'), default_branch: default_branch)
                                 else
                                   target_branches
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -48,7 +48,7 @@ module Fastlane
       end
 
       def self.determine_target_branches(release_version, default_branch)
-        release_branches = Actions.sh('git branch -r -l "origin/release/*"').chomp
+        release_branches = Actions.sh('git', 'branch', '-r', '-l', 'origin/release/*').chomp
 
         all_release_branches_versions = release_branches
                                         .split("\n")
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -103,9 +103,7 @@ module Fastlane
         # if there's a callback, make sure it didn't switch branches
         other_action.ensure_git_branch(branch: "^#{intermediate_branch}/") unless intermediate_branch_created_callback.nil?
 
-        has_commits_between_head_and_base = Fastlane::Helper::GitHelper.has_commits_between?(base_ref: base_branch, head_ref: head_branch)
-
-        unless has_commits_between_head_and_base
+        if Fastlane::Helper::GitHelper.point_to_same_commit?(base_branch, head_branch)
           UI.error("No differences between #{head_branch} and #{base_branch}. Skipping PR creation.")
           return nil
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -134,7 +134,7 @@ module Fastlane
                                        default_value: DEFAULT_BRANCH,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :target_branches,
-                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a version greater than `release_version`. If none are found, it will target `default_branch`',
+                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a `x.y.z` version greater than the version in source branch's name. If none are found, it will target `default_branch`',
                                        optional: true,
                                        default_value: [],
                                        type: Array),
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -43,8 +43,10 @@ module Fastlane
             title: "Merge #{release_branch} into #{target_branch}",
             head_branch: release_branch,
             base_branch: target_branch,
+            labels: labels,
             milestone: target_milestone&.number,
-            labels: labels
+            reviewers: reviewers,
+            team_reviewers: team_reviewers
           )
         end
       end
@@ -63,7 +65,7 @@ module Fastlane
         target_branches
       end
 
-      def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, milestone:, labels:)
+      def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, labels:, milestone:, reviewers:, team_reviewers:)
         intermediate_branch = "merge/#{head_branch.gsub('/', '-')}-into-#{base_branch.gsub('/', '-')}"
         Fastlane::Helper::GitHelper.create_branch(intermediate_branch)
 
@@ -90,7 +92,9 @@ module Fastlane
           head: intermediate_branch,
           base: base_branch,
           labels: labels,
-          milestone: milestone
+          milestone: milestone,
+          reviewers: reviewers,
+          team_reviewers: team_reviewers
         )
       end
 
@@ -107,7 +111,7 @@ module Fastlane
       end
 
       def self.return_value
-        'The list of backmerge PRs created'
+        'The list of the created backmerge Pull Request URLs'
       end
 
       def self.details
@@ -148,6 +152,14 @@ module Fastlane
                                        description: 'The title of the milestone to assign to the created PRs',
                                        optional: true,
                                        type: String),
+          FastlaneCore::ConfigItem.new(key: :reviewers,
+                                       description: 'An array of GitHub users that will be assigned to the pull request',
+                                       optional: true,
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :team_reviewers,
+                                       description: 'An array of GitHub team slugs that will be assigned to the pull request',
+                                       optional: true,
+                                       type: Array),
           Fastlane::Helper::GithubHelper.github_token_config_item,
         ]
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -9,7 +9,7 @@ module Fastlane
       def self.run(params)
         token = params[:github_token]
         repository = params[:repository]
-        release_branch = params[:release_branch]
+        source_branch = params[:source_branch]
         default_branch = params[:default_branch]
         target_branches = params[:target_branches]
         labels = params[:labels]
@@ -17,31 +17,27 @@ module Fastlane
         reviewers = params[:reviewers]
         team_reviewers = params[:team_reviewers]
 
-        unless release_branch.start_with?('release/')
-          UI.user_error!('`release_branch` must start with `release/`')
-        end
-
-        if target_branches.include?(release_branch)
-          UI.user_error!('`target_branches` must not contain `release_branch`')
+        if target_branches.include?(source_branch)
+          UI.user_error!('`target_branches` must not contain `source_branch`')
         end
 
         github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
         target_milestone = milestone_title.nil? ? nil : github_helper.get_milestone(repository, milestone_title)
 
         final_target_branches = if target_branches.empty?
-                                  determine_target_branches(release_version: release_branch.delete('release/'), default_branch: default_branch)
+                                  determine_target_branches(release_version: source_branch.delete('release/'), default_branch: default_branch)
                                 else
                                   target_branches
                                 end
 
         final_target_branches.map do |target_branch|
-          Fastlane::Helper::GitHelper.checkout_and_pull(release_branch)
+          Fastlane::Helper::GitHelper.checkout_and_pull(source_branch)
 
           create_backmerge_pr(
             token: token,
             repository: repository,
-            title: "Merge #{release_branch} into #{target_branch}",
-            head_branch: release_branch,
+            title: "Merge #{source_branch} into #{target_branch}",
+            head_branch: source_branch,
             base_branch: target_branch,
             labels: labels,
             milestone: target_milestone&.number,
@@ -129,8 +125,8 @@ module Fastlane
                                        description: 'The remote path of the GH repository on which we work',
                                        optional: false,
                                        type: String),
-          FastlaneCore::ConfigItem.new(key: :release_branch,
-                                       description: 'The release branch, in the format `release/x.y.z`',
+          FastlaneCore::ConfigItem.new(key: :source_branch,
+                                       description: 'The source branch to create a backmerge PR from, in the format `release/x.y.z`',
                                        optional: false,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :default_branch,
@@ -139,7 +135,7 @@ module Fastlane
                                        default_value: DEFAULT_BRANCH,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :target_branches,
-                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a `x.y.z` version greater than the version in source branch's name. If none are found, it will target `default_branch`',
+                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a `x.y.z` version greater than the version in source branch\'s name. If none are found, it will target `default_branch`', # rubocop:disable Layout/LineLength
                                        optional: true,
                                        default_value: [],
                                        type: Array),
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -83,15 +83,9 @@ module Fastlane
       # @param team_reviewers [Array<String>] the team reviewers for the pull request.
       # @return [void]
       def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, labels:, milestone:, reviewers:, team_reviewers:)
-        # Check for differences between head_branch and base_branch, using the three-dot diff ('base_branch...head_branch') instead
-        # of the two-dot diff ('base_branch..head_branch').
-        # This ensures that we are checking for differences as if a merge was to happen, providing a more accurate representation of what changes
-        # would be introduced by merging head_branch into base_branch.
-        #
-        # See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons
-        diff_output = Actions.sh('git', 'diff', "#{base_branch}...#{head_branch}").strip
+        commits_between_head_and_base = Fastlane::Helper::GitHelper.count_commits_between(base_ref: base_branch, head_ref: head_branch)
 
-        if diff_output.empty?
+        if commits_between_head_and_base.zero?
           UI.message("No differences between #{head_branch} and #{base_branch}. Skipping PR creation.")
           return nil
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -1,0 +1,163 @@
+require 'fastlane/action'
+require_relative '../../helper/github_helper'
+
+module Fastlane
+  module Actions
+    class CreateReleaseBackmergePullRequestAction < Action
+      DEFAULT_BRANCH = 'trunk'.freeze
+
+      def self.run(params)
+        token = params[:github_token]
+        repository = params[:repository]
+        release_branch = params[:release_branch]
+        default_branch = params[:default_branch]
+        target_branches_param = params[:target_branches]
+        labels = params[:labels]
+        milestone_title = params[:milestone_title]
+
+        unless release_branch.start_with?('release/')
+          UI.user_error!('`release_branch` must start with `release/`')
+        end
+
+        if target_branches_param.include?(release_branch)
+          UI.user_error!('`target_branches` must not contain `release_branch`')
+        end
+
+        github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
+        target_milestone = milestone_title.nil? ? nil : github_helper.get_milestone(repository, milestone_title)
+
+        target_branches = if target_branches_param.empty?
+                            determine_target_branches(release_branch.delete('release/'), default_branch)
+                          else
+                            target_branches_param
+                          end
+
+        pr_urls = []
+
+        target_branches.each do |target_branch|
+          Fastlane::Helper::GitHelper.checkout_and_pull(release_branch)
+
+          pr_urls << create_backmerge_pr(
+            token: token,
+            repository: params[:repository],
+            title: "Merge #{release_branch} into #{target_branch}",
+            head_branch: release_branch,
+            base_branch: target_branch,
+            milestone: target_milestone&.number,
+            labels: labels
+          )
+        end
+
+        pr_urls
+      end
+
+      def self.determine_target_branches(release_version, default_branch)
+        release_branches = Actions.sh('git branch -r -l "origin/release/*"').chomp
+
+        all_release_branches_versions = release_branches
+                                        .split("\n")
+                                        .map { |branch| branch.match(%r{origin/release/([0-9.]*)})&.captures&.first }
+                                        .compact
+
+        target_branches = all_release_branches_versions.select { |branch| Gem::Version.new(branch) > Gem::Version.new(release_version) }
+                                                       .map { |v| "release/#{v}" }
+        target_branches = [default_branch] if target_branches.empty?
+
+        target_branches
+      end
+
+      def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, milestone:, labels:)
+        intermediate_branch = "merge/#{head_branch.gsub('/', '-')}-into-#{base_branch.gsub('/', '-')}"
+        Fastlane::Helper::GitHelper.create_branch(intermediate_branch)
+
+        other_action.push_to_git_remote(tags: false)
+
+        pr_body = <<~BODY
+          Merging `#{head_branch}` into `#{base_branch}`.
+
+          Via intermediate branch `#{intermediate_branch}`, to help fix conflicts if any:
+          ```
+          #{head_branch.rjust(40)}  ----o-- - - -
+          #{' ' * 40}       \\
+          #{intermediate_branch.rjust(40)}        `---.
+          #{' ' * 40}             \\
+          #{base_branch.rjust(40)}  ------------x- - -
+          ```
+        BODY
+
+        other_action.create_pull_request(
+          api_token: token,
+          repo: repository,
+          title: title,
+          body: pr_body,
+          head: intermediate_branch,
+          base: base_branch,
+          labels: labels,
+          milestone: milestone
+        )
+      end
+
+      def self.description
+        'Creates backmerge PRs for a release branch into target branches'
+      end
+
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.return_type
+        :array_of_strings
+      end
+
+      def self.return_value
+        'The list of backmerge PRs created'
+      end
+
+      def self.details
+        <<~DETAILS
+          This action creates backmerge Pull Requests from a release branch into one or more target branches.
+
+          It can be used to ensure that changes from a release branch are merged back into other branches, such as newer release branches or the main development branch (e.g., `trunk`).
+        DETAILS
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :repository,
+                                       env_name: 'GHHELPER_REPOSITORY',
+                                       description: 'The remote path of the GH repository on which we work',
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :release_branch,
+                                       description: 'The release branch, in the format `release/x.y.z`',
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       description: 'The default branch to target if no newer release branches exist',
+                                       optional: true,
+                                       default_value: DEFAULT_BRANCH,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :target_branches,
+                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a version greater than `release_version`. If none are found, it will target `default_branch`',
+                                       optional: true,
+                                       default_value: [],
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :labels,
+                                       description: 'The labels that should be assigned to the backmerge PRs',
+                                       optional: true,
+                                       default_value: [],
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :milestone_title,
+                                       description: 'The title of the milestone to assign to the created PRs',
+                                       optional: true,
+                                       type: String),
+          Fastlane::Helper::GithubHelper.github_token_config_item,
+        ]
+      end
+
+      def self.is_supported?(platform)
+        true
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -16,6 +16,7 @@ module Fastlane
         milestone_title = params[:milestone_title]
         reviewers = params[:reviewers]
         team_reviewers = params[:team_reviewers]
+        intermediate_branch_created_callback = params[:intermediate_branch_created_callback]
 
         if target_branches.include?(source_branch)
           UI.user_error!('`target_branches` must not contain `source_branch`')
@@ -46,7 +47,8 @@ module Fastlane
             labels: labels,
             milestone: target_milestone&.number,
             reviewers: reviewers,
-            team_reviewers: team_reviewers
+            team_reviewers: team_reviewers,
+            intermediate_branch_created_callback: intermediate_branch_created_callback
           )
         end.compact
       end
@@ -82,7 +84,7 @@ module Fastlane
       # @param reviewers [Array<String>] the individual reviewers for the pull request.
       # @param team_reviewers [Array<String>] the team reviewers for the pull request.
       # @return [void]
-      def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, labels:, milestone:, reviewers:, team_reviewers:)
+      def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, labels:, milestone:, reviewers:, team_reviewers:, intermediate_branch_created_callback:)
         commits_between_head_and_base = Fastlane::Helper::GitHelper.count_commits_between(base_ref: base_branch, head_ref: head_branch)
 
         if commits_between_head_and_base.zero?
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+        repository = params[:repository]
+        release_branch = params[:release_branch]
+        default_branch = params[:default_branch]
+        target_branches_param = params[:target_branches]
+        labels = params[:labels]
+        milestone_title = params[:milestone_title]
+
+        unless release_branch.start_with?('release/')
+          UI.user_error!('`release_branch` must start with `release/`')
+        end
```
AliSoftware: While I like the safety net, I am not sure we'd want to enforce this strictly?

i.e. I can imagine us using this action to backmerge other branches in the future (not that I can think of a particular use case we'd current have right now… but it seems a bit too strict to enforce this restriction anyway?)

I know this is a requirement for `determine_target_branches` to work, but for the case when the caller  would provide `target_branches` param we should still allow the source branch to not forcibly be `release/*`?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+  module Actions
+    class CreateReleaseBackmergePullRequestAction < Action
+      DEFAULT_BRANCH = 'trunk'.freeze
+
+      def self.run(params)
+        token = params[:github_token]
+        repository = params[:repository]
+        release_branch = params[:release_branch]
+        default_branch = params[:default_branch]
+        target_branches_param = params[:target_branches]
```
AliSoftware: 💄  I understand that you named this one differently because you used `target_branches` variable name later in the code, but for consistency I'd instead suggest to name all those with the same naming convention (`_param` suffix or not). Personally I'd suggest to name that one `target_branches` and remove the `_param` suffix, and then rename your variables on line 29 to `actual_target_branches` or something similar 🤷

AliSoftware: Also, I wonder if we should make it sure we fallback to empty array if nil (i.e. `= params[:target_branches] || []`).

While I know the `default_value` of this param is already `[]`, I wonder what would happen if the caller end up calling this action with `target_branches: nil` explicitly? (or maybe fastlane already interprets this as falling back to the default value, as opposed to only use the default value if parameter not provided explicitly? I can't remember)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+            repository: params[:repository],
+            title: "Merge #{release_branch} into #{target_branch}",
+            head_branch: release_branch,
+            base_branch: target_branch,
+            milestone: target_milestone&.number,
+            labels: labels
+          )
+        end
+
+        pr_urls
```
AliSoftware: You already have `repository = params[:repository]` defined on line 11, so maybe you should use that instead of repeating `params[:repository]` on this line 42?

Also, this whole `pr_urls = []` + `pr_urls << …` + `return pr_urls` can be replaced by a `map` 😉 

```suggestion
        target_branches.map do |target_branch|
          Fastlane::Helper::GitHelper.checkout_and_pull(release_branch)

          create_backmerge_pr(
            token: token,
            repository: repository,
            title: "Merge #{release_branch} into #{target_branch}",
            head_branch: release_branch,
            base_branch: target_branch,
            milestone: target_milestone&.number,
            labels: labels
          )
        end
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+
+      def self.determine_target_branches(release_version, default_branch)
+        release_branches = Actions.sh('git branch -r -l "origin/release/*"').chomp
+
+        all_release_branches_versions = release_branches
+                                        .split("\n")
```
AliSoftware: 💄  Only a matter of code style taste, but I'd put the `.split("\n")` on the `release_branches = ….chomp.split("\n")` personally. This is because the name of the `release_branches` variable suggests that it already contains an array, not a multi-line string 😉

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+        other_action.create_pull_request(
+          api_token: token,
+          repo: repository,
+          title: title,
+          body: pr_body,
+          head: intermediate_branch,
+          base: base_branch,
+          labels: labels,
+          milestone: milestone
+        )
```
AliSoftware: We need to protect against the case where there's nothing to merge.

 - We don't have that use case in Tumblr, because even if there hasn't been any new PR on the `release/*` branch by the time we create a new beta and a backmerge PR, we always add one more commit to the backmerge PR before creating it (to restore the version file to its state on the target branch), so our PRs there are never empty.
 - Arguably, for other apps, it's also probably very rare to have an empty PRs, because on those other apps, unlike Tumblr, usually we only trigger a new beta (and thus a backmerge PR) if there has been betafix PRs merged in the `release/*` branch, otherwise we don't do a beta at all anyway.

BUT, there can still be cases where this happen, especially if:
 - Either the fix for the bug was already cherry-picked in the target branch by the developers (i.e. they fixed the bug already both on `release/*` and `trunk` and then triggered a beta)
 - Or for some reason the CI failed mid-flight on the first try doing the beta, and the Release Manager is retrying the CI step. So if on first try the (non-empty) backmerge PR was already created… and then merged, but something else failed later in the lane, on second try the backmerge PR would be empty.

---

We could protect against that by wrapping the `create_pull_request` call within a `begin…rescue…end` for example, though it'd be better to be able to detect an empty diff ourselves, even before creating the intermediate branch and doing the GitHub API call.

AliSoftware: We might also want to protect against "a PR already exists for this branch" error (especially in case the CI failed on first try but already created the backmerge PR… and the Release Manager is now doing a retry of the lane in ReleaseV2, without having closed or merged the first backmerge PR first)?

Though maybe in those cases it's better for the action to fail explicitly, rather than to have it close the existing PR automatically before creating one anew (or rebasing the existing branch… but that feels even more risky to automate a rebase)? 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.return_type
+        :array_of_strings
+      end
+
+      def self.return_value
+        'The list of backmerge PRs created'
```
AliSoftware: Might want to indicate if what's returned is just the PR numbers of their URLs? (I'm guessing URL since the `return_type` is `array_of_strings` not `array_of_integers`, but still nice to make it clearer in the doc copy 🙂 )

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+        [
+          FastlaneCore::ConfigItem.new(key: :repository,
+                                       env_name: 'GHHELPER_REPOSITORY',
+                                       description: 'The remote path of the GH repository on which we work',
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :release_branch,
+                                       description: 'The release branch, in the format `release/x.y.z`',
+                                       optional: false,
+                                       type: String),
```
AliSoftware: I'd suggest to name that parameter something like `from` (or `branch_to_merge` or `source_branch`…) instead or `release_branch`.

Especially because when you want to create a backmerge PR from `release/12.3` to `release/12.4`, it might not be clear which of those two release branches you need to use as the `release_branch` parameter.

But also, because as mentioned in the other comment, while 99% of our use cases for this action will be with a source branch indeed being a `release/*` branch and `target_branches` being found implicitly from it, we might not want to close the door on this action also being used to backmerge any arbitrary (and possibly non-`release/`) branch into some explicitly-provided `target_branches` (if only to benefit from the intermediate branch auto-creation and nice ASCII art in PR body)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       description: 'The default branch to target if no newer release branches exist',
+                                       optional: true,
+                                       default_value: DEFAULT_BRANCH,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :target_branches,
+                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a version greater than `release_version`. If none are found, it will target `default_branch`',
```
AliSoftware:
```suggestion
                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a `x.y.z` version greater than the version in source branch's name. If none are found, it will target `default_branch`',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+                                       description: 'The labels that should be assigned to the backmerge PRs',
+                                       optional: true,
+                                       default_value: [],
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :milestone_title,
+                                       description: 'The title of the milestone to assign to the created PRs',
+                                       optional: true,
+                                       type: String),
+          Fastlane::Helper::GithubHelper.github_token_config_item,
+        ]
```
AliSoftware, line 207: We probably want to also add a `ConfigItem` for automatically adding reviewers (user or team) to the created PRs

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+            base_branch: target_branch,
+            milestone: target_milestone&.number,
+            labels: labels
+          )
+        end
+
+        pr_urls
+      end
+
+      def self.determine_target_branches(release_version, default_branch)
```
AliSoftware: Suggestion to use keyword parameters here, to make the call site more readable.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,163 @@
...
+        release_branches = Actions.sh('git branch -r -l "origin/release/*"').chomp
```
AliSoftware: I'd suggest to use the more secure approach of passing each individual command parameter as a separate value.

```suggestion
        release_branches = Actions.sh('git', 'branch', '-r', '-l', 'origin/release/*').chomp
```
The difference is that if you pass a single `String`, this will invoke a shell, which will interpret that string to split it into a command and its argument. That parsing of the command line by the shell not only is unnecessary in our case, but also opens the door to command injections, issues with spaces in arguments, or unwanted interpolations.
While if you pass each argument separately, it doesn't need to invoke a shell at all and can directly call the `git` program/process passing it each argument (already provided as separate ones) as its `argv` list, without involving any parsing of the command line to split it in arguments first.

spec/create_release_backmerge_pull_request_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+  end
+
+  def mock_pr_url(release_branch)
+    number = release_branch.delete('release/').gsub('.', '').to_i
+    "https://github.com/#{test_repo}/pull/#{number}"
+  end
+
+  def stub_git_release_branches(branches)
+    allow(Fastlane::Actions).to receive(:sh)
+      .with('git branch -r -l "origin/release/*"')
```
AliSoftware: Adjust this to the below to match my comment above from the implementation
```suggestion
      .with('git', 'branch', '-r', '-l', 'origin/release/*')
```

AliSoftware (commented):
Haven't finished reviewing the whole spec code but given the fires that have arose on my day with some releases I have to pause my review anyway. So submitting the feedback I have for now.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: > what would happen if the caller end up calling this action with target_branches: nil explicitly?

It seems Fastlane will use the default value instead of `nil`, so `[]` on this case.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,159 @@
...
+                                       description: 'The release branch, in the format `release/x.y.z`',
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :default_branch,
+                                       description: 'The default branch to target if no newer release branches exist',
+                                       optional: true,
+                                       default_value: DEFAULT_BRANCH,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :target_branches,
+                                       description: 'Array of target branches for the backmerge. If empty, the action will determine target branches by finding all `release/x.y.z` branches with a `x.y.z` version greater than the version in source branch's name. If none are found, it will target `default_branch`',
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/ea252d81f5879173f92a376c4a6dd5638604c1a9/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb#L137"></span>Lint/Syntax: unexpected token tIDENTIFIER<br />
(Using Ruby 3.2 parser; configure using <code>TargetRubyVersion</code> parameter, under <code>AllCops</code>)</td>
    </tr>
  </tbody>
</table>

dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/ea252d81f5879173f92a376c4a6dd5638604c1a9/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb#L137"></span>Lint/Syntax: unexpected token tSTRING_BEG<br />
(Using Ruby 3.2 parser; configure using <code>TargetRubyVersion</code> parameter, under <code>AllCops</code>)</td>
    </tr>
  </tbody>
</table>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,159 @@
...
+                                       optional: true,
+                                       default_value: [],
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :labels,
+                                       description: 'The labels that should be assigned to the backmerge PRs',
```
dangermattic, line 142: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/ea252d81f5879173f92a376c4a6dd5638604c1a9/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb#L142"></span>Lint/Syntax: unexpected token tCONSTANT<br />
(Using Ruby 3.2 parser; configure using <code>TargetRubyVersion</code> parameter, under <code>AllCops</code>)</td>
    </tr>
  </tbody>
</table>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,159 @@
...
+                                       optional: true,
+                                       default_value: [],
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :milestone_title,
+                                       description: 'The title of the milestone to assign to the created PRs',
```
dangermattic, line 147: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/ea252d81f5879173f92a376c4a6dd5638604c1a9/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb#L147"></span>Lint/Syntax: unexpected token tCONSTANT<br />
(Using Ruby 3.2 parser; configure using <code>TargetRubyVersion</code> parameter, under <code>AllCops</code>)</td>
    </tr>
  </tbody>
</table>

dangermattic, line 147: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/ea252d81f5879173f92a376c4a6dd5638604c1a9/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb#L147"></span>Lint/Syntax: unterminated string meets end of file<br />
(Using Ruby 3.2 parser; configure using <code>TargetRubyVersion</code> parameter, under <code>AllCops</code>)</td>
    </tr>
  </tbody>
</table>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Implemented on 0576800278293b6c094b5b563da5b5d96b272584

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Implemented on 05c6fcf2345b4bdc01b1b2a98e11a628107ef33d

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: I think `source_branch` makes it more intuitive 👍 implemented on 9ad0ef3c96198229fdc9a46b90596040847cde17

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Added on 05c6fcf2345b4bdc01b1b2a98e11a628107ef33d

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,168 @@
...
+
+        if target_branches.include?(source_branch)
+          UI.user_error!('`target_branches` must not contain `source_branch`')
+        end
+
+        github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
+        target_milestone = milestone_title.nil? ? nil : github_helper.get_milestone(repository, milestone_title)
+
+        final_target_branches = if target_branches.empty?
+                                  determine_target_branches(release_version: source_branch.delete('release/'), default_branch: default_branch)
```
AliSoftware: We'd still want to keep your check that the `source_branch` starts with `release/` in the case where `target_branches.empty?`—since `determine_target_branches(release_version:, …)` expects the `release_version` parameter to be string representing a version number for its implementation logic to work.

_(It's only when `target_branches` is non-empty that we want to allow non-`release/`-prefixed `source_branch` to be ok.)_

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,168 @@
...
+            base_branch: target_branch,
+            labels: labels,
+            milestone: target_milestone&.number,
+            reviewers: reviewers,
+            team_reviewers: team_reviewers
+          )
+        end
+      end
+
+      def self.determine_target_branches(release_version:, default_branch:)
```
AliSoftware: How simple and clearly-named this helper function might seem, I still think it'd be nice to add some YARD to it, especially to document the parameters and their expected types.

Same goes for `self.create_backmerge_pr`

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: > We need to protect against the case where there's nothing to merge.

I think even in cases of a cherry-pick, there's always at least a version bump?
The scenario where I see it happening is when we're talking about non-release branches, without a version bump. Still, that seems quite rare. Doesn't hurt to add a check though (as the call to the API will indeed fail) 🤔 

> Though maybe in those cases it's better for the action to fail explicitly

We have this case of the language fetch on Tumblr, where we force-push the new changes + try to create the PR, which I don't think even fails when a PR already exists (I think it just logs an error?).

AliSoftware: There's still the case of a retry. e.g. `finalize_release` -> creates the backmerge PR, the RM merges that backmerge PR, but then the trampoline build fails for some reason (e.g. ASC or PSC API outage or similar unrelated to our side of the code or something that would need to be fixed in `buildkite-ci` or somewhere else but would not need any commit on the `release/*` branch of the repo to fix it…), and then the RM retries the `finalize_release`, which will try to create a new backmerge PR… but since the first one has already been created and merged and there hasn't been anything new pushed on `release/*` branch since, then PR would be empty and cause a GitHub API error in return.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,189 @@
...
+      # @param labels [Array<String>] the labels to add to the pull request.
+      # @param milestone [String] the milestone to associate with the pull request.
+      # @param reviewers [Array<String>] the individual reviewers for the pull request.
+      # @param team_reviewers [Array<String>] the team reviewers for the pull request.
+      # @return [void]
+      def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, labels:, milestone:, reviewers:, team_reviewers:)
+        intermediate_branch = "merge/#{head_branch.gsub('/', '-')}-into-#{base_branch.gsub('/', '-')}"
+        Fastlane::Helper::GitHelper.create_branch(intermediate_branch)
+
+        other_action.push_to_git_remote(tags: false)
```
AliSoftware: We might want to add a `ConfigItem` of `type: Proc` (hope that's allowed by Fastlane? I don't remember even having seen an action taking a block as parameter so far…) that would be called here if not nil and would allow us to add additional commits to the intermediate branch before finalizing the PR.

The typical use case being for Tumblr, where we'd want to pass a block that restores the version files to their content on the target branch. Though I could imagine some other projects where a similar commit to prevent or anticipate-and-fix conflicts would be useful on other files too.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: TIL, Fastlane does support Proc!

Everything went fine until I hit this issue and realize having Proc parameters given our `spec_helper.rb` is a bit more complicated: https://github.com/wordpress-mobile/release-toolkit/blob/476f94f8cf02847cab18d88e59270c3b0372999f/spec/spec_helper.rb#L73

The lane it generates in the method above looks like this on runtime, and the Proc completely breaks it 😅:
```ruby
lane :test do
  create_release_backmerge_pull_request(
    {:github_token=>"ghp_fake_token", :repository=>"repo-test/project-test", :source_branch=>"release/30.6", :default_branch=>"main", :intermediate_branch_created_callback=>#<Proc:0x00000001158d5200 <etc...>/create_release_backmerge_pull_request_spec.rb:316>}
  )
end
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Done on 5a1d57a34966ff6d18e4e5566863af1f955d3a32

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Indeed! Added the check back on d2d0441445bdba82af8867ef076b40b03125cf30

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Added diff validation on 42b2ad44f83de2869e1b41360b21d40891931c55

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Agreed! Updated on 0576800278293b6c094b5b563da5b5d96b272584.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Updated on 0576800278293b6c094b5b563da5b5d96b272584

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Added the check in the appropriate case on d2d0441445bdba82af8867ef076b40b03125cf30

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
AliSoftware: Ah yeah I'm not sure there's an easy way to make that Spec Helper work with `Proc` indeed. You might need to write the test lane manually for that one… or alternatively pass a `String` with the `Proc`'s intended source code when calling the stub (so that its `inspect` echoes that source code and that's what gets evaluated by ``Fastlane::Fastfile.new.parse(…)`), instead of passing the actual `Proc` 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -83,6 +83,14 @@ def self.determine_target_branches(source_release_version:, default_branch:)
       # @param team_reviewers [Array<String>] the team reviewers for the pull request.
       # @return [void]
       def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_branch:, labels:, milestone:, reviewers:, team_reviewers:)
+        # Check for differences between head_branch and base_branch
+        diff_output = Actions.sh('git', 'diff', "#{base_branch}..#{head_branch}").strip
```
AliSoftware: I'm not 100% sure this will still work for cases where the `base_branch` has evolved (i.e. got new commits) since the head_branch got cut and the PR got created 🤔  We might need to first find the `git merge-base #{base_branch} #{head_branch}` first, [or alternatively use the three-dots syntax instead of two-dots range](https://stackoverflow.com/questions/7251477/what-are-the-differences-between-double-dot-and-triple-dot-in-git-dif).
Though it's late and I'm not completely sure which one of the two we actually need (and which one of the two cases GitHub will consider it to create an empty PR). I _think_ the diff that GitHub shows on PRs is the `...` one not the `..` one, and thus we should probably check the `...` one for emptiness… but will probably be worth experimenting a bit to confirm and double-check (and document via code comments)

Also, since we're only interested if there's an actual diff or not, we don't really need the diff's actual output. So we should instead use `git diff --quiet` and check the exit code, rather than using `git diff` and checking if the output is empty.

AliSoftware: Also, I wonder if we shouldn't start using the `Git` gem, instead of shelling out to `sh('git', …)`… though that is probably a broader discussion for outside this PR, because I know we already use `sh('git', …)` in our git helpers so far too… so we'd probably want to migrate everything to using the `Git` gem to be consistent if we do that.

iangmaia: Nice catch! I just seen [this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons), which suggests to indeed use three-dot diff instead:
> By default, pull requests on GitHub show a three-dot diff.

---

On Git operations:

> So we'd probably want to migrate everything to using the Git gem to be consistent if we do that

That's an interesting point -- I think it would be nice to be consistent. Using the `Git` gem would be nice, but I also see it as a more complex project as I guess we'd have to implement more Ruby code to do things like filtering, besides the changes we'd also have to change quite some tests, etc. I've created https://github.com/wordpress-mobile/release-toolkit/issues/572 so we can have a starting point for discussion.

iangmaia: > So we should instead use git diff --quiet and check the exit code

The problem with that is that `sh()` fails the lane when there's a 1 exit code.

AliSoftware: Also, I had plans to add a helper in release-toolkit that would return if there has been new commits between two points in git. The typical use case would be to check if there had been new commits on `release/*` since the last beta git tag created on that branch, so we can determine with automation if a new beta is necessary or not. This could then allow us to run `new_beta_release` on a daily scheduled build on CI, exiting early and not doing the beta if no new commits were detected.

(I have a roadmap item for this)

---

This PR could actually be the occasion to start implementing that helper method here and use it to detect if the PR will be empty rather than looking at the diff?

Also because now that I'm thinking about it, I wonder if GitHub wouldn't allow a PR with an empty diff as long as the PR would create a non-empty merge commit. A case where this can happen is if you create a commit on your head branch then push another commit reverting the change of the first. The diff would be empty but the git graph after the merge would still show the 2 commits being merged into the target branch in the end so that would still be a valid merge (as opposed to trying to merge A into B when A and B end up pointing to the same commit) 🤔 

Definitively worth some experimentation with those edge cases to see what works.

AliSoftware: > (I have a roadmap item for this)

Found it: https://github.com/orgs/Automattic/projects/841/views/1?filterQuery=Beta&pane=issue&itemId=51038655

The idea I had back then for such detection was to use `git rev-list --count --first-parent ${previous_tag}..HEAD` to count the number of commits between `previous_tag` and `HEAD` (and consider the diff empty if that count is 0).
Haven't tested that idea in practice though.

AliSoftware: > The problem with that is that `sh()` fails the lane when there's a 1 exit code.

Right. IINM we can [pass a block to `sh`](https://github.com/fastlane/fastlane/blob/2c34057215aeabdaac4aea7806a8f166417c8dff/fastlane/lib/fastlane/helper/sh_helper.rb#L22) to catch those error codes and prevent fastlane to `UI.user_error!` on those occasions, though it would still `UI.error` (i.e. print the message in red in the logs, but at least not fail the lane and continue instead).

iangmaia: > Also because now that I'm thinking about it, I wonder if GitHub wouldn't allow a PR with an empty diff as long as the PR would create a non-empty merge commit.

TIL, it's allowed: https://github.com/wordpress-mobile/release-toolkit/pull/573
(and the diff is of course empty https://github.com/wordpress-mobile/release-toolkit/pull/573.diff)

Will look into checking for new commits instead. To start with, I'm thinking of just adding it to `GitHelper`, then we can use it / evolve it to a separate action as suggested in the issue?

AliSoftware: WFM 👍

iangmaia: I updated the code to follow this approach on 8e5aa9e807543fa76bfb705ca080136376d73e8e + 7aa838b7138bff644d7458aa3b846775e7e58fec.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Implemented here: 6a3204f72fde2c1181bd1a4ee6088a555d678f08

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -179,6 +179,21 @@ def self.fetch_all_tags
...
+      #
+      # @param base_ref [String] The base reference (branch name, tag, or commit hash).
+      # @param head_ref [String] The head reference (branch name, tag, or commit hash).
+      #
+      # @return [Integer] The number of commits between base_ref and head_ref.
+      #
+      def self.count_commits_between(base_ref:, head_ref:)
+        git_repo = Git.open(Dir.pwd)
+
+        new_commits = git_repo.log.between(base_ref, head_ref)
```
AliSoftware: 💭  This might return the wrong number of commits if there are merge commits (aka commits with more than one parent) along the path between the two points.

For example, how many commits are there between `A` and `G` here?

```
A---B---C---M---F---G
     \     /
      D---E
```

 - 5 (B, C, M, F, G)?
 - 6 (B, D, E, M, F, G)?
 - 7 (B, C, D, E, M, F, G — counting the total on both paths)?

This is why I added `--first-parent` in the command I suggested to use in the apps-infra-roadmap issue 😉 

You will want to add a unit test for that case too btw.

spec/create_release_backmerge_pull_request_spec.rb
```diff
@@ -0,0 +1,329 @@
...
+    "https://github.com/#{test_repo}/pull/#{number}"
+  end
+
+  def stub_git_release_branches(branches)
+    allow(Fastlane::Actions).to receive(:sh)
+      .with('git', 'branch', '-r', '-l', 'origin/release/*')
+      .and_return("\n" + branches.map { |release| "origin/#{release}" }.join("\n") + "\n")
+  end
+
+  def stub_expected_pull_requests(expected_backmerge_branches:, source_branch:, labels: [], milestone_number: nil, reviewers: nil, team_reviewers: nil, commits_between_head_and_base: 42)
```
AliSoftware: 💄  Just for the sake of conciseness, what about naming that param `nb_new_commits` instead of `commits_between_head_and_base`? 😄

spec/git_helper_spec.rb
```diff
@@ -110,6 +110,74 @@
...
 
+  context('count_commits_between(base_ref:, head_ref:)') do
+    before do
+      # Spec branching setup:
+      #
+      #   (1.0)
+      # A---B---C         main
+      #        / \
+      #       |   D---E   feature-branch
+      #       F---G---H   another-branch / new-branch
```
AliSoftware: Lovely ASCII art 👍

AliSoftware: _(Though you're gonna have to redo it once you add the case for a merge commit aka commit with multiple parents to cover the case discussed in the other comment, sorry 😅 )_

AliSoftware (commented):
Overall looking good… except for the missing case of commits with >1 parent (i.e. merge commits) in the path from head to base

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,207 @@
...
+
+        if commits_between_head_and_base.zero?
+          UI.message("No differences between #{head_branch} and #{base_branch}. Skipping PR creation.")
+          return nil
+        end
+
+        intermediate_branch = "merge/#{head_branch.gsub('/', '-')}-into-#{base_branch.gsub('/', '-')}"
+        Fastlane::Helper::GitHelper.create_branch(intermediate_branch)
+
+        intermediate_branch_created_callback&.call
```
AliSoftware: I'm not sure how this will be useful in practice if we don't provide it enough context, especially since this action can create multiple backmerge PRs (with different target branches), and will call the callback for each, and if we provide a callback we'll definitively want to do something slightly different for each one.

For example when in the case of Tumblr we'll want to use that to restore the version file to it's state on the branch targeted by the backmerge PR, we'll have to know what that branch is (`trunk`? or `release/N+1`?)

So I think we should at the very least pass the `base_branch` to the callback (and maybe optionally the name of the `intermediate_branch` that got created?)

---

And obviously don't forget to document those parameters that are passed to that `Proc` in the corresponding `ConfigItem`.

_Speaking of which, you forgot to add the YARD `@param` entry for the new `intermediate_branch_created_callback` parameter of this function…_

iangmaia: Indeed, makes sense. I guess most of the times `intermediate_branch` wouldn't have a lot of use; I can think of a case of having to go back to it if the callback checked out something else 🤔  I think I'll add it now as more context doesn't hurt. Update on 9977085a34a8ea6da056c762f48fa1a5d27e8ced

spec/create_release_backmerge_pull_request_spec.rb
iangmaia: Updated on 0a8825cf8fa8cf506151848410c23c5c8daafda4

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
iangmaia: I've changed the implementation a bit, making the `--first-parent`-like implementation optional, and adding this case to the Spec setup: 9210a33f910d46d77bb2ffcc883f165b73e1cbe8.

AliSoftware: Sorry for the back and forth, but the more I think about it the more I wonder if there's a use case for counting the number of commits between two SHA… that is anything else than then checking if that count is >0 or not. Be it to decide if we need to do a new beta or not, or decide if a PR is going to be empty because would be asking GitHub to merge a commit into itself… in both cases testing if the two commit resolve to the same sha1 would be enough… so I might have suggested you a direction that was more complicated that it needed to, sorry.

Like, maybe testing if `git rev-parse #{base}` and `git rev-parse `#{head}` (or `head_commit.sha == base_commit.sha1`, in the Git gem parlance) is all we need after all 😓

iangmaia: I've made some changes (4d19508443dbeb7303f289d2efe95dd2e4bd692a) as discussed via DM:
- Updated the ref check to be a boolean (`count_commits_between` to `has_commits_between?`)
- Check whether there's already an intermediate branch
- Move the check `has_commits_between?` to after the callback, to cover cases where the callback added new commits

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -0,0 +1,213 @@
...
+      # @param token [String] the GitHub token for authentication.
+      # @param repository [String] the repository where the pull request will be created.
+      # @param title [String] the title of the pull request.
+      # @param head_branch [String] the source branch for the pull request.
+      # @param base_branch [String] the target branch for the pull request.
+      # @param labels [Array<String>] the labels to add to the pull request.
+      # @param milestone [String] the milestone to associate with the pull request.
+      # @param reviewers [Array<String>] the individual reviewers for the pull request.
+      # @param team_reviewers [Array<String>] the team reviewers for the pull request.
+      # @return [void]
```
AliSoftware:
```suggestion
      # @param team_reviewers [Array<String>] the team reviewers for the pull request.
      # @param intermediate_branch_created_callback [Proc?]
      #        A callback to call after having created the intermediate branch
      #        to allow the caller to e.g. push new commits on it before the PR is created
      # @return [String?] The URL of the created Pull Request, or `nil` if no PR was created.
```

AliSoftware: This made me realize that I think I've used `@param [Type] paramname description` in many places where I've written some YARD documentation, while indeed the expected format/order seems to be ``@param paramname [Type] description` instead (like you used here)… :oops:

Even if we don't really use the `yard` executable to generate HTML docs from our own code and in practice those comments are just for us when reading the code… I guess if we want to be consistent we should fix all the places (`Fastfile` in various repos, etc…) where we (I?) have used the wrong format 😅

lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb
```diff
@@ -179,6 +179,34 @@ def self.fetch_all_tags
...
+        current_commit = head_commit
+        while current_commit
+          break if current_commit.sha == base_commit.sha
+
+          count += 1
+          current_commit = current_commit.parents.first
+        end
+
+        count.positive?
+      end
```
AliSoftware: I'm not sure we'll ever need the differentiation of `first_parent:` parameter anymore now that the goal is to just testing if the value is `.positive?` and return a boolean. So I think we can remove the parameter, and all the code after `unless first_parent` too now. Because instead of counting the total number then testing if it is `positive?`, it would be way more effective to exit early as soon as we detect they are not the same commits, instead of walking up the `current_commit = current_commit.parents.first` chain only to test if `count.positive?` in the end.

Also, I just realized that that walk up the chain (in the case of `first_parent: true`) could risk entering an infinite (well, or super-long-until-we-reach-the-repo-root) loop if we happen to call `has_commits_between?` with refs which are unreachable from one another. For example if we were to call `has_commits_between?(D, F)` in [your example of such a git tree](https://github.com/wordpress-mobile/release-toolkit/pull/570/files#diff-7aed954270eabe63c0117e184d02a920e7c50e090136cd8a33fe121bd7e87ca9R118-R122)

So in conclusion I think this whole code can be simplified to:

```suggestion
      # @return [Boolean] Whether two git refs point to the same commit
      #
      def self.point_to_same_commit?(ref1, ref2)
        git_repo = Git.open(Dir.pwd)

        commit1 = git_repo.gcommit(ref1)
        commit2 = git_repo.gcommit(ref2)

        return commit1.sha == commit2.sha
      end
```

And then in your action implementation, you'd call `if point_to_same_commit?(base_branch, head_branch)` instead of `unless has_commits_between?(base_ref: base_branch, head_ref: head_branch)`

iangmaia: Yeah, I think for the purpose of it, probably this line would've been enough:
```ruby
git_repo.log.between(base_ref, head_ref).count.positive?
```

Though I've updated `point_to_same_commit?`, what it does is a bit more straight forward. Updated on 84f3edd7b8221ab8cab2214f1634ee259aae1ff4

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: Updated on b4f411dfa291269c5b732a0c9704aa8b497f234a

> Even if we don't really use the yard executable to generate HTML docs from our own code and in practice those comments are just for us when reading the code… I guess if we want to be consistent we should fix all the places 

Yeah, since it's mostly internal, I guess we could change them as we move + do the new ones in the expected format 🙂

AliSoftware (approved):

PR #571, opened 2024-07-12
Fix `android_download_translations` issues

### LINKED ISSUE #569: `android_download_translations` issues: need to post-process `plurals` nodes + replicate `tools:ignore` attributes
This issue aggregates various small bugs/issues we've noticed on the `android_download_translations` tool while we were doing the DayOne Android release (See p1720472497193469-slack-CC7L49W13).

## 1️⃣ Plurals are not post-processed

The implementation of the tool runs some post-processing code (like `apply_substitutions`, or the replication of attributes, or `quick_lint`) on `//string` and `//string-array/item` nodes. But it doesn't run the same rules on `//plurals/item`, while it probably should too.

This is likely due to the fact that, until recently, GlotPress export to the Android XML format didn't support `<plurals>`, and that support was added very recently, which means GlotPress only recently started to include `<plurals>` nodes in its exports. So we need to add support for them in our tool as well

## 2️⃣ `tools:ignore` attributes not replicated

When we post-process the translated XML exported by GlotPress, amongst other things we replicate some attributes from the English original XML nodes onto the corresponding nodes in the translated XML. [This is in particular the case for `formatted=false` attribute](https://github.com/wordpress-mobile/release-toolkit/blob/476f94f8cf02847cab18d88e59270c3b0372999f/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L238).

We should extend this replication of attributes to apply to:
 - Any `tools:ignore` attributes (e.g. like `tools:ignore="ImpliedQuantity"`)
 - Replicating not just attributes found on `//strings` nodes of the original XML file, but also from `plurals`, `plurals/item`, `string-array`… and whatever other XML nodes might have this kind of attribute.
    - Note that this might require a small refactoring as currently [we build a Hash of `string/tag name => attributes`](https://github.com/wordpress-mobile/release-toolkit/blob/476f94f8cf02847cab18d88e59270c3b0372999f/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb#L241) but if we want to support separate attributes for `plurals[name=x]/item[quantity=one]` vs `plurals[name=x]/item[quantity=other]` so it might need double-keyed hash or similar?

## 3️⃣ `@string/` references bug

When a string from the original `values/strings.xml` contains references like `@string/…`, it is supposed to also be marked by developers as `translatable=false` so that it's not included in GlotPress.

If they don't the translators will likely just keep the translation copy as `@string/…` unmodified as they should… but when the string file is exported by GlotPress, [that leads to an extra backslash to be added to the export and the `values-{locale}/strings.xml` file to contain `\@string/…` instead](https://github.com/bloom/DayOne-Android/pull/4568/files#diff-a2529e239a99387f33127b7e3027bc8447c2392f428129f616ca7f319dd187fbL1408-R1408).

We have [a Dangermattic rule](https://github.com/Automattic/dangermattic/blob/trunk/lib/dangermattic/plugins/android_strings_checker.rb) which helps prevent people from forgetting about it, but bit all Android repos have it enabled. Besides, even if they do, we probably still want `android_download_translations` to have a safety net for if/when that case still end up happening, just to be on the safe side.

AliSoftware:
Fixes #569 — i.e. the issues that DayOne Android encountered in their last `2024.14` release (See p1720472497193469-slack-CC7L49W13)

## What does it do?

- Replicate all XML attributes (e.g. `tools:ignore="ImpliedQuantity"`) from original `values/strings.xml` on all nodes of translated XMLs exported from GlotPress
- Apply text substitutions (`...`=>`…`, `-`=>`–`, etc) on `plurals/item` nodes too (before it was only on `string` and `string-array` nodes)
- Improve our `quick_lint` check on each copies to (1) be run on _all_ nodes containing a string (including `plurals/item` and `string-array/item` not just `string`) and (2) warn about + auto-fix `\@string/…` in exports (suggesting to add `translatable="false"` on the offending key(s) to fix the issue at its root in the future)

- [x] ~If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to describe how the changes will affect the migration from the previous major version and what the clients will need to change and consider.~

## How to Test

 - Run `bundle exec rspec` and validate all unit tests pass
 - In [the DayOne Android repo](https://github.com/bloom/DayOne-Android):
    - Create a test branch
    - Run `bundle exec fastlane download_translations`, then commit the changes if any (this will serve as the "before" baseline for the rest of the test)
    - Update the `Gemfile` to point release-toolkit to `gem 'fastlane-plugin-wpmreleasetoolkit', git: 'git@github.com:wordpress-mobile/release-toolkit', branch: 'fix/android_download_translations'`
    - Run `bundle install`
    - Run `bundle exec fastlane download_translations` again, and check the diff. Especially validate that any `\@string/` escaped references previously present in translations were properly fixed, and that attributes present in the `values/strings.xml` were now all properly replicated on all the translated XMLs
    - Run `./gradlew lint` and validate it passes
    - Edit the `app/src/main/res/values/strings.xml`:
        - Pick a plural that has been translated in other locales (ideally we'd pick `confirm_entry_trash_multiple` since that's the one that caused the problem in `2024.14` release, but at the time of writing this PR this key is not present in GlotPress anymore), like `daily_prompt_card_menu_view_entries` for example
        - Add the `tools:ignore="ImpliedQuantity"` attribute on one of its `item` child nodes, e.g. `<item quantity="one" tools:ignore="ImpliedQuantity">View one Entry</item>`
        - Don't forget to also add the `tools` XML namespace at the top of the XML: `<resources xmlns:tools="http://schemas.android.com/tools">`
    - Re-run `bundle exec fastlane download_translations`, and verify that this new `tools:ignore="ImpliedQuantity"` attribute has been replicated on the `daily_prompt_card_menu_view_entries` item node of all translated XMLs
    - Run `./gradlew lint` and validate it passes
 - Replicate a similar test on WPAndroid and WCAndroid, to validate we wouldn't break those (I'm starting to wonder why we only copied a specific list of attributes before this PR instead of copying them all… maybe there was a reason that I forgot about?)

cc @danilo04 @AmandaRiu 

## What's Next

 - Do a new release of `release-toolkit`
 - Point DayOne Android to the new version
 - Inform DOAndroid team that they can revert [their workaround](https://github.com/bloom/DayOne-Android/pull/4579) and go back to use `tools:ignore="ImpliedQuantity"` on their `plurals/item` nodes.
     - Any attribute present in `values/strings.xml` will be replicated in the downloaded-and-post-processed `values-{locale}/strings.xml` for each corresponding node now.
     - Removing their workaround which tests `if (entries.size > 1)` and go back to `ImpliedQuantity` will allow to let the OS cover even cases of locales which are using `quantity=one` not just for `1` but for other values (e.g. Bielorussian uses this plural form also for `11`, `21`, `31`, …)
 - Inform DOAndroid that, while testing the change on their codebase, it now rightfully informed of a warning about `daily_prompt_settings_notifications` being a string reference but not having `translatable=false`. As a side effect, while other translators have left it untouched in GlotPress, [the Italian translator did translate `@string/notifications` to `\@stringa/notifiche`](https://translate.wordpress.com/projects/day-one/android/it/default/?filters%5Bterm%5D=daily_prompt_settings_notifications&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc) (and because the prefix is now `\@stringa/` not `\@string/`, this wasn't auto-fixed by the action like it now is for the others).

### DIFF
--- a/spec/test-data/translations/glotpress-download/stubs/fr.xml
+++ b/spec/test-data/translations/glotpress-download/stubs/fr.xml
@@ -71,7 +71,7 @@ Language: fr
         <item>\@string/thursday</item>
         <item>\@string/friday</item>
         <item>\@string/saturday</item>
-        <item>None</item>
+        <item>Aucun</item>
     </string-array>
     <string-array name="progress" formatted="false">
         <item>0%%</item>
--- a/spec/test-data/translations/glotpress-download/stubs/fr.xml
+++ b/spec/test-data/translations/glotpress-download/stubs/fr.xml
@@ -31,6 +31,7 @@ Language: fr
 		<item>En cours…</item>
 		<item>Toutes les commandes...</item>
 	</string-array>
+	<string name="stringref">\@string/shipping_label_woo_discount_bottomsheet_title</string>
 	<string name="threat_fix_description">Les prix vont de 70-120 $ par heure</string>
     <string-array name="settings_jetpackdescription">
         <item>Vous pouvez marquer une adresse IP (ou une série d’adresses IP) comme « Toujours autorisé »</item>
@@ -63,13 +64,13 @@ Language: fr
     <string name="friday">Vendredi</string>
     <string name="saturday">Samedi</string>
     <string-array name="weeks_full">
-        <item>@string/sunday</item>
-        <item>@string/monday</item>
-        <item>@string/tuesday</item>
-        <item>@string/wednesday</item>
-        <item>@string/thursday</item>
-        <item>@string/friday</item>
-        <item>@string/saturday</item>
+        <item>\@string/sunday</item>
+        <item>\@string/monday</item>
+        <item>\@string/tuesday</item>
+        <item>\@string/wednesday</item>
+        <item>\@string/thursday</item>
+        <item>\@string/friday</item>
+        <item>\@string/saturday</item>
         <item>None</item>
     </string-array>
     <string-array name="progress" formatted="false">
@@ -81,7 +82,7 @@ Language: fr
     </string-array>
     <string name="confirm_single_entry_trash">Êtes-vous sûr de vouloir déplacer cette entrée vers la corbeille ?</string>
     <plurals name="confirm_entry_trash">
-        <item quantity="one">@string/confirm_single_entry_trash</item>
+        <item quantity="one">\@string/confirm_single_entry_trash</item>
         <item quantity="other">Etes-vous sûr... de vouloir déplacer %d entrées vers la corbeille ?</item>
     </plurals>
 </resources>
--- a/spec/test-data/translations/glotpress-download/stubs/fr.xml
+++ b/spec/test-data/translations/glotpress-download/stubs/fr.xml
@@ -6,6 +6,7 @@ Generator: GlotPress/2.4.0-alpha
 Language: fr
 -->
 <resources>
+  <string name="app_name">WooCommerce</string>
 	<string name="product_attributes_error_saving">Erreur lors de l’enregistrement de vos attributs</string>
 	<string name="shipping_label_payments_cant_edit_warning">Seul le propriétaire du site peut gérer les moyens de paiement des étiquettes d’expédition. Veuillez contacter le propriétaire de la boutique %1$s (%2$s) pour gérer les moyens de paiement.</string>
 	<string name="shipping_label_shipping_method_free_shipping">Livraison gratuite</string>
@@ -54,4 +55,26 @@ Language: fr
         <item> - o Deux: deuxième point</item>
     </string-array>
     <string name="multi_range_statement">La première plage est 2-100 et la seconde est 123-456.</string>
+    <string name="sunday">Dimanche</string>
+    <string name="monday">Lundi</string>
+    <string name="tuesday">Mardi</string>
+    <string name="wednesday">Mercredi</string>
+    <string name="thursday">Jeudi</string>
+    <string name="friday">Vendredi</string>
+    <string name="saturday">Samedi</string>
+    <string-array name="weeks_full">
+        <item>@string/sunday</item>
+        <item>@string/monday</item>
+        <item>@string/tuesday</item>
+        <item>@string/wednesday</item>
+        <item>@string/thursday</item>
+        <item>@string/friday</item>
+        <item>@string/saturday</item>
+        <item>None</item>
+    </string-array>
+    <string name="confirm_single_entry_trash">Êtes-vous sûr de vouloir déplacer cette entrée vers la corbeille ?</string>
+    <plurals name="confirm_entry_trash">
+        <item quantity="one">@string/confirm_single_entry_trash</item>
+        <item quantity="other">Etes-vous sûr... de vouloir déplacer %d entrées vers la corbeille ?</item>
+    </plurals>
 </resources>
--- a/spec/test-data/translations/glotpress-download/stubs/fr.xml
+++ b/spec/test-data/translations/glotpress-download/stubs/fr.xml
@@ -72,6 +72,13 @@ Language: fr
         <item>@string/saturday</item>
         <item>None</item>
     </string-array>
+    <string-array name="progress" formatted="false">
+        <item>0%%</item>
+        <item>25%</item>
+        <item>50%%</item>
+        <item>75%</item>
+        <item>100%</item>
+    </string-array>
     <string name="confirm_single_entry_trash">Êtes-vous sûr de vouloir déplacer cette entrée vers la corbeille ?</string>
     <plurals name="confirm_entry_trash">
         <item quantity="one">@string/confirm_single_entry_trash</item>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- Fix `android_download_translation` issues reported in #569 [#571]
+   - add post-processing of `plurals` nodes too.
+   - detect and fix `\@string/` references escaped in GlotPress exports.
+   - replicate all XML attributes (and xmlns) present in `values/string.xml` on the corresponding nodes in the translated XML.
 
 ### Internal Changes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
@@ -354,18 +354,18 @@ module Fastlane
           end
           # 3. Process copies for `string` nodes
           translated_xml.xpath('//string[@name]').each do |string_node|
-            apply_substitutions(string_node)
+            apply_substitutions!(string_node)
             quick_lint(string_node, locale_code)
           end
           # 4. Process copies for `string-array/item` nodes
           translated_xml.xpath('//string-array[@name]/item').each do |item_node|
-            apply_substitutions(item_node)
+            apply_substitutions!(item_node)
             quick_lint(item_node, locale_code)
           end
           # 5. Replicate attributes + Process copies for `plurals/item` nodes
           translated_xml.xpath('//plurals[@name]/item[@quantity]').each do |item_node|
             copy_orig_attributes.call(item_node, "//*[@name = '#{item_node.parent['name']}']/item[@quantity = '#{item_node['quantity']}']")
-            apply_substitutions(item_node)
+            apply_substitutions!(item_node)
             quick_lint(item_node, locale_code)
           end
         end
@@ -375,7 +375,7 @@ module Fastlane
         #
         # @param [Nokogiri::XML::Node] tag The XML tag/node to apply substitutions to
         #
-        def self.apply_substitutions(tag)
+        def self.apply_substitutions!(tag)
           tag.content = tag.content.gsub('...', '…')
 
           # Typography en-dash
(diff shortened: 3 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

spec/test-data/translations/glotpress-download/stubs/fr.xml
```diff
@@ -54,4 +56,33 @@ Language: fr
...
+        <item>\@string/saturday</item>
+        <item>Aucun</item>
+    </string-array>
+    <string-array name="progress" formatted="false">
+        <item>0%%</item>
+        <item>25%</item>
+        <item>50%%</item>
+        <item>75%</item>
+        <item>100%</item>
+    </string-array>
```
AliSoftware, line 82: Intentional use of `%%` on a `formatted=false` resource of type `string-array` in this GlotPress export stub, so we could write our test on `quick_lint` to validate it'd emit a warning as appropriate.

spec/test-data/translations/glotpress-download/stubs/fr.xml
```diff
@@ -54,4 +56,33 @@ Language: fr
...
+    <string name="confirm_single_entry_trash">Êtes-vous sûr de vouloir déplacer cette entrée vers la corbeille ?</string>
+    <plurals name="confirm_entry_trash">
+        <item quantity="one">\@string/confirm_single_entry_trash</item>
```
AliSoftware, line 85: Intentional use of `\@string/` used on a `plurals/item` node in this GlotPress export stub, to reflect that GlotPress indeed escapes those in its exports, and to thus be able to test that we detect and auto-fix those during `quick_lint` (i.e. notice how those backslashes are removed on that same node in the `expected` fixture of the unit test)

spec/test-data/translations/glotpress-download/stubs/fr.xml
```diff
@@ -54,4 +56,33 @@ Language: fr
...
+    <string-array name="weeks_full">
+        <item>\@string/sunday</item>
+        <item>\@string/monday</item>
+        <item>\@string/tuesday</item>
+        <item>\@string/wednesday</item>
+        <item>\@string/thursday</item>
+        <item>\@string/friday</item>
+        <item>\@string/saturday</item>
+        <item>Aucun</item>
+    </string-array>
```
AliSoftware, line 75: Intentional use of `\@string/` used on those `string-array/item` nodes of this GlotPress export stub, to reflect that GlotPress indeed escapes those in its exports, and to thus be able to test that we detect and auto-fix those during `quick_lint`

spec/test-data/translations/glotpress-download/stubs/fr.xml
```diff
@@ -30,6 +31,7 @@ Language: fr
 		<item>En cours…</item>
 		<item>Toutes les commandes...</item>
 	</string-array>
+	<string name="stringref">\@string/shipping_label_woo_discount_bottomsheet_title</string>
```
AliSoftware, line 34: Intentional use of `\@string/` used on a `string` node of this GlotPress export stub, to reflect that GlotPress indeed escapes those in its exports, and to thus be able to test that we detect and auto-fix those during `quick_lint`

iangmaia:
I ran a few tests and while it worked fine, I noticed a few things specially when running on `WPAndroid`.

I've pushed my `iangmaia/translations-fix-tests` branch to show some concrete examples:
- On the first run with the updated Gem ([f101715ea9f7cd820631503038b428e461421518](https://github.com/wordpress-mobile/WordPress-Android/commit/f101715ea9f7cd820631503038b428e461421518)), it (correctly) added quite a few `a8c-src-lib=` attributes -- which I don't think should be an issue, but I wonder what's the use of them? 🤔 Anyway, perhaps out of the scope of this PR.
- I think this may be a known issue as I remember seeing it before, but mentioning just in case: on my second run ([c627bd769c1b052b1eaa07039d9cbb186c467db8](https://github.com/wordpress-mobile/WordPress-Android/commit/c627bd769c1b052b1eaa07039d9cbb186c467db8)), I didn't expect any other change than my removal of `translatable="false"` from `app_theme_entry_values`, but it changed the `WordPress/src/main/res/values-es-rCL/strings.xml` file 😕 and looking closely, it seems just that the order of some strings changed? I wonder if that's from GlotPress or on the Gem side, and if we could force sorting the xml (which would initially create a big diff, but the subsequent ones would be very small and without these strings changing position).
- Third run ([f402ce4c79000275c794f77764e7a11940c48610](https://github.com/wordpress-mobile/WordPress-Android/commit/16d3ffb0f10091aaef5b643d3210927bafc29623)) and fourth run ([16d3ffb0f10091aaef5b643d3210927bafc29623](https://github.com/wordpress-mobile/WordPress-Android/commit/f402ce4c79000275c794f77764e7a11940c48610)), same thing as above: no other change but some random strings changing order.
- I didn't have this issue on `DOAndroid`, but `WCAndroid` [showed something similar](https://github.com/woocommerce/woocommerce-android/commit/3889cc47ffc793fa01b7e1b55d8d02eed0977187) (see the diff for `WooCommerce/src/main/res/values-de/strings.xml`).
- `WCAndroid` showed a lot of warnings (as expected). But the warnings for the same string, same language showed multiple times:
```console
[20:53:49]: Downloading translations for 'es' from GlotPress (es) [{:status=>"current"}]...
[20:53:55]: Warning: [es] exported translation for 'analytics_date_range_selectors' contains `\@string/`. This is a sign that this entry was not marked as `translatable=false` in the original `values/strings.xml`, and was thus sent to GlotPress, which added the backslash when exporting it back.
[20:53:55]: Warning: [es] exported translation for 'analytics_date_range_selectors' contains `\@string/`. This is a sign that this entry was not marked as `translatable=false` in the original `values/strings.xml`, and was thus sent to GlotPress, which added the backslash when exporting it back.
[20:53:55]: Warning: [es] exported translation for 'analytics_date_range_selectors' contains `\@string/`. This is a sign that this entry was not marked as `translatable=false` in the original `values/strings.xml`, and was thus sent to GlotPress, which added the backslash when exporting it back.
[...]
```

I don't think these are critical, specially the sorting issue that seems to be there for some time, but it's perhaps good to tackle them in a near future.

iangmaia (approved):
Approving to unblock, as the code looks good and [the issues I saw](https://github.com/wordpress-mobile/release-toolkit/pull/571#issuecomment-2234037340) seemed either minor or already known, and can be tackled in a separate PR.

AliSoftware:
Thanks for the thorough testing @iangmaia !

> I think this may be a known issue as I remember seeing it before, but […] looking closely, it seems just that the order of some strings changed?

The non-stable order of the nodes is a known issue (I actually don't remember if this comes from GlotPress not guaranteeing the order during export, or from our post-processing, but I think that's the former?), and one we've long wanted to solve.

> I wonder if that's from GlotPress or on the Gem side, and if we could force sorting the xml

Yep, that would be the idea, by adding some sorting of the strings post-export during our post-processing of the XML files. We just never found the time to get to that (also I think because sorting an XML might not be as trivial as it looks, e.g. making sure we keep the child nodes of the same `<plurals>` and `<string-array>` together, avoiding blank lines that could be inserted by Nokogiri if we remove a node in one place to insert it in another, …). So this will be for another time 😅 

> But the warnings for the same string, same language showed multiple times:

This is kind of expected, because [that string is a `string-array`](https://github.com/woocommerce/woocommerce-android/commit/3889cc47ffc793fa01b7e1b55d8d02eed0977187#diff-62e65ffac1cc0b480156ffda195464c4663c853d1af83b076006de9915567580L3541-L3551) and it was warning you about each element of the array that had that issue.

So if you have:
```xml
 <string-array name="weekday_selection">
        <item>None</item>
        <item>\@string/monday</item>
        <item>\@string/tuesday</item>
        <item>\@string/wednesday</item>
        <item>\@string/thursday</item>
        <item>\@string/friday</item>
        <item>\@string/saturday</item>
        <item>\@string/sunday</item>
        <item>All</item>
</string-array>
```
Then it will warn 7 times: one for each of the `<item>` that contains such a `\@string/`, but not for the 2 `<item>` that contain a plain text like `None` or `All`.

I considered making the error message include the index of the child `item` within the `string-array`, but that would have made the code quite complicated for little benefit. And that would have made it less DRY-able to also cover the similar case of `<plurals>`, for which the interesting info in the error message wouldn't be the index of the `<item>` node, but the value of the corresponding `@quantity` attribute of the offending `<item>`:

```xml
<string name="empty_selection">No item selected</string>
<string name="delete_one">Delete this item?</string>
<plurals name="delete">
  <item quantity="zero">\@string/empty_selection</item>
  <item quantity="one" tools:ignore="ImpliedQuantity">\@string/delete_one</item>
  <item quantity="other">Delete %d items?</item>
</plurals>
```

I also considered de-duplicating the error messages to only print one, but given the examples above I figured it still made sense to warn as many times as there were violations (one per violating `<item>` especially since not all `<item>` children might violate the rule (e.g. `<item>None</item>`, `<item>All</item>` and `<item quantity="other">` in examples above).

So in the end I figured it was not worth the effort and figure it'd be ok to have as many warnings as there are invalid `\@string/` found even if those are on different `<item>` of the same `<string-array>` or `<plurals>`.

AliSoftware:
> * it (correctly) added quite a few `a8c-src-lib=` attributes -- which I don't think should be an issue, but I wonder what's the use of them?

For the record, this is an attribute our tooling adds to strings that got merged from libraries, so we can easily identify them and understand where they come from. For more context, see paqN3M-xk-p2 🙂

PR #591, opened 2024-09-05
`ios_generate_strings_file_from_code`: allow to specify a custom encoding to convert the generated files to

AliSoftware:
## What does it do?

Adds an `output_encoding` parameter to the `ios_generate_strings_file_from_code` action, to allow specifying the encoding we want the generated files to be using.

 - It defaults to `UTF-16LE`, because that's the encoding that `genstrings` is using for the `.strings` files it generates; that way, the current behavior won't change if you don't add this `output_encoding` parameter explicitly to your call sites.
 - But that means that if you prefer to use UTF-8 for the generated files (e.g. for easier diffs on GitHub), you can now add `output_encoding: 'UTF-8'` to your call sites of this action.

Context: this need arose in https://github.com/Automattic/Gravatar-SDK-iOS/pull/380 cc @andrewdmontgomery 

## Why?

By default and by (legacy) design `genstrings` generates files using the `UTF-16LE` encoding, likely because that's the encoding that was suggested to be used for `.strings` files by Apple a while ago, back in ObjC days (I guess for optimization reasons given `NSString`'s backing buffer is using UTF16 internally?).

Since then:
 - Apple and Xcode now recommend using UTF-8 for everything nowadays
 - It shouldn't matter for the final product what encoding the `.strings` files are using, because at compile time, those `.strings` files are converted in binary `plist` format before being included in the final `.app` bundle anyway
 - While it's better to have those files be in UTF-8 than UTF-16 for `diff` purposes; in particular, GitHub diffs consider those files as binary when they are encoded as UTF16, while present them as text and make them easier to compare their diff when encoded as UTF8


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -2,15 +2,32 @@ module Fastlane
...
-        cmd = ['genstrings', '-o', params[:output_dir], *flags, *files]
-        out = Actions.sh_control_output(*cmd, print_command: FastlaneCore::Globals.verbose?, print_command_output: true)
-        out = out.scrub.strip.split("\n")
-        errors = out.select { |line| line.include?('genstrings: error: ') }
-        UI.user_error!(errors.join("\n")) unless !params[:fail_on_error] || errors.empty?
-        out
+        output_encoding = begin
+          Encoding.find(params[:output_encoding])
+        rescue ArgumentError => e
+          UI.user_error!(e.message)
```
iangmaia, line 8: Maybe we could use here a custom message instead, as the reason for failure is clear to us?
```suggestion
          UI.user_error!("Invalid output encoding '#{params[:output_encoding]}'. Please provide a valid encoding supported by Ruby.")
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -2,15 +2,32 @@ module Fastlane
...
+        end
+
+        Dir.mktmpdir('genstrings-output-') do |tmpdir|
+          # Build the command arguments
+          files = files_matching(paths: params[:paths], exclude: params[:exclude])
+          flags = [('-q' if params[:quiet]), ('-SwiftUI' if params[:swiftui]), '-littleEndian'].compact
```
iangmaia: Shouldn't `-littleEndian` be conditional as well? 🤔 (e.g. I don't think this is required or if it even works for utf8?)

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: I thought about it, but also thought that given `rescue ArgumentError` is quite generic,  there's technically no guarantee that the only `ArgumentError` that `Encoding.find` could possibly `raise` will _always_ be about "invalid encoding - {name}" and never about any other `ArgumentError` message…

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: `genstrings` always generates UTF16 output files. The `-littleEndian`/`-bigEndian` optional parameter that you can pass to `genstrings` here is to request `genstrings` to explicitly use a particular endianness for the `UTF-16` output it generates. It is completely unrelated to the `param[:output_encoding]` that the caller requested when calling our action from its `Fastfile`.

If you don't specify `-littleEndian` nor `-bigEndian` when calling `genstrings`, it will still generate output files in UTF16, but using the default endianness of the current OS. Given `genstrings` only runs on Mac, and macOS uses Little Endian, in practice genstrings will always end up generating `UTF-16LE` files even when you don't specify `-littleEndian` explicitly. And it will only generate `UTF-16BE` files if you specify `-bigEndian`, and will never generate `UTF-8` output files.

So in practice that `-littleEndian` was not really strictly necessary (except if one day Apple ports `genstrings` to Linux, and we start being able to run this Fastlane action on Linux machines that use big ending as their default/system endianness…).

But I still preferred to be explicit and provide `-littleEndian` even though `genstrings` run on a Mac will always produce `UTF-16LE` files, **just for consistency** with the fact that, when I later convert the files generated by `genstrings` in `UTF-16` into the `param[:output_encoding]` requested by the Fastlane caller and use `File.read(…, encoding: 'BOM|UTF-16LE')` the encoding that is hardcoded in that `File.read` is guaranteed to match the one endinaness of files produced by `genstrings`.

mokagio: Thanks for the explanation. It wasn't obvious to me.

What do you think of adding a note in the code?

```suggestion
          flags = [
            ('-q' if params[:quiet]),
            ('-SwiftUI' if params[:swiftui]),
            # If no endianness (-bigEndian vs -littleEndian) is specified, genstrings will use endianness of the current OS.
            # Currently, genstrings runs only on macOS, which is little-endian, so this parameter is not strictly necessary.
            # We make it explicit here to raise visibility on the relationship between the endianness of the genstring output and that of the encoding later on.
            '-littleEndian'
          ].compact
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
```diff
@@ -2,15 +2,39 @@ module Fastlane
...
+        Dir.mktmpdir('genstrings-output-') do |tmpdir|
+          # Build the command arguments
+          files = files_matching(paths: params[:paths], exclude: params[:exclude])
+          flags = [
+            ('-q' if params[:quiet]),
+            ('-SwiftUI' if params[:swiftui]),
+            # If no endianness (-bigEndian vs -littleEndian) is specified, genstrings will use endianness of the current OS.
+            # Currently, genstrings runs only on macOS, which is little-endian, so this parameter is not strictly necessary.
+            # We make it explicit here to raise visibility on the relationship between the endianness of the genstring output and that of the encoding later on.
+            '-littleEndian'
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/4d955f19cd2276604d2478dfab63b06c44c0bf94/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb#L20"></span>Style/TrailingCommaInArrayLiteral: Put a comma after the last item of a multiline array.</td>
    </tr>
  </tbody>
</table>

AliSoftware:
```suggestion
            '-littleEndian',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
iangmaia: Thanks for the details @AliSoftware -- though hopefully one day we'll see `genstrings` running on Linux 🙏 😅
Nice suggestion @mokagio 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
iangmaia: Yeah, also works for me.
I was thinking pretty much anything will be about the argument and "invalid encoding" has a generic meaning here, not necessarily "we don't know this encoding". But I'm indifferent in this case as I also expect any error should be clear enough 👍

iangmaia (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb
AliSoftware: > hopefully one day we'll see `genstrings` running on Linux 🙏 😅

I mean tbh it'd be doable to reimplement `genstrings` in swift using SwiftSyntax nowadays 😅 (see my experiment in https://github.com/AliSoftware/swift-syntax-linter-demo), and then have such a tool run on Linux. The catch being that it wouldn't work on ObjC `.m` sources and only `.swift` (but do we still have that many apps still with ObjC code?), and the tricky part not being the parsing of the swift code after all (thanks to `SyntaxVisitor`'s ease of use) but writing the `.strings` output file correctly in the expected OpenStep format (since everything in Foundation to manipulate plist file formats allow for reading the OpenStep format but not writing it anymore)… so we'd have to write code for that ourselves and be sure to handle edge cases (values with newlines, special characters, …) properly …

So probably not worth it, especially given the future is `.xcstrings` (String Catalogs) anyway (though we'll need GlotPress to support Xliff for that first)… 🤷‍♂️

PR #594, opened 2024-09-16
Use a Mermaid graph in the backmerge PR body

mokagio:
## What does it do?

I was waiting for CI to finish on a backmerge PR looking at the ASCII art describing the merge and got side tracked into this nerd indulgence:

image

- Pros: It's neat and colorful and "pro"
- Cons: It's bound to break or change if/when there's a Mermaid syntax 

- [x] Add Unit Tests (aka `specs/*_spec.rb`) if applicable - N.A.
- [x] If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to describe how the changes will affect the migration from the previous major version and what the clients will need to change and consider. - N.A.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -110,17 +110,41 @@ module Fastlane
 
         other_action.push_to_git_remote(tags: false)
 
+        # Live playground to edit this graph
+        # https://mermaid.live/edit#pako:eNqNkU1PwzAMhv9KZanqpWVFwCVHQOKCxoFrLl7itdGaZEoTTWjqf8eLOo2i8eGTPx77TewjKK8JBJTlUbqCzTgTRTEHJ6s6E18C7vtqkc4li8Y9BnSqX6MlBqoYkttV9Y_cW9AUGLxt2_Y7Nfb-8OStNfEVNzQwtcVhpCU1XUJ2p7KU7vzAXNlkmSLQQDjS6v7m4S7nVU9q51O8UsmSv7jzSEuho9Xc3pzaG-Oib_KXlxr_QL8onbuuVy9unvrXbKiBCV645qvme0mIPVmSINjVtMU0RAm8O0YxRf_-4RQIbqca0l5jpGeDXUALIu9_-gRejKzU
+        meramid_git_graph = <<~GRAPH
+          ```mermaid
+          %%{
+            init: {
+              'gitGraph': {
+                'mainBranchName': 'trunk',
+                'mainBranchOrder': 1000,
+                'showCommitLabel': false
+              }
+            }
+          }%%
+          gitGraph
+             branch #{head_branch}
+             checkout #{head_branch}
+             commit
+             commit
+             commit
+             branch #{intermediate_branch}
+             checkout #{intermediate_branch}
+             commit
+             checkout #{base_branch}
+             commit
+             commit
+             merge #{base_branch}
+          ```
+        GRAPH
+
         pr_body = <<~BODY
           Merging `#{head_branch}` into `#{base_branch}`.
 
           Via intermediate branch `#{intermediate_branch}`, to help fix conflicts if any:
-          ```
-          #{head_branch.rjust(40)}  ----o-- - - -
-          #{' ' * 40}       \\
-          #{intermediate_branch.rjust(40)}        `---.
-          #{' ' * 40}             \\
-          #{base_branch.rjust(40)}  ------------x- - -
-          ```
+
+          #{meramid_git_graph}
         BODY
 
         other_action.create_pull_request(
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -139,12 +139,29 @@ module Fastlane
           ```
         GRAPH
 
+        ascii_git_graph = <<~GRAPH
+          ```
+          #{head_branch.rjust(40)}  ----o-- - - -
+          #{' ' * 40}       \\
+          #{intermediate_branch.rjust(40)}        `---.
+          #{' ' * 40}             \\
+          #{base_branch.rjust(40)}  ------------x- - -
+          ```
+        GRAPH
+
         pr_body = <<~BODY
           Merging `#{head_branch}` into `#{base_branch}`.
 
           Via intermediate branch `#{intermediate_branch}`, to help fix conflicts if any:
 
           #{meramid_git_graph}
+
+          <details>
+          <summary>Expand to see an ASCII representation of the Git graph above</summary>
+
+          #{ascii_git_graph}
+
+          </details>
         BODY
 
         other_action.create_pull_request(
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
@@ -117,7 +117,7 @@ module Fastlane
           %%{
             init: {
               'gitGraph': {
-                'mainBranchName': 'trunk',
+                'mainBranchName': '#{base_branch}',
                 'mainBranchOrder': 1000,
                 'showCommitLabel': false
               }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- The backmerge PR body now use a [Mermaid](https://mermaid.js.org) graph to represent the Git tree [#594]
 
 ### Bug Fixes
 


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -110,17 +110,58 @@ def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_bra
...
+          }%%
+          gitGraph
+             branch #{head_branch}
+             checkout #{head_branch}
+             commit
+             commit
+             commit
+             branch #{intermediate_branch}
+             checkout #{intermediate_branch}
+             commit
```
mokagio, line 134: Technically, at this time, there is no commit on the integration branch. However, without the commit, I don't think the Mermaid graph doesn't convey the idea of the integration branch as well:

#### without `commit`

image

#### with `commit`

image

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -110,17 +110,58 @@ def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_bra
...
+          Via intermediate branch `#{intermediate_branch}`, to help fix conflicts if any:
+
+          #{meramid_git_graph}
+
+          <details>
+          <summary>Expand to see an ASCII representation of the Git graph above</summary>
+
+          #{ascii_git_graph}
+
+          </details>
```
mokagio, line 164: Maybe we can mitigate the "disruption" of the graph not rendering correctly in GitHub for whichever reason by having the ASCII as a fallback here.

AliSoftware:
To be honest I remember trying using Mermaid for that a while back (I think at the time I initially introduced the ASCII art in backmerge PRs a while ago actually, as I tried to do that using Mermaid first)

The issues I remember having back then is that some characters in some branch names broke the mermaid interpreter—which was then unable to render the graph at all.

I don't remember what were those problematic characters that caused this during my testing back then (maybe `/` as in `release/12.3`? Or was it `-`? Can't remember) but that's what made be decide to avoid using Mermaid here in case some custom branch names could sometimes cause trouble.

---

That being said, maybe they fixed that bug since (it was quite some time ago, after all); or maybe such issue won't occur in the context of this backmerge-pr action because the branch names are already deterministic (should all be in the form of `release/xx.y` or `release/xx.y.z`, `merge/release-xx.y-into-trunk` and `trunk`/`default`/`main`/…; so as long as characters from `[a-zA-Z0-9/.-]` don't cause trouble (anymore?) that should be ok to use now?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
AliSoftware: Yep, that was also a limitation I remember seeing when I tried it a while ago. It I think it makes sense and is ok to have that commit visually added there anyway.

> Technically, at this time, there is no commit on the integration branch

Besides, this is not entirely true. For example, on Tumblr there is a commit done on the intermediate branch after it is cut from `release/`—the goal of that extra commit being to `git checkout $basebranch -- version.xcconfig` to auto-resolve the conflict that would otherwise be present on that file due to it evolving separately on `release/*` branch (for daily betas) vs `develop` (for daily alphas).

This is also why this Fastlane action has a `ConfigItem` allowing you to pass a `Proc` to add arbitrary commits on the intermediate branch after it's cut but before the PR is created, btw 😉

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -110,17 +110,58 @@ def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_bra
...
-        pr_body = <<~BODY
-          Merging `#{head_branch}` into `#{base_branch}`.
+        # Live playground to edit this graph
+        # https://mermaid.live/edit#pako:eNqNkU1PwzAMhv9KZanqpWVFwCVHQOKCxoFrLl7itdGaZEoTTWjqf8eLOo2i8eGTPx77TewjKK8JBJTlUbqCzTgTRTEHJ6s6E18C7vtqkc4li8Y9BnSqX6MlBqoYkttV9Y_cW9AUGLxt2_Y7Nfb-8OStNfEVNzQwtcVhpCU1XUJ2p7KU7vzAXNlkmSLQQDjS6v7m4S7nVU9q51O8UsmSv7jzSEuho9Xc3pzaG-Oib_KXlxr_QL8onbuuVy9unvrXbKiBCV645qvme0mIPVmSINjVtMU0RAm8O0YxRf_-4RQIbqca0l5jpGeDXUALIu9_-gRejKzU
+        meramid_git_graph = <<~GRAPH
+          ```mermaid
+          %%{
+            init: {
+              'gitGraph': {
+                'mainBranchName': 'trunk',
```
AliSoftware: Shouldn't this be `#{base_branch}` instead?

Even if most of our repos have migrated to all use `trunk` some of our repos still use a different name for their default branch anyway (`develop` for Tumblr-iOS, `master` for Tumblr-Android, `main` for some other apps from OtherBets…)

mokagio: Yes, I missed it when porting the code from the Mermaid live editor.

```suggestion
                'mainBranchName': '#{base_branch}',
```

wzieba:
I just came here to say: nice!

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
```diff
@@ -110,17 +110,58 @@ def self.create_backmerge_pr(token:, repository:, title:, head_branch:, base_bra
...
           #{head_branch.rjust(40)}  ----o-- - - -
           #{' ' * 40}       \\
           #{intermediate_branch.rjust(40)}        `---.
           #{' ' * 40}             \\
           #{base_branch.rjust(40)}  ------------x- - -
           ```
+        GRAPH
+
+        pr_body = <<~BODY
+          Merging `#{head_branch}` into `#{base_branch}`.
```
iangmaia, line 153: wdyt about adding the chart type as a parameter? Something like `:image`, `:ascii`, `:both`.

mokagio: My first instinct was to say thatI'd rather always have both in case the Mermaid one fails, which is something we cannot know at the time we generated the body because it depends on GitHub.

But, it's good to build flexible systems (_unopinionated tools, opinionated golden path_) and the adding a parameter like this does only makes the code a tiny bit more complex (_simple is usually better_).

I'll update 👍

mokagio: > I'll update 👍

Spawned #596 to not slow this down while waiting for that.

iangmaia:
> as long as characters from `[a-zA-Z0-9/.-]` don't cause trouble (anymore?) that should be ok to use now?

This seems to be the case as the images @mokagio generated so far contained these characters (I think the capital letters are the only exception)?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb
iangmaia: IMO the fonts and the overall size of the chart look too big for me (or maybe I'm too used to the ASCII art 😄 ), taking a lot of space of the PR text block. Is there a way to make it a bit smaller?

AliSoftware: It also feels strange that the commits on `trunk` seem to have happened only after the commits from `release/4.53` in the screenshot above; It's like if the 3 commits from `trunk` that were illustrated in this Mermaid graph were only pushed after `release/*` and `merge/release-*-into-*` had been created, instead of suggesting that the `trunk` branch has been existing and living its life for way longer before `release/*` branch… 🤔

AliSoftware:
> > as long as characters from `[a-zA-Z0-9/.-]` don't cause trouble (anymore?) that should be ok to use now?
> 
> This seems to be the case as the images @mokagio generated so far contained these characters (I think the capital letters are the only exception)?

Yeah, seems they fixed the bug since last time I tried it? 🤷

PR #597, opened 2024-09-26
Add Upload Buildkite Pipeline Action

iangmaia:
## What does it do?

Adds the `buildkite_pipeline_upload ` action, which uploads a Buildkite pipeline to a running Buildkite build, by running `buildkite-agent pipeline upload` ([ref](https://buildkite.com/docs/agent/v3/cli-pipeline#uploading-pipelines)).
Uploading the pipeline will add all its steps to the currently running build, without starting a new build altogether.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -2,31 +2,23 @@ module Fastlane
   module Actions
     class BuildkiteUploadPipelineAction < Action
       DEFAULT_ENV_FILE = File.join('.buildkite', 'shared-pipeline-vars').freeze
-      DEFAULT_COMMIT = 'HEAD'.freeze
 
       def self.run(params)
-        env_file = params[:env_file]
         pipeline_file = params[:pipeline_file]
-        branch = params[:branch]
-        commit = params[:commit]
+        env_file = params[:env_file]
+        environment = params[:environment]
 
         UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
-        UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != DEFAULT_COMMIT
         UI.user_error!('This action can only be called from a Buildkite CI build') unless ENV['BUILDKITE'] == 'true'
 
-        UI.message "Adding steps from `#{pipeline_file}` to the current build (#{"branch: `#{branch}`, " if branch}commit: `#{commit}`)"
-
-        env_vars = {
-          'BUILDKITE_BRANCH' => branch,
-          'BUILDKITE_COMMIT' => commit
-        }.compact
+        UI.message "Adding steps from `#{pipeline_file}` to the current build"
 
         if env_file && File.exist?(env_file)
           UI.message(" - Sourcing environment file beforehand: #{env_file}")
 
-          sh(env_vars, "source #{env_file.shellescape} && buildkite-agent pipeline upload #{pipeline_file.shellescape}")
+          sh(environment, "source #{env_file.shellescape} && buildkite-agent pipeline upload #{pipeline_file.shellescape}")
         else
-          sh(env_vars, 'buildkite-agent', 'pipeline', 'upload', pipeline_file)
+          sh(environment, 'buildkite-agent', 'pipeline', 'upload', pipeline_file)
         end
       end
 
@@ -51,17 +43,10 @@ module Fastlane
             type: String
           ),
           FastlaneCore::ConfigItem.new(
-            key: :branch,
-            description: 'The branch you want to run the pipeline on',
-            optional: true,
-            type: String
-          ),
-          FastlaneCore::ConfigItem.new(
-            key: :commit,
-            description: 'The commit hash you want to run the pipeline on',
-            optional: true,
-            default_value: DEFAULT_COMMIT,
-            type: String
+            key: :environment,
+            description: 'Environment variables to load when running `pipeline upload`, to allow for variable substitution in the YAML pipeline',
+            type: Hash,
+            default_value: {}
           ),
         ]
       end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -8,8 +8,9 @@ module Fastlane
         commit = params[:commit]
 
         UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
+        UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != 'HEAD'
 
-        UI.message "Uploading pipeline #{pipeline_file} on branch #{branch}, commit #{commit}"
+        UI.message "Uploading pipeline on #{pipeline_file}, #{"branch #{branch}, " if branch}commit #{commit}"
 
         ENV['BUILDKITE_BRANCH'] = branch
         ENV['BUILDKITE_COMMIT'] = commit
@@ -45,7 +46,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(
             key: :branch,
             description: 'The branch you want to run the pipeline on',
-            optional: false,
+            optional: true,
             type: String
           ),
           FastlaneCore::ConfigItem.new(
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -9,6 +9,7 @@ module Fastlane
 
         UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
         UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != 'HEAD'
+        UI.user_error!('This action can only be called from a Buildkite CI build') unless ENV['BUILDKITE'] == 'true'
 
         UI.message "Uploading pipeline on #{pipeline_file}, #{"branch #{branch}, " if branch}commit #{commit}"
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -1,6 +1,9 @@
 module Fastlane
   module Actions
     class BuildkiteUploadPipelineAction < Action
+      DEFAULT_ENV_FILE = File.join('.buildkite', 'shared-pipeline-vars').freeze
+      DEFAULT_COMMIT = 'HEAD'.freeze
+
       def self.run(params)
         env_file = params[:env_file]
         pipeline_file = params[:pipeline_file]
@@ -44,6 +47,7 @@ module Fastlane
             key: :env_file,
             description: 'The path to a bash file to be sourced before uploading the pipeline',
             optional: true,
+            default_value: DEFAULT_ENV_FILE,
             type: String
           ),
           FastlaneCore::ConfigItem.new(
@@ -56,7 +60,7 @@ module Fastlane
             key: :commit,
             description: 'The commit hash you want to run the pipeline on',
             optional: true,
-            default_value: 'HEAD',
+            default_value: DEFAULT_COMMIT,
             type: String
           ),
         ]
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -1,0 +1,70 @@
+module Fastlane
+  module Actions
+    class BuildkiteUploadPipelineAction < Action
+      def self.run(params)
+        env_file = params[:env_file]
+        pipeline_file = params[:pipeline_file]
+        branch = params[:branch]
+        commit = params[:commit]
+
+        UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
+
+        UI.message "Uploading pipeline #{pipeline_file} on branch #{branch}, commit #{commit}"
+
+        ENV['BUILDKITE_BRANCH'] = branch
+        ENV['BUILDKITE_COMMIT'] = commit
+
+        if env_file && File.exist?(env_file)
+          UI.message("Sourcing environment file: #{env_file}")
+
+          sh(". #{env_file} && buildkite-agent pipeline upload #{pipeline_file}")
+        else
+          sh('buildkite-agent', 'pipeline', 'upload', pipeline_file)
+        end
+      end
+
+      def self.description
+        # https://buildkite.com/docs/agent/v3/cli-pipeline#uploading-pipelines
+        'Uploads a pipeline to Buildkite, adding all its steps to the current build'
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :env_file,
+            description: 'The path to the environment variables file to be sourced before uploading the pipeline',
+            optional: true,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :pipeline_file,
+            description: 'The path to the Buildkite pipeline file',
+            optional: false,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :branch,
+            description: 'The branch you want to run the pipeline on',
+            optional: false,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :commit,
+            description: 'The commit hash you want to run the pipeline on',
+            optional: true,
+            default_value: 'HEAD',
+            type: String
+          ),
+        ]
+      end
+
+      def self.authors
+        ['Automattic']
+      end
+
+      def self.is_supported?(platform)
+        true
+      end
+    end
+  end
+end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -14,7 +14,7 @@ module Fastlane
         UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != DEFAULT_COMMIT
         UI.user_error!('This action can only be called from a Buildkite CI build') unless ENV['BUILDKITE'] == 'true'
 
-        UI.message "Uploading pipeline on #{pipeline_file}, #{"branch #{branch}, " if branch}commit #{commit}"
+        UI.message "Adding steps from `#{pipeline_file}` to the current build (#{"branch: `#{branch}`, " if branch}commit: `#{commit}`)"
 
         env_vars = {
           'BUILDKITE_BRANCH' => branch,
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -21,7 +21,7 @@ module Fastlane
         if env_file && File.exist?(env_file)
           UI.message(" - Sourcing environment file beforehand: #{env_file}")
 
-          sh(env_vars, ". #{env_file.shellescape} && buildkite-agent pipeline upload #{pipeline_file.shellescape}")
+          sh(env_vars, "source #{env_file.shellescape} && buildkite-agent pipeline upload #{pipeline_file.shellescape}")
         else
           sh(env_vars, 'buildkite-agent', 'pipeline', 'upload', pipeline_file)
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -11,7 +11,7 @@ module Fastlane
         commit = params[:commit]
 
         UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
-        UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != 'HEAD'
+        UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != DEFAULT_COMMIT
         UI.user_error!('This action can only be called from a Buildkite CI build') unless ENV['BUILDKITE'] == 'true'
 
         UI.message "Uploading pipeline on #{pipeline_file}, #{"branch #{branch}, " if branch}commit #{commit}"
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
@@ -13,15 +13,17 @@ module Fastlane
 
         UI.message "Uploading pipeline on #{pipeline_file}, #{"branch #{branch}, " if branch}commit #{commit}"
 
-        ENV['BUILDKITE_BRANCH'] = branch
-        ENV['BUILDKITE_COMMIT'] = commit
+        env_vars = {
+          'BUILDKITE_BRANCH' => branch,
+          'BUILDKITE_COMMIT' => commit
+        }.compact
 
         if env_file && File.exist?(env_file)
-          UI.message("Sourcing environment file: #{env_file}")
+          UI.message(" - Sourcing environment file beforehand: #{env_file}")
 
-          sh(". #{env_file} && buildkite-agent pipeline upload #{pipeline_file}")
+          sh(env_vars, ". #{env_file.shellescape} && buildkite-agent pipeline upload #{pipeline_file.shellescape}")
         else
-          sh('buildkite-agent', 'pipeline', 'upload', pipeline_file)
+          sh(env_vars, 'buildkite-agent', 'pipeline', 'upload', pipeline_file)
         end
       end
 
@@ -33,15 +35,15 @@ module Fastlane
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(
-            key: :env_file,
-            description: 'The path to the environment variables file to be sourced before uploading the pipeline',
-            optional: true,
+            key: :pipeline_file,
+            description: 'The path to the YAML pipeline file to upload',
+            optional: false,
             type: String
           ),
           FastlaneCore::ConfigItem.new(
-            key: :pipeline_file,
-            description: 'The path to the Buildkite pipeline file',
-            optional: false,
+            key: :env_file,
+            description: 'The path to a bash file to be sourced before uploading the pipeline',
+            optional: true,
             type: String
           ),
           FastlaneCore::ConfigItem.new(
--- a/spec/buildkite_upload_pipeline_action_spec.rb
+++ b/spec/buildkite_upload_pipeline_action_spec.rb
@@ -87,63 +87,49 @@ describe Fastlane::Actions::BuildkiteUploadPipelineAction do
     end
 
     it 'calls the right command to upload the pipeline without env_file' do
-      expect(Fastlane::Action).to receive(:sh).with('buildkite-agent', 'pipeline', 'upload', pipeline_file)
+      expect(Fastlane::Action).to receive(:sh).with(
+        { 'BUILDKITE_BRANCH' => branch, 'BUILDKITE_COMMIT' => commit_default },
+        'buildkite-agent', 'pipeline', 'upload', pipeline_file
+      )
       expect(Fastlane::UI).to receive(:message).with("Uploading pipeline on #{pipeline_file}, branch #{branch}, commit #{commit_default}")
 
       run_described_fastlane_action(
         pipeline_file: pipeline_file,
         branch: branch
       )
-
-      expect(ENV['BUILDKITE_BRANCH']).to eq(branch)
-      expect(ENV['BUILDKITE_COMMIT']).to eq(commit)
     end
 
     it 'calls the right command to upload the pipeline with env_file' do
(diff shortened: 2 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        UI.message "Uploading pipeline #{pipeline_file} on branch #{branch}, commit #{commit}"
+
+        ENV['BUILDKITE_BRANCH'] = branch
+        ENV['BUILDKITE_COMMIT'] = commit
+
+        if pipeline_env_file && File.exist?(pipeline_env_file)
+          UI.message("Sourcing environment file: #{pipeline_env_file}")
+
+          sh(". #{pipeline_env_file} && buildkite-agent pipeline upload #{pipeline_file}")
```
iangmaia: I could't find a good way of sourcing the `pipeline_env_file`, making the variables it exports available to the `buildkite-agent` call, while also keeping the split arguments.
Since the parameters in this string are files that are validated, I think there's less room for potential issues so I just kept it like this for now.

AliSoftware: > Since the parameters in this string are files that are validated

I don't see what you mean by those being validated? It's still possible to call that action with `buildkite_upload_pipeline(pipeline_env_file: ' false || rm -rf $HOME', pipeline_file: 'pipeline.yml && rm -rf /')` isn't it?

I'm not sure there's a way to use the split arguments syntax here indeed, given the need for `&&` (and I assume you tested having two separate calls of `sh`, but that didn't work probably because each call to `sh` spawns its own shell subprocess so they don't share their environment/context…)

So if we're going to use the single-string syntax instead of split arguments, we should use `shellescape` on them
```suggestion
          sh(". #{pipeline_env_file.shellescape} && buildkite-agent pipeline upload #{pipeline_file.shellescape}")
```
(you might need to add a `require 'shellwords'` at the top of this file for `String#shellescape` to be found… though it might already be required by fastlane itself at that point, so maybe not necessary?)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
...
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :pipeline_env_file,
+            description: 'The path to the environment variables file to be sourced before uploading the pipeline',
+            optional: true,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :pipeline_file,
+            description: 'The path to the Buildkite pipeline file',
```
AliSoftware:
```suggestion
            description: 'The path to the YAML pipeline file to upload',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
...
+        pipeline_file = params[:pipeline_file]
+        branch = params[:branch]
+        commit = params[:commit]
+
+        UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
+
+        UI.message "Uploading pipeline #{pipeline_file} on branch #{branch}, commit #{commit}"
+
+        ENV['BUILDKITE_BRANCH'] = branch
+        ENV['BUILDKITE_COMMIT'] = commit
```
AliSoftware: Wait, does that really result in the behavior you expect? i.e. the `steps` parsed from the `pipeline_file` by `buildkite-agent` and that it then adds as new jobs… that will really doing the cloning of the repo on said branch?

1️⃣  What happens if we provide a value for the `:branch` parameter but not for the `:commit`[^1] (which will likely be the case of most usages where we'd provide an explicit `branch` like `branch: "release/#{release_version_current}"`), or vice-versa?
Is there a risk that the uploaded steps might reference the provided branch (e.g. `release/*`) but also the current commit (e.g. head commit of `trunk`), thus creating an inconsistency in the env vars? Or are we guaranteed that Buildkite will set the missing one based on the only one we provide?

2️⃣  Also, what happens if the `pipeline-shared-vars` script relies on `BUILDKITE_COMMIT` to do some custom logic? If we call `buildkite_upload_pipeline(pipeline_env_file: 'pipeline-shared-vars', pipeline_file: 'foo.yml', branch: 'release/12.3')`, then `BUILDKITE_BRANCH` will be set to `release/12.3` by the above line before sourcing the env file, but `BUILDKITE_COMMIT` will be untouched (or will it be set to the empty string? Or will it be unset?)[^1], so either it will be empty and the `pipeline-shared-vars` file will not expect that, or it will keep the value from the job that called that action, and thus point to a commit SHA that won't match the `BUILDKITE_BRANCH`… 🤔 

3️⃣  Speaking of, should we always set those values, or only do it `unless branch.nil?` / `unless commit.nil?`? (see also footnote[^2])

4️⃣  Finally, is there a risk that the setting of the new values `ENV['BUILDKITE_BRANCH']=…` and `ENV['BUILDKITE_COMMIT']=…` here might keep their values even after executing the action, thus potentially impacting the rest of the code run in the `Fastfile`'s lane? For example if your lane runs from `trunk`, and calls `buildkite_upload_pipeline(…, branch: 'release/12.3')` but then calls other actions afterwards, and those other actions (still running on the caller job and thus on `trunk`) end up relying on `BUILDKITE_COMMIT` and/or `BUILDKITE_BRANCH`, isn't there a risk they'll accidentally use the values you overwrote there? [^2]

[^1]: I wrote that before I realized that your `ConfigItem(key: :commit)` provided a `default_value: 'HEAD'`; so maybe that's not that much of a problem after all; except if we want to mark the parameters as `conflicting_options` like I suggest in my other comment, at which point we might need to make the `commit` `ConfigItem` default to `nil` to allow a `branch` to be provided without conflicting…

[^2]: Note that [`sh` accepts being passed a `Hash` as its first argument that can be used to pass custom env var overrides](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/helper/sh_helper.rb#L116-L119). You should thus be able to use this to pass the `BUILDKITE_COMMIT`/`BUILDKITE_BRANCH` env vars, like `sh({BUILDKITE_BRANCH: branch, BUILDKITE_COMMIT: commit}.compact, "rest of your command here")`; this could potentially solve the solution by ensuring those env vars' values are only overridden for the duration of the `sh` call and nothing more.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
...
+            key: :branch,
+            description: 'The branch you want to run the pipeline on',
+            optional: false,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :commit,
+            description: 'The commit hash you want to run the pipeline on',
+            optional: true,
+            default_value: 'HEAD',
```
AliSoftware: Should the caller be allowed to provide both `commit` and `branch` (while if both values are provided there's a risk they are not consistent with one another)? Shouldn't we use `conflicting_options: [:branch]` here (and `conflicting_options: [:commit]` on the `:branch` ConfigItem)?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
+module Fastlane
+  module Actions
+    class BuildkiteUploadPipelineAction < Action
+      def self.run(params)
```
AliSoftware: I wonder if it'd be worth adding a check for CI here, like

```ruby
UI.user_error!('This action can only be called from a Buildkite CI build') unless ENV['BUILDKITE']=='true'
```

So that if it's accidentally called from someone's Mac (when running a lane in their Terminal.app that ends up calling this action), we'd have a useful error message?

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
...
+          FastlaneCore::ConfigItem.new(
+            key: :pipeline_file,
+            description: 'The path to the Buildkite pipeline file',
+            optional: false,
+            type: String
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :branch,
+            description: 'The branch you want to run the pipeline on',
+            optional: false,
```
AliSoftware: At first when I started reading the code before I reached this point, I thought that `branch` would be optional (and in such case, the steps added by `buildkite-agent pipeline upload` would just run on the same branch as the build on which that action was called). This also explains my questions on L14–L15 above.

While it's likely that we'll want to provide the `branch` more often than not in the main use cases we'll currently have for this action, I think there are also valid use cases for which we'd just want to upload additional steps to the current pipeline on the same branch.

There are actually already [use cases for that (= calling `pipeline upload` from steps but still on the same branch) in Beeper Desktop](https://github.com/beeper/beeper-desktop-new/blob/2af519081993f9552ec429f4c532581a72f55917/.buildkite/pipelines/official-build.yml#L44-L53) for example. _Sure for now they call `buildkite-agent pipeline upload` directly from the `command:`, but if one day they wanted to migrate that logic into a dedicated _fastlane_ lane and start using this action for that…_

spec/buildkite_upload_pipeline_action_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      expect do
+        run_described_fastlane_action(branch: branch)
+      end.to raise_error(FastlaneCore::Interface::FastlaneError, /pipeline_file/)
+    end
+
+    it 'raises an error when branch is not provided' do
+      expect do
+        run_described_fastlane_action(pipeline_file: pipeline_file)
+      end.to raise_error(FastlaneCore::Interface::FastlaneError, /branch/)
+    end
```
AliSoftware: As mentioned above, not sure this should be a requirement 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
+
+        UI.message "Uploading pipeline #{pipeline_file} on branch #{branch}, commit #{commit}"
+
+        ENV['BUILDKITE_BRANCH'] = branch
+        ENV['BUILDKITE_COMMIT'] = commit
+
+        if env_file && File.exist?(env_file)
+          UI.message("Sourcing environment file: #{env_file}")
```
AliSoftware: Nitpick, but I wonder if it might look odd to see in the logs "Uploading pipeline" first then "Sourcing environment file" next, while in practice the sourcing happens first and the upload happens next.

Maybe a simple fix for that is to update the wording to:
```suggestion
          UI.message(" - Sourcing environment file beforehand: #{pipeline_env_file}")
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :env_file,
+            description: 'The path to the environment variables file to be sourced before uploading the pipeline',
+            optional: true,
+            type: String
+          ),
```
AliSoftware: When I saw this `ConfigItem`'s name at first, I thought this was about [the `$BUILDKITE_ENV_FILE` variable exposed by Buildkite itself](https://buildkite.com/docs/pipelines/environment-variables#BUILDKITE_ENV_FILE)… 

Given our conventions I think we might as well call this item `:pipeline_shared_vars`?

Also, given your implementation already takes care of checking if the file exists and ignoring it otherwise, I'd suggest to add `'pipeline-shared-vars'` as a default value for this `ConfigItem`, given that's what our convention uses (and our `local.default.pipeline_upload_steps` on our IaC also hardcodes anyway), so that we don't have to pass it explicitly to all our call sites in our client repo's `Fastfile` every time.

```suggestion
          FastlaneCore::ConfigItem.new(
            key: :pipeline_shared_vars_file,
            description: 'The path to a bash file to be sourced before uploading the pipeline',
            optional: true,
            default_value: 'pipeline-shared-vars',
            type: String
          ),
```

[EDIT] I now see that you force-pushed changes while I was in the middle of this review, and you now renamed it `env_file` instead of `pipeline_env_file`; maybe that new name is ok then. I'd still provide a default value though.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
AliSoftware: Ok after doing the rest of my review + seeing the changes from your force-push done while I was reviewing, if we end up keeping `ConfigItem(key: :commit)` to have a `default_value: 'HEAD'`, maybe only points 3️⃣  and  4️⃣  of the above applies after all…?

And that case, instead of using `conflicting_options`'s built-in feature of `ConfigItem` to detect a user accidentally providing both, maybe we could test such a mistake from inside `def self.run` like `UI.user_error!('You should not provide both branch and commit') if !branch.nil? && commit != 'HEAD'` or something? 🤷

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: > I don't see what you mean by those being validated?

I meant `validated` by the checks to make sure both files exist, so it makes a bit harder to inject commands.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: > I now see that you force-pushed changes

Sorry about that 😞 -- when I created the app PR using this action, I realized the `pipeline_env_file` was a bit convoluted and renamed it back to `env_file`, which was how I started. I think that was the only change after I marked it for Ready For Review?
I assumed it was still "early enough"; but in any case, sorry about that, I should have just gone with a regular commit renaming the param.

iangmaia: The problem with defining a `default_value` for `env_file` is that a default would have to assume a location of the file. Maybe something like `../.buildkite/shared-pipeline-vars` would work, but wouldn't that be too specific to be added in `release-toolkit`? 🤔 

For now I've adopted the suggested description on 7349eb36526b2d7e1d0b95beb822d0399ab4a6f4.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: Agreed! Updated together with other env var related changes here: 07661f76ab61bc32a665e8837c27c68a2a02c5f3

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: True, good point. Updated here: 7349eb36526b2d7e1d0b95beb822d0399ab4a6f4

spec/buildkite_upload_pipeline_action_spec.rb
iangmaia: Updated on 7349eb36526b2d7e1d0b95beb822d0399ab4a6f4

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: Good suggestion! Added on d58c444b323117e40107d1de91cd5033242cba1b.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: Added a validation (not using `conflicting_options`, though) on 7349eb36526b2d7e1d0b95beb822d0399ab4a6f4.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: Nice catch on the impact of modifying `ENV` and setting the vars only on `sh`, thanks! 👍 

Updated the way we set the env vars on 07661f76ab61bc32a665e8837c27c68a2a02c5f3. The handling of `:branch`/`:commit` were updated in other commits as mentioned in other comments.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: > I assume you tested having two separate calls of sh, but that didn't work probably because each call to sh spawns its own shell subprocess so they don't share their environment/context…

🎯 exactly!

For completeness, also mentioning here that I've added the `.shellescape` and did a few other `sh()` / env updates on 07661f76ab61bc32a665e8837c27c68a2a02c5f3.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: Description updated on 07661f76ab61bc32a665e8837c27c68a2a02c5f3. I've also changed the order of the `ConfigItem`s so that the required parameter is the first one and the optional ones come later.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,74 @@
...
+        env_file = params[:env_file]
+        pipeline_file = params[:pipeline_file]
+        branch = params[:branch]
+        commit = params[:commit]
+
+        UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
+        UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != 'HEAD'
+        UI.user_error!('This action can only be called from a Buildkite CI build') unless ENV['BUILDKITE'] == 'true'
+
+        UI.message "Uploading pipeline on #{pipeline_file}, #{"branch #{branch}, " if branch}commit #{commit}"
```
AliSoftware: Technically, we don't "upload a pipeline **on** `release-build.yml`", we "upload a pipeline _named_ `release-build.yml`" 😉 

```suggestion
        UI.message "Uploading pipeline `#{pipeline_file}` with #{"branch `#{branch}`, " if branch}commit `#{commit}`"
```
or
```suggestion
        UI.message "Adding steps from `#{pipeline_file}` to the current build (#{"branch: `#{branch}`, " if branch}commit: `#{commit}`)"
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
AliSoftware: Mmmh that's a good point that it might be too specific to be added in `release-toolkit`…
…But on the other hand our other buildkite-related actions in release-toolkit already assume things about how our pipelines are configured.

For example `buildkite_trigger_build` [sets the `PIPELINE` env var to provide the name of the pipeline file to use](https://github.com/wordpress-mobile/release-toolkit/blob/ee03eb9232597fd812c18d3fb0fbf91130facf97/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb#L9-L15), which only works because this assumes we use `buildkite-agent pipeline upload .buildkite/${PIPELINE:-pipeline.yml}` in our root step the pipeline (i.e. the step defined in `.tf`) and that's how we provide an alternate pipeline file to upload.

So I think it's safe to consider that we'll always call this file `shared-pipeline-vars` too.

---

As for what relative path to use, note that [in the context of fastlane actions, the current working directory is always set by fastlane to be the root of the project](https://docs.fastlane.tools/advanced/fastlane/#directory-behavior). And the call to `sh` within this action's implementation (to call `.` to source the env file and to call `buildkite-agent`) will also make the shell commands be run with the shell's current directory set to the root of the repo.

Which means that the paths that the caller will pass to the `pipeline_file` and `env_file` parameters in the client repo's `Fastfile` will also be paths relative to the root of the repo, namely `pipeline_file: '.buildkite/release-builds.yml', env_file: '.buildkite/shared-pipeline-vars'`. And I don't think it's unreasonable nor too specific for this action to assume the `.buildkite/` subdir name in the context of an action related to Buildkite, even for an action living in `release-toolkit` for which we wouldn't want to be too-specific to our own setups.

Granted, the name of the file itself (`pipeline-shared-vars`) might be a thing that is a bit more specific though; but as I said, we already make similar assumptions in `buildkite_trigger_build` using `PIPELINE` env var to set the pipeline file to use, so I think it'd still be OK (and a reasonable default that will allow us to avoid typing that parameter over and over in all call sites of our various `Fastfiles` 😉 )


TL;DR, I think this should still be fine:
```ruby
  default_value: './buildkite/pipeline-shared-vars',
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        env_vars = {
+          'BUILDKITE_BRANCH' => branch,
+          'BUILDKITE_COMMIT' => commit
+        }.compact
+
+        if env_file && File.exist?(env_file)
+          UI.message(" - Sourcing environment file beforehand: #{env_file}")
+
+          sh(env_vars, ". #{env_file.shellescape} && buildkite-agent pipeline upload #{pipeline_file.shellescape}")
```
AliSoftware: Nitpick: I think that using `source` instead of `.` is more readable.

 - I know that `.` is technically more portable as it's POSIX-standard, but by now most shells (at least bash and zsh) have `source` too, so that shouldn't be a problem
 - [We use `source` and not `.` in the root steps of our pipelines in our `*.tf` files too](https://github.com/Automattic/buildkite-ci/blob/trunk/src/buildkite/pipelines/_pipeline_upload_steps.tftpl.yml#L4) so we might as well be consistent, given one of the goals of this `builkite_upload_pipeline` action is to do the same steps (sourcing a bash file before doing pipeline upload) as when we upload a pipeline at the very start of a build 😉

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: There's some nuance to this as `pipeline.yml` (and even `.buildkite/pipeline.yml`) are possible [defaults on Buildkite side](https://buildkite.com/docs/agent/v3/cli-pipeline#uploading-pipelines-description).
But in general I agree it's fine to build this kind of conventions (e.g. `.buildkite/shared-pipeline-vars`), also due to the fact this this is a novel / non standard mechanism we built. So I'm fine with it 👍  I'll test it and add it.

By the way, I guess you meant this, right?
```yaml
default_value: `.buildkite/shared-pipeline-vars`, # not './buildkite/pipeline-shared-vars'
```

AliSoftware: > There's some nuance to this as `pipeline.yml` (and even `.buildkite/pipeline.yml`) are possible [defaults on Buildkite side](https://buildkite.com/docs/agent/v3/cli-pipeline#uploading-pipelines-description).

True. But to contrast this, the use of e.g the `PIPELINE` env var to specify the name of the YAML file to upload is not a default behavior from Buildkite itself, only an artefact of the fact that our steps use that env var in our call to `buildkite-agent pipeline upload ${PIPELINE:-pipeline.yml}` in the main step of all our pipelines; yet we make such assumption in the implementation of `buildkite_trigger_build` action. So all I'm saying is that some of our release-toolkit actions already rely on our own conventions specific to our setup and not just Buildkite's built-in behavior 🙃


> By the way, I guess you meant this, right?
> 
> ```yaml
> 
> default_value: `.buildkite/shared-pipeline-vars`, # not './buildkite/pipeline-shared-vars'
> 
> ```
Yes, of course. I can never remember the exact name of that file from the top of my head without looking it up 😅

iangmaia: > True. But to contrast this, the use of e.g the PIPELINE env var to specify the name of the YAML file to upload is not a default behavior from Buildkite itself [...]  all I'm saying is that some of our release-toolkit actions already rely on our own conventions specific to our setup and not just Buildkite's built-in behavior

Yep indeed, agreed.
Thanks for the suggestions! I've added the default `env_file` on 2478ff8437e35e1774ff5c2d2c9a1afe792047cd (with some changes for the tests on 2665c7b67998526236c5149f2621bce01295b8c6).

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
```diff
@@ -0,0 +1,78 @@
...
+      DEFAULT_COMMIT = 'HEAD'.freeze
+
+      def self.run(params)
+        env_file = params[:env_file]
+        pipeline_file = params[:pipeline_file]
+        branch = params[:branch]
+        commit = params[:commit]
+
+        UI.user_error!("Pipeline file not found: #{pipeline_file}") unless File.exist?(pipeline_file)
+        UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != 'HEAD'
```
AliSoftware:
```suggestion
        UI.user_error!('You should not provide both `branch` and `commit`') if !branch.nil? && commit != DEFAULT_COMMIT
```

AliSoftware (commented):
Looking good now; only nits left are a suggested an extra use of `DEFAULT_COMMIT`, and the [adjusted wording on the `UI.message`](https://github.com/wordpress-mobile/release-toolkit/pull/597/files#r1777800160), and we'd be good to merge 🙂

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: I switched to `.` while trying out alternatives for running `sh`, but totally agree with the point with consistency and with it being more familiar / easier to read. Updated on b062ef94afd01722f01dafee15b6567b969bc48b.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_upload_pipeline_action.rb
iangmaia: Not sure why I added the `on` later yesterday 🙃
But I really liked the "adding steps from" suggestion, sounds more accurate to what's going to happen. Updated on bf1916a0b4e953698afa854ff23ce7fe2aa9f8d5 & ad7d098f500197dc869e37df5f2318c553bc1fea (tests).

AliSoftware (approved):
Approving the code, but I think it would be worth testing (using a dummy project and some dummy pipelines) that the cases of overriding the commit/branch work as expected and affects the checkouts of all the uploaded steps as expected, before actually merging it

AliSoftware:
Food for thoughts: I wonder if it wouldn't make more sense to call that action `buildkite_pipeline_upload`.
 - On one hand, `buildkite_upload_pipeline` reads close to the English phrase ("upload the pipeline") describing that action, so might feel nicer
 - On the other hand, we're so used to "pipeline upload" being said in that order given we know the underlying command this action calls is `buildkite-agent pipeline upload`.
    - Using that order in the action name also acts as more evident self-documenting of what the action does (even if `buildkite_upload_pipeline` is still pretty clear… but slightly further than the actual command name so less direct correlation)
    - I've also in fact surprised myself almost referencing this action as `buildkite_pipeline_upload` when writing DMs with you as it came more naturally to me given the `pipeline upload` underlying command it calls.

Ultimately this is mostly bikeshedding, and not a big deal, but I figured I'd bring it up anyway before we commit to the final name for that action (would probably not be worth to change it after it has been initially released, so better get the one that feels best to everyone on the first time)

iangmaia:
After some tests and realizing the custom parameters for `BUILDKITE_BRANCH` / `BUILDKITE_COMMIT` wouldn't be forwarded to the actual jobs and only work for variable substitution in the pipeline YAML files, I discussed with @AliSoftware and we decided to [go for another approach](https://github.com/wordpress-mobile/release-toolkit/pull/597/commits/3f932e127d757384a8b1b6422bee4be7f4773b31):
1. Removing the `:branch` / `:commit` lane parameters
2. Adding an `environment` (a `Hash`) parameter instead, that the client of the library can use to pass variables to be used for substitution on pipeline YAML files (thus making it possible to checkout a different branch, if needed)

iangmaia:
> Food for thoughts: I wonder if it wouldn't make more sense to call that action `buildkite_pipeline_upload`.

I had the same thought in the beginning but considered the name following the English syntax would make more sense... but still, I don't feel very strongly about it and the arguments for `buildkite_pipeline_upload` also makes sense, so 🤷

iangmaia:
@AliSoftware I [ran a final test](https://buildkite.com/automattic/dayone-android/builds/16933) to make sure the `environment` parameter we added works for variables substitution in pipelines, and it worked fine 🎉 

Shall we merge this and release a new version?

AliSoftware:
:shipit:

PR #598, opened 2024-09-30
Test `buildkite_pipeline_upload` action in CI

mokagio:
## What does it do?

Calls the new `buildkite_pipeline_upload` action in CI to upload a dedicated pipeline and have an integration-level test for the action.

See https://github.com/wordpress-mobile/release-toolkit/pull/597#pullrequestreview-2333686032

- [x] Make sure you added an entry in [the `CHANGELOG.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md#trunk) to describe your changes under the appropriate existing `###` subsection of the existing `## Trunk` section. – N.A.
- [x] If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to describe how the changes will affect the migration from the previous major version and what the clients will need to change and consider. – N.A.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/pipeline.yml
```diff
@@ -1,27 +1,11 @@
...
-    IMAGE_ID: xcode-15.2
-  plugins:
-    - &ci_toolkit
-      automattic/a8c-ci-toolkit#2.18.2: ~
-    - &docker_plugin
-      docker#v5.8.0:
-        image: &ruby_version "public.ecr.aws/docker/library/ruby:3.2.2"
-        propagate-environment: true
-        environment:
-          - "RUBYGEMS_API_KEY"
```
mokagio, line 15: I got rid of all this because both were used only once.

.buildkite/pipeline.yml
```diff
@@ -39,27 +23,34 @@ steps:
 
           echo "--- :rspec: Run Rspec"
           bundle exec rspec --profile 10 --format progress
-        env: *xcode_image
-        plugins: [*ci_toolkit]
+
+          echo "--- :fastlane: :buildkite: Test pipeline upload action"
+          bundle exec fastlane run buildkite_pipeline_upload pipeline_file:.buildkite/pipeline_for_upload_action_test.yml
```
mokagio, line 28: I considered having a dedicated step for this, and maybe we should, but at this point, before getting more feedback, I decided to leave in the same step as the tests.

What do you think?

A dedicated step might make things a bit clearer, but also seems like a lot (including extracting shared setup) for a simple action like this....

.buildkite/pipeline.yml
```diff
@@ -39,27 +23,34 @@ steps:
...
     if: build.tag != null
     depends_on:
      - test
     # Note: We intentionally call a separate `.sh` script here (as opposed to having all the
     # commands written inline) to avoid leaking a key used in the process in clear in the
     # BUILDKITE_COMMAND environment variable.
     command: .buildkite/commands/gem-push.sh
-    plugins: [*docker_plugin]
+    plugins:
+      - docker#v5.8.0:
```
mokagio, line 50: This version could be moved to the shared vars, too, but given it's used only once I haven't done it yet.

Conversely, the CI toolkit was already setup so I leaned into it rather than removing it.

.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+CI_TOOLKIT_PLUGIN_VERSION="2.18.2"
+RUBY_VERSION=$(cat .ruby-version)
```
mokagio, line 7: I think it'd be good to use the same version in CI as the project uses locally.

But I'm also open to the idea of decoupling them in case we run into some situation where we want to use a version locally for which Docker has no image.

However, if we run into that scenario, maybe we should take a step back and ask why? Given's Ruby adoption, a version with no Docker image seems odd. Maybe if we end up in that scenario we are doing something odd ourselves?

.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+CI_TOOLKIT_PLUGIN_VERSION="3.7.1"
```
mokagio, line 6: Notice this is now the latest version. We were a few versions behind.

.buildkite/pipeline_for_upload_action_test.yml
```diff
@@ -0,0 +1,6 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+steps:
+  - label: ":ok_hand: If this message is shown, the pipeline upload succeeded"
+    command: echo "Acknowledged."
```
mokagio, line 6: We could make this a bit more useful by, for example, uploading with the shared pipeline vars (see https://github.com/bloom/DayOne-Android/pull/4883/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R683) and check that a sample value is available

```
[[ $RUBY_VERSION = $(cat .ruby-version) ]] || false
```

Alternatively we could use a dedicated file so that the test is more straightforward

```
[[ $TEST_VALUE_FOR_PIPELINE_UPLOAD_WITH_ENV_FILE = "this string is defined in the file at .buildkite/test_env" ]] || false
```

.buildkite/pipeline.yml
```diff
@@ -1,27 +1,11 @@
...
 steps:
-  #################
-  # Build and Test
-  #################
-  - group: 🧪 Build and Test"
+  - group: ":test_tube: Build and Test"
     key: test
     steps:
-      - label: "🧪 Build and Test using Ruby {{ matrix.ruby }}"
+      - label: ":test_tube: Build and Test using Ruby {{ matrix.ruby }}"
```
AliSoftware, line 8: Might be worth using the occasion to add a custom check name for that step? (`github_commit_status:` or whatever the attribute is called)

.buildkite/pipeline.yml
AliSoftware: I'm ok with keeping it here

.buildkite/pipeline.yml
AliSoftware: One benefit I like with moving the version values in `shared-pipeline-vars` is that it's immediately visible when we do a wave of updates what things could be updated and what plugins are in use in the pipeline and what version of each plugin is used, without having to parse through the YML files content to determine this.

So it's more to have all the versions grouped in one single place making them easy to locate, even if some of them don't really need DRY-ing because they're only used once.

.buildkite/pipeline_for_upload_action_test.yml
AliSoftware: Yes I think it could be valuable to test those indeed. After all this is a key feature of the action we might want to test via integration testing:
- Make the action load the `shared-pipeline-vars` (which should be the default, btw) and validate that env vars were available
- Provide a Hash of specific env vars to the action's `environment` parameter and validate they were taken into account too. That being said, I'm not sure if it's possible to pass a Hash from the CLI via the `fastlane run` command 🤔

That being said, those features should appear be covered by our unit tests of the action… so we might wonder if it's worth also testing them in integration testing; but that would allow us to ensure Buildkite's behavior with passing env vars via `Action.sh` behaves as expected at least and end up interpolated properly, so maybe still useful.

.xcode-version
```diff
@@ -0,0 +1 @@
+15.2
```
AliSoftware, line 1: To be updated to a more recent image at some point (potentially separate PR)?

.buildkite/shared-pipeline-vars
AliSoftware: I agree we should use the same Ruby version for both

.buildkite/pipeline_for_upload_action_test.yml
```diff
@@ -0,0 +1,6 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+steps:
+  - label: ":ok_hand: If this message is shown, the pipeline upload succeeded"
```
AliSoftware, line 5: Could benefit a custom (and more concise) `github_commit_status`.

Could be done in a separate PR though as for it to be useful we'd also want to disable the default status created by Buildkite on each step (in the `release-toolkit.tf` of `buildkite-ci`) and switch to making all steps have explicit `notify`

iangmaia:
Thanks for this @mokagio! Looks like the build with the pipeline upload worked fine  🎉 
I've merged #597 and released `12.2.0` for `release-toolkit`.

AliSoftware:
@mokagio I just stumbled upon this old PR while looking at the opened PRs for this repo.

It this still relevant?
If so, could benefit fixing conflicts on the pipeline file as well as some freshing-up (account for some new convention decisions made since like wrt to using the emoji in labels after all, adding git_commit_status notify blocks, updating the IMAGE_ID to a more recent one, …) 🙂

mokagio:
@AliSoftware I'd say it's still relevant yet. It doesn't hurt to have these tests does it?

However, just to set expectations, I don't think I'll have time to address this before my support rotation next week subsequent AFK.

mokagio:
> I'd say it's still relevant yet [sic. should have been "yes"]. It doesn't hurt to have these tests does it?

Evidently, as relevant as this might have been, it was also low priority and of little consequence given it sat year for a year ignored by me.

Closing.

PR #609, opened 2024-10-31
Release 12.3.0 into trunk

iangmaia:
Releasing new version 12.3.0.

# What's Next

```
### New Features

- `buildkite_pipeline_upload`: prepend `.buildkite/` to the `pipeline_file` parameter to enforce our conventions [#608]

### Bug Fixes

- `create_release-backmerge_pull_request`: Fix the pre-check logic verifying if a PR is really needed or if there's nothing to backmerge. [#607]

### Internal Changes

- `buildkite_pipeline_upload`: makes sure all values passed in the environment parameter are strings [#608]


```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,16 @@ _None_
 
 ### Bug Fixes
 
+_None_
+
+### Internal Changes
+
+_None_
+
+## 12.3.0
+
+### Bug Fixes
+
 - `create_release-backmerge_pull_request`: Fix the pre-check logic verifying if a PR is really needed or if there's nothing to backmerge. [#607]
 
 ### Internal Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ _None_
 
 ## 12.3.0
 
+### New Features
+
+- `buildkite_pipeline_upload`: prepend `.buildkite/` to the `pipeline_file` parameter to enforce our conventions [#608]
+
 ### Bug Fixes
 
 - `create_release-backmerge_pull_request`: Fix the pre-check logic verifying if a PR is really needed or if there's nothing to backmerge. [#607]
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware (commented):
I was pondering why this was a `12.3.0` and not a `12.2.1` given the only changes are either internal or bug fixes, not new features nor breaking changes…

That made me realize that the fact that the `buildkite_pipeline_upload` action have changed its behavior to now prepend `.buildkite/` on relative paths of `pipeline_file` I think that change (from line 32) should be:
 - Either under Bug Fixes, rather than Internal Changes
 - Or under "New Features", which I think makes a bit more sense to me given the change
 - Or one could even argue it's a breaking change in practice—as it changes the behavior and will require changes in any client repos that were already using that action with a relative path. But I think too few client repos have adopted that action before the change made in this `release-toolkit` propagated too far, and you know exactly which ones they are and are already planning to update them ASAP… so maybe that's not worth marking this as a breaking in that context…

In any case, I think we should move line 32 of the CHANGELOG in an upper section before landing this PR and doing the release 🙂

iangmaia:
> I was pondering why this was a `12.3.0` and not a `12.2.1` given the only changes are either internal or bug fixes, not new features nor breaking changes…

I've changed it manually to `12.3.0`, related to what you pointed out below plus IMO it's a significant behavior change (though without affecting the API).

> That made me realize that the fact that the `buildkite_pipeline_upload` action have changed its behavior to now prepend `.buildkite/` on relative paths of `pipeline_file` I think that change (from line 32) should be:
> 
> * Either under Bug Fixes, rather than Internal Changes
> * Or under "New Features", which I think makes a bit more sense to me given the change
> * Or one could even argue it's a breaking change in practice—as it changes the behavior and will require changes in any client repos that were already using that action with a relative path. But I think too few client repos have adopted that action before the change made in this `release-toolkit` propagated too far, and you know exactly which ones they are and are already planning to update them ASAP… so maybe that's not worth marking this as a breaking in that context…
> 
> In any case, I think we should move line 32 of the CHANGELOG in an upper section before landing this PR and doing the release 🙂

Agreed!

AliSoftware (approved):

PR #612, opened 2024-11-14
Fix bug in `configure_apply` action

AliSoftware:
## What does it do?

When a new encryption key is recently added and pushed to `~/.mobile-secrets` by one developer, other developers will likely not yet have their own `~/.mobile-secrets` repo pulled to that latest commit that includes the new encryption key.

This means that in those scenarios, after a new project is set up for `configure` and a new encryption key is added, any other developer who would be told to "please now run `bundle exec fastlane run configure_apply`" will have the command fail, because on their local machine their `~/.mobile-secrets` would not yet have the new encryption key, and the preflight test we did about this at the start of the action would error early.

The fix is easy: make sure we only check for the presence of the encryption key _after_ we have entered the `prepare_repository` block and thus pulled and checked out the correct commit of `~/.mobile-secrets` first.

Related conversation: p1731606985135599/1731586195.925939-slack-C04PWEZSYFL

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- Fix bug in `configure_apply` that could happen if run on a user machine that hasn't yet `git-pull` the latest commits from `~/.mobile-secrets` to a commit _after_ the new encryption key had been pushed. [#612]
 
 ### Internal Changes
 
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

CHANGELOG.md
```diff
@@ -24,21 +24,21 @@ _None_
 
 ### Bug Fixes
 
-- Fix issue with post-processing of PR urls in the body of GitHub releases created via `create_github_release` [#610]
+- Fix issue with post-processing of PR urls in the body of GitHub releases created via `create_github_release`. [#610]
```
mokagio, line 27: 💅👍

mokagio (approved):
Looks good.

I'd love to have tests for this but it seems like an integration test with various moving parts that would go beyond the scope of this fix PR.

PR #626, opened 2025-01-09
Enable `frozen_strings_literal`

### LINKED ISSUE #625: Should we enable the `FrozenStringLiteralComment` lint?
While working on a private projects, we noticed that `.rubocop.yml` disables the `FrozenStringLiteralComment` lint.

There is a rationale for that setting, but I wonder whether it's time to reconsider it and adopt the standard. 

One case against it is that we don't know what Fastlane might be doing under the hood to the constants we pass it. We even run into at least one instance of a crash due to that, see https://github.com/wordpress-mobile/WordPress-iOS/pull/18417. Even with that, I still think we'd be better off enabling the lint and discover and fix possible issues in the tooling.

What do you think?

mokagio:
## What does it do?

Closes #625

Because most of the changes were done by `bundle exec rubocop -A` I addressed the test failures resulting to the code switching to immutable strings in dedicated commits. I hope it makes it easier to review.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,13 +30,6 @@ Style/EmptyMethod:
 Style/TrailingCommaInArrayLiteral:
   EnforcedStyleForMultiline: consistent_comma
 
-# Enabling the 'frozen_string_literal: true' comment is nice but should be done with extra care
-# because enabling it on code that accidentally mutates string literals will crash at runtime
-# (e.g. `name = 'Hello'` then later `name << ' world'` will crash if the comment is enabled)
-# So we might enable it at some point, be we will need very careful review that this doesn't break everything.
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
 # Unicode is pretty standard by now, and we want to be able to use em-dash, ellipsis, and unicode characters
 # to provide nice Monodraw.app graphs and similar, without being limited only by ascii.
 Style/AsciiComments:
@@ -48,7 +41,7 @@ Style/AsciiComments:
 # which would be incorrect and not what we want
 Style/StringConcatenation:
   Enabled: false
-  
+
 # This rule was enforced after upgrading Rubocop from `1.22.1` to `1.50.2`. We are disabling this rule for the
 # time being so that we don't see any unexpected behavior when running Release Toolkit actions that could be
 # changed by this rule. See https://github.com/wordpress-mobile/release-toolkit/pull/464#pullrequestreview-1396569629
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/encryption_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/encryption_helper.rb
@@ -21,23 +21,19 @@ module Fastlane
 
       def self.encrypt(plain_text, key)
         # Ensure consistent encoding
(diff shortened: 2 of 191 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/encryption_helper.rb
```diff
@@ -19,13 +21,12 @@ def self.cipher(op_type)
...
         # Ensure consistent encoding
-        plain_text.force_encoding(Encoding::UTF_8)
+        sanitized_plain_text = plain_text.dup.force_encoding(Encoding::UTF_8)
 
         cipher = cipher(OperationType::ENCRYPT)
         cipher.key = key
 
-        encrypted = cipher.update(plain_text)
-        encrypted << cipher.final
+        encrypted = cipher.update(sanitized_plain_text) + cipher.final
```
dangermattic, line 29: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/11ffc88428a1928ea85fa822c4bf60dcda5b3224/lib/fastlane/plugin/wpmreleasetoolkit/helper/encryption_helper.rb#L29"></span>Style/RedundantAssignment: Redundant assignment before returning detected.</td>
    </tr>
  </tbody>
</table>

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><code>Gemfile</code> was changed without updating its corresponding <code>Gemfile.lock</code>. Please run <code>bundle install</code> or <code>bundle update &lt;updated_gem&gt;</code>.</td>
    </tr>
  </tbody>
</table>

</p>

.rubocop.yml
```diff
@@ -30,13 +30,6 @@ Style/EmptyMethod:
 Style/TrailingCommaInArrayLiteral:
   EnforcedStyleForMultiline: consistent_comma
 
-# Enabling the 'frozen_string_literal: true' comment is nice but should be done with extra care
-# because enabling it on code that accidentally mutates string literals will crash at runtime
-# (e.g. `name = 'Hello'` then later `name << ' world'` will crash if the comment is enabled)
```
iangmaia, line 35: Let's hope we don't have many of such cases -- I think the more recent actions are probably fine. I guess the concern is more on the older ones, without much testing.

And I think we're all aware that issues caused by this kind of change (or any change in Fastlane related tooling, really) would be caught only during an app release -- I always wonder if we could come up with a way to mitigate this problem or just accept that as inevitable.

mokagio: Yeah... I think this is a situation where it's more about how we react to issues than how we prevent them. Because to fully prevent them we would have to run a copy of each of our app's release scenario, which would take a long time. I'd like to think we can nimbly fix issues that might arise at release time and that it would be a more effective use of our time.

iangmaia (approved):
I ran the updated codebase through Claude (which gave its ok) and checked a few things myself just in case.

:shipit:

PR #631, opened 2025-02-17
Add parameter to iOS Lint to make failing on extra strings optional

iangmaia:
Internal reference: pdnsEh-210-p2

## What does it do?

This PR adds to the `ios_lint_localizations` action a `fail_on_extra_strings` parameter to control whether to report violations when finding strings in translations that are not present in the base language. 

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
@@ -37,7 +37,8 @@ module Fastlane
         helper.run(
           input_dir: resolve_path(params[:input_dir]),
           base_lang: params[:base_lang],
-          only_langs: params[:only_langs]
+          only_langs: params[:only_langs],
+          fail_on_extra_strings: params[:fail_on_extra_strings]
         )
       end
 
@@ -179,6 +180,14 @@ module Fastlane
             key: :check_duplicate_keys,
             env_name: 'FL_IOS_LINT_TRANSLATIONS_CHECK_DUPLICATE_KEYS',
             description: 'Checks the input files for duplicate keys',
+            optional: true,
+            default_value: true,
+            type: Boolean
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :fail_on_extra_strings,
+            env_name: 'FL_IOS_LINT_TRANSLATIONS_FAIL_ON_EXTRA_STRINGS',
+            description: 'Should we report violations when finding strings in translations that are not present in the base language',
             optional: true,
             default_value: true,
             type: Boolean
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
@@ -38,7 +38,7 @@ module Fastlane
           input_dir: resolve_path(params[:input_dir]),
           base_lang: params[:base_lang],
           only_langs: params[:only_langs],
-          fail_on_extra_strings: params[:fail_on_extra_strings]
+          fail_on_strings_not_in_base_language: params[:fail_on_strings_not_in_base_language]
         )
       end
 
@@ -185,8 +185,8 @@ module Fastlane
             type: Boolean
           ),
           FastlaneCore::ConfigItem.new(
-            key: :fail_on_extra_strings,
-            env_name: 'FL_IOS_LINT_TRANSLATIONS_FAIL_ON_EXTRA_STRINGS',
+            key: :fail_on_strings_not_in_base_language,
+            env_name: 'FL_IOS_LINT_TRANSLATIONS_FAIL_ON_STRINGS_NOT_IN_BASE_LANGUAGE',
             description: 'Should we report violations when finding strings in translations that are not present in the base language',
             optional: true,
             default_value: true,
--- a/spec/ios_lint_localizations_spec.rb
+++ b/spec/ios_lint_localizations_spec.rb
@@ -62,8 +62,11 @@ describe Fastlane::Actions::IosLintLocalizationsAction do
     # @param [Bool|nil] check_duplicate_keys If `nil`, the test will run the action with the default `check_duplicate_keys` parameter value.
     #        If a `Bool` value is given, it will pass that.
     #        Using either `Bool` or `nil` adds some cruft, but lets us validate the action default behavior, so it doesn't change unexpectedly.
+    # @param [Bool|nil] fail_on_extra_strings If `nil`, the test will run the action with the default `fail_on_extra_strings` parameter value.
+    #        If a `Bool` value is given, it will pass that.
+    #        Controls whether to report violations when finding strings in translations that are not present in the base language.
     #
-    def run_l10n_linter_test(data_file:, check_duplicate_keys: nil)
+    def run_l10n_linter_test(data_file:, check_duplicate_keys: nil, fail_on_extra_strings: nil)
       # Arrange: Prepare test files
       test_file = File.join(File.dirname(__FILE__), 'test-data', 'translations', 'ios_lint_localizations', "#{data_file}.yaml")
       yml = YAML.load_file(test_file)
@@ -86,6 +89,7 @@ describe Fastlane::Actions::IosLintLocalizationsAction do
         base_lang: 'en'
       }
       parameters[:check_duplicate_keys] = check_duplicate_keys unless check_duplicate_keys.nil?
+      parameters[:fail_on_extra_strings] = fail_on_extra_strings unless fail_on_extra_strings.nil?
       result = run_described_fastlane_action(parameters)
 
       # Assert
@@ -204,6 +208,17 @@ describe Fastlane::Actions::IosLintLocalizationsAction do
       )
 
       expect(result).to eq({ 'fr' => ['`key3` expected placeholders for [Int] but found [] instead.'] })
+    end
+
+    it 'fails on extra strings in translations by default' do
+      run_l10n_linter_test(data_file: 'extra-strings-in-translations-error')
+    end
+
+    it 'do not report extra strings in translations when fail_on_extra_strings is false' do
+      run_l10n_linter_test(
+        data_file: 'extra-strings-in-translations-no-error',
+        fail_on_extra_strings: false
+      )
     end
   end
 end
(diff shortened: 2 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

spec/ios_lint_localizations_spec.rb
```diff
@@ -205,5 +209,16 @@ def run_l10n_linter_test(data_file:, check_duplicate_keys: nil)
 
       expect(result).to eq({ 'fr' => ['`key3` expected placeholders for [Int] but found [] instead.'] })
     end
+
+    it 'fails on extra strings in translations by default' do
+      run_l10n_linter_test(data_file: 'extra-strings-in-translations-error')
+    end
+
+    it 'do not report extra strings in translations when fail_on_extra_strings is false' do
```
mokagio:
```suggestion
    it 'does not report extra strings in translations when fail_on_extra_strings is false' do
```

spec/ios_lint_localizations_spec.rb
```diff
@@ -205,5 +209,16 @@ def run_l10n_linter_test(data_file:, check_duplicate_keys: nil)
...
+      run_l10n_linter_test(
+        data_file: 'extra-strings-in-translations-no-error',
+        fail_on_extra_strings: false
+      )
+    end
```
mokagio, line 229: Worth adding a test for the positive value, just in case?

```suggestion
    end
    
    it 'does report extra strings in translations when fail_on_extra_strings is true' do
      run_l10n_linter_test(
        data_file: 'extra-strings-in-translations-error',
        fail_on_extra_strings: true
      )
    end
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -183,6 +184,14 @@ def self.available_options
             default_value: true,
             type: Boolean
           ),
+          FastlaneCore::ConfigItem.new(
+            key: :fail_on_extra_strings,
+            env_name: 'FL_IOS_LINT_TRANSLATIONS_FAIL_ON_EXTRA_STRINGS',
+            description: 'Should we report violations when finding strings in translations that are not present in the base language',
```
mokagio: How do you feel about being extra verbose here? When reading "extra strings" it wasn't clear to me what the parameter meant.

```ruby
            key: :fail_on_localized_strings_not_in_base_language,
            env_name: 'FL_IOS_LINT_TRANSLATIONS_FAIL_ON_LOCALIZED_STRINGS_NOT_IN_BASE_LANGUAGE',
            description: 'Should we report violations when finding strings in translations that are not present in the base language',
```

If you can find a shorter name for the parameter that does an equal at explaining its function, I'd be happy with that, too.

mokagio: Side note. That looked like a very long env var name, but it works, at least on my machine

image

mokagio (approved):
Looks good. Only a couple of suggestions. Approving to unblock given there is no auto-merge.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
mokagio: > If you can find a shorter name for the parameter that does an equal at explaining its function, I'd be happy with that, too.

Bounced ideas [with ChatGPT](https://chatgpt.com/share/67b3f74e-5f38-8011-9e54-56d2a0242706) and got to a more succinct but clear-enough `fail_on_unmatched_localizations`. What do you think?

spec/ios_lint_localizations_spec.rb
```diff
@@ -205,5 +209,23 @@ def run_l10n_linter_test(data_file:, check_duplicate_keys: nil)
...
+      run_l10n_linter_test(data_file: 'extra-strings-in-translations-error')
+    end
+
+    it 'does not report extra strings in translations when fail_on_extra_strings is false' do
+      run_l10n_linter_test(
+        data_file: 'extra-strings-in-translations-no-error',
+        fail_on_extra_strings: false
+      )
+    end
+
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/1703df701a90952ce1e880ca5a225aec4de9db20/spec/ios_lint_localizations_spec.rb#L223"></span>Layout/TrailingWhitespace: Trailing whitespace detected.</td>
    </tr>
  </tbody>
</table>

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
iangmaia: > succinct but clear-enough fail_on_unmatched_localizations. What do you think?

I like it! Yeah, "missing strings" could potentially be about so many things. I think "unmatched localizations" makes it clearer about what the problem is.

AliSoftware: > If you can find a shorter name for the parameter that does an equal at explaining its function, I'd be happy with that, too.

Given the context, the term "localized" in "localized string" is implied.
So in fact I think that `fail_on_strings_not_in_base_language` would work.

> That looked like a very long env var name

I'd actually argue that we stop adding `env_name` for each of the arguments in our release-toolkit custom actions. We never use them (who ever set those instead of passing the argument explicitly?).

For some arguments that are common to multiple actions (like `GITHUB_TOKEN`) that makes sense, but for very-specific arguments that only make sense for that specific action, they don't provide anything and are kinda useless to have I'd say 🙃

spec/ios_lint_localizations_spec.rb
```diff
@@ -62,8 +62,11 @@
     # @param [Bool|nil] check_duplicate_keys If `nil`, the test will run the action with the default `check_duplicate_keys` parameter value.
     #        If a `Bool` value is given, it will pass that.
     #        Using either `Bool` or `nil` adds some cruft, but lets us validate the action default behavior, so it doesn't change unexpectedly.
+    # @param [Bool|nil] fail_on_extra_strings If `nil`, the test will run the action with the default `fail_on_extra_strings` parameter value.
```
AliSoftware: To double-check in the docs (or ask Cursor), but I think the proper YARD syntax for optional arguments can just be `Bool?` instead of `Bool|nil`; would be a good occasion to fix here and in other places in that same file where we did the same.

spec/test-data/translations/ios_lint_localizations/extra-strings-in-translations-error.yaml
```diff
@@ -0,0 +1,10 @@
+test_data:
+  en:
+    '"base_string" = "This is in the base language";'
+  fr:
+    '"base_string" = "This is in French";
+     "extra_string" = "This string only exists in French";'
```
AliSoftware: nitpick: instead of using single quotes around the whole (multiline) string (which I don't think would keep the literal newlines to begin with), what about using `|-` string block syntax?

```suggestion
  en: |-
    "base_string" = "This is in the base language";
  fr: |-
    "base_string" = "This is in French";
    "extra_string" = "This string only exists in French";
```

image

spec/test-data/translations/ios_lint_localizations/extra-strings-in-translations-no-error.yaml
```diff
@@ -0,0 +1,8 @@
+test_data:
+  en:
+    '"base_string" = "This is in the base language";'
+  fr:
+    '"base_string" = "This is in French";
+     "extra_string" = "This string only exists in French";'
```
AliSoftware: ditto
```suggestion
  en: |-
    "base_string" = "This is in the base language";
  fr: |-
    "base_string" = "This is in French";
    "extra_string" = "This string only exists in French";
```

spec/ios_lint_localizations_spec.rb
iangmaia: `[Boolean, nil]` was the suggestion I found online (confirmed by Cursor -- but it also got confused and suggested `Boolean?` when I asked if it was also possible 😅 ). Updated on c981041094434d430111a6abc895fcfdaa098ee9.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
iangmaia: I've renamed the var to `fail_on_strings_not_in_base_language`, thanks for the suggestions @mokagio & @AliSoftware! Updated on e90cd9b4d8234333982493d502dacd5fa3ec09a6.

AliSoftware (approved):

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
mokagio: > I'd actually argue that we stop adding env_name for each of the arguments in our release-toolkit custom actions. We never use them (who ever set those instead of passing the argument explicitly?).

Sounds good to me!

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -183,6 +184,14 @@ def self.available_options
             default_value: true,
             type: Boolean
           ),
+          FastlaneCore::ConfigItem.new(
+            key: :fail_on_strings_not_in_base_language,
+            env_name: 'FL_IOS_LINT_TRANSLATIONS_FAIL_ON_STRINGS_NOT_IN_BASE_LANGUAGE',
```
AliSoftware, line 189: I thought we said we'd stop introducing useless `env_name` for each of our `ConfigItem`… but seems you merged it anyway 🙃 😛

iangmaia: I didn't think much about it but I considered this a slightly different case (it's not a new action and there would be some value in being consistent with the rest of it; not having it might create questions unnecessarily)... though indeed there's no point of adding new ones, and not having it is probably clear enough  🤷‍♂️ I think I'll just remove it before making a new release (or even in the release PR itself).

Related: do you think it would be worth it to just remove all of them at some point (which would warrant a breaking release)?

AliSoftware: Yeah we can probably make a separate pass to remove all of them before the next major release instead.

PR #630, opened 2025-02-11
Refactor `prototype_build_details_comment` for Firebase App Distribution

AliSoftware:
## What does it do?

Given App Center's imminent deprecation and shutdown, we are migrating our Prototype Builds from App Center to Firebase App Distribution.

This PR:
 - Refactors the `prototype_build_details_comment` action to rely on the data exposed from `firebase_app_distribution` action instead of the ones from the `appcenter_upload` upload
 - Simplifies the action's implementation
 - Updates the unit tests (with the help of Cursor + Claude)

## Internal References

 - 175-gh-Automattic/apps-infra-plans
 - 25739-gh-bloom/DayOne-Apple
 - pdnsEh-1zD-p2

## Checklist before requesting a review

 - [x] 🚧 Proofread the unit tests updated by Claude 
 - [x] 🚧 Potentially continue to work with the AI to ask for more improvements on the implementation, if it has any
 - [x] 🚧 Test the new implementation on DayOne-iOS POC PR (25739-gh-bloom/DayOne-Apple)
 - [x] 🚧 Validate the new implementation can still cover cases of direct download URLs instead of FAD (e.g. Android project with apk uploaded to S3/CloudFront)
- [x] Run `bundle exec rubocop` to test for code style violations and recommendations.
- [x] Run `bundle exec rspec` to run the whole test suite and ensure all your tests pass.
- [x] Make sure you added an entry in [the `CHANGELOG.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md#trunk) to describe your changes under the appropriate existing `###` subsection of the existing `## Trunk` section.
- [x] If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to describe how the changes will affect the migration from the previous major version and what the clients will need to change and consider.

PR #632, opened 2025-02-19
Release `13.0.0` into trunk

iangmaia:
Releasing new version `13.0.0`.

# What's Next

```
### Breaking Changes

- The `prototype_build_details_comment` action now relies on the `lane_context` of `firebase_app_distribution` instead of the one from `appcenter_upload` for the implicit metadata and download URL to use in the comment.
  This is in the context of us moving from App Center (which will be shut down in March 2025) to Firebase App Distribution as our primary mechanism to distribute prototype builds. [#630]

### New Features

- The `ios_lint_localizations` action now accepts a `fail_on_strings_not_in_base_language` parameter (defaulting to `true`) to control whether to report violations when finding strings in translations that are not present in the base language. [#631]

### Internal Changes

- The library now uses immutable literals, via `# frozen_strings_literal: true`. This may result in runtime issues that we will address ASAP once discovered. [#626]


```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,16 +6,30 @@
 
 ### Breaking Changes
 
+_None_
+
+### New Features
+
+_None_
+
+### Bug Fixes
+
+_None_
+
+### Internal Changes
+
+_None_
+
+## 13.0.0
+
+### Breaking Changes
+
 - The `prototype_build_details_comment` action now relies on the `lane_context` of `firebase_app_distribution` instead of the one from `appcenter_upload` for the implicit metadata and download URL to use in the comment.
   This is in the context of us moving from App Center (which will be shut down in March 2025) to Firebase App Distribution as our primary mechanism to distribute prototype builds. [#630]
 
 ### New Features
 
 - The `ios_lint_localizations` action now accepts a `fail_on_strings_not_in_base_language` parameter (defaulting to `true`) to control whether to report violations when finding strings in translations that are not present in the base language. [#631]
-
-### Bug Fixes
-
-_None_
 
 ### Internal Changes
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    fastlane-plugin-wpmreleasetoolkit (12.4.0)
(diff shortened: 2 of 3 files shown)


AliSoftware:
So we're going with a major version that includes my `prototype_build_details_comment` after all, even if that means we'll have to wait until the FAD migration is ready to be adopted for repos (especially WCiOS) to also be able to benefit from `ios_lint_localizations` flexibility to not-crash on unknown translations?

As opposed to reverting my `prototype_build_details_comment` PR, make a `12.5.0` minor version, then restoring my `prototype_build_details_comment` PR to be ready for next major?

I'm ok with this if that was the intent and we consider that this adoption of `ios_lint_localizations` can wait before being adopted in WCiOS; but I just wanted to check what the plan was (especially regarding pdnsEh-210-p2)

iangmaia:
> So we're going with a major version that includes my `prototype_build_details_comment` after all, even if that means we'll have to wait until the FAD migration is ready to be adopted for repos (especially WCiOS) to also be able to benefit from `ios_lint_localizations` flexibility to not-crash on unknown translations?
> 
> As opposed to reverting my `prototype_build_details_comment` PR, make a `12.5.0` minor version, then restoring my `prototype_build_details_comment` PR to be ready for next major?
> 
> I'm ok with this if that was the intent and we consider that this adoption of `ios_lint_localizations` can wait before being adopted in WCiOS; but I just wanted to check what the plan was (especially regarding [pdnsEh-210-p2](https://href.li/?https://wp.me/pdnsEh-210))

Tbh I keep going back and for with this 😓 -- part of me says to just move forward and have WCiOS ready for the FAD migration, which should be soon anyway, and avoid the revert and intermediate releases; part of me wants to be on the more conservative side, play safe and adopt `ios_lint_localizations` sooner -- just by writing this now I'm thinking of following that route instead, so  we can close this issue sooner, which is independent of the FAD, leave the FAD matter to be handled separately.
Then next steps would be to 1) close this PR, 2) create a PR reverting https://github.com/wordpress-mobile/release-toolkit/pull/630 and 3) creating the `12.5.0` release PR.

AliSoftware:
>  just by writing this now I'm thinking of following that route instead, so we can close this issue sooner, which is independent of the FAD, leave the FAD matter to be handled separately.

Let's do this 👍

PR #638, opened 2025-03-11
Introduce monkey-patch to auto-insert collapsible sections in Buildkite logs

AliSoftware:
## What

This PR introduces a monkey-patch of the `Fastlane::Runner#execute_action` method to print a `~~~ :fastlane: [lane :#{lane_name}] #{action}` collapsible group log output for every action called in your `Fastfile`.

> [!NOTE]
> This idea was inspired by https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/fastlane/Fastfile#L156-L179

This will make the Buildkite logs for jobs running fastlane lanes better organized and easier to navigate, as each action will now have its own collapsible section.

## Testing

I have created some dummy branches in `Tumblr-iOS` and `WordPress-iOS`, making their `Gemfile` point to this PR's branch of the `release-toolkit` then triggering a Buildkite build

### Tumblr iOS

Buildkite build: https://buildkite.com/automattic/tumblr-ios/builds/32406

image

### WordPress iOS

Buildkite build: https://buildkite.com/automattic/wordpress-ios/builds/26131

Compare [Buildkite output from _before_ that monkey-patch—but with the ad-hoc patch that has been in WPiOS's Fastfile since a while](https://buildkite.com/automattic/wordpress-ios/builds/26123/summary/annotations?sid=019584a9-01df-4d9a-b5f6-39533bc6e85b#019584a9-0244-49be-91e9-1e5b65c6361f/324)—with [new Buildkite output from the monkey-patch (and with ad-hoc patch in WPiOS's Fastfile removed](https://buildkite.com/automattic/wordpress-ios/builds/26131/canvas?sid=01958585-0113-42a4-a611-20f9eae42474#01958585-01fc-4e33-b4f8-5a1a53fd4120/1261)

image

## What's Next

Once this PR lands, we should remove [the ad-hoc patch code that was implemented in WPiOS's `Fastfile`](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/fastlane/Fastfile#L156-L179), to use this patch from `release-toolkit` instead, to avoid WPiOS having double logging of the collapsible group sections in Buildkite.

PR #648, opened 2025-05-15
Add `buildkite_add_trigger_step` action

AliSoftware:
Closes AINFRA-98

## What does it do?

This implements a new fastlane action that allows to append [a `trigger` step](https://buildkite.com/docs/pipelines/configure/step-types/trigger-step) to the current build so that Buildkite would trigger a new, dependent sub-build of the current build.

<details><summary>How does this differ from other existing <tt>buildkite_*</tt> actions we already have?</summary>

 - This differs from the `buildkite_trigger_build` existing action—which uses the Buildkite REST API to trigger a build from outside a Buildkite build, typically when running a lane from your local terminal
 - This differs from calling `buildkite-agent pipeline upload #{pipeline_file}` to upload the steps of a YML pipeline file directly on the current Build, as instead it actually triggers a child build from the current build. The main different with this is that the triggered child build starts with a brand new build context, especially the `BUILDKITE_BRANCH` that the new build is started on will be the `branch` you asked the trigger to run on (typically the current branch you were checked out at the time you called that action), **even if** the current build started on a different branch and then switched to a new branch since, or pushed new commits since.
</details>

## Why is this needed?

This will be needed especially during the release process of our apps, when we have tasks like the code-freeze that is responsible for creating a new branch + bumping the version, before compiling a beta build from the new branch and commit that just got pushed.

This action will be particularly useful for the case of our repos hosted in private GitHub Enterprise instances (e.g. Tumblr), for which their initial `pipeline upload` step to upload the root YAML pipeline file of a build also takes care of mirroring the latest commit of the `BUILDKITE_BRANCH` that that build starts on to an instance accessible by our other CI agents / builders, so that those CI agents can use that mirror repo as a source to checkout the source code they need to operate. 

<details><summary>Details</summary>

Because of that particular mechanism, if a CI agent that runs in the LAN where the GH:E instance lives (and thus use the real repo) creates a new branch (e.g. `release/*` during code-freeze) and/or pushes a new commit (e.g. version bump), other agents that operate outside the LAN of the GH:E instance and that only use the GitHub.com mirror of the repo won't see the new branch or commits until the mirror is synchronized with the latest GH:E as part of a brand new CI build.

This is why just uploading the steps of the pipeline file directly on the current build won't work for those cases of repos hosted on GH:E; even if the commands that are in charge of building the beta were to `git checkout release/${RELEASE_VERSION}` and `git pull` (like we do for other repos hosted in GitHub.com), since the new branch and commit are only on the GH:E repo but not on the mirror at that time, that wouldn't help. Only triggering a **separate** CI build (as a child build of the current one) would allow our architecture to re-sync the git mirrors and the new jobs running on other agents to see the new branch and commit.
</details>

## Testing

[I created a `use-trigger-action` branch on Tumblr-iOS](https://github.tumblr.net/Tumblr/ios/pull/29788) which:
 - Switched the `Gemfile` to point `release-toolkit` to this PR's branch
 - Update the Fastfile's `trigger_buildkite_build` helper to call this new `buildkite_add_trigger_step` action instead of invoking the `trigger-build.sh` custom script
 - Create a dummy `fake_code_freeze` lane that:
    - Creates a `test/AINFRA-98` branch from the current branch
    - Simulates a dummy version bump and push it
    - Then calls `trigger_buildkite_build` helper method (and this this action) to trigger the `fake-beta-build.yml` pipeline
    - Which in turns just prints some git debug information instead of doing an actual build

Then I used the "New Build" button on Buildkite to trigger a build on that `use-trigger-action` branch using `PIPELINE=release-pipelines/fake-code-freeze.yml`
 - [This is the resulting parent test build](https://buildkite.com/automattic/tumblr-ios/builds/33589/steps) and [its call to the action](https://buildkite.com/automattic/tumblr-ios/builds/33589/steps?sid=0196d9fc-690f-4fa7-8aa1-d0409e44d318#0196d9fc-6947-497b-a9fc-9b02fb3eff8a/221-223)
 - [This is the child build that got triggered, with its git debugging information](https://buildkite.com/automattic/tumblr-ios/builds/33590/steps?sid=0196d9fd-3146-44a5-91b8-d488a6b5d847#0196d9fd-3183-43aa-9578-46a515a4a2f7/347-349)

Notice how:
 - The "Trigger fake-beta-build on test/AINFRA-98" properly shows in the Canvas view as dependent on the job that created it, making it less confusing that the current setup
    
 - The child build that got triggered [ran on the expected commit](https://buildkite.com/automattic/tumblr-ios/builds/33590/steps?sid=0196d9fd-3146-44a5-91b8-d488a6b5d847#0196d9fd-3183-43aa-9578-46a515a4a2f7/347-349), being able to find on the GitHub.com mirror [the commit that was pushed to GH:E by the parent build](https://buildkite.com/automattic/tumblr-ios/builds/33589/steps?sid=0196d9fc-690f-4fa7-8aa1-d0409e44d318#0196d9fc-6947-497b-a9fc-9b02fb3eff8a/195-200)

- [x] ~~If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to describe how the changes will affect the migration from the previous major version and what the clients will need to change and consider.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- Introduce `buildkite_add_trigger_step` action, to insert a `trigger:` step in the current pipeline to trigger a child build from the current build. [#648]
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
@@ -38,10 +38,10 @@ module Fastlane
         }.to_yaml
 
         # Use buildkite-agent to upload the pipeline
-        _stdout, stderr, _status = Open3.capture3('buildkite-agent', 'pipeline', 'upload', stdin_data: trigger_yaml)
+        _stdout, stderr, status = Open3.capture3('buildkite-agent', 'pipeline', 'upload', stdin_data: trigger_yaml)
 
         # Check for errors
-        UI.user_error!("Failed to upload pipeline: #{stderr}") unless stderr.empty?
+        UI.user_error!("Failed to upload pipeline: #{stderr}") unless status.success?
 
         # Log success
         UI.success("Added a trigger step to the current Buildkite build to start a new build for #{pipeline_file} on branch #{branch}")
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
@@ -16,7 +16,7 @@ module Fastlane
         pipeline_file = params[:pipeline_file]
         build_name = File.basename(pipeline_file, '.yml')
         message = params[:message] || build_name
-        branch = params[:branch] || `git rev-parse --abbrev-ref HEAD`.strip
+        branch = params[:branch] || sh('git', 'rev-parse', '--abbrev-ref', 'HEAD').strip
         environment = params[:environment] || {}
         buildkite_pipeline_slug = params[:buildkite_pipeline_slug]
         async = params[:async]
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
@@ -5,7 +5,13 @@ require 'open3'
 module Fastlane
   module Actions
     class BuildkiteAddTriggerStepAction < Action
+      BUILDKITE_ENV_ERROR_MESSAGE = 'This action can only be run from within a Buildkite build'
+
       def self.run(params)
+        unless ENV.key?('BUILDKITE_JOB_ID')
+          UI.user_error!(BUILDKITE_ENV_ERROR_MESSAGE)
+        end
+
         # Extract parameters
         pipeline_file = params[:pipeline_file]
         build_name = File.basename(pipeline_file, '.yml')
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
@@ -1,0 +1,155 @@
+# frozen_string_literal: true
+
+require 'open3'
+
+module Fastlane
+  module Actions
+    class BuildkiteAddTriggerStepAction < Action
+      def self.run(params)
+        # Extract parameters
+        pipeline_file = params[:pipeline_file]
+        build_name = File.basename(pipeline_file, '.yml')
+        message = params[:message] || build_name
+        branch = params[:branch] || `git rev-parse --abbrev-ref HEAD`.strip
+        environment = params[:environment] || {}
+        buildkite_pipeline_slug = params[:buildkite_pipeline_slug]
+        async = params[:async]
+        label = params[:label] || ":buildkite: Trigger #{build_name} on #{branch}"
+
+        # Add the PIPELINE environment variable to the environment hash
+        environment = environment.merge('PIPELINE' => pipeline_file)
+
+        # Create the trigger step YAML
+        trigger_yaml = {
+          'steps' => [
+            {
+              'trigger' => buildkite_pipeline_slug,
+              'label' => label,
+              'async' => async,
+              'build' => {
+                'branch' => branch,
+                'message' => message,
+                'env' => environment
+              }
+            },
+          ]
+        }.to_yaml
+
+        # Use buildkite-agent to upload the pipeline
+        _stdout, stderr, _status = Open3.capture3('buildkite-agent', 'pipeline', 'upload', stdin_data: trigger_yaml)
+
+        # Check for errors
+        UI.user_error!("Failed to upload pipeline: #{stderr}") unless stderr.empty?
+
+        # Log success
+        UI.success("Added a trigger step to the current Buildkite build to start a new build for #{pipeline_file} on branch #{branch}")
+      end
+
+      def self.description
+        'Adds a trigger step to the current Buildkite pipeline to start a new build from this one'
+      end
+
+      def self.details
+        <<~DETAILS
+          This action adds a `trigger` step to the current Buildkite build, to start a separate build from the current one.
+
+          This is slightly different to `buildkite-agent pipeline upload`-ing the YAML steps of the build directly to the current build,
+            as this approach ensures the triggered build starts from a clean and independant context.
+            - This is particularly important if the build being triggered rely on the fact that the current build pushed new commits
+              to the Git branch and we want the new build's steps to start from the new commit.
+            - This is also necessary for cases where we run builds on a mirror of the original Git repo (e.g. for pipelines of repos hosted
+              in a private GitHub Enterprise server, mirrored on GitHub.com for CI building purposes) and we want the new build being triggered
+              to initiate a new sync of the Git repo as part of the CI build bootstrap process, to get the latest commits.
+        DETAILS
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :buildkite_pipeline_slug,
+            env_name: 'BUILDKITE_PIPELINE_SLUG',
+            description: 'The slug of the Buildkite pipeline to trigger. Defaults to the same slug as the current pipeline, so usually not necessary to provide explicitly',
+            type: String,
+            optional: false # But most likely to be auto-provided by the ENV var of the current build and thus not needed to be provided explicitly
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :label,
+            description: 'Custom label for the trigger step. If not provided, defaults to ":buildkite: Trigger {`pipeline_file`\'s basename} on {branch}"',
+            type: String,
+            optional: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :pipeline_file,
+            description: 'The path (relative to `.buildkite/`) to the pipeline YAML file to use for the triggered build',
+            type: String,
+            optional: false
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :branch,
+            description: 'The branch to trigger the build on. Defaults to the Git branch currently checked out at the time of running the action (which is not necessarily the same as the `BUILDKITE_BRANCH` the current build initially started on)',
+            type: String,
+            optional: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :message,
+            description: 'The message / title to use for the triggered build. If not provided, defaults to the `pipeline_file`\'s basename',
+            type: String,
+            optional: true
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :environment,
+            description: 'Environment variables to pass to the triggered build (in addition to the PIPELINE={pipeline_file} that will be automatically injected)',
+            type: Hash,
+            default_value: {},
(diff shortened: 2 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

spec/buildkite_add_trigger_step_action_spec.rb
```diff
@@ -0,0 +1,388 @@
...
+        message: message,
+        environment: environment,
+        buildkite_pipeline_slug: buildkite_pipeline_slug,
+        async: async,
+        depends_on: 'step-1'
+      )
+    end
+
+    it 'includes depends_on when provided with multiple values' do
+      multiple_dependencies = ['step-1', 'step-2', 'step-3']
```
dangermattic, line 313: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/e959bd023b6021da9f05606094ba4d24c4737927/spec/buildkite_add_trigger_step_action_spec.rb#L313"></span>Style/WordArray: Use <code>%w</code> or <code>%W</code> for an array of words.</td>
    </tr>
  </tbody>
</table>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
```diff
@@ -0,0 +1,165 @@
...
+
+module Fastlane
+  module Actions
+    class BuildkiteAddTriggerStepAction < Action
+      def self.run(params)
+        # Extract parameters
+        pipeline_file = params[:pipeline_file]
+        build_name = File.basename(pipeline_file, '.yml')
+        message = params[:message] || build_name
+        branch = params[:branch] || `git rev-parse --abbrev-ref HEAD`.strip
```
iangmaia: I love the convenience and conciseness of using the backticks, but part of me says we'd be better off using the [Git gem](https://github.com/ruby-git/ruby-git), specially thinking about error handling 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
```diff
@@ -0,0 +1,165 @@
...
+                'message' => message,
+                'env' => environment
+              },
+              'depends_on' => depends_on
+            }.compact,
+          ]
+        }.to_yaml
+
+        # Use buildkite-agent to upload the pipeline
+        _stdout, stderr, status = Open3.capture3('buildkite-agent', 'pipeline', 'upload', stdin_data: trigger_yaml)
```
iangmaia, line 47: Maybe good to add some check based on `BUILDKITE_` environment variables if we're indeed running on CI? 🤔

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
AliSoftware: To be honest I've been pulling back a bit about using the Git gem in release-toolkit, because since we started using it I've noticed a couple of things in its implementation details that caused issues.

In particular:
 - Using the Git gem to check if a branch exists locally has a side effect of the gem fetching and catching the list of all the branches in the repo, local and remote, before checking if the branch we asked about was in the array of cached branches
 - Which in some repos can take some time, at least more time than just checking if a local branch of a specific given name exists with a simple `git branch --list "#{branch_name}"`
 - And also causes errors in some cases because the implementation of the Git gem to parse the list of branches from `git branch --list --all` to cache the whole list uses some RegEx string parsing, and some of those fail to account for special character in branch names… including spaces.
 - We actually encountered errors not that long ago about some release automation failing just because it happened that one of the remote branches in a repo contained a space, even if that branch had nothing to do at all with what the release process was manipulating.

So after those experiences, performance issues, and bugs, my trust in this Git gem has lowered a bit since we first started considering it 😓

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
AliSoftware: Good idea. Though if we're not on CI, this command will implicitly fail with either an error that the `buildkite-agent` executable cannot be found, or if you happen to have installed it on your machine anyway (or made it an alias to `docker run … buildkite-agent …` like I did on mine for when I want to try out commands locally 😉 ) and it finds it, it will fail anyway with `fatal: missing job parameter. Usually this is set in the environment for a Buildkite job via BUILDKITE_JOB_ID.` which one might consider clear enough of an error already?

But doesn't hurt to catch it ahead of the call ourselves to customize the error message I guess.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
iangmaia: > and some of those fail to account for special character in branch names… including spaces

Indeed, I've seen it happening as well! I totally understand the losing of confidence in the lib.

I think it's really nice that the Git gem abstracts the details of git calls and the arguments... but at the same time we lose control and it might yield commands or have an internal implementation that might have side effects we don't know / cannot control.

One of my main worries is error handling -- I think using backticks will just "swallow" any eventual error? We can always also use `sh` (which is nice when we have arguments) but for simple cases it may be not strictly needed?

I'm fine on deciding to go in either direction as long as we're sure about the trade offs and keep it consistent 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
```diff
@@ -0,0 +1,165 @@
...
+      def self.description
+        'Adds a trigger step to the current Buildkite pipeline to start a new build from this one'
+      end
+
+      def self.details
+        <<~DETAILS
+          This action adds a `trigger` step to the current Buildkite build, to start a separate build from the current one.
+
+          This is slightly different to `buildkite-agent pipeline upload`-ing the YAML steps of the build directly to the current build,
+            as this approach ensures the triggered build starts from a clean and independant context.
```
iangmaia:
```suggestion
            as this approach ensures the triggered build starts from a clean and independent context.
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
```diff
@@ -0,0 +1,165 @@
...
+
+      def self.description
+        'Adds a trigger step to the current Buildkite pipeline to start a new build from this one'
+      end
+
+      def self.details
+        <<~DETAILS
+          This action adds a `trigger` step to the current Buildkite build, to start a separate build from the current one.
+
+          This is slightly different to `buildkite-agent pipeline upload`-ing the YAML steps of the build directly to the current build,
```
iangmaia: Either way sounds about right to me, but I *think* this is the correct form:

```suggestion
          This is slightly different from `buildkite-agent pipeline upload`-ing the YAML steps of the build directly to the current build,
```

spec/buildkite_add_trigger_step_action_spec.rb
```diff
@@ -0,0 +1,406 @@
...
+    step['depends_on'] = depends_on unless depends_on.nil?
+
+    {
+      'steps' => [step]
+    }.to_yaml
+  end
+
+  before do
+    # Mock the git command to return our test branch
+    allow(described_class).to receive(:`).with('git rev-parse --abbrev-ref HEAD').and_return(branch)
```
iangmaia: ...and it's amazing how even backticks are mockable 😅 (I think I probably even wrote tests mocking it before but this kind of stuff about Ruby never ceases to amaze me).

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- Introduce `buildkite_add_trigger_step` action, to insert a `trigger:` step in the current pipeline to trigger a child build from the current build. [#648]
```
iangmaia, line 13: I like the name 👍 I think we considered a potential confusion with `buildkite_trigger_build`, but this is describing exactly what is happening: we're not really just triggering a build, we're adding a trigger step 🙂

AliSoftware: Yep!

I _did_ consider what we talked about in the past wrt to naming this new action `buildkite_trigger_build` while renaming the old one `buildkite_start_new_build` or whatever… with the drawback of such a breaking change requiring a Major version bump and having to update all our client repos…

But then when I came up with that name I figured it was clear enough to avoid the need to do that breaking change renaming of the other action after all, so I'm happy with it.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
AliSoftware: Yeah that's a fair point. And actually my initial implementation tried to use `Open3.capture3` to run that command just like I use for running the `buildkite-agent` command… but then that overcomplicated the code and I figured it wasn't worth it.

I like the idea of using `sh` though. iinm we can even provide options to Fastlane's `sh` helper to tell it not to print the details of such shell command call in the output nor consider that call as being an action when it prints the summary table at the end of a lane… will check if that works.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
AliSoftware: Fixed in eb52f77f

iangmaia (approved):
Thanks for this!
Just a couple of small comments but LGTM 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_add_trigger_step_action.rb
AliSoftware: Switched to using `sh` in 35685537, and tested how it looks in a real world use case via https://github.tumblr.net/Tumblr/ios/pull/29788/commits/147498ea3cd8f74aa3eddd762a1f018f7678d919 and https://buildkite.com/automattic/tumblr-ios/builds/33625/steps?sid=0196ed6b-2fad-4257-80d5-96ed9e40dab2#0196ed6b-2fe4-4e42-baad-c4a9f64dd40c/212-216:

```
[07:19:38 -0400]: ----------------------------------------
[07:19:38 -0400]: --- Step: buildkite_add_trigger_step ---
[07:19:38 -0400]: ----------------------------------------
[07:19:38 -0400]: $ git rev-parse --abbrev-ref HEAD
[07:19:38 -0400]: ▸ test/AINFRA-98
[07:19:39 -0400]: Added a trigger step to the current Buildkite build to start a new build for fake-beta-build.yml on branch test/AINFRA-98


+-----------------------------------------------------+
|                  fastlane summary                   |
+------+--------------------------------+-------------+
| Step | Action                         | Time (in s) |
+------+--------------------------------+-------------+
| 1    | setup_ci                       | 0           |
| 2    | git checkout -b test/AINFRA-98 | 0           |
| 3    | git_commit                     | 0           |
| 4    | push_to_git_remote             | 1           |
| 5    | is_ci                          | 0           |
| 6    | buildkite_add_trigger_step     | 1           |
+------+--------------------------------+-------------+
```
I guess it's ok for the command and its output to be printed in the logs then, since they are hidden under a collapsible group anyway (thanks to #638) and don't appear in the fastlane summary table 👍

PR #653, opened 2025-06-24
Prevent frozen string errors in promo screenshots action

mokagio:
## What does it do?

See https://linear.app/a8c/issue/AINFRA-827. @jaclync shared a crash when running the promo screenshots generation in Woo. 

These changes address it.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
@@ -24,10 +24,12 @@ module Fastlane
     class PromoScreenshots
       def initialize
         if $skip_magick
-          message = "PromoScreenshots feature is currently disabled.\n"
-          message << "Please, install RMagick if you aim to generate the PromoScreenshots.\n"
-          message << "'bundle install --with screenshots' should do it if your project is configured for PromoScreenshots.\n"
-          message << 'Aborting.'
+          message = <<~MSG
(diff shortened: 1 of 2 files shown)


jaclync (approved):
The fix works from testing with https://github.com/woocommerce/woocommerce-ios/pull/15805 🚢

mokagio:
@wordpress-mobile/apps-infrastructure I'd appreciate a review now that @jaclync and I verified these changes are enough to get the promo screenshots generation over the line. 

Thanks!

AliSoftware (approved):

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -400,10 +402,13 @@ def open_image(path)
       end
 
       def create_image(width, height, background = 'transparent')
-        background.paint.to_hex
+        # The paint method we call below modifies the string in place.
+        # But if the string is frozen, we need to dup it first, otherwise we'll get a frozen string error.
+        working_background = background.frozen? ? background.dup : background
```
AliSoftware, line 407: I _think_ (?) the `.dup` method is mostly inconsequential if `background` were to not be `frozen?`, so I think it'd be ok to just use `wordking_background = background.dup` in all cases without the need for testing `frozen?`?

But OTOH that test on `frozen?` makes it extra explicit why we'd calling the `.dup` here, so I like it and don't mind keeping it. Doesn't hurt anyway 🙃

PR #656, opened 2025-08-04
Add custom prefix configuration to `DerivedBuildCodeFormatter`

iangmaia:
## What does it do?

This PR adds to the class `DerivedBuildCodeFormatter` the possibility to configure a versioning prefix instead of always defaulting to `1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-- Add the possibility to configure DerivedBuildCodeFormatter configure a versioning prefix instead of always defaulting to 1. [#656]
+- Add the possibility to configure in DerivedBuildCodeFormatter a versioning prefix instead of always defaulting to 1. [#656]
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- Add the possibility to configure DerivedBuildCodeFormatter configure a versioning prefix instead of always defaulting to 1. [#656]
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -67,16 +67,9 @@ module Fastlane
           end
 
           # Check if it's a valid integer
-          begin
-            prefix_int = Integer(prefix_str)
-          rescue ArgumentError
-            UI.user_error!("Prefix must be an integer digit (0-9) or empty string, got: '#{prefix_str}'")
-          end
+          return if ('0'..'9').include?(prefix_str)
 
-          # Check if it's within valid range (0-9)
-          return if prefix_int.between?(0, 9)
-
-          UI.user_error!("Prefix must be a single digit (0-9), got: #{prefix_int}")
+          UI.user_error!("Prefix must be an integer digit (0-9) or empty string, got: '#{prefix_str}'")
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -6,9 +6,18 @@ module Fastlane
       # The `DerivedBuildCodeFormatter` class is a specialized build code formatter for derived build codes.
       # It takes in an AppVersion object and derives a build code from it.
       class DerivedBuildCodeFormatter
+        # Initialize the formatter with a configurable prefix.
+        #
+        # @param [String] prefix The prefix to use for the build code. Must be a single digit (0-9). Defaults to '1' for backward compatibility.
+        #
+        def initialize(prefix: '1')
+          validate_prefix!(prefix)
+          @prefix = prefix.to_s
+        end
+
         # Calculate the next derived build code.
         #
-        # This method derives a new build code from the given AppVersion object by concatenating the digit 1,
+        # This method derives a new build code from the given AppVersion object by concatenating the configured prefix,
         # the major version, the minor version, the patch version, and the build number.
         #
         # @param [AppVersion] version The AppVersion object to derive the next build code from.
@@ -19,15 +28,55 @@ module Fastlane
         # @return [String] The formatted build code string.
         #
         def build_code(build_code = nil, version:)
-          format(
-            # 1 is appended to the beginning of the string in case there needs to be additional platforms or
-            # extensions that could then use a different digit prefix such as 2, etc.
-            '1%<major>.2i%<minor>.2i%<patch>.2i%<build_number>.2i',
+          result = format(
+            # The prefix is configurable to allow for additional platforms or
+            # extensions that could use a different digit prefix such as 2, etc.
+            '%<prefix>s%<major>.2i%<minor>.2i%<patch>.2i%<build_number>.2i',
+            prefix: @prefix,
             major: version.major,
             minor: version.minor,
             patch: version.patch,
             build_number: version.build_number
           )
+
+          # Only trim leading zeros when prefix is empty (to avoid removing explicit "0" prefix)
+          if @prefix.empty?
+            result.gsub(/^0+/, '')
+          else
+            result
+          end
+        end
+
+        private
+
+        # Validates that the prefix is a valid single digit (0-9) or empty string.
+        #
+        # @param [String] prefix The prefix to validate
+        #
+        # @raise [StandardError] If the prefix is invalid
+        #
+        def validate_prefix!(prefix)
+          prefix_str = prefix.to_s
+
+          # Allow empty string
+          return if prefix_str.empty?
+
+          # Check if it's longer than 1 character
+          if prefix_str.length > 1
+            UI.user_error!("Prefix must be a single digit or empty string, got: '#{prefix_str}' (length: #{prefix_str.length})")
+          end
+
+          # Check if it's a valid integer
+          begin
+            prefix_int = Integer(prefix_str)
+          rescue ArgumentError
+            UI.user_error!("Prefix must be an integer digit (0-9) or empty string, got: '#{prefix_str}'")
+          end
+
+          # Check if it's within valid range (0-9)
+          return if prefix_int.between?(0, 9)
+
+          UI.user_error!("Prefix must be a single digit (0-9), got: #{prefix_int}")
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -39,12 +39,7 @@ module Fastlane
             build_number: version.build_number
           )
 
-          # Only trim leading zeros when prefix is empty (to avoid removing explicit "0" prefix)
-          if @prefix.empty?
-            result.gsub(/^0+/, '')
-          else
-            result
-          end
+          result.gsub(/^0+/, '')
         end
 
         private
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -19,15 +28,55 @@ class DerivedBuildCodeFormatter
...
             patch: version.patch,
             build_number: version.build_number
           )
+
+          # Only trim leading zeros when prefix is empty (to avoid removing explicit "0" prefix)
+          if @prefix.empty?
+            result.gsub(/^0+/, '')
+          else
+            result
+          end
```
AliSoftware: I'm actually not sure how the Google Play API would react to submitting a build with a `versionCode` that has a leading zero 🤔

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -19,15 +28,55 @@ class DerivedBuildCodeFormatter
...
+          begin
+            prefix_int = Integer(prefix_str)
+          rescue ArgumentError
+            UI.user_error!("Prefix must be an integer digit (0-9) or empty string, got: '#{prefix_str}'")
+          end
+
+          # Check if it's within valid range (0-9)
+          return if prefix_int.between?(0, 9)
+
+          UI.user_error!("Prefix must be a single digit (0-9), got: #{prefix_int}")
```
AliSoftware: I'm not sure how it'd be technically possible to get to that point in the code and fail that test, given we already ensured that `prefix_str` would not be a string longer than 1 and would be a valid `Integer` 😅 🙃

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: iinm, it needs to be an integer in the API so I'd guess they would just be ignored? 🤔 But also considering iOS, perhaps it doesn't make a lot of sense to allow for 0 to be used as prefix (or maybe we should treat it the same as `''`).

AliSoftware: Yeah I think doing the `gsub` unconditionally should be fine.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Indeed 😅 Simplified on 49edcb6621f6918a1159a175ad8ad706ae8be90d

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -6,9 +6,18 @@ module Versioning
       # The `DerivedBuildCodeFormatter` class is a specialized build code formatter for derived build codes.
       # It takes in an AppVersion object and derives a build code from it.
       class DerivedBuildCodeFormatter
+        # Initialize the formatter with a configurable prefix.
+        #
+        # @param [String] prefix The prefix to use for the build code. Must be a single digit (0-9). Defaults to '1' for backward compatibility.
```
AliSoftware:
```suggestion
        # @param [String] prefix The prefix to use for the build code. Must be a single digit (0-9), or the empty string. Defaults to '1' for backward compatibility.
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -6,9 +6,18 @@ module Versioning
...
+        #
+        def initialize(prefix: '1')
+          validate_prefix!(prefix)
+          @prefix = prefix.to_s
```
AliSoftware, line 16: 💡  Could be nice to allow `nil` as an input to the constructor as well, for nicer API 🤔 
```suggestion
          prefix ||= ''
          validate_prefix!(prefix)
          @prefix = prefix.to_s
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Implemented on 75f41271f15fb3c79b92210d7f32f71102b6ef3c.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -19,15 +28,43 @@ class DerivedBuildCodeFormatter
         # @return [String] The formatted build code string.
         #
         def build_code(build_code = nil, version:)
```
AliSoftware, line 31: Probably not for this PR, but mentioning while I saw it:

Note that this `build_code` parameter is unused, but was initially still kept here so that the common API for all the `*BuildCodeFormatter` have the same API (as mentioned in the YARD comment about it), making it easier to switch from one formatter to another without changing the call sites (abstraction layer pattern).

But in practice I just realized that:
1. We don't have an abstract class for all the `*BuildCodeFormatter` classes like we have for `AbstractVersionFormatter`. In fact, the `*BuildCodeFormatter` classes don't inherit any common parent class, unlike the `*VersionFormatter` classes
2. In practice the API for all `*BuildCodeFormatter` classes is not even consistent (which might actually explain point 1). For example this `DerivedBuildCodeFormatter` as well as `FourPartBuildCodeFormatter` indeed both have a signature of `def build_code(build_code = nil, version:)`… but the `SimpleBuildCodeFormatter` uses `def build_code(version = nil, build_code:)` instead…

That makes me think that (1) either one day we should create an abstract class for the `*BuildCodeFormatter` and unify their API for consistency (2) or we should just abandon the idea of having unified API for all (given the needed input parameters might depend on what the formatter does and needs to compute its value), and remove the unused parameters from the `def build_code(…)` signatures for clarity… 🙃 

---

Also, in any case, I am surprised that `rubocop` didn't complain about it and didn't suggest us to rename this parameter with an underscore to indicate that it being unused is intentional (`def build_code(_build_code = nil, version:)` 🤷

spec/derived_build_code_formatter_spec.rb
```diff
@@ -4,16 +4,102 @@
...
+      end
 
-    it 'derives the build code from version numbers that are two digits' do
-      version = Fastlane::Models::AppVersion.new(12, 34, 56, 78)
-      build_code_string = described_class.new.build_code(version: version)
-      expect(build_code_string.to_s).to eq('112345678')
+      it 'rejects symbols and special characters' do
+        expect { described_class.new(prefix: '@') }.to raise_error(/Prefix must be an integer digit/)
+        expect { described_class.new(prefix: '#') }.to raise_error(/Prefix must be an integer digit/)
+      end
```
AliSoftware, line 118: 👌 

(this reminded me of [this meme](https://i.programmerhumor.io/2024/01/programmerhumor-io-testing-memes-programming-memes-d8b14923a839043-758x905.png) 😄 )

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -6,9 +6,19 @@ module Versioning
       # The `DerivedBuildCodeFormatter` class is a specialized build code formatter for derived build codes.
       # It takes in an AppVersion object and derives a build code from it.
       class DerivedBuildCodeFormatter
+        # Initialize the formatter with a configurable prefix.
+        #
+        # @param [String] prefix The prefix to use for the build code. Must be a single digit (0-9), or the empty string. Defaults to '1' for backward compatibility.
+        #
+        def initialize(prefix: '1')
+          prefix ||= ''
```
AliSoftware, line 14: Would be nice to add an extra unit test for the `nil` case 🙃

AliSoftware (approved):
Suggesting one last unit test but otherwise LGTM so approving to unblock.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- Add the possibility to configure DerivedBuildCodeFormatter configure a versioning prefix instead of always defaulting to 1. [#656]
```
iangmaia:
```suggestion
- Add the possibility to configure in DerivedBuildCodeFormatter a versioning prefix instead of always defaulting to 1. [#656]
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: > That makes me thing that (1) either one day we should create an abstract class for the *BuildCodeFormatter and unify their API for consistency (2) or we should just abandon the idea of having unified API for all

Oh, agreed.
I quickly saw the parameter and imagined there would be a base, abstract class like the other versioning formatters.

> I am surprised that rubocop didn't complain about it and didn't suggest us to rename this parameter 

True 🤔

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -6,9 +6,19 @@ module Versioning
       # The `DerivedBuildCodeFormatter` class is a specialized build code formatter for derived build codes.
       # It takes in an AppVersion object and derives a build code from it.
       class DerivedBuildCodeFormatter
+        # Initialize the formatter with a configurable prefix.
+        #
+        # @param [String] prefix The prefix to use for the build code. Must be a single digit (0-9), or the empty string. Defaults to '1' for backward compatibility.
```
AliSoftware: I just realized that funnily enough, I'm not even sure any of our projects use this `DerivedBuildCodeFormatter` yet 🤔 

 - Tumblr uses derived `versionCode` based on `versionName`… but they have their own in-repo implementation (and we've been wanting to migrate to the `release-toolkit` one instead for a while for consistency, but never found the time—see https://linear.app/a8c/issue/AINFRA-155/fix-tumblr-versioning-consistency)
 - I thought for a moment that PocketCasts used that as well because I remember them using some prefix in their `versionCode` to differentiate the Android app, Wear app and Automotive app… but looking at it that's still a `SimpleBuildCodeFormatter` incrementing the versions, and [what I remembered was just an offset between the 3 form factors](https://github.com/Automattic/pocket-casts-android/blob/main/fastlane/Fastfile#L825-L834), not a prefix

… So in practice Gravatar-Android would be the first one to use that class iinm 🙃 

So I wonder if it's really needed to default to `'1'` "for backwards compatibility" given this…
On one hand, better safe that sorry and we can always pass a different value at call site in the product's `Fastfile`. Oh the other hand, if we wanted to change that default prefix, it might actually be now or never 😅

AliSoftware (approved):

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Yeah, that's true. At some point I also realized that Tumblr used their own implementation but wasn't sure about other apps... and anyway this would be a breaking change of the API technically.

But indeed, in practice it would be nicer to not have this default and avoid this issue in the future (especially if more apps adopt this `DerivedBuildCodeFormatter`). Shall we do it? I think I'll do it :feelsgood:

iangmaia: Done in bae9e25c091e5a7904e246faa5bbff4ddee29980

AliSoftware (approved):

PR #657, opened 2025-08-04
Add version components digits configuration to DerivedBuildCodeFormatter

iangmaia:
## What does it do?

This PR adds to the `DerivedBuildCodeFormatter` class the possibility to configure a variable number of digits per version components, so that we have more flexibility to increase the maximum number of builds per a specific major / minor version, for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ _None_
 
 ### New Features
 
-- Add the possibility to configure in DerivedBuildCodeFormatter a versioning prefix instead of always defaulting to 1. [#656]
-- Add configurable number of digits in version components in DerivedBuildCodeFormatter. [#657]
+- Add the possibility to configure in `DerivedBuildCodeFormatter` a versioning prefix instead of always defaulting to 1. [#656]
+- Add configurable number of digits in version components in `DerivedBuildCodeFormatter`. [#657]
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ _None_
 ### New Features
 
 - Add the possibility to configure in DerivedBuildCodeFormatter a versioning prefix instead of always defaulting to 1. [#656]
+- Add configurable number of digits in version components in DerivedBuildCodeFormatter. [#657]
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -45,6 +45,12 @@ module Fastlane
         # @return [String] The formatted build code string.
         #
         def build_code(build_code = nil, version:)
+          # Validate that version components fit within their configured digit limits
+          validate_version_component_fits!(component_value: version.major, digit_limit: @major_digits)
+          validate_version_component_fits!(component_value: version.minor, digit_limit: @minor_digits)
+          validate_version_component_fits!(component_value: version.patch, digit_limit: @patch_digits)
+          validate_version_component_fits!(component_value: version.build_number, digit_limit: @build_digits)
+
           result = [
             @prefix,
             version.major.to_s.rjust(@major_digits, '0'),
@@ -108,6 +114,23 @@ module Fastlane
 
           UI.user_error!("Total digit count (#{total_digits}) exceeds maximum allowed (#{MAX_TOTAL_DIGITS}). " \
                          "Current config: major(#{major_digits}) + minor(#{minor_digits}) + patch(#{patch_digits}) + build(#{build_digits}) digits")
+        end
+
+        # Validates that a version component value fits within its configured digit limit.
+        #
+        # @param [Integer] component_value The version component value to validate
+        # @param [Integer] digit_limit The maximum number of digits allowed for this component
+        #
+        # @raise [StandardError] If the component value exceeds the digit limit
+        #
+        def validate_version_component_fits!(component_value:, digit_limit:)
+          # Calculate the maximum value that fits in the digit limit
+          max_value = (10**digit_limit) - 1
+
+          return if component_value <= max_value
+
+          UI.user_error!("Version component value (#{component_value}) exceeds maximum allowed " \
+                         "for #{digit_limit} digit(s) (max: #{max_value}). Consider increasing the corresponding _digits parameter.")
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -45,17 +45,14 @@ module Fastlane
         # @return [String] The formatted build code string.
         #
         def build_code(build_code = nil, version:)
-          # Build dynamic format string based on configured digit counts
-          format_string = "%<prefix>s%<major>.#{@major_digits}i%<minor>.#{@minor_digits}i%<patch>.#{@patch_digits}i%<build_number>.#{@build_digits}i"
-
-          result = format(
-            format_string,
-            prefix: @prefix,
-            major: version.major,
-            minor: version.minor,
-            patch: version.patch,
-            build_number: version.build_number
-          )
+          # Use manual padding to avoid security risks with dynamic format strings
+          result = [
+            @prefix,
+            version.major.to_s.rjust(@major_digits, '0'),
+            version.minor.to_s.rjust(@minor_digits, '0'),
+            version.patch.to_s.rjust(@patch_digits, '0'),
+            version.build_number.to_s.rjust(@build_digits, '0'),
+          ].join
 
           result.gsub(/^0+/, '')
         end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -19,9 +19,8 @@ module Fastlane
         # @param [Integer] build_digits Number of digits for build number. Must be between 1–3. Defaults to 2.
         #
         def initialize(prefix: nil, major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
-          prefix ||= ''
           validate_prefix!(prefix)
-          @prefix = prefix.to_s
+          @prefix = (prefix || '').to_s
 
           @digit_counts = [major_digits, minor_digits, patch_digits, build_digits]
           @digit_counts.each { |d| validate_digit_count!(d) }
@@ -61,6 +60,10 @@ module Fastlane
         # @raise [StandardError] If the prefix is invalid
         #
         def validate_prefix!(prefix)
+          unless prefix.nil? || prefix.is_a?(String) || prefix.is_a?(Integer)
+            UI.user_error!("Prefix must be a string or integer, got: #{prefix.class}")
+          end
+
           prefix_str = prefix.to_s
 
           # Allow empty string
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -45,7 +45,6 @@ module Fastlane
         # @return [String] The formatted build code string.
         #
         def build_code(build_code = nil, version:)
-          # Use manual padding to avoid security risks with dynamic format strings
           result = [
             @prefix,
             version.major.to_s.rjust(@major_digits, '0'),
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -11,10 +11,10 @@ module Fastlane
         # Initialize the formatter with configurable prefix and digit counts.
         #
         # @param [String] prefix The prefix to use for the build code. Must be a single digit (0-9), or empty string / nil.
-        # @param [Integer] major_digits Number of digits for major version. Defaults to 2.
-        # @param [Integer] minor_digits Number of digits for minor version. Defaults to 2.
-        # @param [Integer] patch_digits Number of digits for patch version. Defaults to 2.
-        # @param [Integer] build_digits Number of digits for build number. Defaults to 2.
+        # @param [Integer] major_digits Number of digits for major version. Must be between 1–3. Defaults to 2.
+        # @param [Integer] minor_digits Number of digits for minor version. Must be between 1–3. Defaults to 2.
+        # @param [Integer] patch_digits Number of digits for patch version. Must be between 1–3. Defaults to 2.
+        # @param [Integer] build_digits Number of digits for build number. Must be between 1–3. Defaults to 2.
         #
         def initialize(prefix: nil, major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
           prefix ||= ''
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -20,7 +20,7 @@ module Fastlane
         #
         def initialize(prefix: nil, major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
           validate_prefix!(prefix)
-          @prefix = (prefix || '').to_s
+          @prefix = prefix.to_s
 
           @digit_counts = [major_digits, minor_digits, patch_digits, build_digits]
           @digit_counts.each { |d| validate_digit_count!(d) }
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -19,17 +19,12 @@ module Fastlane
         def initialize(prefix: nil, major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
           prefix ||= ''
           validate_prefix!(prefix)
-          validate_digit_count!(major_digits)
-          validate_digit_count!(minor_digits)
-          validate_digit_count!(patch_digits)
-          validate_digit_count!(build_digits)
-          validate_total_digits!(major_digits, minor_digits, patch_digits, build_digits)
+
+          @digit_counts = [major_digits, minor_digits, patch_digits, build_digits]
+          @digit_counts.each { |d| validate_digit_count!(d) }
+          validate_total_digits!(@digit_counts)
 
           @prefix = prefix.to_s
-          @major_digits = major_digits
-          @minor_digits = minor_digits
-          @patch_digits = patch_digits
-          @build_digits = build_digits
         end
 
         # Calculate the next derived build code.
@@ -44,22 +39,16 @@ module Fastlane
         #
         # @return [String] The formatted build code string.
         #
-        def build_code(build_code = nil, version:)
-          # Validate that version components fit within their configured digit limits
-          validate_version_component_fits!(component_value: version.major, digit_limit: @major_digits)
-          validate_version_component_fits!(component_value: version.minor, digit_limit: @minor_digits)
-          validate_version_component_fits!(component_value: version.patch, digit_limit: @patch_digits)
-          validate_version_component_fits!(component_value: version.build_number, digit_limit: @build_digits)
-
-          result = [
-            @prefix,
-            version.major.to_s.rjust(@major_digits, '0'),
-            version.minor.to_s.rjust(@minor_digits, '0'),
-            version.patch.to_s.rjust(@patch_digits, '0'),
-            version.build_number.to_s.rjust(@build_digits, '0'),
-          ].join
-
-          result.gsub(/^0+/, '')
+        def build_code(_build_code = nil, version:)
+          formatted_components = version.components.zip(@digit_counts).map do |value, width|
+            comp = value.to_s.rjust(width, '0')
+            if comp.length > width
+              UI.user_error!("Version component value (#{value}) exceeds maximum allowed width of #{width} characters. " \
+                             "Consider increasing the corresponding `*_digits` parameter of your `#{self.class.name}`")
+            end
+            comp
+          end
+          [@prefix, *formatted_components].join.gsub(/^0+/, '')
         end
 
         private
@@ -106,31 +95,14 @@ module Fastlane
 
         # Validates that the total number of digits (excluding prefix) doesn't exceed the maximum for multiplatform compatibility.
         #
-        def validate_total_digits!(major_digits, minor_digits, patch_digits, build_digits)
-          total_digits = major_digits + minor_digits + patch_digits + build_digits
+        def validate_total_digits!(digits_list)
+          total_digits = digits_list.sum
 
           # Limit total digits to 8 (excluding prefix)
           return if total_digits <= MAX_TOTAL_DIGITS
 
           UI.user_error!("Total digit count (#{total_digits}) exceeds maximum allowed (#{MAX_TOTAL_DIGITS}). " \
-                         "Current config: major(#{major_digits}) + minor(#{minor_digits}) + patch(#{patch_digits}) + build(#{build_digits}) digits")
-        end
-
-        # Validates that a version component value fits within its configured digit limit.
-        #
-        # @param [Integer] component_value The version component value to validate
-        # @param [Integer] digit_limit The maximum number of digits allowed for this component
-        #
-        # @raise [StandardError] If the component value exceeds the digit limit
-        #
-        def validate_version_component_fits!(component_value:, digit_limit:)
-          # Calculate the maximum value that fits in the digit limit
-          max_value = (10**digit_limit) - 1
-
-          return if component_value <= max_value
-
-          UI.user_error!("Version component value (#{component_value}) exceeds maximum allowed " \
-                         "for #{digit_limit} digit(s) (max: #{max_value}). Consider increasing the corresponding _digits parameter.")
+                         "Current config: major(#{digits_list[0]}) + minor(#{digits_list[1]}) + patch(#{digits_list[2]}) + build(#{digits_list[3]}) digits")
         end
       end
     end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -4,6 +4,8 @@ module Fastlane
   module Wpmreleasetoolkit
     module Versioning
       MAX_TOTAL_DIGITS = 8
+      MIN_DIGIT_COUNT = 1
+      MAX_DIGIT_COUNT = 3
 
       # The `DerivedBuildCodeFormatter` class is a specialized build code formatter for derived build codes.
       # It takes in an AppVersion object and derives a build code from it.
@@ -87,9 +89,9 @@ module Fastlane
             UI.user_error!("Digit count must be an integer, got: #{digit_count.class}")
           end
 
-          return if digit_count.between?(1, 3)
+          return if digit_count.between?(MIN_DIGIT_COUNT, MAX_DIGIT_COUNT)
 
-          UI.user_error!("Digit count must be between 1 and 3 digits, got: #{digit_count}")
+          UI.user_error!("Digit count must be between #{MIN_DIGIT_COUNT} and #{MAX_DIGIT_COUNT} digits, got: #{digit_count}")
         end
 
         # Validates that the total number of digits (excluding prefix) doesn't exceed the maximum for multiplatform compatibility.
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -19,12 +19,11 @@ module Fastlane
         def initialize(prefix: nil, major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
           prefix ||= ''
           validate_prefix!(prefix)
+          @prefix = prefix.to_s
 
           @digit_counts = [major_digits, minor_digits, patch_digits, build_digits]
           @digit_counts.each { |d| validate_digit_count!(d) }
           validate_total_digits!(@digit_counts)
-
-          @prefix = prefix.to_s
         end
 
         # Calculate the next derived build code.
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -3,7 +3,7 @@
 module Fastlane
   module Wpmreleasetoolkit
     module Versioning
-      MAX_TOTAL_DIGITS = 9
+      MAX_TOTAL_DIGITS = 8
 
       # The `DerivedBuildCodeFormatter` class is a specialized build code formatter for derived build codes.
       # It takes in an AppVersion object and derives a build code from it.
@@ -107,7 +107,7 @@ module Fastlane
         def validate_total_digits!(major_digits, minor_digits, patch_digits, build_digits)
           total_digits = major_digits + minor_digits + patch_digits + build_digits
 
-          # Limit total digits to 9 (excluding prefix)
+          # Limit total digits to 8 (excluding prefix)
           return if total_digits <= MAX_TOTAL_DIGITS
 
           UI.user_error!("Total digit count (#{total_digits}) exceeds maximum allowed (#{MAX_TOTAL_DIGITS}). " \
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -3,6 +3,7 @@
 module Fastlane
   module Wpmreleasetoolkit
     module Versioning
+       # Max total for `*_digits` params, not counting prefix
       MAX_TOTAL_DIGITS = 8
       MIN_DIGIT_COUNT = 1
       MAX_DIGIT_COUNT = 3
@@ -99,6 +100,8 @@ module Fastlane
 
         # Validates that the total number of digits (excluding prefix) doesn't exceed the maximum for multiplatform compatibility.
         #
+        # Since Google Play's max versionCode is ≈ 2_000_000_000, we want to avoid being too close to the limit
+        # as this would then block us from submitting any updates for that app if we reached it.
         def validate_total_digits!(digits_list)
           total_digits = digits_list.sum
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
@@ -3,7 +3,7 @@
 module Fastlane
   module Wpmreleasetoolkit
     module Versioning
-       # Max total for `*_digits` params, not counting prefix
+      # Max total for `*_digits` params, not counting prefix
       MAX_TOTAL_DIGITS = 8
       MIN_DIGIT_COUNT = 1
       MAX_DIGIT_COUNT = 3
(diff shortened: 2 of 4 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -29,10 +45,11 @@ def initialize(prefix: '1')
         # @return [String] The formatted build code string.
         #
         def build_code(build_code = nil, version:)
+          # Build dynamic format string based on configured digit counts
+          format_string = "%<prefix>s%<major>.#{@major_digits}i%<minor>.#{@minor_digits}i%<patch>.#{@patch_digits}i%<build_number>.#{@build_digits}i"
```
AliSoftware: I'm surprised that `rubocop` allows this. Generally generating a dynamic `format` is a red flag for security (think injection, non-validation of parameters, etc).
(For the parallel, that's similar to Swiftlint warning your about `NSString(format: …)` with a `format` that is not a static string)

Obviously in your case you validated the various `@…_digits` variables using `validate_digit_count!` in the constructor, so technically we should hopefully be safe against such security risk and attack. But still:
 - I'd expect `rubocop` to complain about it
 - And suggest you to use [the `*` field width flag](https://docs.ruby-lang.org/en/master/format_specifications_rdoc.html#label-27-2A-27+Flag) to pass the width of the field as a separate parameter to `format` instead of interpolating it into the format string.

Given we've been using [`%<foo>` references by name](https://docs.ruby-lang.org/en/master/format_specifications_rdoc.html#label-Reference+by+Name) so far, I don't think that's compatible with the use of `*` for parametrized field-width though. And at that point I find this format a bit hard to read already (so I can't even imagine if we had to switch to positional parameters instead of named ones…)

So maybe it'd be more readable at that stage to not use `format` but just padding + concatenate the strings ourselves instead? i.e.:

```ruby
[
    prefix,
    version.major.to_s.rjust(major_digits, '0'),
    version.minor.to_s.rjust(minor_digits, '0'),
    version.patch.to_s.rjust(patch_digits, '0'),
    version.build_number.to_s.rjust(build_digits, '0')
].join
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -67,6 +84,35 @@ def validate_prefix!(prefix)
...
+          UI.user_error!("Digit count must be between 1 and 3 digits, got: #{digit_count}")
+        end
+
+        # Validates that the total number of digits (excluding prefix) doesn't exceed the maximum for multiplatform compatibility.
+        #
+        def validate_total_digits!(major_digits, minor_digits, patch_digits, build_digits)
+          total_digits = major_digits + minor_digits + patch_digits + build_digits
+
+          # Limit total digits to 9 (excluding prefix)
+          return if total_digits <= MAX_TOTAL_DIGITS
```
AliSoftware: The max value for `versionCode` is ≈ `2_000_000_000`, right? Which is 10 digits with the prefix.
A max number of digits of 9 seems wayyy too close to that.
That means that if we try to use a `prefix: 3` for example, while `total_digits` is `9`, we'll exceed the max allowed by Google Play by having a `versionCode` like `3_XXX_XXX_XXX` > `2_100_000_000`

So I think we sould only allow a `MAX_TOTAL_DIGITS` of `8` instead. Otherwise we might start with a product that only has an app variant and will fit at first… but then when we'll want to add more form factors to it later in the product's life and might need more prefixes (for Android Wear, Auto, Glasses, …), we'll run out of `versionCode` and won't be able to fit them already.

iangmaia: Yeah, I was pondering about that, wondering if we should accommodate for the case when an app would specifically want to use that limit for some reason... but in practice, that's indeed unlikely and we'll more likely be bound to these restrictions such as Google's `2_100_000_000`, app variant prefixes, etc. So I agree, the 8 digits makes sense 👍 

Fixed in 6e4033bc0fac082506afb23d8a99287552c0289e

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: > I'm surprised that rubocop allows this. Generally generating a dynamic format is a red flag for security (think injection, non-validation of parameters, etc).

Oof, good point.

> it'd be more readable at that stage to not use format but just padding + concatenate the strings ourselves instead

I like that, thanks for the suggestion 👍 Implemented on bf7d3456182fe3e47eae9bc14959dc081fa8d2a4.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -29,16 +45,14 @@ def initialize(prefix: nil)
...
-            # The prefix is configurable to allow for additional platforms or
-            # extensions that could use a different digit prefix such as 2, etc.
-            '%<prefix>s%<major>.2i%<minor>.2i%<patch>.2i%<build_number>.2i',
-            prefix: @prefix,
-            major: version.major,
-            minor: version.minor,
-            patch: version.patch,
-            build_number: version.build_number
-          )
+          # Use manual padding to avoid security risks with dynamic format strings
```
AliSoftware: In the context of this very PR this comment makes sense… but I think a couple of months from now when we might read that comment out of context (i.e. directly in the code not from the PR diff, and not seeing the previous code that was used) this might end up being confusing ("why are we talking about dynamic format strings here?")—including because taken out of context I'm not 100% sure that "dynamic format strings" will immediately make us think of `format` and then of `%<foo>@{digits}` stuff once the old code isn't around anymore…

So I think at that point we might just as well remove that comment 🙃

spec/derived_build_code_formatter_spec.rb
```diff
@@ -81,6 +74,125 @@
...
   end
 
+  describe 'configurable digit counts' do
+    context 'with custom digit counts' do
+      it 'uses 1 digit for each component' do
+        version = Fastlane::Models::AppVersion.new(1, 2, 3, 4)
+        formatter = described_class.new(prefix: '1', major_digits: 1, minor_digits: 1, patch_digits: 1, build_digits: 1)
+        build_code_string = formatter.build_code(version: version)
+        expect(build_code_string.to_s).to eq('11234')
+      end
```
AliSoftware, line 84: What if `minor_digits: 1` but `version.minor >= 10`? Shouldn't we `UI.user_error!` in that case warning that the `verison.minor` component doesn't fit in the number of `minor_digits`?

Because if we don't and this happens, this is going to lead to some mess in the generated `versionCodes`…

 - `AppVersion.new(1, 2, 3, 46)` => `112346`
 - `AppVersion.new(1, 23, 4, 5)` => should be an error because `23` doesn't fit in 1 digit, but imagine if we instead return `112345` (like the current implementation does)… and now we have a `versionCode` that is lower than a previous and quite older version…

If that were to happen one day we should definitively pause and manually make a decision on how to handle it case by case for each product encountering it (probably increasing the `minor_digits` in that example, but only manually doing so and while intentionally pondering about the consequences), rather than having the code silently do something unexpected with potentially impossible-to-revert consequences (e.g. consuming a higher-than-needed `versionCode`)

So I think we should add safeguards against this case in the implementation, `UI.user_error!` if that happens, then add corresponding tests to validate this raise as expected.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Updated on f107df16440d5c3a3c3843054b260b73134f7ec9

spec/derived_build_code_formatter_spec.rb
iangmaia: Good point 👍 Implemented on 31ec3a2b266f55868064d8fc2d893f002dc3059f

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -3,23 +3,39 @@
...
+          validate_digit_count!(minor_digits)
+          validate_digit_count!(patch_digits)
+          validate_digit_count!(build_digits)
+          validate_total_digits!(major_digits, minor_digits, patch_digits, build_digits)
+
           @prefix = prefix.to_s
+          @major_digits = major_digits
+          @minor_digits = minor_digits
+          @patch_digits = patch_digits
+          @build_digits = build_digits
```
AliSoftware: Might be just me nitpicking, but at that point I find the repeating of many lines for each of major, minor, patch and build quite repetitive. I wonder if it wouldn't make the code easier to store those in a 4-item array instead?

Wdyt about something like this instead?
```ruby
@digit_counts = [major_digits, minor_digits, patch_digits, build_digits]
@digit_counts.each { |d| validate_digit_count!(d) }
validate_total_digits!(@digit_counts)
```
And update your `validate_total_digits!` to:
```ruby
def validate_total_digits!(digits_list)
  total_digits = digits_list.sum
  …
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -67,6 +86,52 @@ def validate_prefix!(prefix)
...
+        #
+        def validate_version_component_fits!(component_value:, digit_limit:)
+          # Calculate the maximum value that fits in the digit limit
+          max_value = (10**digit_limit) - 1
+
+          return if component_value <= max_value
+
+          UI.user_error!("Version component value (#{component_value}) exceeds maximum allowed " \
+                         "for #{digit_limit} digit(s) (max: #{max_value}). Consider increasing the corresponding _digits parameter.")
+        end
```
AliSoftware: I think it'd make more sense to validate that the formatting of the digits doesn't exceed the length, so that we actually test the end result of what we want to avoid, not the potential causes for it.

i.e. validate that `.to_s.rjust(@x_digits, '0').length < @xdigits` rather than testing the int value is less that 10^n-1

And at that point and given the suggestion above to switch to use an array instead of repeating each 4 lines, might be better inlined in the `def build_code`? 🤔  (see code suggestion in next comment)

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -29,16 +45,19 @@ def initialize(prefix: nil)
...
+          result = [
+            @prefix,
+            version.major.to_s.rjust(@major_digits, '0'),
+            version.minor.to_s.rjust(@minor_digits, '0'),
+            version.patch.to_s.rjust(@patch_digits, '0'),
+            version.build_number.to_s.rjust(@build_digits, '0'),
+          ].join
 
           result.gsub(/^0+/, '')
         end
```
AliSoftware: Suppose you update `app_version.rb` to add a convenience accessor like this (you could also inline that here directly instead of course but to me it would make sense to expose that as a property of `AppVersion` type):

```ruby
def components
  [@major, @minor, @patch, @build_number]
end
```

Then you could simplify the whole implementation to:
```ruby
def build_code(_build_code = nil, version:)
  formatted_components = version.components.zip(@digit_counts).map do |value, width|
    comp = value.to_s.rjust(width, '0')
    if comp.length > width
      UI.user_error!("Version component value (#{value}) exceeds maximum allowed width of #{width} characters. " \
        "Consider increasing the corresponding `*_digits` parameter of your `#{self.class.name}`")
    end
    comp
  end
  [@prefix, *formatted_components].join.gsub(/^0+/, '')
end
```

And you wouldn't need to repeat lines for each of the 4 components, nor need a separate `validate_version_component_fits!` helper.

iangmaia: > update app_version.rb to add a convenience accessor

That's a great idea 👍  Used it to implement the last few suggestions at 369f423385fd8b809de2e64a91b39c7ba2ba9e53.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Agreed 👍 done as part of 369f423385fd8b809de2e64a91b39c7ba2ba9e53

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Updated on 369f423385fd8b809de2e64a91b39c7ba2ba9e53, the overall code got more compact and less repetitive 👍

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -3,23 +3,34 @@
...
-        def initialize(prefix: nil)
+        def initialize(prefix: nil, major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
           prefix ||= ''
           validate_prefix!(prefix)
+
+          @digit_counts = [major_digits, minor_digits, patch_digits, build_digits]
+          @digit_counts.each { |d| validate_digit_count!(d) }
+          validate_total_digits!(@digit_counts)
+
           @prefix = prefix.to_s
```
AliSoftware: 💄  Nitpick just for the form, to have everything prefix-related grouped together (and at the start since `prefix` is the first parameter so probably the first one that makes sense to assign):

```suggestion
          prefix ||= ''
          validate_prefix!(prefix)
          @prefix = prefix.to_s

          @digit_counts = [major_digits, minor_digits, patch_digits, build_digits]
          @digit_counts.each { |d| validate_digit_count!(d) }
          validate_total_digits!(@digit_counts)
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ _None_
 ### New Features
 
 - Add the possibility to configure in DerivedBuildCodeFormatter a versioning prefix instead of always defaulting to 1. [#656]
+- Add configurable number of digits in version components in DerivedBuildCodeFormatter. [#657]
```
AliSoftware: 💄 
```suggestion
- Add the possibility to configure in `DerivedBuildCodeFormatter` a versioning prefix instead of always defaulting to 1. [#656]
- Add configurable number of digits in version components in `DerivedBuildCodeFormatter`. [#657]
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -3,23 +3,34 @@
...
       # It takes in an AppVersion object and derives a build code from it.
       class DerivedBuildCodeFormatter
-        # Initialize the formatter with a configurable prefix.
+        # Initialize the formatter with configurable prefix and digit counts.
         #
         # @param [String] prefix The prefix to use for the build code. Must be a single digit (0-9), or empty string / nil.
+        # @param [Integer] major_digits Number of digits for major version. Defaults to 2.
+        # @param [Integer] minor_digits Number of digits for minor version. Defaults to 2.
+        # @param [Integer] patch_digits Number of digits for patch version. Defaults to 2.
+        # @param [Integer] build_digits Number of digits for build number. Defaults to 2.
```
AliSoftware: 💡 
```suggestion
        # @param [Integer] major_digits Number of digits for major version. Must be between 1–3. Defaults to 2.
        # @param [Integer] minor_digits Number of digits for minor version. Must be between 1–3. Defaults to 2.
        # @param [Integer] patch_digits Number of digits for patch version. Must be between 1–3. Defaults to 2.
        # @param [Integer] build_digits Number of digits for build number. Must be between 1–3. Defaults to 2.
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -67,6 +75,35 @@ def validate_prefix!(prefix)
...
+        #
+        # @raise [StandardError] If the digit count is invalid
+        #
+        def validate_digit_count!(digit_count)
+          # Check if it's an integer
+          unless digit_count.is_a?(Integer)
+            UI.user_error!("Digit count must be an integer, got: #{digit_count.class}")
+          end
+
+          return if digit_count.between?(1, 3)
```
AliSoftware: 💄  Nit: wonder if it'd make sense to make those `1` and `3` bounds constants, like you did with `MAX_TOTAL_DIGITS`?

spec/derived_build_code_formatter_spec.rb
```diff
@@ -81,6 +74,125 @@
...
+          expect { described_class.new(major_digits: 1, minor_digits: count, patch_digits: 1, build_digits: 1) }.not_to raise_error
+          expect { described_class.new(major_digits: 1, minor_digits: 1, patch_digits: count, build_digits: 1) }.not_to raise_error
+          expect { described_class.new(major_digits: 1, minor_digits: 1, patch_digits: 1, build_digits: count) }.not_to raise_error
+        end
+      end
+
+      it 'accepts mixed valid digit counts within 8 total digits' do
+        # 1 + 2 + 2 + 3 = 8 digits <= 8
+        expect { described_class.new(major_digits: 1, minor_digits: 2, patch_digits: 2, build_digits: 3) }.not_to raise_error
+      end
```
AliSoftware: 💡  Maybe we could create a loop that would test all possible combinations here, instead of just a specific one? 🤔 

Like, at the top of the `describe` block:
```ruby
let (:all_digit_combinations) { Array.new(4, [1,2,3]).reduce(&:product).map(&:flatten) }
```
Then here:
```ruby
context 'with mixed valid digit counts within 8 total digits' do
  all_digit_combinations.select { |comps| comps.sum <= 8 }.each do |major, minor, patch, build|
    it "accepts #{major},#{minor},#{patch},#{build} digit combination" do
      expect { described_class.new(major_digits: major, minor_digits: minor, patch_digits: patch, build_digits: build) }.not_to raise_error
    end
  end
end
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -81,6 +74,125 @@
...
+        expect { described_class.new(minor_digits: 1.0) }.to raise_error(/Digit count must be an integer, got: Float/)
+        expect { described_class.new(patch_digits: nil) }.to raise_error(/Digit count must be an integer, got: NilClass/)
+      end
+
+      it 'rejects configurations exceeding 8 total digits' do
+        # 3 + 3 + 3 + 3 = 12 digits > 8
+        expect { described_class.new(major_digits: 3, minor_digits: 3, patch_digits: 3, build_digits: 3) }.to raise_error(/Total digit count \(12\) exceeds maximum allowed \(8\)/)
+        # 2 + 2 + 2 + 3 = 9 digits > 8
+        expect { described_class.new(major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 3) }.to raise_error(/Total digit count \(9\) exceeds maximum allowed \(8\)/)
+      end
```
AliSoftware: 💡  And here similarly to above you can use `all_digit_combinations` using a `.select` that only keeps the combinations whose `n.sum > 8`:

```ruby
context 'with configurations exceeding 8 total digits' do
  all_digit_combinations.select { |comps| comps.sum > 8 }.each do |major, minor, patch, build|
    it "rejects #{major},#{minor},#{patch},#{build} digit combination" do
      expect { described_class.new(major_digits: major, minor_digits: minor, patch_digits: patch, build_digits: build) }.to raise_error(…)
    end
  end
end
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -81,6 +74,125 @@
...
+
+      it 'accepts configurations within and at 8 total digit limit' do
+        # 2 + 2 + 2 + 2 = 8 digits <= 8 (default config)
+        expect { described_class.new }.not_to raise_error
+        # 2 + 2 + 2 + 2 = 8 digits = 8 (at limit)
+        expect { described_class.new(major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2) }.not_to raise_error
+      end
```
AliSoftware: 🧹  At this point if you go with my suggestion to use `all_digit_combinations` to generate tests for all possible cases, that extra test here would be redundant.

Even if you don't go with my suggestion, in the current state of the test suite, this one seems at the wrong place, since it's currently within `context 'with invalid digit counts'` while it should be within `context 'with valid digit counts'` instead, right?

spec/derived_build_code_formatter_spec.rb
```diff
@@ -99,22 +211,91 @@
     end
 
     context 'with invalid prefixes' do
-      it 'rejects multi-character strings' do
+      it 'rejects invalid prefix formats' do
```
AliSoftware, line 218: 💡 wdyt about adding one more case, about `prefix` not even being a string at all (e.g. an `[]` array, or a `Object.new` or whatever), and ensuring we raise a nice error for those too?

spec/derived_build_code_formatter_spec.rb
```diff
@@ -81,6 +74,125 @@
...
+        # 2 + 2 + 2 + 2 = 8 digits <= 8 (default config)
+        expect { described_class.new }.not_to raise_error
+        # 2 + 2 + 2 + 2 = 8 digits = 8 (at limit)
+        expect { described_class.new(major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2) }.not_to raise_error
+      end
+    end
+  end
+
   describe 'prefix validation' do
     context 'with valid prefixes' do
```
AliSoftware, line 197: 💡 Would be nice to have a case for `nil` prefix as well, since that's also considered valid

spec/derived_build_code_formatter_spec.rb
```diff
@@ -99,22 +211,91 @@
...
+      it 'accepts version components at exactly the digit limit' do
+        formatter = described_class.new(major_digits: 2, minor_digits: 1, patch_digits: 3, build_digits: 2)
+
+        # Edge cases: exactly at limits
+        version = Fastlane::Models::AppVersion.new(99, 9, 999, 99)
+        expect { formatter.build_code(version: version) }.not_to raise_error
+      end
+    end
+
+    context 'with invalid version components exceeding digit limits' do
```
AliSoftware: 💄  I feel the term "invalid version components" might be misleading here. At first I thought you were testing cases like if `major` was a string instead of an integer, or was a negative value, etc…

What makes this invalid in this test case is that a component exceeds the digit limits, which is already stated in the test title, so I think adding "invalid" as an adjective for "version components" adds more confusion there, at least to me 🙃 

```suggestion
    context 'with version components exceeding digit limits' do
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -99,22 +211,91 @@
...
+        expect { described_class.new(prefix: '@') }.to raise_error(/Prefix must be an integer digit/)
       end
+    end
+  end
 
-      it 'rejects numbers outside 0-9 range' do
-        expect { described_class.new(prefix: '10') }.to raise_error(/Prefix must be a single digit or empty string/)
-        expect { described_class.new(prefix: '-1') }.to raise_error(/Prefix must be a single digit or empty string/)
+  describe 'version component validation' do
+    context 'with valid version components within digit limits' do
```
AliSoftware: 💄 
```suggestion
    context 'with version components within digit limits' do
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -99,22 +211,91 @@
...
+        # Edge cases: exactly at limits
+        version = Fastlane::Models::AppVersion.new(99, 9, 999, 99)
+        expect { formatter.build_code(version: version) }.not_to raise_error
+      end
+    end
+
+    context 'with invalid version components exceeding digit limits' do
+      it 'rejects major version exceeding digit limit' do
+        formatter = described_class.new(major_digits: 1, minor_digits: 2, patch_digits: 2, build_digits: 2)
+        version = Fastlane::Models::AppVersion.new(10, 1, 1, 1) # major=10 > max(9) for 1 digit
```
AliSoftware: 💄  nitpick: This `> max(9)` in the comment suggests that we're testing the int value, while in practice we're testing the formatted string's length now, so might as well remove that mention? 🤷 

```suggestion
        version = Fastlane::Models::AppVersion.new(10, 1, 1, 1) # major=10 is longer than 1 digit
```
(Of course, same for the other ones below)

spec/derived_build_code_formatter_spec.rb
```diff
@@ -81,6 +74,125 @@
...
+
+    context 'with empty prefix and custom digits' do
+      it 'trims leading zeros correctly with 1-digit major' do
+        version = Fastlane::Models::AppVersion.new(5, 12, 34, 56)
+        formatter = described_class.new(prefix: '', major_digits: 1, minor_digits: 2, patch_digits: 2, build_digits: 2)
+        build_code_string = formatter.build_code(version: version)
+        expect(build_code_string.to_s).to eq('5123456')
+      end
+
+      it 'trims leading zeros correctly with larger major' do
```
AliSoftware: 💄  The name of this test felt a bit misleading to me, as you're mentioning testing with "larger major", but the major is just `7` (sure, `7` is larger than the `5` used in the previous test… but was confused as to why that would be important).

It took me a minute to finally understand that this was about a larger _number of major digits_. So might be worth updating the test title to make this clearer 😅 

```suggestion
      it 'trims leading zeros correctly with larger number of major digits' do
```

AliSoftware (commented):
Mostly 💄 nitpicks + some suggestions to improve the unit tests. Otherwise LGTM 👍

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: I like it! 👍

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Added to 7dc1acaa7b2a83889f8a2376c88717b6e0d75b11 👍

spec/derived_build_code_formatter_spec.rb
iangmaia: Implemented common all_digit_combinations on 767671f241f5aa48279ef59b41be9fab8b1aaed9.

spec/derived_build_code_formatter_spec.rb
iangmaia: Added on 767671f241f5aa48279ef59b41be9fab8b1aaed9 👍

spec/derived_build_code_formatter_spec.rb
iangmaia: 👍 Yep, added `all_digit_combinations` and cleaned up here on 767671f241f5aa48279ef59b41be9fab8b1aaed9

spec/derived_build_code_formatter_spec.rb
iangmaia: Added on 767671f241f5aa48279ef59b41be9fab8b1aaed9 together with the remaining test improvements.

spec/derived_build_code_formatter_spec.rb
iangmaia: Added on 767671f241f5aa48279ef59b41be9fab8b1aaed9 together with the other test improvements.

iangmaia:
@AliSoftware I believe I addressed all the test suggestions!

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -52,6 +60,10 @@ def build_code(build_code = nil, version:)
         # @raise [StandardError] If the prefix is invalid
         #
         def validate_prefix!(prefix)
+          unless prefix.nil? || prefix.is_a?(String) || prefix.is_a?(Integer)
+            UI.user_error!("Prefix must be a string or integer, got: #{prefix.class}")
+          end
+
           prefix_str = prefix.to_s
```
AliSoftware, line 68: 👍 Today I re-learned that `nil.to_s` returns the `''` empty string (as opposed to crashing).

This is why even if now you call `validate_prefix!` from the `def initialize` without doing the `prefix ||= ''` anymore first, this `prefix_str = prefix.to_s` still doesn't crash 👍

spec/derived_build_code_formatter_spec.rb
```diff
@@ -81,6 +74,125 @@
...
+        formatter_new = described_class.new(prefix: '1', major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
+
+        expect(formatter_old.build_code(version: version)).to eq(formatter_new.build_code(version: version))
+      end
+    end
+  end
+
+  describe 'digit count validation' do
+    # Generate all combinations of [1,2,3] for each of the 4 digit count parameters
+    all_digit_combinations = Array.new(4, [1, 2, 3]).reduce(&:product).map(&:flatten)
```
AliSoftware, line 150: 💭 I'm surprised that `rubocop` doesn't complain about you using a local variable here instead of using RSpec's `let(:all_digit_combinations) { … }` syntax (which I believed was more idiomatic to Rspec)…

I don't mind either approach in practice (should do the same in the end, maybe except some memoizing of the value when using `let`?). I just wouldn't want a `rubocop` warning to appear because of this?

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -3,23 +3,34 @@
...
+        # @param [Integer] minor_digits Number of digits for minor version. Must be between 1–3. Defaults to 2.
+        # @param [Integer] patch_digits Number of digits for patch version. Must be between 1–3. Defaults to 2.
+        # @param [Integer] build_digits Number of digits for build number. Must be between 1–3. Defaults to 2.
         #
-        def initialize(prefix: nil)
-          prefix ||= ''
+        def initialize(prefix: nil, major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
           validate_prefix!(prefix)
-          @prefix = prefix.to_s
+          @prefix = (prefix || '').to_s
```
AliSoftware: 💡  Given what I re-learned about `nil.to_s` already returning `''`, I guess we could just write this now:
```suggestion
          @prefix = prefix.to_s
```
The only kind of cases I could think of where `(prefix || '').to_s` would do something different than `prefix.to_s` is if `prefix` was a falsy value other than nil… like if it had the actual boolean value `false`, in which case `false.to_s` returns `"false"` while `(false || '').to_s` returns `''`… but `prefix` being the `false` boolean would not be possible thanks to `validate_prefix!` forbidding it 🤷

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -3,23 +3,34 @@
 module Fastlane
   module Wpmreleasetoolkit
     module Versioning
+      MAX_TOTAL_DIGITS = 8
```
AliSoftware, line 7: 💄 Adding comment for extra context
```suggestion
       # Max total for `*_digits` params, not counting prefix
      MAX_TOTAL_DIGITS = 8
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -67,6 +79,35 @@ def validate_prefix!(prefix)
...
+            UI.user_error!("Digit count must be an integer, got: #{digit_count.class}")
+          end
+
+          return if digit_count.between?(MIN_DIGIT_COUNT, MAX_DIGIT_COUNT)
+
+          UI.user_error!("Digit count must be between #{MIN_DIGIT_COUNT} and #{MAX_DIGIT_COUNT} digits, got: #{digit_count}")
+        end
+
+        # Validates that the total number of digits (excluding prefix) doesn't exceed the maximum for multiplatform compatibility.
+        #
```
AliSoftware, line 102: 💄  Adding extra context in comment about the reason us wanting to limit the total number of digits being about Google Play's ≈`2_000_000_000` limit.
```suggestion
        # Validates that the total number of digits (excluding prefix) doesn't exceed the maximum for multiplatform compatibility.
        #
        # Since Google Play's max versionCode is ≈ 2_000_000_000, we want to avoid being too close to the limit
        # as this would then block us from submitting any updates for that app if we reached it.
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -96,25 +208,108 @@
...
+        version = Fastlane::Models::AppVersion.new(10, 1, 1, 1) # major=10 is longer than 1 digit
+
+        expect { formatter.build_code(version: version) }.to raise_error(
+          /Version component value \(10\) exceeds maximum allowed width of 1 characters/
+        )
+      end
+
+      it 'rejects minor version exceeding digit limit' do
+        formatter = described_class.new(major_digits: 2, minor_digits: 1, patch_digits: 2, build_digits: 2)
+        version = Fastlane::Models::AppVersion.new(1, 23, 4, 5) # minor=23 > max(9) for 1 digit
```
AliSoftware: I initially suggested to update all those comments but I see you only updated the first one 🙃 

Adding GitHub suggestions for all of them below to make it easier for you to apply them in batch directly from the GitHub UI
```suggestion
        version = Fastlane::Models::AppVersion.new(1, 23, 4, 5) # minor=23 is longer than 1 digit
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -96,25 +208,108 @@
...
+
+        expect { formatter.build_code(version: version) }.to raise_error(
+          /Version component value \(23\) exceeds maximum allowed width of 1 characters/
+        )
+      end
+
+      it 'rejects patch version exceeding digit limit' do
+        formatter = described_class.new(major_digits: 2, minor_digits: 2, patch_digits: 1, build_digits: 2)
+        version = Fastlane::Models::AppVersion.new(1, 2, 34, 5) # patch=34 > max(9) for 1 digit
```
AliSoftware:
```suggestion
        version = Fastlane::Models::AppVersion.new(1, 2, 34, 5) # patch=34 is longer than 1 digit
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -96,25 +208,108 @@
...
+
+        expect { formatter.build_code(version: version) }.to raise_error(
+          /Version component value \(34\) exceeds maximum allowed width of 1 characters/
+        )
+      end
+
+      it 'rejects build number exceeding digit limit' do
+        formatter = described_class.new(major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 1)
+        version = Fastlane::Models::AppVersion.new(1, 2, 3, 46) # build_number=46 > max(9) for 1 digit
```
AliSoftware:
```suggestion
        version = Fastlane::Models::AppVersion.new(1, 2, 3, 46) # build_number=46 is longer than 1 digit
```

spec/derived_build_code_formatter_spec.rb
```diff
@@ -96,25 +208,108 @@
...
+
+        expect { formatter.build_code(version: version) }.to raise_error(
+          /Version component value \(46\) exceeds maximum allowed width of 1 characters/
+        )
+      end
+
+      it 'provides helpful error messages with different digit limits' do
+        formatter = described_class.new(major_digits: 2, minor_digits: 2, patch_digits: 2, build_digits: 2)
+        version = Fastlane::Models::AppVersion.new(123, 4, 5, 6) # major=123 > max(99) for 2 digits
```
AliSoftware:
```suggestion
        version = Fastlane::Models::AppVersion.new(123, 4, 5, 6) # major=123 is longer than 2 digits
```

AliSoftware (approved):
Just nitpicks to fix code comments, otherwise LGTM so approving to unblock you

spec/derived_build_code_formatter_spec.rb
iangmaia: Yeah I did it first with a `let`, but had to re-learn that it won't work if we use it as part of the `context`:

```
  `all_digit_combinations` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).
```

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
iangmaia: Good one 👍

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb
```diff
@@ -3,23 +3,35 @@
 module Fastlane
   module Wpmreleasetoolkit
     module Versioning
+       # Max total for `*_digits` params, not counting prefix
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/wordpress-mobile/release-toolkit/blob/5e40646c91c5225f8d7199a3f461b9b352acfcf1/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb#L6"></span>Layout/CommentIndentation: Incorrect indentation detected (column 7 instead of 6).</td>
    </tr>
  </tbody>
</table>

PR #667, opened 2025-10-17
Extract GlotPress download retry logic and apply to all cases

iangmaia:
Related to p1760606044394999-slack-CC7L49W13
Fixes AINFRA-1419

## What does it do?

This PR extracts the GlotPress download code with retry logic originally from `metadata_download_helper.rb` to a separate helper `GlotPressDownloader`, and use that same helper also in `android_localize_helper.rb` and `ios_l10n_helper.rb`.

## Testing

I've created a draft PR where I ran lanes that download GlotPress: https://github.com/wordpress-mobile/WordPress-Android/pull/22291 and triggered two builds:
- First, running the lanes without the change: https://buildkite.com/automattic/wordpress-android/builds/23657
- Then running the lanes with the changes in this PR: https://buildkite.com/automattic/wordpress-android/builds/23659 


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
```diff
@@ -285,18 +286,15 @@ def self.download_from_glotpress(res_dir:, glotpress_project_url:, locales_map:,
...
-            UI.error "Error downloading #{locale} - #{e.message}"
-            retry if e.is_a?(OpenURI::HTTPError) && UI.confirm("Retry downloading `#{locale}`?")
-            nil
+          downloader = Fastlane::Helper::GlotPressDownloader.new(auto_retry: true)
+          result = nil
+          downloader.download(url, locale) do |response_body|
+            # Replace tabs with spaces (GlotPress uses tabs, but we prefer spaces)
+            result = Nokogiri::XML(response_body.gsub("\t", '    '), nil, Encoding::UTF_8.to_s)
           end
+          result
```
AliSoftware: Make the `downloader.download` method return the return value of the inner block it calls, so that you can then write this like this, with implicit return of the return value of `downloader.download`:

```suggestion
          downloader.download(url, locale) do |response_body|
            # Replace tabs with spaces (GlotPress uses tabs, but we prefer spaces)
            Nokogiri::XML(response_body.gsub("\t", '    '), nil, Encoding::UTF_8.to_s)
end
```

AliSoftware: Actually I see you already have implemented a `download_and_return` helper for that in the downloader class… so you probably want to use it here?

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_download_helper.rb
```diff
@@ -0,0 +1,119 @@
...
+      #
+      # @param [String] url The URL to download from
+      # @param [String] locale The locale being downloaded (for logging purposes)
+      # @return [String, nil] The response body if successful, nil otherwise
+      #
+      def download_and_return(url, locale)
+        result = nil
+        download(url, locale) { |body| result = body }
+        result
+      end
```
AliSoftware: Are there any use cases / call sites where we'd need to differentiate `download` vs `download_and_return`? e.g. does `download` return a `true`/`false` result saying download succeeeded, while `download_and_return` returns the value of the yielded block?!

If all our uses cases are either "we don't care about the returned value" or "we do want the return value of the yielded block", I think we might as well make `download` always return the result of the yielded block (like `download_and_return` currently does) and get rid of `download_and_return`?

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,119 @@
...
+      def make_request(uri)
+        http = Net::HTTP.new(uri.host, uri.port)
+        http.use_ssl = (uri.scheme == 'https')
+        request = Net::HTTP::Get.new(uri.request_uri)
+        request['User-Agent'] = Wpmreleasetoolkit::USER_AGENT
+        http.request(request)
+      rescue StandardError => e
+        # Network errors, connection errors, etc.
+        UI.error("Error downloading locale `#{@current_locale}` — #{e.message} (#{uri})")
+        retry if !FastlaneCore::Helper.is_ci? && UI.confirm("Retry downloading `#{@current_locale}`?")
```
AliSoftware: This means that if we're on CI it won't retry though, right?

While I think we want to auto-retry if CI is `true`, _or_ if CI is false but we `UI.confirm`'d we want to retry?
So shouldn't this be:

```suggestion
        retry if FastlaneCore::Helper.is_ci? || UI.confirm("Retry downloading `#{@current_locale}`?")
```

AliSoftware: Or is this only for retries that are unrelated to 429 and other cases that you handle in `handle_response`?

AliSoftware (commented):
Preliminary review, I didn't check the whole code in details nor the unit tests just yet, but leaving my initial high-level thoughts still as a first pass.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
iangmaia: Yeah I think this is a case where we want CI to fail, unrelated to 429.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_download_helper.rb
iangmaia: 👍 This was a leftover but indeed it's clearer to always return it. Done in 901a62c63f678ba55ccde774a987ef41230cf09a.

lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_localize_helper.rb
iangmaia: Done in 901a62c63f678ba55ccde774a987ef41230cf09a.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      # Downloads data from GlotPress
+      #
+      # @param [String] url The URL to download from
+      # @param [String] locale The locale being downloaded (for logging purposes)
+      # @yield [String] The response body if the download was successful
+      # @return The result of the block if provided, or true/false indicating success
+      #
+      def download(url, locale)
+        @current_locale = locale # Store for error handling
```
AliSoftware: Is there no way to propagate that locale to each method call that might need it, instead of using an instance variable, to avoid that risk altogether?
For example I think just passing it as a parameter to `make_request` (the only method which seems to reference that `@current_locale` apparently?) would be better?

Especially, I don't think we use `Parallel.map` right now to download multiple locales in parallel—but instead we just use a `.each` serial loop, right?—…but if we ever do, and thus one `GlotPressDownloader` might be reused for multiple loop iterations in parallel, storing that in an instance variable would lead to race condition issues.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_download_helper.rb
```diff
@@ -0,0 +1,110 @@
...
+        @auto_retry = auto_retry
+        @auto_retry_attempt_counter = 0
+      end
+
+      # Downloads data from GlotPress
+      #
+      # @param [String] url The URL to download from
+      # @param [String] locale The locale being downloaded (for logging purposes)
+      # @yield [String] The response body if the download was successful
+      # @return The result of the block if provided, or true/false indicating success
```
AliSoftware:
```suggestion
      # @return The result of the block if provided, or true/false indicating success if no block provided
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,110 @@
...
+        when '429'
+          # Rate limited
+          handle_rate_limiting(locale: locale, uri: original_uri) do |body|
+            yield body if block_given?
+          end
+        else
+          # Unexpected status code (including 404, 500, etc.)
+          status_line = "#{response.code} #{response.message}"
+          UI.error("Error downloading locale `#{locale}` — #{status_line} (#{original_uri})")
+          if !FastlaneCore::Helper.is_ci? && UI.confirm("Retry downloading `#{locale}`?")
```
AliSoftware: I wonder if `is_ci?` is the right condition for it here, as opposed to "is the current terminal interactive"… which I think Fastlane's `UI` module already have a helper for?

[EDIT] [Yes, `UI.interactive?` is the one](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb#L125) I had in mind 💭

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
AliSoftware: Also, the fact that we have to remember to reset the auto-retry counter once we get a `200` response (line 64) so that we're ready to handle the next call to `download` in the next caller loop iteration… makes it brittle too as well given the reuse context.

So I wonder if we shouldn't instead require to create a new instance for every single download, i.e. not have a single `GlotpressDownloader` instance created outside of the caller's loop and calling `downloader.download(url, locale)` on the same instance inside each loop iteration, but instead have the `url` + `locale` be on the constructor (`def initialize`) and `download` take no argument at all, to enfoce each loop iteration having to create a separate instance, so that we're guaranteed that there won't be any instance variables with old values we forgot to reset between loops, and that we wouldn't risk having race conditions from reuse (especially if we migrate to `Parallel.map` one day…)

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
AliSoftware, line 1: Update the file name to match class?

iangmaia: Done on cc05d07c2394fcae14b0e5ca56caa6527273921a.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
iangmaia: Good one! 👍 Updated on 5f11e781e6495dd538eb5ac2fe5d245419c07b90.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
iangmaia: Great suggestion, thanks! 👍 it makes the state management easier to follow. Updated on 19fc7118e20b751b702d0b59d30e2fe0ec92925d.

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,114 @@
...
+      def handle_rate_limiting(url:)
+        if @auto_retry && @auto_retry_attempt_counter < MAX_AUTO_RETRY_ATTEMPTS
+          UI.message("Received 429 for `#{@locale}`. Auto retrying in #{AUTO_RETRY_SLEEP_TIME} seconds... (attempt #{@auto_retry_attempt_counter + 1}/#{MAX_AUTO_RETRY_ATTEMPTS})")
+          sleep(AUTO_RETRY_SLEEP_TIME)
+          @auto_retry_attempt_counter += 1
+          download_from_url(url) { |body| yield body if block_given? }
+        elsif UI.interactive? && UI.confirm("Retry downloading `#{@locale}` after receiving 429 from the API?")
+          download_from_url(url) { |body| yield body if block_given? }
+        else
+          UI.error("Abandoning `#{@locale}` download as requested.")
```
AliSoftware: 💄  That use case can also happen when you don't _request_ abandoning per se, but also when `@auto_retry = false` but you're also on CI (`UI.interactive? == false`)… in that case you wouldn't have been asked and would not technically have _requested_ to stop… 🤷 
```suggestion
          UI.error("Abandoning `#{@locale}` download.")
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb
```diff
@@ -0,0 +1,114 @@
...
+        @auto_retry = auto_retry
+        @auto_retry_attempt_counter = 0
+      end
+
+      # Downloads data from GlotPress
+      #
+      # @yield [String] The response body if the download was successful
+      # @return The result of the block if provided, or true/false indicating success if no block provided
+      #
+      def download(&)
```
AliSoftware, line 46: Given that all call sites will now basically always call `downloader = GlotpressDownloader.new(…)` followed immediately by `downloader.download`, I wonder if it would be worth adding a convenience class method too to make all the call sites that use that pattern even simpler?

```ruby
def self.download(url:, locale:, auto_retry: false, &)
  downloader = self.new(url: url, locale: locale, auto_retry: auto_retry)
  downloader.download(&)
end
```

AliSoftware (approved):
Approving to unblock, as the 2 comments are nitpicks not blockers.

PS: Probably worth testing this on a real project (e.g. making `wordpress-android` point its `Gemfile` to this PR's branch and run a test lane calling this to validate it works in real world use case) before merging, just to be extra sure.

iangmaia:
> PS: Probably worth testing this on a real project

Done, updated the description with a reference to a [draft PR](https://github.com/wordpress-mobile/WordPress-Android/pull/22291).

PR #685, opened 2026-01-16
Promo screenshots action refinements

mokagio:
## What does it do?

I tried running the screenshot annotation on my end and it failed. Internal ref p1768453313053599/1768288277.528279-slack-CC7L49W13.

The reason it failed is that I didn't have screenshots images to annotate. That's obvious in hindsight, but it took me a bit to understand in between the dense logs.

This PR tracks the changes I made to cleanup the logs.

> [!NOTE]  
> ~~As discussed in the thread, I haven't been able to test the changes yet in terms of actually calling down to ImageMagick. Still, I'd appreciate your feedback.~~
>
> Update: This has now been tested locally, but I closed the session with the successful outcome so I don't have a screenshot (pun not intended) to share.
>
> I sampled the generated annotated screenshots and they LGTM.

- [x] Add Unit Tests (aka `specs/*_spec.rb`) if applicable. — N.A.


lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -33,7 +33,7 @@ def initialize
           UI.user_error!(message)
         end
 
-        UI.user_error!('`drawText` not found – install it using `brew install automattic/build-tools/drawText`.') unless system('command -v drawText')
+        UI.user_error!('`drawText` not found – install it using `brew install automattic/build-tools/drawText`.') unless system('command -v drawText > /dev/null')
```
mokagio: Otherwise, we'd get a long list of `command -v drawText` output.

As a side note, this line seems to be called _many_ times. We should cache, memoize, or otherwise implement a way to run this check only once.

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -287,7 +287,16 @@ def draw_text_to_canvas(canvas, text, width, height, x_position, y_position, fon
...
+          system(
+            'drawText',
+            "html=#{text}",
+            "maxWidth=#{width}",
+            "maxHeight=#{height}",
+            "output=#{temp_text_file.path}",
+            "fontSize=#{font_size}",
+            "stylesheet=#{stylesheet_path}",
+            "alignment=#{position}"
+          )
```
mokagio, line 306: Fastlane's `sh` should support `logs: false` to silence the call, but it did not parse.

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -407,8 +416,8 @@ def create_image(width, height, background = 'transparent')
         working_background = background.frozen? ? background.dup : background
         working_background.paint.to_hex
 
-        Image.new(width, height) do
-          self.background_color = working_background
+        Image.new(width, height) do |info|
+          info.background_color = working_background
         end
       end
```
mokagio, line 429: This addresses a deprecation warning. See commit message for more details.

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -432,8 +441,12 @@ def resolve_path(path)
           return resolved_path if !resolved_path.nil? && resolved_path.exist?
         end
 
-        message = "Unable to locate #{path}"
-        UI.crash!(message)
+        message = <<~MESSAGE
+          Unable to locate #{path}.
+
+          Did you run the automation to generate the screenshots?
```
mokagio, line 454: Just in case I go through the same process of pondering why annotating non existing screenshots fails again.

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -432,8 +441,12 @@ def resolve_path(path)
...
+        MESSAGE
+        UI.user_error!(message)
```
mokagio, line 456: Used `user_error!` instead of `crash!` because it makes less noise in the logs as it doesn't give a stacktrace.

iangmaia: Good one 👍

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
iangmaia: Shouldn't it be just `log: false`, per [docs](https://docs.fastlane.tools/actions/sh/) and [source](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/sh.rb)?

mokagio: ~~Thanks @iangmaia ! Sometimes the issue is just a small letter...~~

~~Addressed in 12abcb0d.~~

Update: I thought I tested 12abcb0d via my WooCommerce iOS using the local version, but I had discarded that `Gemfile` change so my edits did not run.

Using `log` did not fix the issue. Here's the stacktrace:

```
bundler: failed to load command: fastlane (/Users/gio/Developer/a8c/wcios/vendor/bundle/ruby/3.2.0/bin/fastlane)
/Users/gio/Developer/a8c/wcios/vendor/bundle/ruby/3.2.0/gems/fastlane-2.231.0/fastlane/lib/fastlane/helper/sh_helper.rb:28:in `sh_control_output': [!] unknown keyword: :log (ArgumentError)
	from /Users/gio/.rbenv/versions/3.2.2/lib/ruby/3.2.0/forwardable.rb:240:in `sh'
	from /Users/gio/Developer/a8c/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb:290:in `draw_text_to_canvas'
	from /Users/gio/Developer/a8c/release-toolkit/lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb:126:in `draw_caption_to_canvas'
```

mokagio: > Sometimes the issue is just a small letter...

Just a small letter indeed.

`Action.sh` fails. `Actions.sh` works.

From https://github.com/wordpress-mobile/release-toolkit/pull/685/commits/05875c7484cb7630180f6968adc4060b51a7157b:

`Action.sh` delegates directly to `Actions.sh_control_output,` which accepts
`print_command:` and `print_command_output:`, not `log:`.

The delegation occurs in `fastlane/lib/fastlane/action.rb` using Ruby's
Forwardable module:

```
def_delegator(Actions, :sh_control_output, :sh)
```

`Actions.sh` is a wrapper that accepts `log:` and translates it to `print_command:` and
`print_command_output:` when calling `sh_control_output`.

See:

- https://github.com/fastlane/fastlane/blob/2.231.0/fastlane/lib/fastlane/action.rb#L32-L39
- https://github.com/fastlane/fastlane/blob/2.231.0/fastlane/lib/fastlane/helper/sh_helper.rb#L4-L28

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
```diff
@@ -444,6 +458,7 @@ def resolve_text_into_path(text, locale)
         elsif can_resolve_path(localized_file)
           resolve_path(localized_file).realpath.to_s
         else
+          UI.important("Could not identify '#{localized_file}' as a file path or the file was not found. Will use its value as a raw string. This may result in undesired annotations.")
```
mokagio, line 467: In testing, I was surprised to find annotated screenshots with a path instead of of text as the title.

The reason was misconfiguration of the path on my end, but the fact remains that the tool, in being flexible, might hide missing files. This can be dangerous if automation runs unsupervised.

We might want to rethink the tool's behavior. But in the meantime, here's a warning in the logs.

Image

(_Not sure yet why it's printed twice_)

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
iangmaia: So I guess we should consider something like:

```ruby
UI.user_error!('`drawText` not found – install it using `brew install automattic/build-tools/drawText`.') unless self.drawText_available?
```

And then:

```ruby
def self.drawText_available?
  @drawText_available ||= system('command -v drawText')
end
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
iangmaia: > The reason was misconfiguration of the path on my end, but the fact remains that the tool, in being flexible, might hide missing files. This can be dangerous if automation runs unsupervised.

👍 I agree.

---

> Not sure yet why it's printed twice

I see that if both paths are misconfigured and can't be resolved, we'll then have two warnings as `resolve_text_into_path` is called twice (`draw_caption_to_canvas` and `draw_text_attachment_to_canvas`).

iangmaia (approved):
Left a [non blocking comment](https://github.com/wordpress-mobile/release-toolkit/pull/685#discussion_r2705048509) about avoiding `command -v` being called multiple times and there's also the [CHANGELOG.md update](https://github.com/wordpress-mobile/release-toolkit/pull/685#issuecomment-3757520138).

Approving to unblock (I've disabled auto-merge in case you want to address anything before merging).

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/promo_screenshots_action.rb
```diff
@@ -93,7 +93,7 @@ def self.confirm_directory_overwrite(path, description)
             FileUtils.rm_rf(path)
             Dir.mkdir(path)
           else
-            UI.user_error!("Exiting to avoid overwriting #{description}.")
+            UI.abort_with_message!("Exiting to avoid overwriting #{description}.")
```
mokagio, line 98: Seems more appropriate as an exit call resulting from a `confirm` user input than `user_error!`.

I'd actually argue we should not crash and simply terminate the execution (`return`, `next`). But that's out of scope at the moment because it would require restructuring the entire method to bubble up the intention of terminating. Simply calling `return` here would only result in the code continuing, which is not what we desire.

lib/fastlane/plugin/wpmreleasetoolkit/helper/promo_screenshots_helper.rb
mokagio: > I see that if both paths are misconfigured

Oh, of course! Thanks @iangmaia !

In my tests, I had hardcoded only one language, but forgot Woo is configured to create promo screenshots for iPhone and iPad. Given the copy is the same, we'll obviously attempt two reads of that localization file. All clear now.

mokagio:
@iangmaia thanks again for the review. I implemented your suggestions and tested the code locally via the integration PR on WooCommerce iOS, https://github.com/woocommerce/woocommerce-ios/pull/16517

I'm going to merge to keep the ball rolling. Then, I'll merge `trunk` into https://github.com/wordpress-mobile/release-toolkit/pull/684 to sort out the CHANGELOG conflicts that will arise.

PR #687, opened 2026-01-21
Add Claude skill for publishing releases

mokagio:
> [!CAUTION]
> Closed. We discussed internally and concluded that since the process is already mostly automated, the skill is redundant.

I used this skill to publish https://github.com/wordpress-mobile/release-toolkit/releases/tag/14.0.0

To be precise, I iterated on the skill live while publishing the release.

I particularly like the bit where the agent polls the PR status and then goes on with the GitHub release creation. 

One could argue that it might have been more robust to add further automation instead of having an resource intense LLM agent put the steps together. I hear you. This is just an experiment, happy to iterate or drop altogether.


.claude/skills/release/SKILL.md
```diff
@@ -0,0 +1,190 @@
...
+  --trailer "Co-Authored-By: <USER_NAME> <<USER_EMAIL>>"
+git push --force-with-lease
+```
+
+### Step 2: Create the Pull Request
+
+After the rake task pushes the branch, create the PR using `gh`:
+
+```bash
+/opt/homebrew/bin/gh pr create \
```
mokagio, line 69: I have `gh` aliased to `git checkout` and have instructed my Claude to always use the Homebrew path.

I think it's fine to leave as is. If not found at that path, Claude should try other likely locations itself.

Or, maybe we can use `gh` here and see what my agent does...

mokagio:
Closing. We discussed it internally, p1768971412519689-slack-C08CUT5NT88, and concluded that since the process is already mostly automated, the skill is unnecessary.

PR #696, opened 2026-02-24
Add `exclude` parameter to `find_previous_tag` action

iangmaia:
Fixes AINFRA-2056

## What does it do?

Adds an optional `exclude` parameter to the `find_previous_tag` action that maps to git's `--exclude` flag, allowing callers to skip beta/pre-release tags when searching for the previous stable release tag — e.g. `find_previous_tag(pattern: 'v*', exclude: '*beta*')`.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_previous_tag.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_previous_tag.rb
@@ -8,7 +8,7 @@ module Fastlane
     class FindPreviousTagAction < Action
       def self.run(params)
         tag_pattern = params[:pattern]
-        exclude_patterns = Array(params[:exclude])
+        exclude_patterns = params[:exclude] || []
 
         # Make sure we have all the latest tags fetched locally
         Actions.sh('git', 'fetch', '--tags', '--force') { nil }
@@ -41,7 +41,7 @@ module Fastlane
           reachable from the current commit and that matches a specific naming pattern
 
           e.g. `find_previous_tag(pattern: '12.3.*.*')`, `find_previous_tag(pattern: '12.3-rc-*')`,
-          `find_previous_tag(pattern: 'v*', exclude: '*beta*')`,
+          `find_previous_tag(pattern: 'v*', exclude: ['*beta*'])`,
           `find_previous_tag(pattern: 'v*', exclude: ['*alpha*', '*beta*'])`
         DETAILS
       end
@@ -54,11 +54,10 @@ module Fastlane
                                        default_value: nil,
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :exclude,
-                                       description: 'One or more _fnmatch_-style patterns of tags to exclude from the search (maps to `git describe --exclude`). ' \
-                                                    'Can be a single string or an array of strings',
+                                       description: 'An array of _fnmatch_-style patterns of tags to exclude from the search (maps to `git describe --exclude`)',
                                        optional: true,
                                        default_value: nil,
-                                       is_string: false),
+                                       type: Array),
         ]
       end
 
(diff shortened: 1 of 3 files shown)


AliSoftware (approved):
Approving to unblock but while we're here I wonder if we shouldn't allow **multiple** exclude patterns (by allowing to pass an array to the argument/ConfigItem), for example if we want to exclude both `-alpha*` and `-beta*` patterns 🤔

iangmaia:
> Approving to unblock but while we're here I wonder if we shouldn't allow **multiple** exclude patterns (by allowing to pass an array to the argument/ConfigItem), for example if we want to exclude both `-alpha*` and `-beta*` patterns 🤔

Great idea! Added on 4208ed79c78617ebc1e41fc4e1d01c4789fe480a.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_previous_tag.rb
```diff
@@ -49,6 +53,12 @@ def self.available_options
                                        optional: true,
                                        default_value: nil,
                                        type: String),
+          FastlaneCore::ConfigItem.new(key: :exclude,
+                                       description: 'One or more _fnmatch_-style patterns of tags to exclude from the search (maps to `git describe --exclude`). ' \
+                                                    'Can be a single string or an array of strings',
+                                       optional: true,
+                                       default_value: nil,
+                                       is_string: false),
```
AliSoftware: The use of `is_string:` is deprecated in `ConfigItem`, and has been replaced by `type:` instead.


```suggestion
                                       type: Array),
```

iangmaia: Ah indeed, though the type could be a string as well (single string or Array)... will that work with `type`? 🤔  Or better to restrict it to be always an array?

AliSoftware: I believe if the `ConfigItem`'s `type` is `Array` but you pass a `String` [it gets auto-converted into an array](https://github.com/fastlane/fastlane/blob/737aae7ce100404630bef42bd9302228bf49c478/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L269-L273) and [validation should thus pass in those cases too](https://github.com/fastlane/fastlane/blob/737aae7ce100404630bef42bd9302228bf49c478/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L219-L228).

Though wouldn't hurt to add a small spec example if you want to ensure this case is covered as expected anyway.

iangmaia: Updated on d09c29d6e660919c3acce557cba86ad8ec4d0771, enforcing it to be always array (thinking about it, it's fine as it's a new parameter anyway).

spec/find_previous_tag_spec.rb
```diff
@@ -65,6 +65,70 @@ def stub_main_command(expected_command, stdout:, success: true)
...
+      stdout: 'v1.7.3'
+    )
+    # Act — Fastlane's ConfigItem auto-converts a String to Array when type is Array
+    tag = run_described_fastlane_action(
+      pattern: 'v*',
+      exclude: '*beta*'
+    )
+    # Assert
+    expect(tag).to eq('v1.7.3')
+  end
```
AliSoftware, line 114: Nice to have that confirmed by a unit test 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_previous_tag.rb
iangmaia: ...and the auto-conversion works! 4121687e17373727785961ee14e797da0bbbd014 (*chef's kiss*)

AliSoftware (approved):

PR #699, opened 2026-02-26
Add common pitfalls and quick start to `AGENTS.md`

mokagio:
## What does it do?

Documents the bootstrap sequence and instructs the agent to not inspect `vendor/` unless necessary.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,12 @@ Release Toolkit is a _fastlane_ plugin  implemented as a Ruby gem (`fastlane-plu
 - **Main branch**: `trunk`
 - **Gem version**: defined in `lib/fastlane/plugin/wpmreleasetoolkit/version.rb`
 
+## Quick Start
+
+```sh
+bundle install && gem build && bundle exec rubocop && bundle exec rspec
+```
+
 ## Build and Development Commands
 
 ```sh
@@ -215,3 +221,7 @@ Runtime and development dependencies with version constraints are defined in `fa
 - **gettext** — i18n/PO files
 - **google-cloud-storage** — GCS uploads
 - **buildkit** — Buildkite API
+
+## Common Pitfalls
+
+- **Don't inspect or edit `vendor/bundle/` unless necessary** — The bundled gems live there and are managed by Bundler.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,7 +15,7 @@ Release Toolkit is a _fastlane_ plugin  implemented as a Ruby gem (`fastlane-plu
 ## Quick Start
 
 ```sh
-bundle install && gem build && bundle exec rubocop && bundle exec rspec
+bundle install && bundle exec rubocop && bundle exec rspec
 ```
 
 ## Build and Development Commands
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

AGENTS.md
```diff
@@ -12,6 +12,12 @@ Release Toolkit is a _fastlane_ plugin  implemented as a Ruby gem (`fastlane-plu
 - **Main branch**: `trunk`
 - **Gem version**: defined in `lib/fastlane/plugin/wpmreleasetoolkit/version.rb`
 
+## Quick Start
+
+```sh
+bundle install && gem build && bundle exec rubocop && bundle exec rspec
```
mokagio: This could have been a script. But scripting this would have brought up the question: where to we put the script? 

- Can't use `rake` because it should be used via Bundler
- Using `make` would create an inconsistency with `rake`
- Adding a `bin/` folder with a `bootstrap` script would also raise the question: why this a script and that via `rake` or `bundle exec`?

AliSoftware: I don't mind it being directly in `AGENTS.md` rather than a separate script.

---

I think it'd make more sense to lint and run tests first before building the gem though, i.e. swap the order to:

```suggestion
bundle install && bundle exec rubocop && bundle exec rspec && gem build
```

I also wonder if `gem build` is actually useful / actually makes sense in the context of a "Quick Start". Personally, I never ran this command locally ever when I worked on `release-toolkit` so far. I make changes to the code, run `rubocop` and `rspec` to validate syntax and tests pass, and that's it. I don't need a `fastlane-plugin-wpmreleasetoolkit.gem` package to be generated locally.

The only benefit from having that `.gem` would be to:
 - Either install it locally with `gem install fastlane-plugin-wpmreleasetoolkit.gem`—but if my goal is to run the gem commands to test them locally, I can just `bundle exec` locally from the local working copy without having to generate the `.gem` package and install it on my system for that
 - Or if I intended to then `gem push` that `.gem` package to RubyGems for publishing. But that's something the agent shouldn't do itself, as that's done by CI (triggered on a tag) instead

The only remaining benefit of running `gem build` (other that to actually get a `.gem` archive/package produced) would be to validate that the gem _can_ build without error (e.g. validate that there isn't a syntax error in the `.gemspec` that would cause issues building the packaged gem). But while that can be useful if the AI actually makes changes to the `.gemspec` file, I think this is a very rare case and probably not a situation for what I'd call a "Quick Start" use case?

So maybe just this is enough:

```suggestion
bundle install && bundle exec rubocop && bundle exec rspec
```
Which would also prevent creating the `.gem` artifact then it risking creating confusion to the AI further down the road (e.g. having the AI thing that since the instructions suggest to call `gem build` and generate the `.gem` package, that's how it's supposed to test any changes when it wants to call any action from the gem, instead of calling an action directly from `bundle exec` without packaging + installing the gem first)

mokagio: I'm fine to remove `gem build`, too. In this context, it was just a lint of the `.gemspec` file.

twstokes (approved):

PR #701, opened 2026-03-02
Add update_apps_cdn_build_metadata action

iangmaia:
Fixes AINFRA-2102

## Summary

- Adds a new `update_apps_cdn_build_metadata` Fastlane action that updates metadata of existing builds on the Apps CDN without re-uploading files
- Uses the dedicated Apps CDN update endpoint (`POST /wpcom/v2/sites/{site_id}/a8c-cdn/builds/{post_id}`) with a JSON body that accepts the same string-based format as the upload flow (e.g. `visibility: 'Internal'`)
- Accepts an array of `post_ids` to update multiple builds (one request per post)
- Supports updating `visibility` (`:internal` / `:external`) and `post_status` (`publish` / `draft`)
- This enables a two-phase release flow: upload builds as Internal during finalize, then flip to External at publish time

## Context

The companion [Studio PR](https://github.com/Automattic/studio/pull/2724) uses this action in its `publish_release` lane to flip CDN build visibility from Internal to External.

The existing `upload_build_to_apps_cdn` action cannot be used to change visibility because visibility is part of the dedup matching criteria in the CDN plugin — re-uploading with a different visibility creates a new build rather than updating the existing one.

> [!NOTE]
> An earlier version of this PR used the standard WP REST API v2 post endpoint with taxonomy term ID lookups. Standard REST writes are now blocked for builds (`rest_build_write_blocked`) in favor of the dedicated update endpoint added in the CDN plugin (docs: PCYsg-15tP-p2), so the action was reworked accordingly — which also simplified it (string-based values, no term ID lookups).

## Test plan

- [x] Unit tests updated and passing (`bundle exec rspec spec/update_apps_cdn_build_metadata_spec.rb` — 16 examples, 0 failures)
- [x] `bundle exec rubocop` — no offenses on changed files
- [x] End-to-end test against the CDN test site (`sandboxcdn.wordpress.com`, see AINFRA-2171): uploaded test builds as Internal, flipped single + batch `post_ids` to External and back, updated `post_status` alone and combined with `visibility`, verified each change via wp/v2 reads, confirmed the legacy wp/v2 write path returns `403 rest_build_write_blocked`, and confirmed a bogus post ID fails cleanly


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
```diff
@@ -0,0 +1,149 @@
...
+          result = JSON.parse(response.body)
+          post_id = result['ID']
+
+          UI.success("Successfully updated Apps CDN build metadata for post #{post_id}")
+
+          { post_id: post_id }
+        else
+          UI.error("Failed to update Apps CDN build metadata: #{response.code} #{response.message}")
+          UI.error(response.body)
+          UI.user_error!('Update of Apps CDN build metadata failed')
```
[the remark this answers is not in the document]
iangmaia: This follows the same pattern as `upload_build_to_apps_cdn` and other HTTP-calling actions, relying on Fastlane's built-in exception handling instead. Rescuing on `StandardError` would also risk masking non-network bugs with misleading error messages.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
```diff
@@ -0,0 +1,181 @@
...
+
+        request = Net::HTTP::Get.new(uri.request_uri)
+        request['Accept'] = 'application/json'
+        request['Authorization'] = "Bearer #{api_token}"
+
+        response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: uri.scheme == 'https') do |http|
+          http.open_timeout = 10
+          http.read_timeout = 30
+          http.request(request)
+        end
```
AliSoftware, line 91: Given the call site will typically call this `update_apps_cdn_build_metadata` action in a loop ([here](https://github.com/Automattic/studio/pull/2695/changes#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R850), this means that the fetching of the visibility term id here would be done many times during the `make_cdn_builds_public` lane that calls those.

Wouldn't it be better to have `update_apps_cdn_build_metadata` be able to take an array of `post_id` values to update the visibility/post_status metadata of, so that it does this `lookup_visibility_term_id` API call only once while looping on all the posts to update them with the same visibility ID.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
iangmaia: Updated it to use the array `post_ids` on 3276a267e7277013e3d6047d3987d5512055951a.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
```diff
@@ -0,0 +1,193 @@
...
+require 'fastlane/action'
+require 'net/http'
+require 'uri'
+require 'json'
+
+module Fastlane
+  module Actions
+    class UpdateAppsCdnBuildMetadataAction < Action
+      VALID_VISIBILITIES = %i[internal external].freeze
+      VALID_POST_STATUS = %w[publish draft].freeze
```
AliSoftware: 💭 Between those constants and the URL templates used for AppsCDN API calls and the code doing those API calls all having some stuff in common between this new action and the existing `upload_build_to_apps_cdn` action, I think maybe it'd make sense to DRY some of those commonalities in a helper then have the actions reference those?

This way for example if the list of valid visibilities ever changes in the Apps CDN wpcom plugin and it now accept a different set of `VALID_VIAIBILITIES` values, we'd only have to update it in a single place / source of truth within the release-toolkit codebase.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
```diff
@@ -0,0 +1,193 @@
...
+        post_ids = params[:post_ids]
+        UI.message("Updating Apps CDN build metadata for #{post_ids.size} post(s): #{post_ids.join(', ')}...")
+
+        # Build the base JSON body for the WP REST API v2
+        body = {}
+        body['status'] = params[:post_status] if params[:post_status]
+
+        if params[:visibility]
+          term_id = lookup_visibility_term_id(site_id: params[:site_id], api_token: params[:api_token], visibility: params[:visibility])
+          body['visibility'] = [term_id]
```
AliSoftware, line 21: Probably more of a question for Hannah but it feels surprising to me that the API would technically allow an array if values for this metadata instead of a single one 🤔

Like, what would happen at the API level if we called the API with `{"visibility":[1345, 1367]}` body (where, say, 1345 is internal and 1367 is external, for example) 😅

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
AliSoftware: This would also allow to have more consistency in the base URL endpoint used for both actions (while currently one uses `v1.1` while this new one uses `v2`)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
```diff
@@ -0,0 +1,193 @@
...
+        end
+
+        case response
+        when Net::HTTPSuccess
+          result = JSON.parse(response.body)
+          updated_id = result['id']
+
+          UI.message("  Updated post #{updated_id}")
+
+          { post_id: updated_id }
```
AliSoftware: Why return a `Hash` with a single key instead of returning the `updated_id` value directly?

(Btw would be nice to add some YARD doc on this helper method, especially a `# @return [Integer]` info assuming you switch to get rid of that hash)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
```diff
@@ -0,0 +1,193 @@
...
+          UI.error("Failed to update Apps CDN build metadata for post #{post_id}: #{response.code} #{response.message}")
+          UI.error(response.body)
+          UI.user_error!("Update of Apps CDN build metadata failed for post #{post_id}")
+        end
+      end
+
+      # Look up the taxonomy term ID for a visibility value (e.g. :internal -> 1316)
+      def self.lookup_visibility_term_id(site_id:, api_token:, visibility:)
+        slug = visibility.to_s.downcase
+        api_endpoint = "https://public-api.wordpress.com/wp/v2/sites/#{site_id}/visibility?slug=#{slug}"
```
AliSoftware: Again an opportunity to DRY the base URL of that endpoint into a helper, typically one common to this action and the other existing action

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
iangmaia: Good idea 👍 Added a `AppsCdnHelper` on f0a30219935b63e4c95c8dab7b3e29308890f106.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
iangmaia: 👍 Also updated on f0a30219935b63e4c95c8dab7b3e29308890f106.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
iangmaia: Updated on 6df3e3e013ea37e50b4fb53cd2270b6e3fe60a4b (YARD doc added on d4d539d07e93859c1d8bd56082fd4d8676459ec7)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
```diff
@@ -0,0 +1,207 @@
...
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :post_status,
+            description: "The new post status ('publish' or 'draft')",
+            optional: true,
+            type: String,
+            verify_block: proc do |value|
+              UI.user_error!("Post status must be one of: #{VALID_POST_STATUS.join(', ')}") unless VALID_POST_STATUS.include?(value)
+            end
+          ),
```
AliSoftware, line 171: I was wondering if it would be worth moving that into the `Helper` too, since those parameters are needed for both actions, and that way we'd DRY the `verify_block` for each… but I realized that the `description` for each would be slightly different ("new visibility" here for example).

But maybe still worth DRY-ing the `verify_block` of each in the `Helper`, at least?
 - For example this could be a `def verify_post_status(value)` method on the `Helper` whose function body would just consists of the `UI.user_error!(…) unless …` line, or maybe the helper function would return the `proc` directly so you can do something like `verify_block: &:Helper::verify_post_status`
 - This would also allow you to remove lines 12-13 where you alias the constants to the helper into this action

lib/fastlane/plugin/wpmreleasetoolkit/helper/apps_cdn_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      # Builds a WordPress.com WP REST API v2 URL scoped to a site.
+      #
+      # @param site_id [String] the WordPress.com site ID
+      # @param path [String] the API path relative to the site (e.g. 'a8c_cdn_build/123')
+      # @return [String] the full API URL
+      def self.wp_v2_url(site_id:, path:)
+        "#{API_BASE_URL}/wp/v2/sites/#{site_id}/#{path}"
+      end
```
AliSoftware, line 52: Given all the call sites end up calling `URI.parse` on the returned string, maybe it'd make more sense to have the helper function directly return a parsed `URI` instead of a `String`?

(and same for the v1.1 one obviously)

lib/fastlane/plugin/wpmreleasetoolkit/helper/apps_cdn_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+      VALID_POST_STATUS = %w[publish draft].freeze
+
+      # Builds a WordPress.com REST API v1.1 URL scoped to a site.
+      #
+      # @param site_id [String] the WordPress.com site ID
+      # @param path [String] the API path relative to the site (e.g. 'media/new')
+      # @return [String] the full API URL
+      def self.rest_v1_1_url(site_id:, path:)
+        "#{API_BASE_URL}/rest/v1.1/sites/#{site_id}/#{path}"
+      end
```
AliSoftware, line 23: I was about to suggest that maybe we should migrate `upload_build_to_apps_cdn` to use the `v2/` endpoint instead of the `v1.1` endpoint, to simplify the code by being consistent in the API used in both actions and avoiding 2 different URLs.

But then I asked LibreChat what this would involve and it seems it would not be as simple as I thought to migrate the media upload process (`/media/new`): https://chat.a8c.com/share/c7IckxiQzEs9VS0VxvG_O

So probably simpler to keep the `/rest/v1.1` API for `upload_build_to_apps_cdn` after all, even if that means using 2 different APIs for `upload_*` and `update_*` actions 🤷

spec/update_apps_cdn_build_metadata_spec.rb
```diff
@@ -0,0 +1,364 @@
...
+    end
+
+    it 'handles visibility term lookup failure' do
+      stub_request(:get, visibility_term_url)
+        .with(query: { 'slug' => 'external' })
+        .to_return(
+          status: 200,
+          body: [].to_json,
+          headers: { 'Content-Type' => 'application/json' }
+        )
```
AliSoftware, line 263: I'm curious, is it really the case in the real world that the API would return an `HTTP 200` (with `[]` empty list as response) in such case, as opposed to, say, a `404` or something?

The first time I read this code I focused on the `200` without really seeing the `body: [].to_json` and thought it was not correct because we were supposed to test an error case. But in retrospect it probably makes sense that the API is indeed returning a `200` (like this test suggests) for such cases, so this test code is probably correct after all, but just wanted to double-check

AliSoftware (approved):
2 small suggestions for also DRY-ing the `ConfigItems` or at least their `verify_block:` procs into the `Helper`, as well as integrating the `URI.parse(url)` calls into the `{rest_v1_1,wp_v2}_url` helpers directly.
But LGTM apart from that, so approving to unblock.

lib/fastlane/plugin/wpmreleasetoolkit/helper/apps_cdn_helper.rb
iangmaia: 👍 Updated on 185f5ed6266e6f714c3be1d1d14c61e72a941333

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
iangmaia: Yep, I think it makes sense to centralize in the helper 👍 Updated on a79094ea1ebcaecdafc8bdf50ee90f73c9316da9

spec/update_apps_cdn_build_metadata_spec.rb
iangmaia: Yes, confirmed by testing against the real API.
The WP REST API v2 taxonomy endpoint returns HTTP 200 with an empty array for non-existent slugs:
```
GET /wp/v2/sites/{site_id}/visibility?slug=nonexistent                                                                                                                                                                            
→ 200 OK                                           
[]                                               
```

 vs. a valid slug:

```
GET /wp/v2/sites/{site_id}/visibility?slug=external
→ 200 OK
[{"id":21293,"count":2483,"description":"","name":"External","slug":"external","taxonomy":"visibility",...}]
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/apps_cdn_helper.rb
```diff
@@ -0,0 +1,55 @@
...
+      # @return [URI] the parsed full API URI
+      def self.wp_v2_url(site_id:, path:)
+        URI.parse("#{API_BASE_URL}/wp/v2/sites/#{site_id}/#{path}")
+      end
+
+      # Returns a proc that validates a visibility value against {VALID_VISIBILITIES}.
+      # Intended for use as a `verify_block` in Fastlane ConfigItem definitions.
+      #
+      # @return [Proc] a proc that raises FastlaneError if the value is invalid
+      def self.verify_visibility
```
AliSoftware: 💄 Maybe name this `verify_visibility_param` (to not make the helper method be confused with a name that could be interpreted as something that would verify the visibility of a given post using an API call or something)?

(Same suggestion for `verify_post_status` obviously)

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
iangmaia: Just double checked using the API. It indeed accepts multiple term IDs and assigns both:

```
POST /wp/v2/sites/{site_id}/a8c_cdn_build/737
{"visibility": [1316, 21293]}

→ 200 OK
{"id": 737, "visibility": [21293, 1316], ...}
```

So a post would end up being both "internal" and "external" at the same time 😅 it just treats visibility as a taxonomy that supports multiple terms.

lib/fastlane/plugin/wpmreleasetoolkit/helper/apps_cdn_helper.rb
iangmaia: 👍 Updated on 1a4e9d3bfce72ab9c596555d0ca7aa81ad5f5f96

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
AliSoftware: Interesting! I wonder why we didn't have to pass an array but [just a string in the API call that creates the build in `upload_build_to_apps_cdn`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_build_to_apps_cdn.rb#L65) then 🤔 🤷 

Also, this doesn't match the table in the FieldGuide documentation (interal ref: PCYsg-15tP-p2#How-to-upload-a-build-to-a-CDN-blog), so I guess one of the 2 should probably be updated.
I'd vote for only updating the AppsCDN plugin for accepting only one visibility (both in the API call and on the visibility metadata stored for a post taxonomy), but I'm not sure if that's technically feasible/easy (vs if this is an inherent technical limitation of how taxonomies work in WordPress? cc @hannahtinkler).

Anyway, I think ultimately it's not really critical (given the way we use the API and this visibility taxonomy we always set only one value in practice in all our call sites); as long as we're sure that calling `update_build_to_apps_cdn` with a one-item array `[term_id]` on a post that already had a different value set for `visibility` indeed _replaces_ the visibility of that post with a single visibility value of `term_id`—as opposed to adding it to the existing ones.

AliSoftware (approved):
:shipit:

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/update_apps_cdn_build_metadata.rb
hannahtinkler: Hmm, I didn’t realise how annoying uploading via that endpoint would be with having to use the internal IDs, pass arrays etc. I looked at customising the update functionality for this post type, but it got complex to the point it seemed better to add a dedicated upload endpoint - you can find the docs for that here: PCYsg-15tP-p2 🙂

I also blocked edits via the standard REST post endpoints so the validation can be enforced - this should all be available in prod now! FYI, I also added a test site so we don't need to clean up testing calls in the prod CDN sites - you can find the details for that in AINFRA-2171.

iangmaia: I've updated the action to use the new `/wpcom/v2/sites/{site_id}/a8c-cdn/builds/{post_id}` endpoint, which ends up being much simpler now (string values, no term-ID lookups) (5ca786686c500dda537a65db7395b68e25a81ae5), and checked it against `sandboxcdn` and confirmed the standard REST write path now returns `rest_build_write_blocked`.

PR #702, opened 2026-03-03
Add list_apps_cdn_builds action

iangmaia:
Fixes AINFRA-2102

## Summary

Adds a new `list_apps_cdn_builds` Fastlane action that queries the WP REST API v2 to list builds on the Apps CDN. Supports server-side filtering by `post_status` (`draft` / `publish`) and `version` (via taxonomy term ID lookup). Returns an array of build metadata (post_id, title, status, version, visibility, platform, build_type).

## Context

This enables Studio's `publish_release` lane to query the CDN directly for draft builds by version, then publish them — rather than embedding post IDs in the GitHub release body (which is fragile). See [companion Studio PR](https://github.com/Automattic/studio/pull/2695) for the consumer side.

Note: The v1.1 API returns 500 for `a8c_cdn_build` custom post types, which is why this action uses the WP v2 endpoint (`/wp/v2/sites/{site_id}/a8c_cdn_build`). Additionally, CDN posts with `visibility: internal` are completely hidden from all listing queries by the CDN plugin, so the draft/publish approach is used instead of internal/external visibility for gating builds before release.

## Test plan

- [x] `bundle exec rspec spec/list_apps_cdn_builds_spec.rb` — 12 examples, 0 failures
- [x] `bundle exec rubocop` — no offenses on new files
- [x] Manual test: verified against the real CDN API that listing by post_status and version works correctly, including the full draft→publish roundtrip

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-- Added new `list_apps_cdn_builds` action to list builds on the Apps CDN with optional filtering by visibility (server-side) and version (client-side). This enables querying CDN builds directly by version instead of embedding post IDs in GitHub releases. [#xxx]
+- Added new `list_apps_cdn_builds` action to list builds on the Apps CDN with optional filtering by visibility (server-side) and version (client-side). This enables querying CDN builds directly by version instead of embedding post IDs in GitHub releases. [#702]
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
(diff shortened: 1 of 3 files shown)


AliSoftware:
> Note: The v1.1 API returns 500 for a8c_cdn_build custom post types, which is why this action uses the WP v2 endpoint (/wp/v2/sites/{site_id}/a8c_cdn_build).

Should we then [update `upload_build_to_apps_cdn` action](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_build_to_apps_cdn.rb#L58-L59) to use the `/v2` API endpoint too, for consistency?

iangmaia:
No longer needed. The new Studio PR (Automattic/studio#2724) uses the internal visibility approach with post IDs embedded in the GitHub release body, so there's no need to query the CDN API for builds. Only the `update_apps_cdn_build_metadata` action (#701) is required.

PR #707, opened 2026-03-30
Release 14.2.1 into trunk

iangmaia:
Releasing new version 14.2.1.

# What's Next

```
### Bug Fixes

- Added optional `next_version` parameter to `android_update_release_notes` and `ios_update_release_notes` actions, allowing callers to provide the next version directly instead of relying on the built-in calculator that assumes minor version caps at 9. This fixes incorrect version bumps (e.g., 8.9 → 9.0 instead of 8.10) for apps using semantic versioning. [#706]


```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,17 @@ _None_
 
 ### Bug Fixes
 
-- Added optional `next_version` parameter to `android_update_release_notes` and `ios_update_release_notes` actions, allowing callers to provide the next version directly instead of relying on the built-in calculator that assumes minor version caps at 9. This fixes incorrect version bumps (e.g., 8.9 → 9.0 instead of 8.10) for apps using semantic versioning. [#706]
+_None_
 
 ### Internal Changes
 
 _None_
+
+## 14.2.1
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,12 +14,18 @@ _None_
...
 
 ### Internal Changes
 
 _None_
 
+## 14.2.1
+
+### Bug Fixes
+
+- Added optional `next_version` parameter to `android_update_release_notes` and `ios_update_release_notes` actions, allowing callers to provide the next version directly instead of relying on the built-in calculator that assumes minor version caps at 9. This fixes incorrect version bumps (e.g., 8.9 → 9.0 instead of 8.10) for apps using semantic versioning. [#706]
```
AliSoftware, line 27: A bit pedantic of me, but while this is a bugfix (which is why it was considered being a patch semver bump)… in practice the fix for the bug consisted of adding a new parameter to the action and thus changing the API (even if in a backwards-compatible way), so IMHO that should be a minor version bump, not a patch for that reason.

iangmaia: Haha, I thought exactly about it right after I ran `rake new_version` and wondered the same 😄 oh well. Let's do it properly.

iangmaia:
Closing in favor of a minor version bump (14.3.0) since the change adds a new parameter.

iangmaia:
Moved to https://github.com/wordpress-mobile/release-toolkit/pull/708

PR #709, opened 2026-03-30
Remove activesupport as a runtime dependency and clean up RuboCop config

iangmaia:
## Summary

- Remove `activesupport` as a runtime dependency. The only production usage was `deep_dup` in `PromoScreenshotsAction`, replaced with `Marshal.load(Marshal.dump(...))`. Moved to a dev dependency (`~> 7.2`) for specs that use ActiveSupport methods (`Time.use_zone`, `String#to_time`, `Hash#slice!`).
- Fix RuboCop violations: resolve `Style/FileOpen` (use `File.foreach` instead of `File.open` without a block), fix obsolete `Naming/PredicateName` → `Naming/PredicatePrefix` rename.
- Move permanent style choices from `.rubocop_todo.yml` to `.rubocop.yml` (e.g. `Naming/PredicatePrefix`, `Naming/PredicateMethod`, `Style/Documentation`, `RSpec/MessageSpies`, `RSpec/SpecFilePathFormat`, etc.), keeping only actionable tech debt in the todo file.

## Test plan

- [x] `bundle exec rspec` — 800 examples, 0 failures
- [x] `bundle exec rubocop` — no offenses
- [x] Verified `Marshal.load(Marshal.dump(entry))` behaves identically to `deep_dup` for the hash structures used in `PromoScreenshotsAction`


spec/spec_helper.rb
```diff
@@ -23,6 +23,7 @@
 module SpecHelper
 end
 
+require 'active_support/all' # ActiveSupport methods (e.g. Time.use_zone, String#to_time) used by some specs
```
mokagio, line 26: This surprised me. I checked with GPT 5.4 and learned a bit about ActiveSupport's architecture and agreed that `/all` makes sense in the context of `spec_helper.rb` to simplify downstream usages.

mokagio (approved):
Makes sense @iangmaia . Thanks for the ping.

I looked at the Git history and that upper bound was introduced to address a `gem build` warning:

```
WARNING:  open-ended dependency on activesupport (>= 6.1.7.1) is not recommended
  if activesupport is semantically versioned, use:
    add_runtime_dependency "activesupport", "~> 6.1", ">= 6.1.7.1"
```

Given there's no need to expose `activesupport` as an external dependency, the fix in this PR is better indeed.

Just in case, I run `gem build` from here and it gave no warnings.

---

Aside. `activesupport` comes as a CocoaPods dependency now. That made me realize that we are not that far from being able to remove CocoaPods itself from here. Without discarding the amazing service CP has done for us over the years, it will be a good day when we'll get rid of what is now essentially dead weight.

spec/spec_helper.rb
iangmaia: btw, this is also required on `actions/common/promo_screenshots_action.rb`. Given we still directly depend  on it, maybe it makes sense to keep the explicit dependency and just remove the upper bound 🤔

iangmaia: ...and removing the upper bound gives you this warning:

```console
 WARNING:  open-ended dependency on activesupport (>= 6.1) is not recommended
       if activesupport is semantically versioned, use:
         add_runtime_dependency 'activesupport', '~> 6.1'
```

iangmaia: I've now [fixed](https://github.com/wordpress-mobile/release-toolkit/pull/709/commits/a9de2f1ab530a12ef6d8cbc8c064501a23b8b7e6) some Rubocop violations ~and [added](https://github.com/wordpress-mobile/release-toolkit/pull/709/commits/22b0d177ae442d6c239bcf08a2d4db4707bc508c) a higher upper bound limit~, and got no warnings on my side so I think we're good?

iangmaia: Update: removed activesupport as a runtime dependency and added it only as a dev dependency: https://github.com/wordpress-mobile/release-toolkit/pull/709/commits/0bc62744527f56e51baacff7ccef33fed028cf36

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/promo_screenshots_action.rb
```diff
@@ -187,7 +187,7 @@ def self.build_entries(config_entries, languages, output_directory, params)
         config_entries
           .flat_map do |entry|
           languages.map do |language|
-            new_entry = entry.deep_dup
+            new_entry = Marshal.load(Marshal.dump(entry))
```
AliSoftware: Worth adding a comment to explain that workaround/trick
```suggestion
            new_entry = Marshal.load(Marshal.dump(entry)) # trick to do a _deep_ copy (`.dup` would only do shallow one)
```

CHANGELOG.md
```diff
@@ -18,7 +18,8 @@ _None_
 
 ### Internal Changes
 
-_None_
+- Remove explicit `activesupport` version constraint (`< 8`) from gemspec and re-add it as an explicit runtime dependency (`>= 6.1, < 9`) since production code uses it. [#709]
```
AliSoftware: This is not true anymore with latest commits

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -46,6 +45,7 @@ Gem::Specification.new do |spec|
   # `google-cloud-storage` is required by fastlane, but we pin it in case it's not in the future
   spec.add_dependency 'google-cloud-storage', '~> 1.31'
 
+  spec.add_development_dependency 'activesupport', '~> 7.2'
```
AliSoftware: At that point I'm guessing it's probably OK to bump to `8.x` when used only for running specs the risk is smaller than if it was used in prod code?

iangmaia: Ah yes, maybe good to add a comment here as well: `cocoapods` enforces `< 8`.

AliSoftware: Wait, do we need `cocoapods` as a dependency still? 🤔 
(Maybe something to ponder for a follow-up PR though if you don't want to drag this PR too long)

CHANGELOG.md
iangmaia: Updated on https://github.com/wordpress-mobile/release-toolkit/pull/709/commits/864ce97ea891ea486211e28f43fa152fc99c1c59

AliSoftware (approved):

fastlane-plugin-wpmreleasetoolkit.gemspec
iangmaia: Hmm I assumed so, but a search shows `ios_build_preflight.rb` which calls `other_action.cocoapods`. So maybe we could remove it, as it seems we're using it via _fastlane_ and not directly?

iangmaia: Okay 😄 https://github.com/wordpress-mobile/release-toolkit/pull/709/commits/2bc392c7ced7cf4397b11015cd846b6b2dcd9332 🧹 🧹 🧹

PR #717, opened 2026-05-12
Release 14.4.2 into trunk

jaclync:
Releasing new version 14.4.2.

# What's Next

```
### New Features

- `openai_ask`: support OpenAI tool-use (function calling) and model overrides. New optional parameters: `tools`, `tool_handlers`, and `max_tool_iterations` enable a multi-turn loop where the model invokes locally-defined tools and receives `role: tool` results until it produces a plain text response or hits the iteration cap; `model` overrides the previously hardcoded `gpt-4o`. Tool-handler keys may be strings or symbols (normalized internally), exceptions raised inside a handler are surfaced back to the model as structured `{ error:, exception: }` tool results (carrying only the exception class — the message and backtrace are logged locally but not forwarded, since handler errors can contain secrets) rather than crashing the lane, and the new parameters validate their inputs (non-empty Array, callable handlers, iteration cap >= 1). Backwards-compatible — when `tools` is omitted, behavior is unchanged. Also fixes a long-standing `NameError` when generating action documentation (documented examples are now non-interpolated heredocs) and a `EXEMPLE` typo in the example HEREDOC tags. [#716]


```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-- `openai_ask`: support OpenAI tool-use (function calling) and model overrides. New optional parameters: `tools`, `tool_handlers`, and `max_tool_iterations` enable a multi-turn loop where the model invokes locally-defined tools and receives `role: tool` results until it produces a plain text response or hits the iteration cap; `model` overrides the previously hardcoded `gpt-4o`. Tool-handler keys may be strings or symbols (normalized internally), exceptions raised inside a handler are surfaced back to the model as structured `{ error:, exception: }` tool results (carrying only the exception class — the message and backtrace are logged locally but not forwarded, since handler errors can contain secrets) rather than crashing the lane, and the new parameters validate their inputs (non-empty Array, callable handlers, iteration cap >= 1). Backwards-compatible — when `tools` is omitted, behavior is unchanged. Also fixes a long-standing `NameError` when generating action documentation (documented examples are now non-interpolated heredocs) and a `EXEMPLE` typo in the example HEREDOC tags. [#716]
+_None_
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ _None_
 
 _None_
 
+## 14.4.2
+
+### New Features
+
+- `openai_ask`: support OpenAI tool-use (function calling) and model overrides. New optional parameters: `tools`, `tool_handlers`, and `max_tool_iterations` enable a multi-turn loop where the model invokes locally-defined tools and receives `role: tool` results until it produces a plain text response or hits the iteration cap; `model` overrides the previously hardcoded `gpt-4o`. Tool-handler keys may be strings or symbols (normalized internally), exceptions raised inside a handler are surfaced back to the model as structured `{ error:, exception: }` tool results (carrying only the exception class — the message and backtrace are logged locally but not forwarded, since handler errors can contain secrets) rather than crashing the lane, and the new parameters validate their inputs (non-empty Array, callable handlers, iteration cap >= 1). Backwards-compatible — when `tools` is omitted, behavior is unchanged. Also fixes a long-standing `NameError` when generating action documentation (documented examples are now non-interpolated heredocs) and a `EXEMPLE` typo in the example HEREDOC tags. [#716]
```
mokagio, line 27: What do you think of this slightly less verbose entry?

```suggestion
- `openai_ask`: support OpenAI tool-use (function calling) and model overrides. New optional parameters: `tools`, `tool_handlers`, and `max_tool_iterations` enable a multi-turn loop where the model invokes locally-defined tools and receives `role: tool` results until it produces a plain text response or hits the iteration cap; `model` overrides the previously hardcoded `gpt-4o`. Exceptions raised inside a handler are surfaced back to the model as structured `{ error:, exception: }` tool results rather than crashing the lane. [#716]
```

lib/fastlane/plugin/wpmreleasetoolkit/version.rb
```diff
@@ -3,6 +3,6 @@
 module Fastlane
   module Wpmreleasetoolkit
     NAME = 'fastlane-plugin-wpmreleasetoolkit'
-    VERSION = '14.4.1'
+    VERSION = '14.4.2'
```
mokagio, line 6: Given it's a new feature, this should be 14.5.0:

```suggestion
    VERSION = '14.5.0'
```

@jaclync did you run the release Rake task or did this manually? I'm asking because if the task did this version bump, then there's something we need to fix there.

mokagio (changes requested):
I think this should be 14.5.0 instead of 14.4.2. 

@jaclync if you are still around to address this, take this as an approval to merge and ship afterward. I'll check back here in a couple of hours and if nothing has changed, I'll assume you were offline already and will do the change myself to keep the RSM project moving along.

lib/fastlane/plugin/wpmreleasetoolkit/version.rb
jaclync: Thanks for the review Gio! It was from the release Rake task, `rake new_release`, and I typed in `14.4.2` thinking that this is a minor backward compatible change. I will close this PR and create a new one from the Rake task + your changelog suggestions.

jaclync:
Closing in favor of https://github.com/wordpress-mobile/release-toolkit/pull/718 from feedback https://github.com/wordpress-mobile/release-toolkit/pull/717#pullrequestreview-4268480121

CHANGELOG.md
jaclync: Great point. Updated the changlog in https://github.com/wordpress-mobile/release-toolkit/pull/718/changes/a922e08b89cca15f22aefca95460f90147608cf8.

PR #719, opened 2026-05-12
Improve openai_ask tool-use behavior

iangmaia:
## What does it do?

Improves `openai_ask` follow-up behavior after the initial tool-use support:

- Uses `max_completion_tokens` for Chat Completions requests.
- Adds `store: false` to opt out of OpenAI request storage.
- Validates function tool definitions before sending them to OpenAI.
- Keeps malformed or unsupported tool calls from invoking local handlers.
- Caps local tool execution without dropping the final allowed model response.
- Avoids logging raw tool arguments, handler exception messages, or backtraces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- `openai_ask`: improve tool-use handling by validating supported function tools, using `max_completion_tokens`, avoiding secret-bearing debug logs, and refusing to execute additional tool calls after `max_tool_iterations`.
 
 ### Internal Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-- `openai_ask`: improve tool-use handling by requiring named function tools, using `max_completion_tokens`, opting out of OpenAI request storage, omitting sensitive tool diagnostics from logs, and refusing to execute additional tool calls after `max_tool_iterations`.
+- `openai_ask`: improve tool-use handling by requiring named function tools, using `max_completion_tokens`, opting out of OpenAI request storage, omitting sensitive tool diagnostics from logs, and refusing to execute additional tool calls after `max_tool_iterations`. [#719]
 
 ### Internal Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-- `openai_ask`: improve tool-use handling by requiring named function tools, using `max_completion_tokens`, omitting sensitive tool diagnostics from normal logs, and refusing to execute additional tool calls after `max_tool_iterations`.
+- `openai_ask`: improve tool-use handling by requiring named function tools, using `max_completion_tokens`, opting out of OpenAI request storage, omitting sensitive tool diagnostics from logs, and refusing to execute additional tool calls after `max_tool_iterations`.
 
 ### Internal Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,11 @@ _None_
 
 ### Bug Fixes
 
-- `openai_ask`: improve tool-use handling by requiring named function tools, using `max_completion_tokens`, opting out of OpenAI request storage, omitting sensitive tool diagnostics from logs, and refusing to execute additional tool calls after `max_tool_iterations`. [#719]
+- `openai_ask`: avoid logging sensitive tool diagnostics and refuse to execute additional tool calls after `max_tool_iterations`. [#719]
 
 ### Internal Changes
 
-_None_
+- `openai_ask`: validate named function tools, use `max_completion_tokens`, and opt out of OpenAI request storage. [#719]
 
 ## 14.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ _None_
 
 ### Internal Changes
 
-- `openai_ask`: validate named function tools, use `max_completion_tokens`, and opt out of OpenAI request storage. [#719]
+- `openai_ask`: validate named function tools, default to `gpt-4.1`, use `max_completion_tokens`, and opt out of OpenAI request storage. [#719]
 
 ## 14.5.0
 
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
```diff
@@ -32,19 +33,22 @@ def self.run(params)
...
           Authorization: "Bearer #{api_token}"
         }
 
         # Backwards-compatible single-shot path when no tools are provided.
-        if tools.nil? || tools.empty?
+        if tools.nil?
           body = request_body(prompt: prompt, question: question, model: model)
           response = Net::HTTP.post(OPENAI_API_ENDPOINT, body, headers)
           return parse_text_response(response)
         end
```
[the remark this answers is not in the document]
iangmaia: Fixed in c3bab1ed by adding direct `.run` validation for `tools` before `empty?`/`validate_tools!`, with a spec covering non-array input.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
```diff
@@ -32,19 +33,22 @@ def self.run(params)
         # callers can register handlers with either string or symbol keys without surprises.
         tool_handlers = (params[:tool_handlers] || {}).transform_keys(&:to_s)
         max_tool_iterations = params[:max_tool_iterations] || DEFAULT_MAX_TOOL_ITERATIONS
+        validate_max_tool_iterations!(max_tool_iterations)
```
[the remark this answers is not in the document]
iangmaia: Fixed in fd139b11 by moving `validate_max_tool_iterations!` into the tool-use path, after `tools` is known to be present and valid. Added a regression spec for single-shot calls with an otherwise invalid `max_tool_iterations` value.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
```diff
@@ -8,6 +8,7 @@ module Fastlane
   module Actions
     class OpenaiAskAction < Action
       OPENAI_API_ENDPOINT = URI('https://api.openai.com/v1/chat/completions').freeze
+      DEFAULT_MAX_COMPLETION_TOKENS = 2048
```
mokagio, line 12: It would be good to document where this number comes from. What's its rationale?

I feel the 5 in the max tool iterations is much easier to accept at face value than 2048 tokens.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
```diff
@@ -39,12 +40,15 @@ def self.run(params)
...
-        if tools.nil? || tools.empty?
+        if tools.nil?
           body = request_body(prompt: prompt, question: question, model: model)
           response = Net::HTTP.post(OPENAI_API_ENDPOINT, body, headers)
           return parse_text_response(response)
         end
 
+        validate_tools_array!(tools)
+        validate_max_tool_iterations!(max_tool_iterations)
+        validate_tools!(tools)
```
mokagio, line 52: Nitpick: What do you think of an empty line to visually separate the validation from the execution?

```suggestion
        validate_tools!(tools)

```

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
```diff
@@ -140,8 +152,49 @@ def self.parse_assistant_message(response)
...
+        return if invalid_tools.empty?
+
+        UI.user_error!(
+          'Parameter `tools` only supports OpenAI function tools with a non-empty `function.name`. ' \
+          "Invalid tool definitions: #{invalid_tools.join(', ')}"
+        )
+      end
+
+      def self.tool_type(tool)
+        return '' unless tool.is_a?(Hash)
```
mokagio: I wonder if returning `nil` would be clearer here.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- `openai_ask`: improve tool-use handling by requiring named function tools, using `max_completion_tokens`, opting out of OpenAI request storage, omitting sensitive tool diagnostics from logs, and refusing to execute additional tool calls after `max_tool_iterations`. [#719]
```
mokagio: Is this a bug fix then? All of the entries seem improvements to me. Should this be an internal changes or new feature entry?

iangmaia: I think it's both 😅 `omitting sensitive tool diagnostics from logs` and not running `additional tool calls after max_tool_iterations` are actually bug fixes. But indeed other points are improvements.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
iangmaia: Applied in 1e9c94c8.

CHANGELOG.md
iangmaia: Updated to split in 1e9c94c8.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
iangmaia: Agreed. Changed `tool_type` to return `nil` for missing/non-hash definitions in 1e9c94c8.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/openai_ask_action.rb
iangmaia: Addressed in 1e9c94c8 by documenting that `DEFAULT_MAX_COMPLETION_TOKENS` preserves the previous `max_tokens` ceiling while moving to the current API field.

mokagio: Thanks! Also, only later in the review I realized that this was a previously introduced magic number, I should have acknowledge that in the comment.

mokagio (approved):

jaclync:
Thanks for the enhancements 🙏

PR #723, opened 2026-05-20
Populate process ENV from EnvManager by default

mokagio:
## What does it do?

Flips `EnvManager`'s default back to populating the process `ENV` from the loaded `.env` file (no-override semantics — pre-existing `ENV` wins).
Adds a `mutate_env: false` opt-out for callers who want to preserve the parse-only / instance-isolation semantics introduced in #578.

### Why

#578 switched `EnvManager` from `Dotenv.load` to `Dotenv.parse` to make instances independent of process `ENV` and to give `reset!` something to undo.

Alas, fastlane actions look up their `default_value:` via `ENV.fetch(...)`, so with `ENV` pristine, calls like `app_store_connect_api_key` can't see anything loaded from the `.env` file.

This PR keeps the structure of #578 (parse + per-instance dict) but layers values into `ENV` by default.
`reset!` tracks the keys it added and removes only those, so pre-existing `ENV` entries are never disturbed.

### Migration

This is technically a breaking behavior change for anyone who upgraded to a release that included #578 and depended on `ENV` staying pristine.
Mitigation: pass `mutate_env: false` to `set_up` (or `new`) to restore the previous behavior.

It's unfortunate to have to do a major version bump for this little used feature, but versions are cheap so we're better off keeping semantic integrity, IMHO.

- [x] If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) — pending a decision on whether this ships in a major or minor bump.


lib/fastlane/plugin/wpmreleasetoolkit/env_manager/env_manager.rb
```diff
@@ -31,10 +43,29 @@ def initialize(
...
+          @mutated_keys << key
+        end
+      end
+
+      # Remove from `ENV` any keys this instance added via `mutate_env: true`.
+      # Idempotent — calling more than once is safe. Used by `reset!` and
+      # available for callers that want to roll back manually.
+      def restore_env!
+        @mutated_keys.each { |key| ENV.delete(key) }
+        @mutated_keys = []
```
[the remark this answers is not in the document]
mokagio: Good catch — fixed in 97a20022.

`@mutated_keys` is now `@mutations` (a `key => value` hash recording what this instance wrote), and `restore_env!` only deletes when `ENV[key]` still equals the value we set. Any later overwrite by another caller is left in place. Added a regression test in `spec/env_manager_spec.rb`.

*Posted by Claude (Opus 4.7) on behalf of @mokagio with approval.*

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-_None_
+- `EnvManager`: populate the process `ENV` from the loaded `.env` file by default (no-override semantics — pre-existing `ENV` values win), so fastlane actions that read their `default_value:` from `ENV` find loaded secrets without callers having to thread them through explicitly. Pass `mutate_env: false` to opt out and keep the parse-only / instance-isolation semantics introduced in [#578]. `EnvManager.reset!` now also rolls back the keys it added.
```
```suggestion
- `EnvManager`: populate the process `ENV` from the loaded `.env` file by default (no-override semantics — pre-existing `ENV` values win) [#578].
```

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-_None_
+- `EnvManager`: populate the process `ENV` from the loaded `.env` file by default (no-override semantics — pre-existing `ENV` values win) [#578].
```
```suggestion
- `EnvManager`: populate the process `ENV` from the loaded `.env` file by default (no-override semantics — pre-existing `ENV` values win) [#723].
```

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-_None_
+- `EnvManager`: populate the process `ENV` from the loaded `.env` file by default (no-override semantics — pre-existing `ENV` values win), so fastlane actions that read their `default_value:` from `ENV` find loaded secrets without callers having to thread them through explicitly. Pass `mutate_env: false` to opt out and keep the parse-only / instance-isolation semantics introduced in [#578]. `EnvManager.reset!` now also rolls back the keys it added. [#XXX]
```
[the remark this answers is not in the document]
mokagio: Already addressed in the current PR head: the changelog entry now references `#723` instead of the placeholder.

*Posted by Codex (GPT-5) on behalf of @mokagio with approval.*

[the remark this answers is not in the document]
mokagio: Already addressed in the current PR head: the changelog entry now references `#723` instead of the placeholder.

*Posted by Codex (GPT-5) on behalf of @mokagio with approval.*

mokagio:
Thanks for picking this up @iangmaia !

Looks like Copilot has nothing to add to its previous reviews, which I had already addressed.

image

What do you think of this? For live validation, see 4-ghe-Automattic/readablog/files#diff-61bacd7fc8b74e42763f36b953e5dfd00182b336a71446ea1c8ddcfbfcb8f147R10-R15 and 147-ghe-Automattic/everyfirst-app/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR10-R15

iangmaia:
@mokagio I've disabled auto-merge just in case we want to merge @oguzkocer 's https://github.com/wordpress-mobile/release-toolkit/pull/733 first and release a new version before adding this breaking change on `trunk`.

MIGRATION.md
```diff
@@ -1,5 +1,17 @@
 # Migration Instructions for Major Releases
 
+## From 14.x to 15.0.0
+
+### `EnvManager` populates the process `ENV` by default
+
+`EnvManager` now layers the values from the loaded `.env` file into the process `ENV` when you call `set_up`/`new`, so fastlane actions that resolve their `default_value:` via `ENV.fetch(...)` can see them. Pre-existing `ENV` entries always win (no-override), and `reset!` removes only the keys this instance added.
```
iangmaia: This mentions `set_up`/`new`, then says `reset!` removes keys this instance added. Correct me if I'm wrong, but `reset!` only restores the default instance created by `set_up` and callers using `new` need `restore_env!` instead?

lib/fastlane/plugin/wpmreleasetoolkit/env_manager/env_manager.rb
```diff
@@ -31,10 +43,34 @@ def initialize(
           @print_warning_lambda.call("Warning: env file not found at #{@env_path}. Environment variables may not be loaded.")
         end
 
-        # Parse rather than load so we don't mutate the global ENV. Each instance
-        # gets its own view, and values from the process environment (e.g. set by
-        # CI) still take precedence — see `env_value`.
+        # Always parse rather than load: it lets us track exactly which keys we
+        # add to `ENV` so `reset!` can undo only those, without disturbing keys
+        # that pre-existed in the process environment.
         @loaded_env = File.exist?(@env_path) ? Dotenv.parse(@env_path) : {}
```
iangmaia: 💭 One concern, maybe a bit of an edge case: if I understand correctly, it seems `Dotenv.parse` + manual `ENV` mutation does not fully match `Dotenv.load` semantics for interpolation.
If `ENV['A']` already exists and the .env has `A=from_file` plus `B=${A}`, `Dotenv.load` resolves `B` from the pre-existing `ENV['A']`, but this implementation resolves it from the file value.

I wonder if a possible solution could be to use `Dotenv.load` when `mutate_env: true`, record the `ENV` delta for rollback, and keep `Dotenv.parse` only for `mutate_env: false`? 🤔

iangmaia (approved):
Approving to unblock.
I left a couple of comments, I'll leave them up to you. I'd just be careful when finally merging this as it is a breaking version, so perhaps good to keep an eye and coordinate with other PRs we have in the queue.

MIGRATION.md
mokagio: You are right — fixed in ea2c7090.

`EnvManager.reset!` only rolls back the default instance created by `set_up`; an instance you built with `new` rolls back via `restore_env!`. MIGRATION.md now says exactly that.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

lib/fastlane/plugin/wpmreleasetoolkit/env_manager/env_manager.rb
mokagio: Good catch, and confirmed against dotenv 2.8.1 — `Dotenv::Substitutions::Variable` flips precedence on the `is_load` flag: `env.merge(ENV)` for `load`, `ENV.to_h.merge(env)` for `parse`. So with `ENV['A']` set, `parse` really did resolve `B=${A}` from the file value.

Fixed in fbe42534 along the lines you suggested: `mutate_env: true` now uses `Dotenv.load` and records the `ENV` delta around the call for rollback, while `mutate_env: false` keeps `Dotenv.parse` (that mode is about staying out of the process environment entirely). Two specs cover both directions; the `mutate_env: true` one fails against the previous implementation with `expected "from_env", got "from_file"`.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

mokagio:
Thanks again @iangmaia for the review. I finally got onto this because I wanted to ship David's #772 to close the loop on https://github.com/wordpress-mobile/GutenbergKit/issues/591 , but given #773 landed since, and it's a breaking change, I thought I might as well land this and ship two breaking changes in one major version bump.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 - The `ios_build_preflight` and `android_build_preflight` actions have been removed. [#773]
+- `EnvManager`: populate the process `ENV` from the loaded `.env` file by default (no-override semantics — pre-existing `ENV` values win) [#723].
```
```suggestion
- `EnvManager` now populates the process `ENV` from the loaded `.env` file by default [#723].
```

PR #729, opened 2026-06-11
Use Ruby 3.4.9 and re-bundle

mokagio:
I wanted to ship a new version after merging #757 , but run into the nokogiri-3.2.2-Xcode native extension clash.

Given adopting #757 is not super urgent, I decided to detour and (finish the work already started to) support Ruby 3.4.9 in this repo.


Gemfile.lock
```diff
@@ -436,7 +436,6 @@ PLATFORMS
 DEPENDENCIES
   activesupport (~> 8.1)
   buildkite-test_collector (~> 2.3)
-  bundler (~> 2.0)
```
mokagio, line 441: I think this came from the old gem generator. But we don't need to define an explicit dependency on Bundler. As far as I know, we don't use any of its API directly.

.buildkite/shared-pipeline-vars
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-export IMAGE_ID="xcode-26.1.1"
+export IMAGE_ID="xcode-26.6"
 export CI_TOOLKIT="automattic/a8c-ci-toolkit#5.7.0"
 export DOCKER_PLUGIN="docker#v5.13.0"
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-export RUBY_VERSION="$(cat "$SCRIPT_DIR/../.ruby-version")"
+RUBY_VERSION="$(cat "$SCRIPT_DIR/../.ruby-version")"
+export RUBY_VERSION
```
mokagio, line 8: ShellCheck told me about this.

fastlane-plugin-wpmreleasetoolkit.gemspec
```diff
@@ -42,18 +42,4 @@ Gem::Specification.new do |spec|
...
-  # Use at least Fastlane 2.210.0 to ensure compatibility with the Xcode 14 toolchain
-  # See https://github.com/fastlane/fastlane/releases/tag/2.210.0
-  spec.add_development_dependency 'fastlane', '~> 2.210'
-  spec.add_development_dependency 'pry', '~> 0.12.2'
-  spec.add_development_dependency 'rmagick', '~> 5.3'
-  spec.add_development_dependency 'rspec', '~> 3.8'
-  spec.add_development_dependency 'rspec_junit_formatter', '~> 0.4.1'
-  spec.add_development_dependency 'rubocop', '~> 1.65'
-  spec.add_development_dependency 'rubocop-rspec', '3.0'
-  spec.add_development_dependency 'simplecov', '~> 0.16.1'
```
mokagio, line 58: `add_development_dependency` is a valid setting and the Bundler team [rejected a proposal to deprecated it](https://github.com/ruby/rubygems/discussions/5065).

[The docs](https://guides.rubygems.org/make-your-own-gem/) are somewhat ambiguous:

> Gemfile: Manages gem dependencies for development. Contains a gemspec line meaning that Bundler will include dependencies specified in foodie.gemspec too. It’s best practice to specify all dependencies in the gemspec.

Happy to revert if we find this confusing. In which case, though, we should also move the remaining gems from `Gemfile` to here.

AliSoftware (approved):

PR #732, opened 2026-06-15
Bump pinned Ruby to 3.3.4

mokagio:
## What does it do?

Bumps `.ruby-version` from 3.2.2 to 3.3.4.

That file feeds `.buildkite/shared-pipeline-vars`, which derives `RUBY_VERSION` (the Build & Test matrix) and `DOCKER_RUBY_IMAGE` (the gem-publish step), so this moves both CI and release builds onto 3.3.4 in one place.

- [x] Run `bundle exec rubocop` — 198 files, no offenses (locally, under 3.3.4).
- [x] Add Unit Tests if applicable — N/A, config-only change.
- [x] Run `bundle exec rspec` — 893 examples, 0 failures (locally, under 3.3.4); `Gemfile.lock` unchanged after `bundle install`.
- [x] `CHANGELOG.md` entry added under `## Trunk` → `### Internal Changes`.
- [ ] `MIGRATION.md` — N/A, no breaking change.

## How to test

CI itself is the test: this PR's Build & Test step now runs on 3.3.4. Confirm it goes green.


mokagio:
Converting to **draft** — CI surfaced a blocker.

### What happened

The Mac queue ran the full suite green on 3.3.4 (🧪 Build and Test, 2m02s), but both linter-cluster jobs failed instantly:

```
rbenv: version `3.3.4' is not installed (set by .../.ruby-version)
Gem automattic-buildkite-linter is not installed.
```

- Rubocop — failed (rbenv, before any lint ran)
- Danger – PR Check — failed (same rbenv error; **not** a label/milestone issue)

### Root cause

The shared linter-cluster AMI (`buildkite-ci/src/agents/linter/`) provisions a fixed rbenv list — currently **3.2.2, 3.2.7, 3.3.7** — with 3.3.7 as the global default. 3.3.4 isn't in that set, so rbenv aborts before the linter gem loads.

### Why I'm parking it rather than retargeting

Rather than chase a one-off 3.3.4 (or add 3.3.4 to the linter AMI), this should fold into the existing effort: [**Support Ruby 3.4 and 4.0**](https://linear.app/a8c/project/support-ruby-34-and-40-af9813ece5f2). A piecemeal patch bump here would likely be superseded by that work. If we do want an interim bump, 3.3.7 is the version the linter fleet already blesses.

Leaving the branch up for reference.

---

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

PR #733, opened 2026-06-17
Add `find_or_create_pull_request` action

oguzkocer:
## What does it do?

Adds a generic action that returns the URL of the **open** Pull Request for a head branch, creating one only if none exists yet. Backed by a new `GithubHelper#find_pull_request` (list-by-head).

Useful for "rolling" automations (e.g. a daily translations or dependency-update job) that force-push the same head branch each run — GitHub auto-refreshes the existing PR's diff, so a PR only needs to be opened the first time. Token handling uses the standard `github_token_config_item`, and PR parameters are forwarded to `create_pull_request`.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_or_create_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_or_create_pull_request_action.rb
@@ -42,7 +42,7 @@ module Fastlane
 
       def self.details
         <<~DETAILS
-          Looks for an open Pull Request whose head is the given branch (optionally targeting a specific base),
+          Looks for an open Pull Request whose head is the given branch and which targets the given base,
           and returns its URL if found. Otherwise, creates a new Pull Request and returns its URL.
 
           This is useful for "rolling" automations (e.g. a daily translations or dependency-update job) that
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_or_create_pull_request_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_or_create_pull_request_action.rb
@@ -1,0 +1,123 @@
+# frozen_string_literal: true
+
+require 'fastlane/action'
+require_relative '../../helper/github_helper'
+
+module Fastlane
+  module Actions
+    class FindOrCreatePullRequestAction < Action
+      def self.run(params)
+        github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
+
+        existing_pr = github_helper.find_pull_request(
+          repository: params[:repository],
+          head: params[:head],
+          base: params[:base]
+        )
+
+        unless existing_pr.nil?
+          UI.message("An open Pull Request already exists for `#{params[:head]}`: #{existing_pr.html_url}")
+          return existing_pr.html_url
+        end
+
+        other_action.create_pull_request(
+          api_url: params[:api_url],
+          api_token: params[:github_token],
+          repo: params[:repository],
+          title: params[:title],
+          body: params[:body],
+          head: params[:head],
+          base: params[:base],
+          labels: params[:labels],
+          assignees: params[:assignees],
+          reviewers: params[:reviewers],
+          team_reviewers: params[:team_reviewers],
+          milestone: params[:milestone]
+        )
+      end
+
+      def self.description
+        'Returns the URL of the open Pull Request for a head branch, creating one if none exists yet'
+      end
+
+      def self.details
+        <<~DETAILS
+          Looks for an open Pull Request whose head is the given branch (optionally targeting a specific base),
+          and returns its URL if found. Otherwise, creates a new Pull Request and returns its URL.
+
+          This is useful for "rolling" automations (e.g. a daily translations or dependency-update job) that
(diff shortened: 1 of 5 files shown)


spec/github_helper_spec.rb
```diff
@@ -46,6 +46,46 @@ def download_file_from_tag(download_folder:)
...
+        'auto_paginate=': nil
+      )
+    end
+
+    before do
+      allow(Octokit::Client).to receive(:new).and_return(client)
+    end
+
+    it 'qualifies an unqualified head with the repository owner and forwards the base' do
+      expect(client).to receive(:pull_requests).with(test_repo, { state: 'open', head: 'repo-test:my-branch', base: 'trunk' })
```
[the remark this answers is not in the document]
oguzkocer: Claude: Leaving this as-is. These tests rely on the `instance_double(Octokit::Client, pull_requests: [found_pr], …)` default return, so `client.pull_requests(...)` returns `[found_pr]` and `.first` never hits `nil` (the suite is green — it would raise otherwise). This matches the existing pattern in this file, e.g. `expect(client).to receive(:list_milestones)` paired with an `instance_double` default. Adding `.and_return` onto the `expect` would also conflict with rubocop's `RSpec/StubbedMock`.

spec/github_helper_spec.rb
```diff
@@ -46,6 +46,46 @@ def download_file_from_tag(download_folder:)
...
+      find_pull_request(head: 'my-branch', base: 'trunk')
+    end
+
+    it 'uses an already-qualified head as-is and omits the base when not provided' do
+      expect(client).to receive(:pull_requests).with(test_repo, { state: 'open', head: 'someone:other-branch' })
```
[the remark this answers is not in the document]
oguzkocer: Claude: Leaving this as-is. These tests rely on the `instance_double(Octokit::Client, pull_requests: [found_pr], …)` default return, so `client.pull_requests(...)` returns `[found_pr]` and `.first` never hits `nil` (the suite is green — it would raise otherwise). This matches the existing pattern in this file, e.g. `expect(client).to receive(:list_milestones)` paired with an `instance_double` default. Adding `.and_return` onto the `expect` would also conflict with rubocop's `RSpec/StubbedMock`.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_or_create_pull_request_action.rb
```diff
@@ -0,0 +1,123 @@
...
+require 'fastlane/action'
+require_relative '../../helper/github_helper'
+
+module Fastlane
+  module Actions
+    class FindOrCreatePullRequestAction < Action
+      def self.run(params)
+        github_helper = Fastlane::Helper::GithubHelper.new(github_token: params[:github_token])
+
+        existing_pr = github_helper.find_pull_request(
```
iangmaia, line 12: `api_url` is used on `create_pull_request` but not here. I guess we'd need to change the helper for that but it probably makes sense, for consistency?

oguzkocer: My understanding is that adding `api_url` requires changing a shared helper used across ~14 call sites, and a behavior change for [create_release_backmerge_pull_request_action](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb) which will now call the custom host for `@client.user` calls. That, in my opinion, requires its own PR.

Since this PR follows the existing [create_release_backmerge_pull_request_action](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb) from the repo, I'd prefer to leave that to a follow up if you don't mind.

iangmaia: Gotcha, makes sense 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_or_create_pull_request_action.rb
```diff
@@ -0,0 +1,123 @@
...
+          head: params[:head],
+          base: params[:base]
+        )
+
+        unless existing_pr.nil?
+          UI.message("An open Pull Request already exists for `#{params[:head]}`: #{existing_pr.html_url}")
+          return existing_pr.html_url
+        end
+
+        other_action.create_pull_request(
```
iangmaia, line 23: 💭 nit: this call supports creating drafts, we could forward a `draft` parameter to allow for creating a draft PR. Not a big deal and we can address it in a follow-up.

iangmaia (approved):
Thanks for implementing this, sounds like a useful addition 👍 

Apologies for not already approving it yesterday but I was already at the end of my day.

lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_or_create_pull_request_action.rb
oguzkocer: Addressed in dd96bb56.

@iangmaia If that looks good, please feel free to merge the PR, so it doesn't block that PR you mentioned me on.

PR #735, opened 2026-06-18
Add `ContinuousBuildCodeFormatter` for continuous-trunk Android `versionCode`s

oguzkocer:
## What does it do?

Adds a new build-code formatter, `ContinuousBuildCodeFormatter`, that derives an Android Play Store `versionCode` for a "continuous trunk" release model where the low-order term is a high-cardinality, monotonically increasing build number (e.g. a Buildkite build number).

The build code is computed as:

```
versionCode = (major * 10 + minor) * 10^build_digits + build_number
```

- `build_digits` is configurable and defaults to `6` (multiplier = `1_000_000`). For example, `major: 26, minor: 9, build_number: 84231` → `269_084_231`.
- Takes explicit `build_code(major:, minor:, build_number:)` keyword arguments rather than an `AppVersion`. There is deliberately no `patch` input: in a continuous-trunk model hotfixes are disambiguated by the globally monotonic build number rather than a patch digit, so there is no patch to ignore.
- The `build_number` is any high-cardinality, monotonically increasing counter (e.g. a Buildkite build number); the formatter is source-agnostic.
- Returns a raw `Integer` (Android `versionCode` is itself an integer).

### Validation / guards

- Raises (`UI.user_error!`) when `minor > 9`, since `major * 10 + minor` is only unambiguous while `minor` is a single digit.
- Raises (`UI.user_error!`) when the derived code would exceed the Play Store maximum `versionCode` (`2_100_000_000`).
- The constructor validates that `build_digits` is a positive integer.

### Why a new formatter (instead of changing `DerivedBuildCodeFormatter`)

`DerivedBuildCodeFormatter` is fixed-width string concatenation: it caps the total at 8 digits, caps each component at 3 digits (so build ≤ 999), and always includes a patch digit. It therefore cannot hold a Buildkite build number (already ~27,000) and isn't suited to continuous builds. That formatter is correct for the conventional release model, so its contract is left unchanged; this new formatter targets the continuous-trunk case.

Because the build number is globally monotonic and the version prefix only ever increases, the resulting code is always strictly increasing — even if the build number eventually exceeds `10^build_digits` (which only costs human-readability, not ordering). The only hard correctness constraint is staying at or below the Play Store cap.

- [x] If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to describe how the changes will affect the migration from the previous major version and what the clients will need to change and consider. _(N/A — additive change, no breaking changes.)_

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
@@ -51,6 +51,12 @@ module Fastlane
         # @return [Integer] The derived `versionCode`.
         #
         def build_code(major:, minor:, build_number:)
+          # Validate up front so bad input (e.g. strings from env vars or file reads) raises a
+          # user-friendly error rather than an opaque `TypeError` from the arithmetic below.
+          validate_component!('major', major)
+          validate_component!('minor', minor)
+          validate_component!('build_number', build_number)
+
           # `major * 10 + minor` is only unambiguous while minor is a single digit.
           if minor > 9
             UI.user_error!("Minor version (#{minor}) must be 9 or lower to derive an unambiguous build code with `#{self.class.name}`")
@@ -67,6 +73,23 @@ module Fastlane
         end
 
         private
+
+        # Validates that a version component is a non-negative integer.
+        #
+        # @param [String] name The component name, used in the error message
+        # @param [Integer] value The value to validate
+        #
+        # @raise [StandardError] If the value is not a non-negative integer
+        #
+        def validate_component!(name, value)
+          unless value.is_a?(Integer)
+            UI.user_error!("`#{name}` must be an integer, got: #{value.class}")
+          end
+
+          return unless value.negative?
+
+          UI.user_error!("`#{name}` must be a non-negative integer, got: #{value}")
+        end
 
         # Validates that `build_digits` is a positive integer.
         #
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
@@ -14,8 +14,14 @@ module Fastlane
       #
       #   versionCode = (major * 10 + minor) * 10^build_digits + build_number
       #
-      # There is deliberately no patch input: in a continuous-trunk model hotfixes are disambiguated
-      # by the (globally monotonic) build number, not by a patch digit.
+      # It takes `major`, `minor`, and `build_number` as explicit arguments rather than an
+      # `AppVersion`, because the inputs come from different sources and an `AppVersion` does not
+      # model them: the marketing `major`/`minor` come from the parsed version, while `build_number`
+      # is an independent CI counter (e.g. `BUILDKITE_BUILD_NUMBER`). Notably, `AppVersion#build_number`
+      # means something else in this domain (the RC/beta iteration counter, e.g. `-rc-1`), so taking an
+      # `AppVersion` here would invite reading the wrong field. There is also no `patch`: in a
+      # continuous-trunk model the build number strictly orders every build and subsumes patch's
+      # ordering role (hotfixes get a new build number, not a patch digit in the code).
       #
       # Because the build number is globally monotonic and the version prefix only ever increases,
       # the resulting code is always strictly increasing — even if the build number eventually
@@ -40,7 +46,7 @@ module Fastlane
         # @param [Integer] major The major (marketing) version number.
         # @param [Integer] minor The minor (marketing) version number. Must be 9 or lower.
         # @param [Integer] build_number A high-cardinality, monotonically increasing build number
-        # (e.g. a Buildkite build number).
+        # (e.g. a Buildkite build number). This is a CI counter, not `AppVersion#build_number`.
         #
         # @return [Integer] The derived `versionCode`.
         #
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
@@ -65,6 +65,11 @@ module Fastlane
           prefix = (major * 10) + minor
           code = (prefix * (10**@build_digits)) + build_number
 
+          # Sanity check: Play Store versionCodes must be positive integers.
+          if code <= 0
+            UI.user_error!("Derived build code (#{code}) must be a positive integer")
+          end
+
           if code > MAX_PLAY_STORE_VERSION_CODE
             UI.user_error!("Derived build code (#{code}) exceeds the maximum allowed Play Store versionCode (#{MAX_PLAY_STORE_VERSION_CODE})")
           end
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
@@ -30,7 +30,7 @@ module Fastlane
       #
       # Unlike `DerivedBuildCodeFormatter` (fixed-width string concatenation capped at 8 total digits
       # and 3 digits per component, i.e. build <= 999), this formatter can hold a large build number.
-      # The two target different release models; this one does not replace the other.
+      # The two formatters target different release models; this one does not replace the other.
       class ContinuousBuildCodeFormatter
         # @param [Integer] build_digits Number of digits reserved for the build number, which sets the
         # multiplier applied to the `major * 10 + minor` prefix (multiplier = 10^build_digits).
--- a/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+module Fastlane
+  module Wpmreleasetoolkit
+    module Versioning
+      # Google Play Store's maximum allowed versionCode.
+      MAX_PLAY_STORE_VERSION_CODE = 2_100_000_000
+
+      # The `ContinuousBuildCodeFormatter` derives an Android Play Store `versionCode` for a
+      # "continuous trunk" release model, where the low-order term is a high-cardinality,
+      # monotonically increasing build number (e.g. a Buildkite build number).
+      #
+      # The build code is computed as:
+      #
+      #   versionCode = (major * 10 + minor) * 10^build_digits + build_number
+      #
+      # There is deliberately no patch input: in a continuous-trunk model hotfixes are disambiguated
+      # by the (globally monotonic) build number, not by a patch digit.
+      #
+      # Because the build number is globally monotonic and the version prefix only ever increases,
+      # the resulting code is always strictly increasing — even if the build number eventually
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
```diff
@@ -0,0 +1,83 @@
...
+      # by the (globally monotonic) build number, not by a patch digit.
+      #
+      # Because the build number is globally monotonic and the version prefix only ever increases,
+      # the resulting code is always strictly increasing — even if the build number eventually
+      # exceeds `10^build_digits` (which only costs human-readability, not ordering). The only hard
+      # correctness constraint is staying at or below the Play Store's max versionCode.
+      #
+      # Unlike `DerivedBuildCodeFormatter` (fixed-width string concatenation capped at 8 total digits
+      # and 3 digits per component, i.e. build <= 999), this formatter can hold a large build number.
+      # The two target different release models; this one does not replace the other.
```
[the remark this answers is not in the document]
oguzkocer: Addressed in f01b6b87

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
```diff
@@ -0,0 +1,83 @@
...
+        #
+        def build_code(major:, minor:, build_number:)
+          # `major * 10 + minor` is only unambiguous while minor is a single digit.
+          if minor > 9
+            UI.user_error!("Minor version (#{minor}) must be 9 or lower to derive an unambiguous build code with `#{self.class.name}`")
+          end
+
+          prefix = (major * 10) + minor
+          code = (prefix * (10**@build_digits)) + build_number
+
```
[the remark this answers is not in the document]
oguzkocer: Addressed in c906a6b9

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
```diff
@@ -0,0 +1,112 @@
...
+      #   versionCode = (major * 10 + minor) * 10^build_digits + build_number
+      #
+      # It takes `major`, `minor`, and `build_number` as explicit arguments rather than an
+      # `AppVersion`, because the inputs come from different sources and an `AppVersion` does not
+      # model them: the marketing `major`/`minor` come from the parsed version, while `build_number`
+      # is an independent CI counter (e.g. `BUILDKITE_BUILD_NUMBER`). Notably, `AppVersion#build_number`
+      # means something else in this domain (the RC/beta iteration counter, e.g. `-rc-1`), so taking an
+      # `AppVersion` here would invite reading the wrong field. There is also no `patch`: in a
+      # continuous-trunk model the build number strictly orders every build and subsumes patch's
+      # ordering role (hotfixes get a new build number, not a patch digit in the code).
```
oguzkocer, line 24: I've shared this PR with @jkmassel and he suggested that we should still use `AppVersion`. In my opinion, that would be incorrect because even though the shape of the type matches, the contract - specifically how the `AppVersion#build_number` property is utilized in the client - is different.

I'll leave the decision to infra folks.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
```diff
@@ -0,0 +1,112 @@
...
+      #
+      # Because the build number is globally monotonic and the version prefix only ever increases,
+      # the resulting code is always strictly increasing — even if the build number eventually
+      # exceeds `10^build_digits` (which only costs human-readability, not ordering). The only hard
+      # correctness constraint is staying at or below the Play Store's max versionCode.
+      #
+      # Unlike `DerivedBuildCodeFormatter` (fixed-width string concatenation capped at 8 total digits
+      # and 3 digits per component, i.e. build <= 999), this formatter can hold a large build number.
+      # The two formatters target different release models; this one does not replace the other.
+      class ContinuousBuildCodeFormatter
```
oguzkocer, line 34: Another suggestion from Jeremy was to modify one of the existing formatters, most likely `DerivedBuildCodeFormatter`. I don't think that's the right fit: it's fixed-width (caps the build number at 999 and always encodes `patch`), so loosening it to hold a Buildkite-sized build number and drop `patch` would change the contract the existing consumers rely on.

Same as the other comment, I'll leave the decision to infra folks.

oguzkocer (commented):
I asked Jeremy to look at this PR before marking it as ready for review, specifically to make sure that the version calculation would work for our continuous releases project for WordPress/Jetpack apps.

He has shared a couple implementation suggestions with me and I've included them for transparency. We didn't get a chance to settle our discussion yesterday and I believe he might be AFK today, so I figured posting his suggestions alongside the PR would be the best way to keep things moving.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
```diff
@@ -0,0 +1,112 @@
...
+          end
+
+          prefix = (major * 10) + minor
+          code = (prefix * (10**@build_digits)) + build_number
+
+          if code > MAX_PLAY_STORE_VERSION_CODE
+            UI.user_error!("Derived build code (#{code}) exceeds the maximum allowed Play Store versionCode (#{MAX_PLAY_STORE_VERSION_CODE})")
+          end
+
+          code
```
iangmaia, line 77: Maybe worth it to add a guard against `code <= 0` just in case?

oguzkocer: Addressed in 59bf8640c49ad49217847ece23f691849c5b64e2.

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
iangmaia: Indeed avoiding another formatter would be ideal, if the existing abstraction can naturally grow.

Though indeed as it currently is, it feels like a different release model rather than an extension of the same one?
So in principle a separate formatter is cleaner given `DerivedBuildCodeFormatter`'s current constraints (fixed-width, contract including patch, etc), but maybe it could be refactored/broken down to be more flexible?

But to be honest, I'm also not quite sure what's the best route in both questions. Given we have:
1. A working implementation, 
2. Not too many potential clients using the new Version / formatter classes,
3. Probably not too many versions and formatters we want to add,
4. The fact that this entire versioning part of the toolkit doesn't change a lot,

iterating / making changes in the future if needed (or if reassessed) doesn't sound like a big problem, so I'm personally leaning towards just going with the current code as it seems to fix the problem and the current abstractions don't seem to accommodate the new changes very well.

_[pinging @mokagio just to get another AppsInfra opinion just in case you have some time during in your day]_

lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/continuous_build_code_formatter.rb
iangmaia: Hmm that's an interesting one and totally understand where @jkmassel is coming from: it would keep this formatter closer and consistent with the existing formatter APIs and avoid introducing a one-off calling convention. I think we probably even have Version / Formatters that don't quite follow the canonical examples.

At the same time, while `AppVersion` kinda fits, I believe the semantics of `AppVersion#build_number` already means the parsed version/build component in this toolkit, while this formatter needs an external monotonic CI counter 🤔 So reusing `AppVersion` would make the API more consistent, but it would also make it easier to pass the wrong value without noticing.

I'm not fully convinced either way and wrote a bit more about it in the comment [below](https://github.com/wordpress-mobile/release-toolkit/pull/735#discussion_r3454872029).

iangmaia (approved):
Approving to unblock.

I left a [comment](https://github.com/wordpress-mobile/release-toolkit/pull/735#discussion_r3454872029) and I'm in two minds about it (would be curious to hear @mokagio's opinion), but I'm leaning towards going with the existing code for now.

mokagio (commented):
@oguzkocer @iangmaia 

> In my opinion, that would be incorrect because even though the shape of the type matches, the contract - specifically how the `AppVersion#build_number` property is utilized in the client - is different.

> At the same time, while `AppVersion` kinda fits, I believe the semantics of `AppVersion#build_number` already means the parsed version/build component in this toolkit, while this formatter needs an external monotonic CI counter 🤔 So reusing `AppVersion` would make the API more consistent, but it would also make it easier to pass the wrong value without noticing.

Makes sense to me. Two different jobs. `ContinuousBuildCodeFormatter` encapsulates a very specific version code logic. `AppVersion` "represents a version of an app with major, minor, patch, and build number components."

On top of what @iangmaia already observed in regards to [iterating on this](https://github.com/wordpress-mobile/release-toolkit/pull/735/changes#r3454872029), nothing stops us from building an `AppVersion` implementation that uses `ContinuousBuildCodeFormatter` under the hood.

@oguzkocer I looked on WordPress-Android for a PR using this code and I haven't found it yet. Have you considered trying to integrate first? Just in case it reveals some rought edges that are worth addressing before merging? Not a big deal anyway: any change to this object won't affect other consumers, given it's new.

One more question: Is there a particular advantage in composing the version code this way, rather than, say injecting the Buildkite build number only? Initially, I thought it might be because of Google's requirement on ever-increasing build numbers (remembering the early version management headaches with hotfixes and beta builds) but then I read this in the header docs:

> in a continuous-trunk model the build number strictly orders every build and subsumes patch's ordering role (hotfixes get a new build number, not a patch digit in the code)

That makes me think that we'll embrace trunk-based-development and therefore won't really hotfix builds (as in a fix on top of a shipped release ignoring what went on trunk since) anymore.

I can see how including the major and minor in the version code makes puts all information in a single place. In my experience with Apple, however, version name and code often travel together, so a single number containing all the info is not as useful as one might expect—IMHO, of course.

I can also see how having a bit of logic between Buildkite and what goes in the app decouples us the vendor, which is nice to have.

mokagio (approved):
_I mean to approve earlier_

oguzkocer:
@iangmaia @mokagio Thank you both for the review!

I've "integrated" it to WPAndroid, but didn't test it, so... that probably doesn't count. The reason being that once I do test, the internal track can never go back to a lower `versionCode` value. So, I wanted both @jkmassel & infra to sign off on this before I tested it. I'd be happy to leave this unmerged and test it end to end tomorrow before merging it in.

As for whether we should use the standalone Buildkite build number - I think that's a reasonable option and the one @jkmassel initially proposed. I suggested we encode the marketing name, even if in this slightly clipped form, so that:
1. It's easier to match a version code to marketing version name in Play Store without having to track it down
2. To partially de-couple it from Buildkite in case of a future CI provider switch. This is an addressable issue of course if we go with the standalone Buildkite build number, but it is cleaner when the most important digits in the `versionCode` are not tied to an external system.

Those were my arguments, but I don't feel particularly strong about them and would be happy to make the switch if anyone does. So, please let me know if any of you prefer the standalone version.

mokagio:
@oguzkocer 

> The reason being that once I do test, the internal track can never go back to a lower versionCode value.

Right! I didn't consider that.

> It's easier to match a version code to marketing version name in Play Store without having to track it down

I see... I can see how Google sometimes doesn't keep the two together, e.g. in the AAB file name here (the version name is visible in the UI, but once you download the file, you only have the version code).

image

Sounds good then. Ship it.

_(If we ever change our mind, it should be "easy" to pad a build so that it becomes something like `1000<CI build number>` and then essentially have only the CI value drive the number. Just an escape hatch idea to lower the stakes and be more relaxed about merging sooner.)_

Issue #736, opened 2026-06-19
New action: localization placeholder-compatibility check (catch placeholder-breaking changes to existing keys)


jkmassel:
## Summary

Add a reusable **placeholder-shape primitive** plus an action that **fails when an existing localization key's source-language value changes its format placeholders** — count, position, or argument type — between two versions of the source strings. Such a change silently breaks every existing translation for that key.

This is a **different axis** from `ios_lint_localizations`, which compares each *translation* against the *base* language at a single point in time. This compares **base ↔ base across versions** (temporal).

> Filed for the WordPress-iOS continuous-localization work ("Faster Releases" RFC). Reference implementation: wordpress-mobile/WordPress-iOS#25675 — `fastlane/helpers/string_placeholders.rb` + the `validate_string_placeholders` lane. WPiOS will adopt the toolkit version once it lands and delete the project-side copy.

## Motivation

The continuous-localization model uploads regenerated source strings on every trunk merge and downloads whatever human translations exist daily — so **whatever translations exist ship as-is**. That is only safe if a translation key is **never reused for placeholder-incompatible copy**:

- GlotPress (and most TMS) key translations by the source string's **key**, not its current English text.
- If a key's English changes from `"%1$@ liked your post"` to `"%1$d people liked your post"` (object → int), existing translations stay under the same key, and the app formats the new call site's `Int` argument with a translation that still expects an object → wrong output or a crash.

Today this invariant is **convention-only**. We need an automated guardrail in CI.

### The invariant

For every key present in **both** the previous and the newly-generated source-language strings, the set/sequence of format specifiers — accounting for **positional args** and **argument type** — must be identical. New and removed keys are ignored: copy that needs a fresh translation is expected to land under a **new** key (which shows up as remove-old + add-new).

### 1. Reusable primitive (the important part — also consumed by the companion AI-translation backfill issue)

A platform-agnostic helper operating on raw strings / `{ key => value }` hashes:

- `placeholder_signature(value) -> String` — canonical signature of the placeholders in a string (`''` if none).
- `placeholders_compatible?(a, b) -> Boolean` — `signature(a) == signature(b)`.
- `incompatible_placeholder_changes(old_hash, new_hash) -> Array<Hash>` — keys present in both whose signature differs, each `{ key:, old:, new:, old_signature:, new_signature: }`.

### 2. Action

[...]

PR #738, opened 2026-06-19
Add `ios_lint_localization_placeholder_changes` action

jkmassel:
> **Depends on #741.** The `.strings` parsing and duplicate-key detection this action builds on were extracted there, so this PR is now just the primitive + the action. Review/merge #741 first; GitHub will retarget this PR to `trunk` once it lands.

## What does it do?

Adds a placeholder-compatibility guardrail for localization source strings, as requested in #736.

Existing translations are keyed by a string's **key**, not its English text. If an existing key's source value changes its format placeholders — count, position, or argument type — every translation filed under that key silently breaks (or crashes) at the new call site. This is a **different axis** from `ios_lint_localizations`, which compares each _translation_ against the base language at a single point in time; this compares **base ↔ base across versions** (temporal).

### 1. Reusable primitive — `Fastlane::Helper::StringPlaceholdersHelper`

Pure Ruby, no file I/O, platform-agnostic (the `%@` / `%1$d` placeholder syntax is shared by iOS and Android):

- `placeholder_signature(value)` — canonical signature of a value's placeholders (`''` if none), e.g. `"1:int,2:object"`.
- `placeholders_compatible?(a, b)` — `signature(a) == signature(b)`.
- `incompatible_placeholder_changes(old_hash, new_hash)` — the keys present in **both** whose signature differs. New/removed keys are ignored on purpose: copy that needs a fresh translation is expected to land under a new key.
- `mixed_operators?(value)` — whether a value invalidly mixes positional (`%1$@`) and non-positional (`%@`) specifiers.

The signature is invariant to benign copy edits and to reordering equivalent positional args, but sensitive to count/position/type changes. `%d`↔`%i` are compatible; `%%` is skipped; the `,`/`'` grouping flags are recognized; `*` dynamic width/precision and the write-back `%n` are deliberately left unrecognized. A single literal `%` followed by whitespace is treated as a literal percent, not a specifier, so ordinary copy like `"50% off"` doesn't fabricate a phantom placeholder.

### 2. iOS action — `ios_lint_localization_placeholder_changes`

Takes `old_file` / `new_file` (the previous and newly-generated base `.strings`), parses them via the existing `L10nHelper.read_strings_file_as_hash` (plutil), runs the primitive, and reports. Mirrors `ios_lint_localizations`'s `abort_on_violations` (default `true`).

It **fails closed** on any input where the comparison would be unreliable:

- **`check_duplicate_keys`** (default `true`) — aborts when either file defines a key more than once (`plutil` silently keeps only the *last* value, which could otherwise hide a real placeholder change). Delegates to the shared `StringsFileValidationHelper.scan_for_duplicate_keys` from #741.
- **Always aborts** (independently of `abort_on_violations`) when a value mixes positional and non-positional placeholders — that's invalid and leaves the signature undefined.

## Worked examples (match the issue spec — verified by tests)

| Value | Signature |
|---|---|
| `Just text` | (empty) |
| `Hello %@` | `1:object` |
| `%1$d items in %2$@` | `1:int,2:object` |
| `%2$@ told by %1$@` | `1:object,2:object` |
| `100%% sure about %@` | `1:object` |
| `%d likes` | `1:int` |

## Notes / scope

- **Builds on #741**, which carries the `.strings` parsing (unquoted keys/values, inter-token comments) and the `scan_for_duplicate_keys` helper this action calls.
- **Android wrapper deferred to a follow-up**, as the issue permits — the primitive is already platform-agnostic, so the follow-up is a thin `android_*` parse-and-call wrapper.
- **Reuses `L10nHelper.read_strings_file_as_hash`** rather than adding a second `.strings` parser — the action is therefore macOS-only by design, like `ios_lint_localizations`.
- `.stringsdict` / ICU plurals are out of scope (likely v2).
- No `MIGRATION.md` entry — purely additive.

## Test Plan

- [x] `bundle exec rubocop` — no violations.
- [x] `bundle exec rspec` — all green. Helper specs cover every worked example + edge cases (`%%`, a literal `%` in copy, nil/whitespace, non-ASCII, escapes, width/precision/length modifiers, `%d`↔`%i`, the `,`/`'` grouping flags, `*` dynamic width/precision, `%n`, `%.f`, and the invalid positional/non-positional mix). Action specs write real `.strings` fixtures parsed by `plutil`: no-op, abort, `abort_on_violations: false`, add/remove ignored, duplicate-key abort, fail-closed on an un-tokenizable file, mixed-operator abort, and missing file.

## Related issues

- Implements the primitive + iOS wrapper from #736.
- Foundation (`.strings` parsing + `scan_for_duplicate_keys`) extracted to #741.
- Reference implementation: wordpress-mobile/WordPress-iOS#25675 (`fastlane/helpers/string_placeholders.rb` + the `validate_string_placeholders` lane). WPiOS will adopt the toolkit version and delete the project-side copy once this lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ _None_
 ### New Features
 
 - Added `find_or_create_pull_request` action and `GithubHelper#find_pull_request`: returns the URL of the open Pull Request for a head branch, creating one only if none exists yet. Useful for "rolling" automations (e.g. a daily translations or dependency-update job) that force-push the same head branch on every run. [#733]
+- Added the platform-agnostic `StringPlaceholdersHelper` primitive (`placeholder_signature`, `placeholders_compatible?`, `incompatible_placeholder_changes`, `mixed_operators?`) and the `ios_lint_localization_placeholder_changes` action: fails when an existing localization key's source-language value changes its format placeholders — count, position, or argument type — between two versions of the base `.strings` file. Such a change silently breaks every existing translation filed under that key. Complements `ios_lint_localizations` on the temporal (base↔base across versions) axis. The action also aborts (by default, via `check_duplicate_keys`) when either input file defines a key more than once, since `plutil` silently keeps only the last value and a duplicate could otherwise hide a real placeholder change, and always aborts when a source value mixes positional (`%1$@`) and non-positional (`%@`) placeholders, which is invalid and makes the placeholder shape impossible to compare reliably. [#738]
 
 ### Bug Fixes
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localization_placeholder_changes.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localization_placeholder_changes.rb
@@ -1,0 +1,275 @@
+# frozen_string_literal: true
+
+require 'fastlane/action'
+require_relative '../../helper/string_placeholders_helper'
+require_relative '../../helper/ios/ios_l10n_helper'
+require_relative '../../helper/ios/ios_strings_file_validation_helper'
+
+module Fastlane
+  module Actions
+    class IosLintLocalizationPlaceholderChangesAction < Action
+      def self.run(params)
+        # Parse and validate both inputs first, so a genuinely invalid file (a non-dictionary plist, an
+        # unparseable file) surfaces an accurate input error rather than tripping the duplicate-key
+        # scanner below and reporting it as a duplicate-check failure.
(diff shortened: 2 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #741, opened 2026-06-23
Improve `.strings` handling: unquoted strings, inline comments, and merge prefixing

jkmassel:
## Summary

Related improvements to the existing `.strings` handling — the duplicate-key scanner (`StringsFileValidationHelper`) and the `merge_strings` helper — extracted from #738 so the new `ios_lint_localization_placeholder_changes` action, which builds on these, can be reviewed on its own. This PR stands alone and is independently useful.

The common thread: each change accepts more of the unquoted-string grammar `plutil` itself allows, so legitimate `InfoPlist.strings`-style input — unquoted keys/values, inter-token comments — is handled instead of rejected.

### 1. Parse unquoted `.strings` values and widen the unquoted grammar

`StringsFileValidationHelper` already tokenized unquoted *keys*; it now also tokenizes unquoted *values* and accepts the full character set `plutil` allows for unquoted strings (alphanumerics plus `_ . - $ : /`). So an `InfoPlist.strings`-style file like `CFBundleName = WordPress;` parses instead of raising `Invalid character`. The scanner only ever runs on input `plutil` has already accepted, so matching its grammar keeps a parseable file from tripping it; a genuinely un-tokenizable plist (e.g. a nested-dictionary value) still fails closed.

### 2. Accept comments placed *between* a statement's tokens

`StringsFileValidationHelper` recognized comments only in its top-level `:root` context, so a comment between a statement's tokens — `"key" /* note */ = "value";`, `"key" = /* note */ "value";`, or `"key" = "value" /* note */;` — raised `Invalid character`, even though `plutil` accepts all three. The scanner now threads a `resume_context` so a comment returns to the state it interrupted instead of always dropping back to `:root`. The lone ambiguous spot — a `/` right after `=`, which could begin a comment **or** a `/usr/bin`-style unquoted value — is disambiguated one character later via a new `:maybe_comment_or_value` state, so `/`-leading unquoted values still parse. ✅

### 3. Centralize duplicate-key detection behind `scan_for_duplicate_keys`

`ios_lint_localizations` open-coded a `strings_file_type == :text ? scan : warn` gate. That logic now lives in `StringsFileValidationHelper.scan_for_duplicate_keys`, which detects the file format and returns a `[:scanned | :unsupported_format | :unscannable, payload]` tri-state; each caller maps it to its own policy. As a side effect, `ios_lint_localizations`' `check_duplicate_keys` **no longer crashes the lane** on a valid-but-not-flat-`.strings` file — it warns via `UI.important` and skips it.

### 4. `assume_valid:` to skip a redundant `plutil -lint`

`L10nHelper.strings_file_type` gains an opt-in `assume_valid:` flag that skips the `plutil -lint` validity check when the caller has already parsed the file. Default behavior is unchanged for every existing caller. (Used by #738.)

### 5. Apply the `merge_strings` key prefix to unquoted keys and values

`L10nHelper.merge_strings` rewrites each key to add the caller's prefix, but its line matcher only handled `[A-Z0-9_]` keys with a **quoted** value. An unquoted key containing `. - $ : /` (e.g. a reverse-DNS key), or any line with an **unquoted** value (`CFBundleName = WordPress;`), was written to the merged file **without** the prefix — while still being bookkept *with* it. That inconsistency can resurface the very collisions the prefix exists to avoid and break downstream key extraction. The matcher now accepts the full unquoted-string set `plutil` allows; the unquoted-value case keys on the fact that an unquoted string can't contain spaces, so it won't mistake comment prose like `and = a red herring …` for a key/value pair (a case the fixtures deliberately guard). ✅

## Test Plan

- [x] `bundle exec rubocop` — no violations.
- [x] `bundle exec rspec` — 914 examples, 0 failures; new coverage for unquoted values, the widened key set, `scan_for_duplicate_keys`, the graceful warn-and-skip, `assume_valid`, inter-token comments, and `merge_strings` unquoted-key/value prefixing.
- [x] CHANGELOG updated.
- [ ] MIGRATION.md — N/A (additive + backward-compatible).

## Related

- #738 will be rebased to depend on this.


lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -56,15 +56,20 @@ def self.find_duplicated_keys(params)
           language = File.basename(File.dirname(file), '.lproj')
           path = File.join(params[:input_dir], file)
 
-          file_type = Fastlane::Helper::Ios::L10nHelper.strings_file_type(path: path)
-          if file_type == :text
-            duplicates = Fastlane::Helper::Ios::StringsFileValidationHelper.find_duplicated_keys(file: path)
-            duplicate_keys[language] = duplicates.map { |key, value| "`#{key}` was found at multiple lines: #{value.join(', ')}" } unless duplicates.empty?
-          else
+          status, payload = Fastlane::Helper::Ios::StringsFileValidationHelper.scan_for_duplicate_keys(file: path)
```
mokagio, line 59: Nice appraoch.

lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb
```diff
@@ -56,15 +56,20 @@ def self.find_duplicated_keys(params)
...
+          case status
+          when :scanned
+            duplicate_keys[language] = payload.map { |key, value| "`#{key}` was found at multiple lines: #{value.join(', ')}" } unless payload.empty?
+          when :unsupported_format
             UI.important <<~WRONG_FORMAT
-              File `#{path}` is in #{file_type} format, while finding duplicate keys only make sense on files that are in ASCII-plist format.
-              Since your files are in #{file_type} format, you should probably disable the `check_duplicate_keys` option from this `#{action_name}` call.
+              File `#{path}` is in #{payload} format, while finding duplicate keys only make sense on files that are in ASCII-plist format.
```
mokagio: Nitpick, even though it was in the previous version "only makes sense" is a bit condescending IMHO.

```suggestion
              File `#{path}` is in #{payload} format, while finding duplicate keys can only occurr on files that are in ASCII-plist format.
```

lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_l10n_helper.rb
```diff
@@ -96,9 +102,17 @@ def self.merge_strings(paths:, output_path:)
...
                   line.gsub!(/^(\s*")/u, "\\1#{prefix}") # Lines starting with a quote are considered to be start of a key; add prefix right after the quote
-                  line.gsub!(/^(\s*)([A-Z0-9_]+)(\s*=\s*")/ui, "\\1\"#{prefix}\\2\"\\3") # Lines starting with an identifier followed by a '=' are considered to be an unquoted key (typical in InfoPlist.strings files for example)
+                  # Lines starting with an unquoted key followed by a `=` and a *quoted* value (typical in `InfoPlist.strings`).
+                  line.gsub!(%r{^(\s*)([a-zA-Z0-9_.$:/-]+)(\s*=\s*")}u, "\\1\"#{prefix}\\2\"\\3")
+                  # Lines starting with an unquoted key followed by a `=` and an *unquoted* value (e.g. `CFBundleName = WordPress;`).
+                  # The value is required to be a single unquoted token ending in `;` — an unquoted string can't contain spaces, so
+                  # this won't mistake comment prose like `and = a red herring for when…` (spaces in the "value") for a key/value pair.
+                  line.gsub!(%r{^(\s*)([a-zA-Z0-9_.$:/-]+)(\s*=\s*)([a-zA-Z0-9_.$:/-]+\s*;)}u, "\\1\"#{prefix}\\2\"\\3\\4")
                 end
```
mokagio: My AI noticed that `merge_strings` doesn't handle inter-token comments.

Could this logic handle the same inter-token comment placements that `StringsFileValidationHelper` now accepts? (https://github.com/wordpress-mobile/release-toolkit/pull/741/changes#diff-6e464dfe8ccd3974026649f8aaa02f9f8518261ef391f13bdb9535ef46d2b908R111-R116)

For example, `CFBundleName /* c */ = WordPress;` is accepted by `plutil` and by the duplicate-key scanner after this PR, but `merge_strings` still writes it unprefixed while bookkeeping it as prefixed.
That can reintroduce collisions in the merged file.

I added failing specs in #742 to pin the gap.

AI suggests: _A tokenizer-backed rewrite would be ideal, but a comment-aware interim rewrite would also close the regression._

jkmassel:
@mokagio – I merged #742 then applied https://github.com/wordpress-mobile/release-toolkit/pull/741/commits/0957623a3df3025d9cc9169df553a7b1fe66480b, which fixed the underlying issue

mokagio:
Nicely done @jkmassel ! I got my agent to poke around more and it found this https://github.com/wordpress-mobile/release-toolkit/pull/750

To be honest, I can't tell whether that's a legit concern for our use case because `"k" = { a = b; };`, while a valid `plist` apparently, is not in a shape I've ever seen working in our domain. So, I guess it's a matter of deciding whether to investigate how to improve the logic in the abstract, or to stay focused on the practical use case we have.

If you need this addressed ASAP, then by all means let's merge and ship. But if you have the time to investigate further, then maybe it'd be nice to address?

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

jkmassel:
@mokagio – I merged https://github.com/wordpress-mobile/release-toolkit/pull/750 with the test and a fix. It properly handles nested dictionaries (where "properly" means `"k" = { a = b; };` becomes `"prefix.k" = { a = b; };`. I tried to keep it the same as `plutil`. 

Apparently it doesn't handle binary or malformed data, but I think we can live with that for the purposes of this PR.

mokagio (approved):
> Apparently it [the implementation] doesn't handle binary or malformed data, but I think we can live with that for the purposes of this PR.

@jkmassel sounds good. One step at a time.

PR #746, opened 2026-06-26
Add `values-car` pruning regression test

mokagio:
## What does it do?

Adds test-only coverage showing that `android_prune_orphaned_translations` currently treats `values-car` as a locale directory and prunes car UI mode resources.
No production fix is included; this is stacked on #734 to make the review finding executable.

Expected failure:

```sh
bundle exec rspec spec/android_prune_orphaned_translations_spec.rb
```

The new example, `leaves car UI mode qualifier directories untouched`, fails with `expected: 0, got: 1`.

## Checklist before requesting a review

- [x] Run `bundle exec rubocop` to test for code style violations and recommendations.
- [x] Add Unit Tests (aka `specs/*_spec.rb`) if applicable.
- [ ] Run `bundle exec rspec` to run the whole test suite and ensure all your tests pass.
- [ ] Make sure you added an entry in [the `CHANGELOG.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md#trunk) to describe your changes under the appropriate existing `###` subsection of the existing `## Trunk` section.
- [x] If applicable, add an entry in [the `MIGRATION.md` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to describe how the changes will affect the migration from the previous major version and what the clients will need to change and consider.

PR #747, opened 2026-06-29
Skip Android UI-mode `values-*` qualifier dirs (e.g. `values-car`) when pruning

oguzkocer:
> Heads up: this is _a_ solution that passes the regression test and is correct for WordPress-Android, but I don't have full context on GlotPress export conventions or every consuming app. Happy for someone with more context (`@wordpress-mobile/apps-infra-tooling`) to take over or steer this differently.

## The situation

#746 (by `@mokagio`, already merged into this branch) added a regression test showing that `android_prune_orphaned_translations` treats `values-car` as a locale and would prune its resources. That test currently fails on this branch.

Root cause: the locale-detection regex's 2–3 letter language branch (`[a-z]{2,3}`) matches `car`. But `values-car` is the Android **UI-mode** qualifier ([docs](https://developer.android.com/guide/topics/resources/providing-resources#UiModeQualifier)), not a locale. The catch is that `car` is *also* a valid 3-letter ISO 639 language code, so there's **no purely syntactic way** for a regex to tell `values-car`-the-UI-mode from `values-car`-the-language — Android itself can't either.

## What this PR does

- Keeps the 2–3 letter locale grammar and explicitly excludes Android's documented UI-mode qualifiers via a `UI_MODE_QUALIFIERS` list. In practice `car` is the only one short enough to collide today; the rest are listed to capture the documented set.
- Adds a `values-kmr` (Northern Kurdish) test so 3-letter legacy locales keep getting pruned.

## Alternatives considered

- **Restrict legacy locales to 2 letters** — simpler, but rejected: WordPress-Android ships `values-kmr`, a real 3-letter legacy locale that must still be pruned. A 2-letter rule would silently skip it.
- **Validate against a real ISO 639 language list** — doesn't help: `car` *is* a valid ISO 639 code, so a language list would still classify `values-car` as a locale.

## Open question (not a blocker)

Could the GlotPress → Android export ever emit *other* folder names that collide with a non-locale qualifier? If so, we'd just extend the same exclusion list. Confirming the export's folder-naming conventions separately; not a blocker for WordPress-Android (no `values-car`).

## Notes

- Targets the #734 branch (`add/prune-orphaned-translations`) so the regression + fix stay contained in one reviewable PR rather than folded into the larger feature PR.
- Full spec + RuboCop green locally.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_prune_orphaned_translations_action.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_prune_orphaned_translations_action.rb
@@ -6,10 +6,18 @@ require 'nokogiri'
 module Fastlane
   module Actions
     class AndroidPruneOrphanedTranslationsAction < Action
-      # Matches an Android `values-<qualifier>` directory whose qualifier is a locale (a language code, optional
-      # region, or a BCP-47 `b+` form), so non-locale qualifier dirs (e.g. `values-night`, `values-v21`,
-      # `values-land`) are left untouched.
+      # Matches an Android `values-<qualifier>` directory whose qualifier is a locale: a 2- or 3-letter language
+      # code (`fr`, `kmr`), an optional region (`pt-rBR`, `es-r419`), or a BCP-47 `b+` form (`b+sr+Latn`), so
+      # non-locale qualifier dirs (e.g. `values-night`, `values-v21`, `values-land`) are left untouched.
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

mokagio (approved):
Thank you for following up on this!

PR #753, opened 2026-07-08
Fix GitHub release asset lookup on upload_github_release_assets

iangmaia:
## What does it do?

Fixes `upload_github_release_assets` release lookup when the releases list misses an existing published release.

The helper still checks `client.releases(repository)` first so draft releases continue to work. If that list does not include the requested tag, it falls back to GitHub's direct release-by-tag endpoint once. A missing release still fails with the existing clear error.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
@@ -200,7 +200,7 @@ module Fastlane
       # @raise [Fastlane::UI::Error] UI.user_error! if the release does not exist.
       #
       def get_release(repository:, version:)
-        release = client.releases(repository).find { |candidate| candidate.tag_name == version }
+        release = find_release(repository: repository, version: version)
         return release unless release.nil?
 
         UI.user_error!("Could not find GitHub Release for tag #{version} in #{repository}")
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
```diff
@@ -244,6 +244,22 @@ def upload_release_assets(repository:, version:, assets:, replace_existing: true
...
+        return release unless release.nil?
+
+        release_for_tag(repository: repository, version: version)
+      end
+
+      def release_for_tag(repository:, version:)
+        client.release_for_tag(repository, version)
+      rescue Octokit::NotFound
+        nil
+      end
```
mokagio, line 258: What about other errors that might get thrown? Are we okay to crash on them? Not sure how strong of a concern it is, but seeing a single `rescue` made me wonder...

spec/github_helper_spec.rb
```diff
@@ -714,6 +717,34 @@ def create_release(is_draft:, assets: [], name: nil)
       end
     end
 
+    it 'uses the release list without calling the direct release-by-tag lookup' do
```
mokagio: "It does something without doing something else" comes across a bit oblique to me. How about:

```suggestion
    it 'uses the release list if available' do
```

mokagio (approved):

lib/fastlane/plugin/wpmreleasetoolkit/helper/github_helper.rb
iangmaia: Followed up in https://github.com/wordpress-mobile/release-toolkit/pull/755

PR #757, opened 2026-07-19
Add `macos_verify_code_signing` action

mokagio:
## What does it do?

Extracts the `verify_code_signing` lane written for the WordPress Contributor Toolkit (WordPress/experimental-wp-dev-env#22) into a reusable action, so every app we build for macOS can assert its artifacts are signed and notarized. `electron-builder`, in particular, only *warns* when it can't find a valid signing identity — it skips signing and produces an artifact that looks fine until a user tries to launch it.

Two deliberate additions over the original lane:

- `xcrun stapler validate`, so a missing notarization ticket fails the job (thanks @iangmaia for [the suggestion](https://github.com/WordPress/experimental-wp-dev-env/pull/22#discussion_r3586443617)).
- `verify_notarization` (default `true`), so callers can verify at a point where the app is signed but not yet notarized — e.g. from an `electron-builder` `afterSign` hook.

See https://linear.app/a8c/issue/AINFRA-2709.

## How to test

`bundle exec rspec spec/macos_verify_code_signing_spec.rb`.

I also smoke-ran it against a real notarized app, which exercises the actual `codesign`/`spctl`/`stapler` invocations the specs mock:

```
bundle exec fastlane run macos_verify_code_signing app_path:/Applications/Claude.app \
  "expected_authority:Developer ID Application: Anthropic PBC (Q6L2SF6YDW)"
```

It passes, and fails as expected when given a different `expected_authority`.

### DIFF
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
@@ -46,7 +46,7 @@ module Fastlane
       def self.verify_disk_image(path:, expected_authority:, verify_notarization:)
         # Said up front because `sh` logs a non-zero exit in red regardless of it being handled,
         # which reads as a broken build in the CI log of an otherwise passing job.
-        UI.message("Checking whether #{path} is signed. Disk images usually aren't, so a `codesign` failure below is expected and tolerated.")
+        UI.message("Checking whether #{path} is signed. Disk images usually aren't, so a `codesign` failure due to the image being unsigned is expected and tolerated (other signature failures will still fail).")
 
         if signed?(path)
           verify_authority!(path: path, expected_authority: expected_authority) unless expected_authority.nil?
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
@@ -29,7 +29,7 @@ module Fastlane
       end
 
       def self.verify_app_bundle(path:, expected_authority:, verify_notarization:)
-        verify!("The code signature of #{path} is not valid", 'codesign', '--verify', '--deep', '--strict', '--verbose=2', path)
+        UI.user_error!("#{path} is not signed at all") unless signed?(path, deep: true)
         verify_authority!(path: path, expected_authority: expected_authority) unless expected_authority.nil?
 
         return unless verify_notarization
@@ -61,8 +61,14 @@ module Fastlane
       # Distinguishes an artifact that carries no signature at all from one whose signature is
       # broken: the former is expected for a disk image, the latter always a failure.
       #
-      def self.signed?(path)
-        exitstatus, output = sh('codesign', '--verify', '--strict', '--verbose=2', path) { |status, result, _| [status.exitstatus, result] }
+      # @param deep [Boolean] Whether to also verify the nested code an app bundle embeds.
+      #
+      def self.signed?(path, deep: false)
+        command = ['codesign', '--verify']
+        command << '--deep' if deep
+        command += ['--strict', '--verbose=2', path]
+
+        exitstatus, output = sh(*command) { |status, result, _| [status.exitstatus, result] }
 
         return true if exitstatus.zero?
         return false if output.include?('not signed at all')
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
@@ -84,22 +84,13 @@ module Fastlane
 
       def self.details
         <<~DETAILS
-          Verify that the given macOS artifacts are signed, and optionally notarized, so that a build
-          that is unsigned or signed with the wrong identity fails the CI job instead of shipping.
-
-          `electron-builder`, in particular, only warns when it can't find a valid signing identity: it
-          skips signing and produces an artifact that looks fine until users try to launch it.
+          Verify that the given macOS artifacts are signed, and optionally notarized.
 
           The checks that apply are picked from the artifact's extension:
 
           - `.app` — the signature is valid and satisfies its designated requirement, Gatekeeper accepts
             the bundle for execution, and a notarization ticket is stapled to it.
-          - `.dmg` — a notarization ticket is stapled to the image. Disk images are frequently left
-            unsigned (`electron-builder` signs only the app inside), in which case the signature checks
-            are skipped with a warning rather than failing.
-
-          Run with `verify_notarization: false` at a point in the build where the artifact has been
-          signed but not notarized yet, such as from an `electron-builder` `afterSign` hook.
+          - `.dmg` — a notarization ticket is stapled to the image.
         DETAILS
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
@@ -97,7 +97,7 @@ module Fastlane
 
           - `.app` — the signature is valid and satisfies its designated requirement, Gatekeeper accepts
             the bundle for execution, and a notarization ticket is stapled to it.
-          - `.dmg` — a notarization ticket is stapled to the image.
+          - `.dmg` — if the image is signed, the signature is valid (and can be checked against the expected authority) and Gatekeeper accepts opening it; when `verify_notarization` is true, a notarization ticket is stapled to the image.
         DETAILS
       end
 
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
@@ -52,7 +52,7 @@ module Fastlane
           verify_authority!(path: path, expected_authority: expected_authority) unless expected_authority.nil?
           verify!("#{path} was rejected by Gatekeeper", 'spctl', '--assess', '--type', 'open', '--context', 'context:primary-signature', '--verbose=2', path) if verify_notarization
         else
-          UI.important("#{path} is not signed — skipping its signature checks. Only the app it contains carries a signature.")
+          UI.important("#{path} is not signed — skipping its signature checks. We expect that the app it contains is the one carrying the signature.")
         end
 
         verify!("#{path} has no notarization ticket stapled to it", 'xcrun', 'stapler', 'validate', path) if verify_notarization
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'fastlane/action'
+require 'fastlane_core/ui/ui'
+
 module Fastlane
   module Actions
     class MacosVerifyCodeSigningAction < Action
--- a/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
+++ b/lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
@@ -4,35 +4,74 @@ module Fastlane
   module Actions
     class MacosVerifyCodeSigningAction < Action
       def self.run(params)
-        app_paths = Array(params[:app_path])
-        UI.user_error!('No app bundle to verify: `app_path` is empty') if app_paths.empty?
+        paths = Array(params[:artifact_path])
+        UI.user_error!('No artifact to verify: `artifact_path` is empty') if paths.empty?
 
-        app_paths.each do |app_path|
-          UI.user_error!("There is no app bundle at #{app_path}") unless File.exist?(app_path)
+        paths.each do |path|
+          UI.user_error!("There is no artifact at #{path}") unless File.exist?(path)
 
-          UI.message("Verifying code signing of #{app_path}")
+          UI.message("Verifying #{path}")
 
-          verify!("The code signature of #{app_path} is not valid", 'codesign', '--verify', '--deep', '--strict', '--verbose=2', app_path)
-          verify_authority!(app_path: app_path, expected_authority: params[:expected_authority]) unless params[:expected_authority].nil?
-
-          next unless params[:verify_notarization]
-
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,142 @@
...
+        'Verify that macOS artifacts are properly code signed and notarized'
+      end
+
+      def self.details
+        <<~DETAILS
+          Verify that the given macOS artifacts are signed, and optionally notarized, so that a build
+          that is unsigned or signed with the wrong identity fails the CI job instead of shipping.
+
+          `electron-builder`, in particular, only warns when it can't find a valid signing identity: it
+          skips signing and produces an artifact that looks fine until users try to launch it.
```
mokagio:
```suggestion
          Verify that the given macOS artifacts are signed, and optionally notarized.
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,142 @@
...
+
+          The checks that apply are picked from the artifact's extension:
+
+          - `.app` — the signature is valid and satisfies its designated requirement, Gatekeeper accepts
+            the bundle for execution, and a notarization ticket is stapled to it.
+          - `.dmg` — a notarization ticket is stapled to the image. Disk images are frequently left
+            unsigned (`electron-builder` signs only the app inside), in which case the signature checks
+            are skipped with a warning rather than failing.
```
mokagio:
```suggestion
          - `.dmg` — a notarization ticket is stapled to the image.
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,142 @@
...
+
+          Run with `verify_notarization: false` at a point in the build where the artifact has been
+          signed but not notarized yet, such as from an `electron-builder` `afterSign` hook.
```
mokagio:
```suggestion
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,142 @@
...
+            key: :artifact_path,
+            description: 'The path, or list of paths, to the `.app` bundle(s) or `.dmg` disk image(s) to verify',
+            is_string: false,
+            verify_block: proc do |value|
+              UI.user_error!('`artifact_path` must be a String or an Array of Strings') unless value.is_a?(String) || value.is_a?(Array)
+            end
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :expected_authority,
+            description: 'The signing authority the artifact is expected to be signed by, e.g. `Developer ID Application: Automattic, Inc. (ABCDE12345)`. ' \
```
mokagio:
```suggestion
            description: 'The signing authority the artifact is expected to be signed by, e.g. `Developer ID Application: ACME, Inc. (ABCDE12345)`. ' \
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,143 @@
...
+          - `.dmg` — if the image is signed, the signature is valid (and can be checked against the expected authority) and Gatekeeper accepts opening it; when `verify_notarization` is true, a notarization ticket is stapled to the image.
+        DETAILS
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(
+            key: :artifact_path,
+            description: 'The path, or list of paths, to the `.app` bundle(s) or `.dmg` disk image(s) to verify',
+            is_string: false,
```
AliSoftware: `is_string:` us deprecated. Can't we use something like `type: Array` instead? 🤔 
(iirc, if you pass a `String` to a `ConfigItem` expecting an `Array`, then fastlane automatically wraps that string into a single-item array for you? To be double-checked with live testing that theory though)

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,143 @@
...
+              next if value.is_a?(String)
+              next if value.is_a?(Array) && value.all?(String)
+
+              UI.user_error!('`artifact_path` must be a String or an Array of Strings')
+            end
+          ),
+          FastlaneCore::ConfigItem.new(
+            key: :expected_authority,
+            description: 'The signing authority the artifact is expected to be signed by, e.g. `Developer ID Application: ACME, Inc. (ABCDE12345)`. ' \
+                         + 'When omitted, any valid signature is accepted',
```
AliSoftware, line 123: 💡 It would be nice to provide some guidance in that `description` on how to get that signing authority (e.g. provide a command to run on a certificate to extract and print that value, or how to get it from your Xcode project, or whatever)

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,143 @@
...
+          UI.message("Verifying #{path}")
+
+          case File.extname(path).downcase
+          when '.app'
+            verify_app_bundle(path: path, expected_authority: params[:expected_authority], verify_notarization: params[:verify_notarization])
+          when '.dmg'
+            verify_disk_image(path: path, expected_authority: params[:expected_authority], verify_notarization: params[:verify_notarization])
+          else
+            UI.user_error!("Don't know how to verify #{path}. Supported artifacts are `.app` bundles and `.dmg` disk images")
+          end
```
AliSoftware, line 25: 💡  Should we also support `.ipa` files and iOS artifacts?

I know you named that action `macos/macos_verify_code_signing.rb` / `MacosVerifyCodeSigningAction`, and that the main use case we have today is for Electron macOS apps, but I figured since the main thing it does it call `codesign --verify` with the proper arguments, maybe it could be useful to support `.ipa` iOS products as well? 🤔 

I know that's not a super strong argument, but that would also avoid the strange case of if we don't rename that action but later need to support iOS/`.ipa` too after all, the `macos_` action name would be awkward to keep and we'd have to either rename the action or implement that is a separate `ios_verify_code_signing` action with very similar logic 🙃 

Anyway, probably not that important, but I still figured I'd leave this as food for thoughts.

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,143 @@
...
+      def self.verify_disk_image(path:, expected_authority:, verify_notarization:)
+        # Said up front because `sh` logs a non-zero exit in red regardless of it being handled,
+        # which reads as a broken build in the CI log of an otherwise passing job.
+        UI.message("Checking whether #{path} is signed. Disk images usually aren't, so a `codesign` failure due to the image being unsigned is expected and tolerated (other signature failures will still fail).")
+
+        if signed?(path)
+          verify_authority!(path: path, expected_authority: expected_authority) unless expected_authority.nil?
+          verify!("#{path} was rejected by Gatekeeper", 'spctl', '--assess', '--type', 'open', '--context', 'context:primary-signature', '--verbose=2', path) if verify_notarization
+        else
+          UI.important("#{path} is not signed — skipping its signature checks. Only the app it contains carries a signature.")
```
AliSoftware: 💄  This message mentioning "Only the app it contains carries a signature" makes it looks like the action actually checked the signature of the app _inside_ the disk image and validated it was signed. While in practice this is more of an assumption. So the wording should probably read something more like:


```suggestion
          UI.important("#{path} is not signed — skipping its signature checks. We expect that the app it contains is the one carrying the signature.")
```

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,143 @@
...
+        end
+
+        verify!("#{path} has no notarization ticket stapled to it", 'xcrun', 'stapler', 'validate', path) if verify_notarization
+      end
+
+      # Distinguishes an artifact that carries no signature at all from one whose signature is
+      # broken: the former is expected for a disk image, the latter always a failure.
+      #
+      def self.signed?(path)
```
AliSoftware: 💭  The name of this helper makes sense on its own, but looking at the rest of the actions' code I see other places where we also call `codesign --verify …` to check a signature (e.g. [line 32 in `verify_app_bundle`](https://github.com/wordpress-mobile/release-toolkit/pull/757/changes#diff-b33e72f67698afdb38c549d07cc891ef64be03d4e35c3dce3d5b2c798290c662R32)). which made me wonder why that line didn't also call `signed?` and what was the difference between those two call sites.

❓ I can see there are subtle differences between L32 and here indeed, but I wonder it would be worth it to make the call site on L32 rely on this helper rather than itself also passing the `codesign …` command and arguments? 🤔 💭

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
```diff
@@ -0,0 +1,143 @@
...
+        exitstatus, output = sh('codesign', '--verify', '--strict', '--verbose=2', path) { |status, result, _| [status.exitstatus, result] }
+
+        return true if exitstatus.zero?
+        return false if output.include?('not signed at all')
+
+        UI.user_error!("The code signature of #{path} is not valid:\n#{output}")
+      end
+
+      def self.verify!(error_message, *command)
```
AliSoftware, line 79: 💄  Pure style nitpick:
While I like the Ruby splat operator and its ability to make a function take arbitrary parameters and convert them into an array for you, in this particular case when I'm looking at the call sites for `verify!` in this code (e.g. `verify!("#{path} was rejected by Gatekeeper", 'spctl', '--assess', '--type', 'execute', '--verbose=2', path)`) for some reason it feels a bit weird to me.

I think it's a combination of having those `*command` parameters being mixed inline with the `error_message` parameter + the fact that neither of the `error_message` nor the `*command` have a parameter label / is a keyword param.

So instead I think one of those alternatives would read better:
```ruby
verify!('spctl', '--assess', '--type', 'execute', '--verbose=2', path, error_message: "#{path} was rejected by Gatekeeper")
verify!("#{path} was rejected by Gatekeeper", command: ['spctl', '--assess', '--type', 'execute', '--verbose=2', path])
verify!(
  command: ['spctl', '--assess', '--type', 'execute', '--verbose=2', path],
  error_message: "#{path} was rejected by Gatekeeper"
)
```
wdyt?

AliSoftware (approved):
Left some nitpicks / cosmetic feedbacks / questions / suggestions, but none of them are blockers (except maybe the question on the action name if we want to later support iOS but don't want to have to rename?), so approving to unblock you.

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
mokagio: Good point. The error message up front followed by all the arguments felt iffy to me too...

"verify" as a verb for what the method does is confusing, also. In the end, it's just a convenience wrapper for 

```ruby
sh(*command, error_callback: ->(_) { UI.user_error!(error_message) })
```

I think maybe a combination of your suggestion to explicitly pass `command:` and `error_message:` plus a rename, `run!` or `execute!`?, would make the code read better.

Given this is an internal refactor and that I'm keep to ship the action to then use it officially in https://github.com/Automattic/wp-calypso/pull/112767, I'm going to punt working on this in a follow up, though. Thanks!

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
mokagio: On the one hand, Apple code signing is rather similar between platforms, so it makes sense to have a generic action that can handle any mix bag of Apple binaries. On the other, the presence of notarization in macOS represent a substantial difference. An API that allows to check code signing for an `.ipa` with `verify_notarization: true` would feel odd, even if the action would print an helpful log such as "Ignoring notarization on IPA input".

If we'll feel the need for an iOS counterpart, I think we could address the code duplication concern by extracting helper objects.

AliSoftware: WFM 👍

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
mokagio: Opus 4.8 input on the naming:

Thread context read. `verify!` is just `sh` with a custom failure message; every call site is an assertion. Here's a draft reply:

> Some options for the follow-up, roughly in increasing order of how much they change:
>
> **1. Name it after the mechanism** — it's `sh` with a custom failure message, nothing more:
>
> ```ruby
> sh_or_fail!(command: ['xcrun', 'stapler', 'validate', path], error_message: "#{path} has no notarization ticket stapled to it")
> run!(command: [...], error_message: "...")
> execute!(command: [...], error_message: "...")
> ```
>
> `sh_or_fail!` is my pick of the three: `run!`/`execute!` don't say what the `!` is for, and plain `sh` already raises — the only thing this wrapper adds is the message.
>
> **2. Name it after what the call sites actually do** — assert:
>
> ```ruby
> assert_command!(command: [...], error_message: "...")
> fail_unless!(command: [...], message: "#{path} was rejected by Gatekeeper")
> ```
>
> `fail_unless!` reads as a sentence at the call site: fail unless this command succeeds.
>
> **3. Drop the generic helper entirely** and give each check its own method:
>
> ```ruby
> assert_gatekeeper_accepts!(path: path, type: 'execute')
> assert_notarization_stapled!(path: path)
> ```
>
> The message stops being a parameter — it lives next to the command it describes — and the ordering/labelling problem disappears. Downside is two more small methods, but they'd sit right alongside `verify_authority!`, which is already shaped this way.
>
> I'd go 3 for the two `stapler`/`spctl` checks and keep something like `sh_or_fail!` around only if a third caller shows up.

Option 3, drop generic helper entirely is attractive because a method named can be clearer than an array of command parameters. It doesn't solve the repetition of the `sh... UI.user_error!...` pattern, but that's a small price to pay and could anyway be addressed anyway with one of our suggestions above, or, my favorite at this time, `fail_unless!(command: [...], failure_message: ...)`.

lib/fastlane/plugin/wpmreleasetoolkit/actions/macos/macos_verify_code_signing.rb
mokagio: Addressed via fd9fe788cad0fedf9637303d40c47921516d1691 adding a `deep` parameter.

